Issue #671, opened 2014-08-21
page.rank with personalized vector


richardbeare:
Hi,
appologies if this is the wrong place to ask this.

I have a query regarding the behaviour of personalization relative to a single node - i.e. a vector with a single 1 and all other entries 0.

Previous versions would fail if the node had no outgoing edges, which makes some sense ("teleporting" to a node with no exiting links should cause problems). The current version (0.7.1) does not appear to suffer issues if the only non zero probability node in the personalization vector has no outgoing links. Are you able to explain the mechanism used to solve the problem? I don't recall anything about this situation in the relative pagerank papers I've read.

I'm working in R, and using the CRAN version.

Thanks

eamanj:
Hi, 

I used to use python 0.6.5 version. My personalized pageranks were sometimes wacky, but in general they were OK. After upgrade to 0.7.0 I see the following error:

    >>> import igraph
    >>> igraph.__version__
    '0.7.0'
    >>> g = igraph.Graph.Formula("A --> B --> C, B --> D")
    >>> g.personalized_pagerank(reset_vertices=["C"])
    [0.13234721681588163, 0.24484235110938102, 0.3864052160373686, 0.23640521603736858]

Personalized pagerank of a node with no path from reset vertex in many graphs is non-zero. Moreover, the regular pagerank is also vastly different than old ARPACK values.
I bet this must be an issue with bad dependencies. Any idea?

dgleich:
Your graph has dangling nodes, these need to be adjusted in a PageRank
computation. The standard adjustment is either -- link them to the reset
vector, or link them to the uniform vector. In this case, you are getting
the uniform vector. I believe there is an option to set them to the reset
vector, but I don't know it off the top of my head.

(So it looks like this is the same issue as before with u != v discussed
above).

David

eamanj:
Thanks for the response.

Sorry, I should have clarified a bit more. From the thread above it looks like the decision was to provide a parameter for this and set the default to u=v.
But the python man page does not mention this. Is it not implemented yet?
http://igraph.org/python/doc/igraph.GraphBase-class.html#personalized_pagerank

ntamas:
This is not implemented yet; the pull request referenced above has not been merged into the master branch yet (and even if it were, it does not contain any updates to the Python interface yet). In igraph 0.7.0 and 0.7.1, it works like this:

- with the ARPACK implementation, the random walker teleports according to the distribution of the reset vector when it encounters a dangling node
[...]

PR #673, opened 2014-08-26
Added support for personalization vector for dangling nodes

richardbeare:
Hi,
Here is my attempt at adding the support for personalization vectors for dangling nodes. The default behaviour is implemented via R default function flags, although there is also some #ifdef code in prpack.cpp to provide a copy of the reset vector if the resetdangling vector is absent.

Testing is not yet thorough, but everything compiles and installs, checks pass and documentation updated.

Thanks for your help, and I hope this patch makes life easier for you.


vtraag:
@ntamas just a question for clarification. This does not seem to be merged either with `master` or `develop`, is that correct? This fixes #671, but we have now deferred it until 0.9.0.

ntamas:
Yes, that's correct, it wasn't merged. We need to address the comments raised above in the review  and then it can be merged (or, rather, a new branch could be merged as the original repo is now gone).

vtraag:
OK, clear, thanks.

Issue #632, opened 2014-06-11
Longest path for directed acyclic graphs


pyrovski:
It would be great if igraph could compute the longest path for weighted directed acyclic graphs.  Computing this in R is a bit slow.

szhorvat:
This can be accomplished using currently existing functionality. Simply negate edge weights and use the Bellman-Ford algorithm to find a shortest path. This works in acyclic graphs, as no negative cycles will be present.

I'll leave this issue open anyway as more efficient approaches are possible.

lucasalavapena:
@szhorvat A modified Bellman-Ford (this could already be what you mean) can be O(|V| +|E|), [0, 1, 2] (also chapter 24.2 of CLRS) has some information on that matter. A topological sort step is performed first, so then you can just perform just one iteration of the relaxation step using the topological sort ordering of vertices, because then the edges of any paths are relaxed in the order that they would appear in the path. Lastly given that DAGs do not have cycles, the negative edges cycles check in Bellman-Ford can be removed. 

I think networkx also has a somewhat similar implementation from the other sources I linked, see [3]. I would be willing to contribute a PR.

[0] https://www.algotree.org/algorithms/single_source_shortest_path/bellman_ford_on_dag/
[1] https://en.wikipedia.org/wiki/Longest_path_problem#Acyclic_graphs
[2] https://courses.csail.mit.edu/6.006/fall10/lectures/lecture16.pdf
[3] https://networkx.org/documentation/stable/_modules/networkx/algorithms/dag.html#dag_longest_path

szhorvat:
@lucasalavapena Thanks for the references. Collecting these is actually quite useful for us.

It is now clear that it's worth having a specialized implementation for DAGs, for performance reasons, even if we have Bellman-Ford for general graphs.

If you're also up for contributing an implementation, let us know. I should point out that we are overwhelmed at the moment and that it will take longer than usual to review any contribution. However, this appears to be relatively simple, so hopefully it won't take too long.

lucasalavapena:
@szhorvat I have an initially look at it and setup igraph correctly on my machine. However, I am very busy for the next 2 weeks, would it be okay if I work on it in 2 weeks time? I think once I have time it should not take too long. If you want to just do it yourself then that is fair of course.

PR #818, opened 2015-03-22
add igraph_get_shortest_paths_bellman_ford;

pyrovski:
This code will compute shortest paths for graphs with negative edge weights.
Merging this will resolve [issue #632](https://github.com/igraph/igraph/issues/632), at least in the C library. I will also create a pull request for the R interface.


gaborcsardi:
Thanks, I'll look over it, probably in 1-2 weeks on so only, sorry about the delay.

ntamas:
I took a quick look at it and it seems to be okay; we will need test cases for `igraph_get_shortest_paths_bellman_ford`, though.

pyrovski:
I renamed the getPath function, and will work on adding tests.

dHannasch:
Hi,
I know I'm late to the party here, but I'm working on a igraph_dag_single_source_longest_paths that should take advantage of a topological ordering to find longest paths in O( s(|V| + |E|) ).
I have a question. igraph_topological_sorting gives out the result as a igraph_vector_t, a vector of floating-point numbers, rather than a igraph_vector_long_t.
Yet, in the code of igraph_topological_sorting, it only ever places longs in the result vector.
Is this just a legacy from a previous version where you used various nonintegers as error codes or somesuch, or is there a reason for this?
I am planning to have an (optional) igraph_vector_long_t* predecessors to allow tracing back the longest paths if desired; should this be a igraph_vector_t* instead?

Thank you,
David Hannasch

ntamas:
`igraph_topological_sorting` uses an `igraph_vector_t` because it was added way before we have implemented `igraph_vector_long_t` (or any other non-floating-point vector type). So yes, it is just a legacy feature, but non-integers were never used as error codes in igraph so you can be sure that the result vector will always contain integers only (despite the type of each element being a floating-point number). Feel free to use `igraph_vector_t` or `igraph_vector_long_t` in your code for the predecessor vector, whichever is more convenient to you.

gaborcsardi:
What's the status of this? Can I look at it, or is somebody working on it?

ntamas:
I haven't been looking at it for a while so feel free to go ahead.

gaborcsardi:
Yeah, what I meant, is that there are still no tests, so I guess it is not ready to merge. Anyway, I'll close it now, and please feel free to reopen/resubmit this when you think it is ready.

wlandau:
Any update on this feature? I was really hoping to find a shortest or longest path algorithm in `igraph`.

szhorvat:
igraph already has many shortest path related functions:

http://igraph.org/c/doc/igraph-Structural.html#idm470927281168

What do you need specifically that is not already available?

wlandau:
That's great to hear, thanks for the link. I will check it out.

szhorvat:
I think I misunderstood. It seems that the existing functions can calculate the path *length* even with negative weights (which you can use for longest paths), but they cannot calculate the actual path.

Issue #804, opened 2015-01-06
igraph_clique_number and igraph_largest_cliques should use the new maximal clique implementation


ntamas:
Right now they rely on the obsolete `igraph_i_maximal_cliques` function, which is slower than the new implementation. The net result is that it is usually slower to find the clique number of the graph than to find all the maximal cliques.

iosonofabio:
Hi guys, this one is references in #1192 which itself ends with a reference to #1220, which is tagged 0.9. So is this 0.8 or 0.9?

szhorvat:
I want to get in a _partial_ fix for 0.8 because I already use it in IGraph/M. I just need to clean up the naming of the functions (which is a bit of trouble because the names are assembled with macros). I'll deal with it.

iosonofabio:
That was 3+ months ago and there has been no change since. I propose we move this to 0.9. I'm sorry if that means @szhorvat has to revert some changes in the mathematica interface.

szhorvat:
@iosonofabio 

It cannot be reverted because the functionality is already released and people are using it. It needs to be in 0.8.

You have not seen any changes that happened because they are on my machine and not on this branch. When it is complete, it will be sent as a separate PR.

szhorvat:
This should have been moved to the machinery in `maximal_cliques.c` and `maximal_cliques_template.c`, not to Cliquer.  I'd like to reopen temporarily, to at least check what's the best solution ...

PR #824, opened 2015-05-04
modified igraph_clique_number to use the new maximal clique implementati...

aparent:
From issue #804.  I can also do the other half of the issue (igraph_largest_cliques), I just wanted feedback on this first.

### DIFF
--- a/src/maximal_cliques_template.h
+++ b/src/maximal_cliques_template.h
@@ -113,6 +113,19 @@
 #define FOR_LOOP_OVER_VERTICES_PREPARE
 #endif
 
+#ifdef IGRAPH_MC_SINGLE_LARGEST
+#define RESTYPE igraph_vector_t *res
+#define RESNAME res
+#define SUFFIX _single_largest
+#define RECORD do { \
+  if (igraph_vector_size(res) < clsize) \
+    igraph_vector_update(res,R); \
+  } while (0)
+#define FINALLY igraph_vector_init(res,0);
+#define FOR_LOOP_OVER_VERTICES for (i=0; i<no_of_nodes; i++) {
+#define FOR_LOOP_OVER_VERTICES_PREPARE
+#endif
+
 #ifdef IGRAPH_MC_ORIG
 void igraph_i_maximal_cliques_free(void *ptr) {
   igraph_vector_ptr_t *res=(igraph_vector_ptr_t*) ptr;
--- a/src/maximal_cliques_template.h
+++ b/src/maximal_cliques_template.h
@@ -125,12 +125,11 @@
 #define PREFIX i_
 #define RECORD do { \
   if (igraph_vector_size(res) < clsize){ \
-    int j = 0; \
-    igraph_vector_t *cl = igraph_Calloc(1, igraph_vector_t); \
-    igraph_vector_init(cl, clsize); \
-    for (j=0; j<clsize; j++) { VECTOR(*cl)[j] = VECTOR(*R)[j]; } \
-      igraph_vector_update(res,cl); \
-      igraph_vector_destroy(cl); \
+    igraph_vector_init(res, clsize); \
+    int j; \
+    for (j=0; j<clsize; j++) { \
+      VECTOR(*res)[j] = VECTOR(*R)[j]; \
+    } \
   } \
   } while (0)
 #define FINALLY igraph_vector_init(res,0);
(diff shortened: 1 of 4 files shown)


ntamas:
Looks great - I only have two comments:

1. The naming of `igraph_maximal_cliques_number` and `igraph_maximal_cliques_single_largest`. You mentioned that this is enforced by `maximal_cliques_template.h`. It is true, but similarly to the `#define SUFFIX` part in `maximal_cliques_template.h`, we could have a `#define PREFIX` part as well so the name of the final function would be composed from `igraph_`, `PREFIX`, `maximal_cliques` and `SUFFIX`. Then we could use `igraph_i_maximal_cliques_number` as the name of the function to indicate that this is an internal function.

2. Is it necessary to place `igraph_maximal_cliques_number` and `igraph_maximal_cliques_single_largest` in `include/igraph_cliques.h`? I would prefer to put only those functions in `include/igraph_cliques.h` that are part of the public API. You can create a new header in `src` if needed - the headers in `src` are considered private.

gaborcsardi:
@aparent Do you think you can update this?

aparent:
Sure.  Sorry I forgot about this.  Will try to get to it this week or weekend.

aparent:
I use the function  `igraph_largest_single_clique` in a separate project.  Could it be included in the public interface?

The function returns a single clique from the set of largest maximal cliques.  This allows to keep in memory only one representative clique of a given size.  This means it can find a maximally sized clique on larger graphs without using very large amounts of memory.

gaborcsardi:
I think that could be fine, but then you need to document it, and ideally it should not duplicate the code.

aparent:
I added some documentation.

What do you mean by not duplicating code?

gaborcsardi:
Can you also add test cases? Thanks.

aparent:
Sure.  Will do that sometime in the next day or so.

gaborcsardi:
Any update on this?

src/maximal_cliques_template.h
```diff
@@ -100,6 +104,39 @@
...
+#endif
+
+#ifdef IGRAPH_MC_SINGLE_LARGEST
+#define RESTYPE igraph_vector_t *res
+#define RESNAME res
+#define SUFFIX _single_largest
+#define PREFIX i_
+#define RECORD do { \
+  if (igraph_vector_size(res) < clsize){ \
+    igraph_vector_init(res, clsize); \
```
szhorvat: Isn't this a memory leak?  It would seem that a new clique vector is allocated every time a larger clique is found, but the old one is never freed.  It's possible that I am misunderstanding something.

src/maximal_cliques_template.h
```diff
@@ -100,6 +104,39 @@
...
+    int j; \
+    for (j=0; j<clsize; j++) { \
+      VECTOR(*res)[j] = VECTOR(*R)[j]; \
+    } \
+  } \
+  } while (0)
+#define FINALLY igraph_vector_init(res,0);
```
szhorvat: Shouldn't this be `igraph_vector_clear` instead?  igraph functions typically expect already initialized vectors, which are then resized as necessary.  This way *both* allocation and destruction are left to the user (and not igraph), which avoids surprises.

szhorvat:
Is there any reason why `igraph_largest_cliques` is not also included in this patch?

src/maximal_cliques_template.h
ntamas: Good catch - most likely it is; I think it should be `igraph_vector_resize(res, clsize)` instead. We simply have to assume that `res` is initialized by the user to an empty (or non-empty) vector already when `igraph_i_maximal_cliques_single_largest()` is called.

src/maximal_cliques_template.h
ntamas: True, it should be `igraph_vector_clear()` instead.

axrodriguez1:
I have a question I need to develop my own function of the edge of betweenness centrality and modularity within calling the python function any idea?

gaborcsardi:
@axrodriguez1 This has nothing to do with the issue you are commenting on. Please stop commenting on random issues. This is an issue tracker, not a help forum.

See http://igraph.org/r/#help if you need help with igraph.

vtraag:
This PR cannot be edited further because the original fork was removed. PR continues in #1192.

Issue #128, opened 2013-05-04
Speed up igraph_rewire


gaborcsardi:
igraph_rewire is slow because it is adding and removing edges using igraph_add_edges and igraph_remove_edges instead of working on an adjacency list. Check whether it can be speeded up and also examine whether it is possible to keep the edge attributes.

---

Imported from Launchpad using lp2gh.
- date created: 2012-12-21T11:18:50Z
- owner: ntamas
- the launchpad url was https://bugs.launchpad.net/bugs/1092868

szhorvat:
Done already, please close.

PR #826, opened 2015-05-10
Add support for a faster adjacency-list implementation of rewiring.

smowton:
This improves graph_rewire performance if you pass the USE_ADJLIST flag. The improvement factor is huge with test graphs that are quite sparse, so this should address #128.


smowton:
I only kept the old version on the suspicion that it's probably faster for a very small number of iterations against a big graph, and to ease comparative testing

smowton:
Would you prefer me to amend this branch to just replace the implementation btw?

ntamas:
I was thinking a bit about whether we should keep both implementations or not, and I realized that there is indeed one use-case where a small number of iterations might be performed on a large network - this is the case when the user is trying to study the rewiring process and thus iteratively rewires a graph, making a single step with each invocation of `igraph_rewire`. It might be beneficial to keep the old implementation to cater for this use-case, but in that case, I'd rather prefer `igraph_rewire` to choose between the two implementations transparently as the end user usually does not want to deal with implementation details. @gaborcsardi - please make a decision here as I do not want to decide on my own. I think that @smowton's contribution is a really useful addition and it should most likely be the default in the next version of igraph, but I'm unsure how the old implementation should be preserved (if it should be preserved at all).

gaborcsardi:
Thanks! I am sorry this took so long. I am reviewing it now and will have some comments.

gaborcsardi:
Can you please add a test case?

smowton:
How about this? I've made the check automatic, and altered the test to check the permutations of directed / undirected and standard / adjlist methods. I've also pushed the adjlist functions over to adjlist.c as requested.

smowton:
(Only tested so far using this test case, not in general usage)

gaborcsardi:
Btw. if you resubmit this, can you please squash your changes?  Thanks.

smowton:
Evidently the test packs up on your servers too! Will take a look...

smowton:
Test failure fixed, functions renamed and commits squashed as requested

gaborcsardi:
Great, thanks!

smowton:
:) Thanks for merging!

PR #850, opened 2015-09-01
igraph_isomorphic_bliss: warn about directed graphs

szhorvat:
The current behaviour if igraph_isomorphic_bliss is:

 - it treats all graphs as undirected
 - it aborts with an error if the directedness of the arguments differs, but it does accept two directed graphs
 - it does not warn when it received directed graphs

This patch removes the error and adds a warning for directed graphs.


gaborcsardi:
Thanks! The warning for directed graphs is fine, but please keep the error if one is directed and the other is undirected. All other isomorphism functions give errors in this case afair.

szhorvat:
Does it look fine now?  Sorry, I'm not experienced with GitHub pull requests, I might have messed something up.

gaborcsardi:
Looks great, thanks!

PR #867, opened 2015-09-15
Make test 159 compatible with Windows

### LINKED ISSUE #863: Test 159,  Maximum flow value (igraph_maxflow_value), doesn't compile on Windows
Test 159, `flow.c`, does not compile on Windows because it includes `<sys/time.h>` and `<sys/resource.h>`. It uses these for a timing function. The result of the timing function is not actually used (it's commented out).  Consider removing these to allow the test to run on Windows.

https://github.com/igraph/igraph/blob/master/examples/simple/flow.c#L58

(Otherwise the test passes with the MinGW-w64 gcc 5.2.0 I used.)

szhorvat:
Fixes #863

### DIFF
--- a/examples/simple/flow.c
+++ b/examples/simple/flow.c
@@ -23,17 +23,6 @@
 
 #include <igraph.h>
 
-#include <sys/time.h>
-#include <sys/resource.h>
-
-float timer ()
-{
-  struct rusage r;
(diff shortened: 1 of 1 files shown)


gaborcsardi:
Thanks!

PR #873, opened 2015-09-21
Update Bliss to version 0.73

szhorvat:
This patch updates bliss from 0.35 to 0.73

Why?  Because I need to be able to get the automorphism group.

What appears to be new since 0.35?

 - generators for the automorphism group
 - directed graphs
 - vertex colouring

### Description of the patch and questions

This patch only updates bliss to version 0.73, but it does not yet add the new functionality, except for support for directed graphs.

I tried to keep modifications to the bliss source minimal in order to ease any future updates.  In this spirit, the bliss files are now in their own `bliss` folder.  I also documented all modifications made to the bliss code in `igraph-changes.md`.  If you don't see a modification mentioned there, it didn't happen.

There is a `fatal_error` function in `defs.cc` which calls `exit(1)` and thus isn't safe for igraph.  I did not touch this.  I checked where it is used, and it seems it will only be called if an invalid splitting heuristic value is passed to Bliss.  I added a check for this on the igraph side to make sure it will never happen.

**What was in the previous bliss adaptation that is not done for 0.73:**

I did not include the status reporting that was previously in there (`igraph_statusf`).  There were two occurrences if I'm not mistaken, one of which used the Timer class.  I also removed the Timer class as it uses platform specific headers, and it seemed like a very minor feature.  (The Bliss verbosity level is set to 0 to make sure it won't print anything.)

There were two global variables in `partition.cc` marked as `IGRAPH_THREAD_LOCAL`. These variables are no longer there in 0.73.  I do not know much about thread-safety, so I didn't attempt to figure out what else that is new may need to be marked this way.  Note that there are several other global variables in different files (e.g. `graph.cc`), some of which were already there in 0.35 and *not originally marked* as `IGRAPH_THREAD_LOCAL`.

The original adaptation used the `igraph` namespace.  I didn't change this and left it as `bliss`.  Perhaps 0.35 didn't use namespaces at all?

So far I only tested this on OS X, with and without GMP.

Tamás, Gábor: Before making further changes, I need your input on this.


szhorvat:
Guys, I know reviewing the whole thing takes longer, but could you give a few preliminary comments based on my description above?  I would like to know if I can proceed with adding the interface for the new functionality (i.e. retrieve automorphism group generators).

ntamas:
I'm on the road now but I can probably take a look at it tonight (i.e. in a few hours).

szhorvat:
This patch is certainly going to need some modifications, but let me summarize my concerns more concisely:

 * Is my assumption correct that *every* not obivously read-only global variable should be marked as `IGRAPH_THREAD_LOCAL`?  In the last version there were some anonymous structs not marked as such; were these oversights?

 * How big a deal is status reporting?  Do we really need to add it?  (It might involve quite a few modifications I'm afraid).

ntamas:
Sorry, I still couldn't take a proper look at the patch (still on the road), but at least I'll try to respond to the questions:

* Yes, the assumption is correct. What you should strive for is that BLISS should not crash or give incorrect results if its functions are being used concurrently from two separate threads. Global variables obviously break this rule as they could be manipulated from two threads at the same time, *unless* you make them thread-local, in which case each thread gets its own copy from the variable. If there were some structs in the previous versions that were global and not thread-local, they were probably oversights, although it's really hard to test this. When in doubt, I think it's better to make a variable thread-local than to leave it global.

* Status reporting is nice but not super crucial. You can leave it to the end or to a separate pull request.

szhorvat:
Do you know if it's possible at all to get thread local storage working on OS X, either with clang or gcc?  I'm finding some conflicting information, and the igraph documentation currently says it doesn't work on OS X.

ntamas:
I finally managed to go through it and it seems good to me. Status reporting is not a big deal. However, it would be nice if you could identify places in the BLISS code where the algorithm spends a lot of time and add a few `IGRAPH_ALLOW_INTERRUPTION()` calls there so the algorithm becomes interruptible with Ctrl-C. It might require subtler changes, though, as `IGRAPH_ALLOW_INTERRUPTION()` assumes that the function it is being placed in returns an igraph error code.

If it is not possible or too difficult to implement, let me know and I think we can take the pull request as is - the previous BLISS implementation did not have interruption support either.

Either way, I'd like to wait and hear @gaborcsardi's opinion as well - this is a pretty big patch and I don't want to merge it without his approval.

szhorvat:
About interruption support:

I think the best way would be not through the macro `IGRAPH_ALLOW_INTERRUPTION()`, but using a custom method based on exceptions (given that Bliss is C++). The top-level function that catches the exception can then return the proper error code.

But a note first: I haven't really found any examples yet where Bliss would hang for an unreasonable time.  On the (otherwise naive) stress tests that I managed to devise, Bliss is much much faster than either VF2 or LAD.  Take e.g. a 200 by 200 lattice graph (40000 vertices), and randomly shuffle its vertex order, then test isomorphism with the original.  Bliss takes under a second to decide isomorphism.  VF2 takes 2.5 seconds *without* shuffling. With shuffling I waited for 10 minutes then shut it down.

Of course if the graph is large enough, it will always be slow, but isomorphism testing for very large graphs can be considered a specialty application where people will anticipate having to kill the system.

There's also a risk of breaking things when adding interruption support to someone else's code (e.g. `igraph_community_optimal_modularity` appears to crash sometimes when interrupted, will report when I can reproduce it consistently).

So maybe this functionality is not really needed for Bliss as much as for VF2.

Nevertheless I'll give it a go and will let you know how it went during the weekend.  I will make this a separate commit to make it easier to track the changes.  If you know of any difficult benchmark graphs on which Bliss is slow, let me know!

szhorvat:
Just for fun, here's some more benchmarking of Bliss vs Mathematica's builtin (which I suspect to be Nauty).  Bliss is doing very well and perhaps it should be the default algorithm for all large graphs once this is merged and it gets directed graph support.

I also tried to benchmark against MATLAB which got some graph support in its last release, but it crashes on graphs of this size.

```
In[2]:= g = GridGraph[{200, 200}];
shuffle[g_] := 
 Graph[RandomSample@VertexList[g], EdgeList[g]] (* randomly permute vertices *)

sg = shuffle[g];

In[5]:= IGBlissIsomorphicQ[g, 
  sg] // AbsoluteTiming (* includes time to convert from Mathematica to igraph format *)

Out[5]= {0.813759, True}

In[6]:= IsomorphicGraphQ[g, sg] // AbsoluteTiming (* Mathematica builtin *)
Out[6]= {0.368906, True}

In[7]:= AbsoluteTiming[
 ig = IGraphM`Private`igMake[g];
 sig = IGraphM`Private`igMake[sg];
 ] (* just the conversion to igraph format, I need to improve this *)

Out[7]= {0.564219, Null}

In[8]:= (* benchmarking only the igraph Bliss call, with 6 possible splitting heuristics *)
Table[
 ig@"blissIsomorphic"[ManagedLibraryExpressionID[sig], i] // AbsoluteTiming,
 {i, 0, 5}
 ]

Out[8]= {{0.215889, True}, {0.181879, True}, {0.159921, True}, 
         {0.197488, True}, {0.223436, True}, {0.157901, True}}
```

ntamas:
Thanks for the benchmarks and for the info re the interruption support. Let's wait and see what @gaborcsardi says and I think we are ready to merge if he approves.

gaborcsardi:
Sorry guys, travelling again today, but from tomorrow I'll have more time for this.

szhorvat:
Here's a second commit adding an interface for the new functionality that comes with the Bliss update.  This is just partial, please do not merge yet, but please do review.

 * The old docs stated that directed graphs are not supported. This is now not true anymore.
 * The old docs stated that multiple edges are ignored. They are definitely not ignored in this version, and in a few simple tests Bliss does give correct results for multigraphs.  We should ask the Bliss author about this. @gaborcsardi  Can you contact him please, since he will know you as one of the igraph authors?
 * About tests: I'm not sure how to add tests for this. In principle there's more than one minimal set of generators, and a new version of Bliss might return a different result (there's a warning in the Bliss docs that different versions give different canonical labelings).  In a few simple tests it gave me the same generators as Mathematica.
 * The generators are stored as a pointer vector of `igraph_vector_t`. Other igraph functions check for out-of-memory conditions when allocating such structures and abort as necessary. I didn't yet do this because: 1. Bliss's callback function doesn't support aborts, so Bliss itself may try to keep allocating memory anyway ... probably proper interruption support should be added first (with C++ exceptions or otherwise). 2. Unlike isomorphisms/automorphisms, of which there are a huge number, there are usually only a few group *generators*. Maybe running out of memory is not such a big concern here.

As far as I can tell, Bliss is the fastest available algorithm in igraph, now it supports directed graphs, and it appeared to work with multigraphs (not sure if it always does).  Should it be the default method for `isomorphic()`?

This patch is not done, I still need to add vertex colouring.

szhorvat:
In fact Bliss does not handle graphs with multiple edges correctly.

image

There's still the question whether self-loops are handled well.

szhorvat:
I received a response from Tommi Junttila:  Bliss supports self-loops, but it does not support multiple edges.  These could be handled using the vertex coloring feature, by introducing coloured vertices where multiple edges are present.

szhorvat:
@ntamas @gaborcsardi Guys, could you let me know how to proceed with exposing the coloured graph isomorphism functionality?  Do we change the existing functions (probably best), or add a coloured counterpart for each existing one?

This doesn't just replicate what the VF2 functions already do, as Bliss can also compute the automorphism group.

szhorvat:
I added vertex colouring support, and changed some function signatures in the process.

This is the final version of the patch.

ntamas:
Thanks for your hard work @szhorvat, it seems okay for me - still waiting for @gaborcsardi to review it.

szhorvat:
@ntamas  I'm I using `IGRAPH_FINALLY` correctly?  In short, I did something like this, in C++

```C++
void free_thing(Thing *thing) { delete thing; }

...
Thing t = new Thing;
IGRAPH_FINALLY(free_thing, t);

...
delete t; // manually delete here
IGRAPH_FINALLY_CLEAN(1); // this does not call free_thing, just signals the manual delete
```

I was a bit confused about the correct use of these macros ...

ntamas:
Yes, that's correct.

szhorvat:
Please don't merge yet. I found a problem while writing unit tests, but don't have time to investigate tonight.

szhorvat:
Question: for undirected graphs, is it guaranteed that `IGRAPH_FROM(graph, i) < IGRAPH_TO  (graph, i)`?  `igraph_add_edges` does ensure that, but perhaps other functions don't.

What went wrong was that I forgot to remove the edge canonicalization (sorting its vertices) from `igraph_isomorphic_bliss` for directed graphs.  Now I don't know if I should keep it for undirected graphs (definitely safe) or if it's safe to remove it altogether.

ntamas:
>
> Question: for undirected graphs, is it guaranteed that IGRAPH_FROM(graph,
> i) < IGRAPH_TO (graph, i)? igraph_add_edges does ensure that, but perhaps
> other functions don't.
>
I would say that it is true *right now*, but we make no guarantees that it
will stay this way in the future. So, the safest option is to keep the
canonicalization.

szhorvat:
The problem is fixed and I added a few tests. It can be merged.

szhorvat:
Please let me know if I can help with anything to make this patch easier to review!

gaborcsardi:
It looks good. 

One question. As I see the colors are only checked in the post-processing. This is not very efficient, actually, ideally they could be used during the search. But this is Bliss's business of course. Anyway, could you please add a note about this in the docs?

szhorvat:
@gaborcsardi  Sorry, it took me a while to understand what you mean.  I thought you were referring to something inside of the Bliss code.  You are referring to this, right?

https://github.com/szhorvat/igraph/commit/d4e9cc2bca26a4dccb3b10a2b45336e0d7a28c4d#diff-d474511a31889e0925e1c6f3c2c1ce02R2933

The colours are in fact used by Bliss itself too, but Bliss does not check isomorphism.  It just computes a canonical ordering of the vertices.  The in some igraph (non-Bliss) code we apply this ordering to the graphs, and check that: 1. the edge list is now identical 2. the vertex colours are identical too.

These checks on the igraph side would not be sufficient to correctly check coloured graph isomorphism unless the colours are taken into account for the canonical ordering as well (because often there will be multiple possible mappings between isomorphic graphs, but not all those mappings leave the colours invariant).

szhorvat:
But do check you are happy with the new file I put some tests in.  I can move them into already existing files instead, or rename the test file, if that is better.

gaborcsardi:
Hmmm. Maybe I misunderstood then.

> The colours are in fact used by Bliss itself too, but Bliss does not check isomorphism. It just computes a canonical ordering of the vertices. The in some igraph (non-Bliss) code we apply this ordering to the graphs, and check that: 1. the edge list is now identical 2. the vertex colours are identical too.

If Bliss uses colours, and computes the canonical ordering based on them, then 1. and 2. always hold, by definition, no?

szhorvat:
I mean, to decide whether two graphs, g1 and g2, are isomorphic, we do this:

A. Bring both graph to canonical representations
B. Check that their canonical representations are identical

To carry out A, we use Bliss.  Bliss takes a single graph as input, then it returns a canonical permutation of vertices.  To construct the canonical representation of a graph, 

 * we first apply this permutation to its vertices (relabel them), 
 * then for undirected graphs we make sure that each edge is sorted (i.e. 2-1 is transformed to 1-2 because 1<2)
 * then finally we sort the edge list (`igraph_vector_order`)

We do this for both graphs.  If the graphs are isomorphic, ignoring colours, then the two canonicalized edge lists will be identical, element for element.  If they are not isomorphic, they will be different.

But with colours, we may have these two graphs:

    g1: RED -- GREEN -- BLUE

and

    g2: RED -- RED -- RED

These are not isomorphic as coloured graphs.  Yet they may potentially have the same canonicalized edge list, using Bliss's canonical ordering, *even with taking the colours into account*.  That's because the colours in g2 are all the same so they do not have any effect on the canonical vertex ordering.  Thus we must do an extra check to ensure that all colours match, *in addition* to checking that the edge lists match.

Unfortunately at first I neglected to do this and learned the lesson the hard way (the program gave me wrong results) :smile:

gaborcsardi:
Yes, sorry, I understand now. Great!

gaborcsardi:
Thanks much, great work!

PR #874, opened 2015-09-21
Triad census for undirected graphs

### LINKED ISSUE #872: triad_census returns bad output for undirected graph
For an undirected graph, `triad_census` returns a meaningless result and issues a warning.  This is true both for C and R.  The result is meaningless partly because some of it comes from uninitialized memory and partly because the assignment to MAN labels is inconsistent.  This situation is not ideal.  I propose one of the following solutions:

 - Turn the warning into an error instead, do not return a bad result.
 - Return zeros for triads containing asymmetric dyads, assign correctly to 300, 201, 102 and 003.  This would be consistent with `dyad_census`, which does return a meaningful result.

I can send a pull request, just let me know which solution you prefer.

szhorvat:
Fixes #872


ntamas:
Can you add a few tests for `igraph_triad_census`, at least in the undirected case? (I know that we were being lazy and there are no tests for the directed case, but it is tested later on in the R and Python interfaces).

szhorvat:
I don't really understand this testing framework, but I tried to add something.  Can you check it?

szhorvat:
Also: I turned off igraph warnings in the test because I wasn't sure how to properly test for them.

ntamas:
Seems good - I'll have a more detailed look at it tomorrow and then I'll merge it. Thanks!

szhorvat:
Sorry about that, fixed now.

ntamas:
Awesome, thanks a lot!

PR #877, opened 2015-09-24
Add missing break; in igraph_layout_drl_options_init()

szhorvat:
Without this, using `IGRAPH_LAYOUT_DRL_FINAL` causes an error.


ntamas:
Thanks!

Issue #880, opened 2015-10-03
igraph_lcf_vector hangs with [2, -2]


szhorvat:
`igraph_lcf_vector` hangs with `n=2`, `shifts={2, -2}` and `repeats=1`.

ntamas:
It also crashes for `n=2`, `shifts = [2]` and `repeats=1`. Clearly, the problem is caused by `shifts = [2]` - this would create a loop edge and `igraph_lcf_vector` prevents this by only creating edges for which `from < to`.

I'm not that familiar with the LCF notation, though. Are loop edges allowed in graphs created with the LCF notation?

szhorvat:
I'm also unfamiliar with LCF notation, I only heard about it from the igraph docs. None of the example I see online have either self loops or multiple edges ...

PR #885, opened 2015-10-10
Fix igraph_lcf_vector

szhorvat:
For some inputs, `igraph_lcf_vector` was returning results that did not agree with the description at [MathWorld](http://mathworld.wolfram.com/LCFNotation.html), or with networkx's implementation.  One example is `shifts=[3,-2]`, `repeats=4`, `n=8`.  It also crashed for some inputs, see #880 

This patch fixes these problems and ensures that no multiple or loop edges are returned.

### DIFF
--- a/examples/simple/igraph_lcf.c
+++ b/examples/simple/igraph_lcf.c
@@ -28,6 +28,7 @@ int main() {
   igraph_t g, g2;
   igraph_bool_t iso;
   
+  // Franklin graph
   igraph_lcf(&g, 12, 5, -5, 6, 0);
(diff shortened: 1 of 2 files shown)


ntamas:
Thanks, great! Can you add a regression test for #880 (i.e. extend `examples/simple/igraph_lcf.c` with a test case that tests specifically the combination of parameters that you have reported in #880)? I'll be happy to merge this once we have a regression test.

szhorvat:
I added the checks.

ntamas:
Thanks!

PR #912, opened 2016-01-13
fix error messages

edzer:


ntamas:
The function names you use in the error messages are from the R interface. igraph has higher level interfaces in other languages (e.g., Python) where the function names are different. Furthermore, the C core of igraph is a library on its own right, so it makes the most sense to keep the C function names in the warnings.

On the other hand, these warning messages should not be visible to the user when they use the higher level interfaces (e.g., igraph from R or Python) because the interfaces should have already been updated not to use deprecated features from the C core. Therefore, if you see these warnings from R, please file a bug report to the [rigraph repo](https://github.com/igraph/rigraph) or send a pull request there that gets rid of these warnings.

edzer:
Thanks for the explanation!

PR #922, opened 2016-02-22
Additional argument check igraph_bipartite_game_gnm()

szhorvat:
igraph_bipartite_game_gnm() should report an error when n1 == 0 or n2 == 0 and m > 0.


ntamas:
Thanks!

PR #926, opened 2016-02-26
igraph_maximum_cardinality_search: handle empty graphs

### LINKED ISSUE #925: igraph_maximum_cardinality_search crashes when called with empty graph
`igraph_maximum_cardinality_search(&graph, &res, NULL)` crashes if `graph` is empty, possibly due to a call to `free` with an invalid pointer (based on console messages).

I see the crash on Linux only so far.  I haven't gotten to the bottom of it yet.

szhorvat:
fixes #925


gaborcsardi:
Great, thanks!

PR #941, opened 2016-04-18
More checks for igraph_static_fitness_game

### LINKED ISSUE #939: static_fitness_game is prone to infinite loops
`igraph_static_fitness_game` is prone to infinite loops.  This is dangerous because it can hang the user's interactive environment (e.g. R session).  The host process (e.g. R) will need to be killed so people can lost some of their work ...

Example (in R):

    static.fitness.game(4,c(1,1,1))

This of course hangs because we requested 4 edges in a 3-vertex graph.  But the problem is not always this obvious and mistakes can happen.

Another example:

    static.fitness.game(3, c(1,1,0))

`static_fitness_game` should try to avoid this situation by:

 - Being interruptible (currently it is not)
 - Try to determine if too many edges are requested and if so, abort with an error.

szhorvat:
This patch has three changes:

 1. It verifies that fitness_in and fitness_out are of the same size.

 2. It attempts to verify that the user didn't request more edges than what are possible to add.  Please verify this, as there are different cases depending directed/loop-free/multi-edge graph choices.

 3. Better interruptibility.  Before this function allowed interruption after each 10000 edges added.  Not every edge insertion trial is successful.  It is possible to get stuck trying to add more edges but never succeeding (or only succeeding with very small probabilities).  Instead of counting the edges that were added, now we count the number of trials (whether successful or not).

I did not check interactions with other functions that might call `igraph_static_fitness_game`

Fixes #939

### DIFF
--- a/src/games.c
+++ b/src/games.c
@@ -3530,6 +3530,7 @@ int igraph_static_fitness_game(igraph_t *graph, igraph_integer_t no_of_edges,
   igraph_real_t x, max_in, max_out;
   igraph_bool_t is_directed = (fitness_in != 0);
   float num_steps;
+  igraph_integer_t step_counter = 0;
   long int from, to, pos;
 
   if (fitness_out == 0) {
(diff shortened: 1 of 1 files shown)


ntamas:
Seems okay to me - I'll make a few stylistic adjustments after merging, though. MSVC does not like variable declarations in the middle of a function body so I'll move them to the top of the function. Thanks!

szhorvat:
Thanks!  I thought that variables at the beginning of blocks should always be fine in C.

```
void f() {
  printf("foo\n");
  int i=0; /* error */
  printf("%d\n", i);
}
```

```
void f() {
  printf("foo\n");
  {
    int i=0; /* valid */
    printf("%d\n", i);
  }
}
```

But I only tried with the latest MSVC (2015)

ntamas:
Not sure, I remember that we've had problems with that in the past and we don't do regular testing on MSVC (due to lack of access to Windows machines) so I prefer to be on the safe side.

PR #944, opened 2016-05-20
Clarify documentation on mean geodesic lengths

szhorvat:
This is based on user feedback (for IGraph/M).

For a directed graph, it is not true that the mean shortest paths within connected components will be taken.  Consider the two-vertex directed graph `1 -> 2`.  The two components are `{1}` and `{2}` and the `1 -> 2` edge is not part of either component.  Yet it is taken into account during the calculation of the average.

It also doesn't exactly compute the average geodesic length, i.e. average of the lengths of all geodesics. There may be more than one geodesic between the same two vertices.

Yes, this is splitting hairs, but useful when one is in doubt about what exactly gets calculated.


ntamas:
Thanks!

PR #1012, opened 2017-07-05
Clarify the doc of param start_from of igraph_barabasi_game.

xuhdev:
Clarify what it means by having start_from being a null pointer.

---

The same issue also exists in the Python API documentation.


ntamas:
Thanks (and sorry that it took so long)!

PR #1015, opened 2017-07-16
GCC not tolerating empty variadic argument without ## token before __VA_ARGS__

jannick0:
In preprocessor macros [GCC](https://gcc.gnu.org/onlinedocs/cpp/Variadic-Macros.html#Variadic-Macros) does not tolerate an __empty__ variadic argument `__VA_ARGS__` without a preceding `##` token, since in this case the preceding comma is **not** removed.

The proposed patch singles out GCC compliant compilers to add the required `##` token. In case this syntax is fine for any other compiler, then clearly the `#ifdef` branches can be removed and only the `##` token before `__VA_ARGS__` can be added. I checked that MSVC does not complain about the additional `##`.

### DIFF
--- a/src/foreign-dl-lexer.l
+++ b/src/foreign-dl-lexer.l
@@ -57,8 +57,13 @@
 #define YY_EXTRA_TYPE igraph_i_dl_parsedata_t*
 #define YY_USER_ACTION yylloc->first_line = yylineno;
 /* We assume that 'file' is 'stderr' here. */
-#define fprintf(file, msg, ...) \
-  igraph_warningf(msg, __FILE__, __LINE__, 0, __VA_ARGS__)
+#ifdef __GNUC__
+	#define fprintf(file, msg, ...) \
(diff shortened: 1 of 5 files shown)


gaborcsardi:
I don't understand. We have always used GCC to compile igraph. Why is this a problem now?

jannick0:
Well, the issue is raised if `flex` is asked to produce a scanner file .yy.c  __with__ debug mode enabled (`flex -d`) which then has some debug lines like `fprintf( stderr, "--(end of buffer or a NUL)\n" );`. 

Given the error macro as is now this is the very case with an empty variadic argument gcc complains about (because after preprocessing there is in the argument list a lost comma directly followed by the closing bracket. More details in the link I added to the first posting).

Thus the macro definition is not purely GCC compliant and it looks like the scanners were not run in debug mode for some time though. Separately, today I checked that `clang` and `MSVC` are smart enough to remove the single comma if the variadic argument is empty.

PR #1034, opened 2017-09-29
Implemented k-truss

alexperrone:
The k-truss algorithm was implemented. This is a C++ implementation since it needs a hashtable (add, delete, find an element, where delete is not supported by igraph_hashtable.c) as well as an ordered vector of these hashtables. 

Guidance is needed on how to compile the C++ code to yield an object file as part of the igraph build system. This should be analagous to how the walktrap code (e.g. walktrap_communities.cpp) is compiled.


ntamas:
Thanks for the pull request and sorry for the delay. I will add a few comments soon.

src/truss.cpp
```diff
@@ -0,0 +1,242 @@
...
+ * indicates the highest k-truss that the edge occurs in.
+ *
+ * \param graph The input graph.
+ * \param truss Pointer to initialized vector of truss values that will
+ * indicate the highest k-truss each edge occurs in. It will be resized as
+ * needed.
+ *
+ * \example examples/simple/igraph_truss.cpp
+ */
+int igraph_truss(const igraph_t* graph, igraph_vector_int_t* truss){
```
ntamas: The return value of igraph functions should be of type `igraph_error_t`, not `int`.

src/truss.cpp
```diff
@@ -0,0 +1,242 @@
...
+  igraph_vector_int_t triangles, support;
+  igraph_vector_t eid, unpacked_triangles;
+
+  // List the triangles as vertex triplets.
+  igraph_vector_int_init(&triangles, 0);  // will be resized
+  igraph_list_triangles(graph, &triangles);
```
ntamas: `igraph_list_triangles` (and any other function call that returns an igraph error code) should be wrapped in `IGRAPH_CHECK()` to ensure that any errors that happen during the function call (e.g., out-of-memory conditions) are handled properly. This applies to invocations of `igraph_vector_int_init()` and similar as well - in general, only `igraph_vector_*_destroy()` and functions that return `void` or are guaranteed not to bail out with an error should not be wrapped in `IGRAPH_CHECK`.

src/truss.cpp
```diff
@@ -0,0 +1,242 @@
...
+  IGRAPH_FINALLY(igraph_vector_int_destroy, &triangles);
```
ntamas: `IGRAPH_FINALLY` must be called right after `igraph_vector_int_init` to ensure that igraph's memory management routine is already aware about the allocated memory for `triangles` even before `igraph_list_triangles()` is called (otherwise this memory would not be freed when `igraph_list_triangles()` returns with an error).

src/truss.cpp
```diff
@@ -0,0 +1,242 @@
...
+  return 0;
+}
+
+// Unpack the triangles as a vector of vertices to be a vector of edges.
+// So, instead of the triangle specified as vertices [1, 2, 3], return the
+// edges as [1, 2, 1, 3, 2, 3] so that the support can be computed.
+void unpack(const igraph_vector_int_t *tri, igraph_vector_t *unpacked_tri) {
+  int j;
+  for (long int i = 0; i < igraph_vector_int_size(tri); i = i + 3){
+    j = i * 2;
```
ntamas: It would be more efficient to add `j = 0` in the initializer of the `for` loop and use `i += 3, j += 2` instead of performing a multiplication in every iteration. (The compiler might be able to optimize it away, but I'm not sure).

src/truss.cpp
```diff
@@ -0,0 +1,242 @@
...
+  IGRAPH_FINALLY_CLEAN(1);
+
+  return 0;
+}
+
+// Unpack the triangles as a vector of vertices to be a vector of edges.
+// So, instead of the triangle specified as vertices [1, 2, 3], return the
+// edges as [1, 2, 1, 3, 2, 3] so that the support can be computed.
+void unpack(const igraph_vector_int_t *tri, igraph_vector_t *unpacked_tri) {
+  int j;
```
ntamas: `j` should be a `long int`, not an `int` since `i` is a `long int` and `j` will be equal to `i * 2`.

src/truss.cpp
```diff
@@ -0,0 +1,242 @@
...
+  // List the triangles as vertex triplets.
+  igraph_vector_int_init(&triangles, 0);  // will be resized
+  igraph_list_triangles(graph, &triangles);
+  IGRAPH_FINALLY(igraph_vector_int_destroy, &triangles);
+
+  // Unpack the triangles from vertex list to edge list.
+  long int num_triangles = igraph_vector_int_size(&triangles);
+  igraph_vector_init(&unpacked_triangles, 2 * num_triangles);
+  unpack(&triangles, &unpacked_triangles);
+  IGRAPH_FINALLY(igraph_vector_int_destroy, &unpacked_triangles);
```
ntamas: Again, `IGRAPH_FINALLY` should be moved to right after the allocation. This applies to all the other occurrences of `IGRAPH_FINALLY` in your code so I won't mention it again.

src/truss.cpp
```diff
@@ -0,0 +1,242 @@
...
+  igraph_vector_int_destroy(&triangles);
+  IGRAPH_FINALLY_CLEAN(1);
```
ntamas: Here we have a problem. This line discards the last entry from igraph's "finally stack" which keeps track of the memory chunks that were allocated during the execution of a function. However, the last entry in the finally stack at this point is not `triangles` (which we have just deallocated) but `unpacked_triangles`. If any error would happen after this point and we added `IGRAPH_CHECK()` calls everywhere, igraph would attempt to free `triangles` again (because it still stayed in the finally stack) and leak `unpacked_triangles` (which is not in the finally stack any more).

The solution is typically to allocate all the vectors that you would need right at the beginning of the function, with zero length, then put them in the finally stack in the order they will be deallocated during the execution. Then, whenever you figure out what the real size of a vector should be, you can call `IGRAPH_CHECK(igraph_vector_resize(...))` or `IGRAPH_CHECK(igraph_vector_int_resize(...))` to grow the vector to the right size.

src/truss.cpp
```diff
@@ -0,0 +1,242 @@
...
+    VECTOR(*unpacked_tri)[j+1] = VECTOR(*unpacked_tri)[j+4] = VECTOR(*tri)[i+1];
+    VECTOR(*unpacked_tri)[j+3] = VECTOR(*unpacked_tri)[j+5] = VECTOR(*tri)[i+2];
+  }
+}
+
+// Compute the edge support, i.e. number of triangles each edge occurs in.
+// Time complexity: O(m), where m is the number of edges listed in eid.
+void compute_support(const igraph_vector_t *eid, igraph_vector_int_t *support) {
+  for (long int i = 0; i < igraph_vector_size(eid); ++i){
+    VECTOR(*support)[(int) VECTOR(*eid)[i]] += 1;
```
ntamas: Casting to `long int` here would probably be safer - `int` might be only 16 bits on some (very old) systems.

src/truss.cpp
```diff
@@ -0,0 +1,242 @@
...
+    VECTOR(*unpacked_tri)[j]   = VECTOR(*unpacked_tri)[j+2] = VECTOR(*tri)[i];
+    VECTOR(*unpacked_tri)[j+1] = VECTOR(*unpacked_tri)[j+4] = VECTOR(*tri)[i+1];
+    VECTOR(*unpacked_tri)[j+3] = VECTOR(*unpacked_tri)[j+5] = VECTOR(*tri)[i+2];
+  }
+}
+
+// Compute the edge support, i.e. number of triangles each edge occurs in.
+// Time complexity: O(m), where m is the number of edges listed in eid.
+void compute_support(const igraph_vector_t *eid, igraph_vector_int_t *support) {
+  for (long int i = 0; i < igraph_vector_size(eid); ++i){
```
ntamas: I typically put `igraph_vector_size(eid)` in a separate variable before the iteration so it won't be evaluated once during every iteration (after all, the size of `eid` won't change). Some compilers are smart enough to figure it out on their own, but it does not hurt to use a separate variable.

src/truss.cpp
```diff
@@ -0,0 +1,242 @@
...
+ *
+ *
+ * Time complexity: It should be O(|E|^1.5). See Algorithm 2 in:
+ * Wang, Jia, and James Cheng. "Truss decomposition in massive networks."
+ * Proceedings of the VLDB Endowment 5.9 (2012): 812-823.
+ *
+ *
+ * \example examples/simple/igraph_truss.c
+ */
+void trussness(const igraph_t *graph, igraph_vector_int_t *support,
```
ntamas: This function needs to return an `igraph_error_t` and its invocation should be wrapped in `IGRAPH_CHECK()`. This is because the function itself may run into out-of-memory errors due to internal memory allocations, and we need to wrap those internal calls in `IGRAPH_CHECK()` as well.

src/truss.cpp
```diff
@@ -0,0 +1,242 @@
...
+  igraph_vector_destroy(&eid);
+  IGRAPH_FINALLY_CLEAN(1);
+
+  // Compute the truss of the edges.
+  igraph_vector_int_init(truss, igraph_ecount(graph));
+  trussness(graph, &support, truss);
+  igraph_vector_int_destroy(&support);
+  IGRAPH_FINALLY_CLEAN(1);
+
+  return 0;
```
ntamas: For sake of clarity, it's better to return `IGRAPH_SUCCESS` (even if we know that its value is currently 0 and it is not likely to change in the future).

src/truss.cpp
```diff
@@ -0,0 +1,242 @@
...
+ */
+void trussness(const igraph_t *graph, igraph_vector_int_t *support,
+  igraph_vector_int_t *truss){
+
+  // Get max possible value = max entry in support.
+  int max = 0;
+  for (long int i = 0; i < igraph_vector_int_size(support); ++i) {
+    int value = VECTOR(*support)[i];
+    if (max < value) { max = value; }
+  }
```
ntamas: I think we have `igraph_vector_int_max()` to get the maximum value from a vector - just use that.

src/truss.cpp
```diff
@@ -0,0 +1,242 @@
...
+
+  // The vector of levels. Each level of the vector is a set of edges initially
+  // at that level of support, where support is # of triangles the edge is in.
+  vector< unordered_set<long int> > vec(max + 1);
+
+  // Initialize completed edges.
+  igraph_vector_bool_t completed;
+  igraph_vector_bool_init(&completed, igraph_ecount(graph));
```
ntamas: We need `IGRAPH_CHECK()` and `IGRAPH_FINALLY()` here as well. Also, we need `IGRAPH_FINALLY_CLEAN()` at the end of the function after the deallocations.

src/truss.cpp
```diff
@@ -0,0 +1,242 @@
...
+
+      if (igraph_vector_size(&q1) > igraph_vector_size(&q2)){
+        // case: #fromNeighbors > #toNeigbors, so make q1 the smaller set.
+        q1 = toNeighbors;
+        q2 = fromNeighbors;
+      }
+
+      // Sort the neighbors.
+      igraph_vector_sort(&q1);
+      igraph_vector_sort(&q2);
```
ntamas: Lots of extra legwork is being done here because you are querying and sorting the neighbors of a single vertex many times during the iterations. Consider using an `igraph_adjlist_t` data structure in this function, followed by `igraph_adjlist_sort()` to speed things up - basically the adjacency list will contain the pre-sorted neighbor vectors so you can simply query them later.

examples/simple/igraph_truss.cpp
```diff
@@ -0,0 +1,81 @@
...
+
+  igraph_real_t edges[] = { 0,1, 0,2, 0,3, 0,4,
+    1,2, 1,3, 1,4, 2,3, 2,4, 3,4, 3,6, 3,11,
+    4,5, 4,6, 5,6, 5,7, 5,8, 5,9, 6,7, 6,10, 6,11,
+    7,8, 7,9, 8,9, 8,10 };
+  igraph_vector_view(&v, edges, sizeof(edges)/sizeof(double));
+  igraph_create(&graph, &v, 0, IGRAPH_UNDIRECTED);
+
+  FILE *output;
+  output = fopen("igraph_truss.out", "w");
```
ntamas: You should not need to create files from the test case - instead of that, print the result to the standard output and _provide_ a file named `igraph_truss.out` that the output will be compared with. The test case will pass if the printed output is identical to the provided `igraph_truss.out` file.

examples/simple/igraph_truss.cpp
```diff
@@ -0,0 +1,81 @@
...
+  igraph_vector_int_init(&truss, igraph_ecount(&graph));
+
+  // Time truss.
+  struct timespec begin, end;
+  clock_gettime(CLOCK_MONOTONIC_RAW, &begin);
+  igraph_truss(&graph, &truss);
+  clock_gettime(CLOCK_MONOTONIC_RAW, &end);
+  uint64_t delta = (end.tv_sec - begin.tv_sec) * 1000000 + (end.tv_nsec - begin.tv_nsec) / 1000;
+  double elapsed = (double) delta / 1000000;
+  printf("Truss (seconds): %f\n", elapsed);
```
ntamas: Don't add the timing information to the output because it will almost surely make the test case fail as the output won't be identical to the expected output provided in `igraph_truss.out`.

szhorvat:
@alexperrone  Are you still working on this?

alexperrone:
@szhorvat Yes, I'm still working on this. Not at this moment but I haven't forgotten and it's definitely on my to-do list and it will get off there -- it may be up to a couple months.

szhorvat:
@alexperrone Any updates on this?

alexperrone:
@szhorvat Please keep it open.

szhorvat:
Any news on this @alexperrone ? It'd be great to have it merged ...

alexperrone:
@szhorvat Still working on it... I'm struggling to get it built successfully with igraph as a whole even though the algorithm itself works. Thanks for the update, I'm doing my best on it.

szhorvat:
> I'm struggling to get it built successfully with igraph as a whole

What kinds of problems are you running into?  Did you manage to build igraph itself, _without_ your additions?  What OS are you using, and what compiler?  Maybe I can help.

src/truss.cpp
alexperrone: @ntamas In everything I've checked, it seems `int` is used, such as throughout `src/vector.c`, `src/centrality.c`, and `src/iterators.c` (to name a few). What are some example files for code standards I should follow?

src/truss.cpp
alexperrone: done

src/truss.cpp
alexperrone: I agree addition is faster. However, I don't see how `j += 2` accomplishes what is needed here because the first iteration is `i = 0, j = 0`, the second iteration needs to be `i = 3, j = 6`, the third iteration needs to be `i = 6, j = 12`, etc.

alexperrone:
@szhorvat Yes, I just built igraph successfully as far as I can tell (`make check` looked good and it seemed to install without errors) using a clone of this forked repository which has the truss code. However, I can't seem to compile `truss.cpp` on its own anymore. Originally, it worked when I developed on Linux Ubuntu. Now I'm on Mac OS 10.12 machine. 

To attempt to compile this code, I first ran:

```
> pkg-config --libs --cflags igraph
-I/usr/local/include/igraph -L/usr/local/lib -ligraph
```

so I issued the following compile command, yet I could not run it: 

```
> g++ -c -std=c++11 truss.cpp -I/usr/local/include/igraph -I . -L/usr/local/lib -ligraph -o truss
clang: warning: -ligraph: 'linker' input unused [-Wunused-command-line-argument]
clang: warning: argument unused during compilation: '-L/usr/local/lib' [-Wunused-command-line-argument]
> ./truss
zsh: exec format error: ./truss
> file ./truss
./truss: Mach-O 64-bit object x86_64
```

src/truss.cpp
ntamas: Okay, sorry, my bad then.

src/truss.cpp
ntamas: Ah, sorry, I meant `j += 6` instead.

ntamas:
It is not finding the new variant of  `libigraph` that includes `igraph_truss()` -- it finds the "old" one installed somewhere else on your system. You need `DYLD_LIBRARY_PATH=some-folder-that-contains-the-new-libigraph0.dylib ./truss` - the folder given in the `DYLD_LIBRARY_PATH` envvar will take precedence over the standard library paths.

szhorvat:
@ntamas  I am not familiar with `pkg-config`, so I'm not sure what that line does ... but would it not be still necessary to add `-ligraph` to the compilation command?

@alexperrone Side comment: It may not be the best of ideas to install the dev version you're working on in a global location (`/usr/local`) where all other tools will find it.

Personally, I install it in a private location, then when compiling a program that uses it, I simply do

    cc myprog.c -I/path/to/include/igraph -L/path/to/lib -ligraph

If your code requires C++, use `c++` instead of `cc`.

Another note: Since igraph is a C library, it would be good to have the example file (`igraph_truss.cpp`) use only C and no C++. After a quick glance, it seems that this is already the case. Thus, compile it with a C compiler.

alexperrone:
@szhorvat How do you suggest I install the dev version? With `configure --prefix`, or `make install --prefix=` or `DESTDIR`? There are several options (https://stackoverflow.com/questions/11307465/destdir-and-prefix-of-make). I will work on a local version. I want to make sure everything compiles and runs as-is on the present commit before addressing the valuable comments on code.

szhorvat:
Sorry, I forgot to include that bit. Personally, I have been using `./configure --prefix=...`. I am not an expert on this so there might be better ways than my workflow.

alexperrone:
@szhorvat 
1. The `pkg-config` came from here: https://igraph.org/c/doc/igraph-tutorial.html
2. I've installed a clone of this forked repository (`git clone git@github.com:alexperrone/igraph.git`) locally in its own directory called `install` with `./configure --disable-graphml --prefix=/path/to/forked/igraph/install` (not sure if it's a bad idea to install it inside the repository itself, just wanted to keep everything together in one folder). 
3. After installing, the `install` folder has `include` and `lib` directories, and I can see `install/lib/libigraph.dylib`. I do note that the `include` does not have `truss.h` (not sure if it needs to). 
4. I still cannot compile the example code with `g++ -std=c++11 igraph_truss.cpp -I . -I .. -I ../../src/ -I ../../install/include/igraph -L ../../install/lib -ligraph -o truss`. @szhorvat I can set the `DYLD_LIBRARY_PATH` upon running the compiled example, but I can't compile it yet. It fails with same error as above: 
```
Undefined symbols for architecture x86_64:
  "_igraph_truss", referenced from:
      _main in igraph_truss-6b1f23.o
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```
Thus, I believe `igraph_truss` is not making it into the `libigraph.dylib`. How can I ensure that it gets compiled when I `make` igraph? Or is there some way I can develop outside of the `make` process?

ntamas:
Did you add `src/truss.cpp` to `src/Makefile.am` to the `SOURCES` variable?

szhorvat:
@alexperrone I suggest joining https://gitter.im/igraph/igraph to sort these details out. Feel free to ping me there with questions.

src/truss.cpp
```diff
@@ -0,0 +1,242 @@
...
+
+#include "truss.h"
+#include <iostream>
+#include <vector>
+#include <unordered_set>
+
+using namespace std;
+
+/**
+ * \function trussness
```
alexperrone: From @szhorvat : " In the doc comment, the \function name must match the name of your function.

alexperrone:
Thanks, I'll re-open when I continue this work.

iosonofabio:
guys what do you want to do with this? It's 14 months of no update and we are moving to int64, but it seems almost done except for some fairly minor things. Shall I try and fix the PR according to the review suggestions and see if we can have it?

ntamas:
@iosonofabio I'd be grateful if you can pick this up and make it up-to-date with the current `master` first. Then we can review it once again to see if there is anything to be done before we merge it. Thanks!

szhorvat:
`Copyright 2017 The Johns Hopkins University Applied Physics Laboratory LLC. All Rights Reserved.`

I am wondering if it would be wise to request a transfer of copyright first ...

iosonofabio:
@alexperrone I'd like to finish this one up so igraph has it. Would you mind confirming that you authorize igraph folks (including me) to make changes to your PR, integrate it into the codebase, and thereby take over the copyright?

The new note will say you are the original author and the code was copyrighted by Johns Hopkins in 2017, but that it was edited by us and copyright is owned by us in 2021.

It should be a formality since it's GPL anyway.

If that's granted I can start pushing commits onto this PR I think.

Thank you!

iosonofabio:
Actually, I'll wait a day or two but it looks like Alex has done 0 contributions over the last year, so perhaps he's not going to see this at all. Since the code is GPL I think we can just go ahead and edit it while making it clear he was the one writing the initial version.

I've made a branch in case you are interested:

https://github.com/iosonofabio/igraph/tree/ktruss

it addresses @ntamas 's review points except for an output file in the test, which I'm going to try and do now.

alexperrone:
@iosonofabio Yes, the edits needed seemed relatively minor. I would kindly request that the original copyright under Johns Hopkins to remain the same since the the core algorithm was published under that copyright, and they allowed me to publish the code under GPL under the condition that the particular copyright is included. Thank you so much for your effort to close this, please let me know any questions at all, including output file, testing, or the algorithm.

iosonofabio:
Thank you Alex, of course, the original copyright line will stay, you coded the thing up.

I've just pushed directly to this PR.

@ntamas: the compiler cannot find `igraph_truss` but I'm not sure why. Could you please have a look? Context:

- I removed `truss.h` and put function declarations (renamed as `igraph_truss_i_...`) at the top of `truss.cpp`
- There is a single function declaration for `igraph_truss` in `include/igraph_community.h` (slightly edited to use `IGRAPH_EXPORT`)
- I moved `truss.cpp` into `src/centrality`, where the k-cores are too I believe
- I moved the test to `tests/unit/truss.c` and am including only `igraph.h`, which should include `igraph_community.h`, which *should* include `igraph_truss`, but perhaps is does not?

Thanks! (I'm not so familiar with CPP/C cross-compiling and with compiler units or whatever they are called)

tests/unit/truss.c
```diff
@@ -0,0 +1,77 @@
...
+  
+    // Compute the truss of the edges.
+    igraph_vector_int_init(&truss, igraph_ecount(&graph));
+    igraph_truss(&graph, &truss);
+    print_and_destroy(&graph, &truss);
+  
+    VERIFY_FINALLY_STACK();
+  
+    return 0;
+}
```
iosonofabio, line 77: haven't added an `.out` file yet, but will do once it compiles

szhorvat:
> @ntamas: the compiler cannot find `igraph_truss` but I'm not sure why. Could you please have a look?

You need to include the header containing the prototype of `igraph_truss` into `truss.cpp` (the file where `igraph_truss` is defined).

Since it is not included, `igraph_truss` appears as a C++ function (not a C function) in the object file. The caller expects a C function (`extern "C"`).

iosonofabio:
Thanks @szhorvat I'm a little unsure how you would do it.

The prototype of `igraph_truss` is currently only in `include/igraph_community.h`, are you suggesting to include that whole header in `truss.cpp`? Or to go back to a separate header file for the truss function located in `src/centrality`?

szhorvat:
Just add `#include "igraph_community.h"`.

szhorvat:
You also need to add `truss.cpp` to `src/CMakeLists.txt` and add the appropriate `#include`s to `truss.cpp` for every igraph function that is used in that file.

ntamas:
Converted this PR into a draft until it is ready for review. @iosonofabio please mark it as ready when you are done.

iosonofabio:
This one should be ready, just waiting for CI but I recomputed the `.out` file by hand and it should be alright, the test passes locally on my machine.

src/centrality/truss.cpp
```diff
@@ -0,0 +1,271 @@
...
+  Foundation, Inc.,  51 Franklin Street, Fifth Floor, Boston, MA
+  02110-1301 USA
+
+*/
+
+#include "igraph_error.h"
+#include "igraph_adjlist.h"
+#include "igraph_interface.h"
+#include "igraph_motifs.h"
+#include "igraph_community.h"
```
szhorvat, line 33: Let's order these alphabetically, and put the header containing functions defined in this file as the first one (this is the convention we use recently).

src/centrality/truss.cpp
```diff
@@ -0,0 +1,271 @@
...
+#include <vector>
+#include <unordered_set>
+
+using namespace std;
+
+/* helper functions */
+void igraph_truss_i_unpack(const igraph_vector_int_t *tri, igraph_vector_t *unpacked_tri);
+void igraph_truss_i_compute_support(const igraph_vector_t *eid, igraph_vector_int_t *support);
+int igraph_i_trussness(const igraph_t *graph, igraph_vector_int_t *support,
+  igraph_vector_int_t *truss);
```
szhorvat, line 45: If these are not used in other files, they should be marked `static`.

src/centrality/truss.cpp
```diff
@@ -0,0 +1,271 @@
...
+ * \param truss This will indicate the highest k-truss each edge occurs in.
+ * \return Error code.
+ *
+ *
+ * Time complexity: It should be O(|E|^1.5). See Algorithm 2 in:
+ * Wang, Jia, and James Cheng. "Truss decomposition in massive networks."
+ * Proceedings of the VLDB Endowment 5.9 (2012): 812-823.
+ *
+ *
+ * \example test/unit/truss.cpp FIXME: can we make a C file instead?
```
szhorvat, line 155: Yes, tests should be pure C. Do not add tests to the docs (only examples).

src/centrality/truss.cpp
```diff
@@ -0,0 +1,271 @@
+/*
+
+  Copyright 2017 The Johns Hopkins University Applied Physics Laboratory LLC. All Rights Reserved.
```
szhorvat, line 3: All rights aren't really reserved ;-) Quite a few rights are granted perpetually through the GPL.

src/centrality/truss.cpp
```diff
@@ -0,0 +1,271 @@
...
+using namespace std;
+
+/* helper functions */
+void igraph_truss_i_unpack(const igraph_vector_int_t *tri, igraph_vector_t *unpacked_tri);
+void igraph_truss_i_compute_support(const igraph_vector_t *eid, igraph_vector_int_t *support);
+int igraph_i_trussness(const igraph_t *graph, igraph_vector_int_t *support,
+  igraph_vector_int_t *truss);
+
+/**
+ * \function trussness
```
szhorvat, line 48: The name here must match the function.

Wouldn't `igraph_trussness` be a better name for this function? Opinions?

tests/unit/truss.c
```diff
@@ -0,0 +1,73 @@
...
+
+#include "test_utilities.inc"
+
+void print_and_destroy(igraph_t *graph, igraph_vector_int_t *truss) {
+    long int i;
+    igraph_integer_t from, to;
+    printf("fromNode,toNode,truss\n");
+    for (i=0; i < igraph_vector_int_size(truss); i++) {
+        igraph_edge(graph, i, &from, &to);
+        printf("%d,%d,%li\n", from, to, (long int) VECTOR(*truss)[i]);
```
szhorvat, line 40: Instead of `"%d"`, use `IGRAPH_PRId`

src/centrality/truss.cpp
```diff
@@ -0,0 +1,271 @@
...
+ *
+ *
+ * Time complexity: It should be O(|E|^1.5). See Algorithm 2 in:
+ * Wang, Jia, and James Cheng. "Truss decomposition in massive networks."
+ * Proceedings of the VLDB Endowment 5.9 (2012): 812-823.
+ *
+ *
+ * \example test/unit/truss.cpp FIXME: can we make a C file instead?
+ */
+int igraph_i_trussness(const igraph_t *graph, igraph_vector_int_t *support,
```
szhorvat, line 157: This function is documented, but it is now internal. @alexperrone is this meant as a public function?

src/centrality/truss.cpp
```diff
@@ -0,0 +1,271 @@
...
+ * \function trussness
+ * \brief Find the "trussness" for every edge in the graph. This
+ * indicates the highest k-truss that the edge occurs in.
+ *
+ * \param graph The input graph.
+ * \param truss Pointer to initialized vector of truss values that will
+ * indicate the highest k-truss each edge occurs in. It will be resized as
+ * needed.
+ *
+ * \example examples/simple/igraph_truss.cpp
```
szhorvat, line 57: Missing example. Example should be implemented, or this line removed.

szhorvat:
Thanks for the cleanup @iosonofabio ! There seems to be a memory leak, and it's not immediately obvious to me where it comes from.

This is now added to the directory `centrality`, but the header used is `igraph_community.h`. We should decide where it fits best, and be consistent in the header and source location used.

@alexperrone :

 - There are two documented functions, but only one is public. Which one did you mean to make public?
 - Can you write a brief explanation of the concept in the documentation?

src/centrality/truss.cpp
```diff
@@ -0,0 +1,271 @@
...
+  igraph_vector_destroy(&unpacked_triangles);
+  IGRAPH_FINALLY_CLEAN(1);
+
+  // Compute the support of the edges.
+  igraph_truss_i_compute_support(&eid, &support);
+  igraph_vector_destroy(&eid);
+  IGRAPH_FINALLY_CLEAN(1);
+
+  // Compute the truss of the edges.
+  IGRAPH_CHECK(igraph_vector_int_init(truss, igraph_ecount(graph)));
```
szhorvat, line 92: Do not allocate any vector passed to the function. The function should expect to get a pre-allocated vector. Use `vector_int_resize` to make it the correct size.

This is the cause of the test failures (memory leak error).

src/centrality/truss.cpp
```diff
@@ -0,0 +1,271 @@
...
+
+  return IGRAPH_SUCCESS;
+}
+
+// Unpack the triangles as a vector of vertices to be a vector of edges.
+// So, instead of the triangle specified as vertices [1, 2, 3], return the
+// edges as [1, 2, 1, 3, 2, 3] so that the support can be computed.
+void igraph_truss_i_unpack(const igraph_vector_int_t *tri, igraph_vector_t *unpacked_tri) {
+  long int i, j;
+  for (i = 0, j = 0; i < igraph_vector_int_size(tri); i += 3, j += 6){
```
szhorvat, line 105:
```suggestion
  long int i, j, n = igraph_vector_int_size(tri);
  for (i = 0, j = 0; i < n; i += 3, j += 6) {
```

Because there is a cost to non-inlineable function calls in a tight-loop.

src/centrality/truss.cpp
```diff
@@ -0,0 +1,271 @@
...
+      vec[level].erase(seed);  // remove the first element
+
+      /* Find the vertices of this edge */
+      igraph_edge(graph, seed, &fromVertex, &toVertex);
+
+      /* Find neighbors of both vertices. If they run into each other,
+       * there is a triangle. Because we sorted the adjacency list already,
+       * we don't need to sort it for every edge here */
+      fromNeighbors = adjlist.adjs[fromVertex];
+      toNeighbors = adjlist.adjs[toVertex];
```
szhorvat, line 217: Let's stick to the documented API and use `igraph_adjlist_get` here instead of directly accessing the innards of the the adjlist.

src/centrality/truss.cpp
iosonofabio: I had the same thought. The truss is the subgraph, but because of inclusivity of course one can reconstruct the subgraphs from this function.

I suggest we add two functions to the API:
- `igraph_trussness` -> the current `igraph_truss`, but with the last argument renames to `trussness`
- `int igraph_truss(const igraph_t *graph, long int k, igraph_t *truss)`

And also let's use the convention of everyone else without subtracting -2 from the trussness

src/centrality/truss.cpp
iosonofabio: probably not, it's where the heavy lifting happens after support has been computed...

src/centrality/truss.cpp
iosonofabio: good catch!

src/centrality/truss.cpp
iosonofabio: Hmm, maybe this idea is not so good after all. If we want to give the user the actual subgraph, we need to initialize a new graph, but then vertex ids will not match unless we include a bunch of singletons (then it's not the actualy truss, technically). For k-cores, we only have one function `igraph_coreness`, let's do the same here.

iosonofabio:
Thank you @szhorvat for the thorough review. See above for my preference as of how to rename the function to `igraph_trussness`.

As of which `src/` subfolder versus header, I did the same that was there for `igraph_coreness`. In principle, this function does not compute a community so it should be moved to another header, but then we should move `igraph_coreness` too. I'm not sure we can do it without breaking API?

src/centrality/truss.cpp
```diff
@@ -0,0 +1,254 @@
...
+  IGRAPH_VECTOR_INIT_FINALLY(&unpacked_triangles, 0);
+  IGRAPH_VECTOR_INT_INIT_FINALLY(&triangles, 0);  // will be resized
+
+  // List the triangles as vertex triplets.
+  IGRAPH_CHECK(igraph_list_triangles(graph, &triangles));
+
+  // Unpack the triangles from vertex list to edge list.
+  long int num_triangles = igraph_vector_int_size(&triangles);
+  IGRAPH_CHECK(igraph_vector_resize(&unpacked_triangles, 2 * num_triangles));
+  igraph_truss_i_unpack(&triangles, &unpacked_triangles);
```
ntamas, line 91: Minor nitpicking: I would move the `igraph_vector_resize` call to inside `igraph_truss_i_unpack()`. That would eliminate the pre-condition of `igraph_truss_i_unpack()`; right now the function assumes that there is enough space in `unpacked_triangles` for all the edges.

src/centrality/truss.cpp
```diff
@@ -0,0 +1,254 @@
...
+ */
+int igraph_trussness(const igraph_t* graph, igraph_vector_int_t* truss){
+  igraph_vector_int_t triangles, support;
+  igraph_vector_t eid, unpacked_triangles;
+
+  /* Manage the stack to make it memory safe */
+  IGRAPH_VECTOR_INT_INIT_FINALLY(&support, igraph_ecount(graph));
+  IGRAPH_VECTOR_INIT_FINALLY(&eid, 0);
+  IGRAPH_VECTOR_INIT_FINALLY(&unpacked_triangles, 0);
+  IGRAPH_VECTOR_INT_INIT_FINALLY(&triangles, 0);  // will be resized
```
ntamas, line 83: Add a comment here that mentions that `triangles` must be the last one as it will be the first vector to be destroyed and popped off the `FINALLY` stack later. In case someone feels the urge to reorder the variable initializations in alphabetic order ;)

src/centrality/truss.cpp
```diff
@@ -0,0 +1,254 @@
...
+  long int num_triangles = igraph_vector_int_size(&triangles);
+  IGRAPH_CHECK(igraph_vector_resize(&unpacked_triangles, 2 * num_triangles));
+  igraph_truss_i_unpack(&triangles, &unpacked_triangles);
+  igraph_vector_int_destroy(&triangles);
+  IGRAPH_FINALLY_CLEAN(1);
+
+  // Get the edge ids of the unpacked triangles. Note: a given eid can occur
+  // multiple times in this list if it is in multiple triangles.
+  IGRAPH_CHECK(igraph_vector_resize(&eid, num_triangles));
+  IGRAPH_CHECK(igraph_get_eids(graph, &eid, &unpacked_triangles, 0, 0, 1));
```
ntamas, line 98: I think that `igraph_get_eids()` will resize the `eid` vector as needed so no need to resize it in advance.

src/centrality/truss.cpp
```diff
@@ -0,0 +1,254 @@
...
+  IGRAPH_FINALLY_CLEAN(1);
+
+  // Compute the support of the edges.
+  igraph_truss_i_compute_support(&eid, &support);
+  igraph_vector_destroy(&eid);
+  IGRAPH_FINALLY_CLEAN(1);
+
+  // Compute the truss of the edges.
+  IGRAPH_CHECK(igraph_vector_int_resize(truss, igraph_ecount(graph)));
+  IGRAPH_CHECK(igraph_i_trussness(graph, &support, truss));
```
ntamas, line 109: Again, let's make the resizing of `truss` part of `igraph_i_trussness`. This is also consistent with the rest of the library in the sense that almost all functions can take vectors of arbitrary length and take care of resizing them as needed on their own.

tests/unit/truss.c
```diff
@@ -0,0 +1,76 @@
...
+    igraph_vector_view(&v, edges, n);
+
+    igraph_create(&graph, &v, 0, IGRAPH_UNDIRECTED);
+  
+    /* Compute the trussness of the edges. */
+    igraph_vector_int_init(&trussness, igraph_ecount(&graph));
+    igraph_trussness(&graph, &trussness);
+    print_and_destroy(&graph, &trussness);
+
+    /* Compute the actual subgraph */
```
ntamas, line 71: I think this line can be removed.

src/centrality/truss.cpp
```diff
@@ -0,0 +1,254 @@
...
+
+  // Get max possible value = max entry in support.
+  int max = igraph_vector_int_max(support);
+
+  // The vector of levels. Each level of the vector is a set of edges initially
+  // at that level of support, where support is # of triangles the edge is in.
+  vector< unordered_set<long int> > vec(max + 1);
+
+  // Initialize completed edges.
+  IGRAPH_VECTOR_BOOL_INIT_FINALLY(&completed, igraph_ecount(graph));
```
ntamas, line 161: I was wondering whether we should use an `std::vector<bool>` here if we are in C++-land already. `igraph_vector_bool_t` is not optimized at all; it uses one byte per item, while I would expect that `std::vector<bool>` is space-optimized in any reasonably implemented standard C++ library.

@szhorvat Any opinions here?

ntamas:
Done with my part of the review as well; let's discuss the points I raised and I think we are good to go afterwards.

iosonofabio:
Thank you @ntamas 

I've implemented the requested changes and renamed the output variable as `trussness` since that's what t is.

I've converted `completed` to a `vector<bool>`, a quick googling indicated we lose nothing and it will be bit-optimized on some platforms. The cost is that when you address pointers along the vector things might not be as clean because of course you are not looking at a specific memory location anymore, but a proxy therefor that is cut to a single bit. Since we only use the values and don't do much with the pointers themselves, we are just fine I think.

szhorvat:
The final thing would be to go through this:

https://github.com/igraph/igraph/wiki/Checklist-for-new-%28and-old%29-functions

There are several relevant points, such as:

 - Catching exceptions
 - Add a test for edge cases (e.g. null graph, singleton, etc.)
 - Does it work for various kinds of non-simple graphs? If not, can it be made to work? If not, check input / document situation.

iosonofabio:
I can add a test case for the trivial graph. On the other hand, I don't have a lot of time atm to check support for multigraphs. Would be great if you could take a look.

ntamas:
Re exceptions: basically just wrap all C++ code in an `IGRAPH_HANDLE_EXCEPTIONS()` macro from `core/exceptions.h` and you are OK. I can do it if you are about to go to sleep now.

iosonofabio:
I've wrapped a few things, but it's a little unclear to me what exactly you guys are requesting here. C++ can raise exceptions at every single function call in principle, but wrapping every `vector.begin()` seems silly.

@ntamas why don't you finish up the wrapping the way you like it and then directly push here and merge? That'd be more efficient than me trying to read your mind.

ntamas:
Sure, hang on, let me drink my coffee and then I'll get to it.

iosonofabio:
haha, sure, thank you! Then I just opened this one to feel the heck overwhelmed with all this:

https://github.com/igraph/python-igraph/issues/389

;-)

ntamas:
Looking good here. I want to add some more tests to cover the cases with loop / multiple edges and then we're good to go.

iosonofabio:
Thank you! I wasn't sure if you could wrap the whole function call with `IGRAPH_CHECK` and all with the exception-catching macro of if that led to conflicts.

Almost there!

src/centrality/truss.cpp
```diff
@@ -0,0 +1,266 @@
...
+*/
+
+#include "igraph_adjlist.h"
+#include "igraph_community.h"
+#include "core/exceptions.h"
+#include "igraph_error.h"
+#include "igraph_interface.h"
+#include "igraph_motifs.h"
+
+#include <iostream>
```
szhorvat, line 36: This header should not be needed (and such code should do no I/O).

szhorvat:
With C++ code that is not borrowed, I normally use explicit `try ... catch` blocks. With code that is written by us, we should know precisely what exceptions to expect, and how to best translate them to igraph error codes. I originally wrote `IGRAPH_HANDLE_EXCEPTIONS` to deal with C++ libraries where we are not fully in control of what may be thrown.

@iosonofabio You only need one try/catch block for every public function. Wrap the entire contents. No `extern "C"` function should throw exceptions. This is a recurring issue, and quite important to be aware of.

iosonofabio:
@szhorvat Tamas is in charge now, perhaps you two have a slightly different opinion on this one, so please feel free to connect with him and suggest him what you think is best

szhorvat:
All that said, I see no good reason for this code to use C++. As far as I can tell, originally it only used sets with integer elements. We do have such a set data structure in igraph.

I would not spend the time to rewrite this to use `igraph_set_t`, but neither would I entangle it more with C++ constructs. In particular the special-cased `vector<bool>` is widely considered to have been a mistake, and most people wish it could go away.

szhorvat:
> @szhorvat Tamas is in charge now, perhaps you two have a slightly different opinion on this one, so please feel free to connect with him and suggest him what you think is best

I pointed out the issue with exceptions as you need to be aware of this any time you work with C++ code in igraph (this code or any other in the future).

iosonofabio:
Sure, we all need to be aware, but there doesn't seem to be a single opinion on the table. Perhaps we could discuss it at our next zoom meeting and decide together what is the best practice.

szhorvat:
Fabio, you misunderstood. There are no differences of opinion, not is there any reason to dwell on this point. One way or another, you must ensure that any functions that gets called from C will never throw exceptions. Please do this in the future for any function that you write in C++.

iosonofabio:
Then I think I had understood just fine ;-)

ntamas:
> We do have such a set data structure in igraph.

Yes, we do, but I don't think it's optimized for performance at all. It simply uses a sorted list, which kinda sucks if you do frequent insertions and removals (which is the case here). I believe that `std::set` should have better performance; most of the STL implementations use a red-black tree or something similar.

I propose leaving the C++ constructs as they are now in the latest revision; we can rewrite this later in C if we feel the urge to do so. I am not going to introduce any further C++ constructs, though.

One final thing that's left here is to test whether the function works fine with multigraphs; I'll take care of this soon, hopefully today.

ntamas:
The function does not work with multigraphs as is, so I have added a check for that. It is good to merge now. Thanks for all your work on this @alexperrone and @iosonofabio !

alexperrone (commented):
> All that said, I see no good reason for this code to use C++. As far as I can tell, originally it only used sets with integer elements. We do have such a set data structure in igraph.
> 
> I would not spend the time to rewrite this to use `igraph_set_t`, but neither would I entangle it more with C++ constructs. In particular the special-cased `vector<bool>` is widely considered to have been a mistake, and most people wish it could go away.

I believe at the time of the original implementation, either there was no `igraph_set_t` or I couldn't figure out how to get what was needed which was actually a `vector` of a `set` of `long int`s (i.e. equivalent to the type `vector< unordered_set<long int> >`. If you can have a vector consisting of `igraph_set_t`s then that should work, in theory.

szhorvat:
> @iosonofabio Yes, the edits needed seemed relatively minor. I would kindly request that the original copyright under Johns Hopkins to remain the same since the **the core algorithm was published under that copyright**, and they allowed me to publish the code under GPL under the condition that the particular copyright is included.

Can you please clarify if you mean the algorithm (which cannot be copyrighted at all) or a specific implementation (actual code)?  When writing this code, did you re-use previous code whose copyright you did not own?

alexperrone:
> > @iosonofabio Yes, the edits needed seemed relatively minor. I would kindly request that the original copyright under Johns Hopkins to remain the same since the **the core algorithm was published under that copyright**, and they allowed me to publish the code under GPL under the condition that the particular copyright is included.
> 
> Can you please clarify if you mean the algorithm (which cannot be copyrighted at all) or a specific implementation (actual code)? When writing this code, did you re-use previous code whose copyright you did not own?

I meant the code. As far as re-use, I'm not sure how to give a clear yes/no... what I did is that I adapted/translated Java code into C++ code that formed this PR. Now, the Java code is under the same copyright (Johns Hopkins) that I do not own and also proprietary anyway. I got permission from Johns Hopkins to publish the C++ code here under a GPL license with the condition the copyright is included as a header.  Does this answer your questions?

szhorvat:
@alexperrone Yes, thank you. Let's keep it as it is.

Issue #1042, opened 2017-11-19
Wishlist: Graph colouring algorithms


szhorvat:
This is a feature request to implement graph colouring (vertex or edge colouring) algorithms.

As a start, it would be nice to have the Brélaz algorithm

szhorvat:
Here's a proposal for an interface:

    igraph_vertex_coloring_greedy(
        const igraph_t *graph, 
        igraph_vector_int_t *colors,
        igraph_greedy_coloring_t heuristic);

This function would implement greedy colouring with a selectable heuristic. Greedy colouring works by colouring vertices one by one with the smallest-indexed colour that differs from all already coloured neighbours. The performance depends on the order the vertices are considered. This is chosen based on the `heuristic` argument.  This method does not necessarily find the mimimum colouring.

Later we can have a separate function `igraph_vertex_coloring_minimum`, which may need to have a different signature (e.g. it may be able to find *all* minimum colourings).  This is to be decided when an implementation is ready.

Does this sound good to everyone (before I send a PR for the greedy algorithm)?

ntamas:
Sounds good to me!

szhorvat:
Fixed by #1043

euler314:
@szhorvat We've discussed various aspects of this before, and I'm also very interested in adding more support for graph coloring to igraph. Did you have any plans of adding more heuristics? I'd be happy to contribute too.

szhorvat:
@euler314 I did not have immediate plans (primarily due to lack of time), but I am willing to work with you on this, especially if you can also contribute some code. I can help you get through the initial technical hurdles (i.e. the igraph-specific quirks).

Which specific method do you have in mind?

euler314:
@szhorvat That would actually be quite helpful. I was thinking I could do RLF (recursive largest first), which in general is a bit slower than the sequential method you implemented. I could start out with writing a first implementation using igraph, and we can then see how to adjust it.

szhorvat:
Can you give a link to a concise description of the algorithm (or provide a quick summary)?

euler314:
There should be many resources describing RLF. The following is one of my first search hits: https://www.gerad.ca/~alainh/RLFPaper.pdf (see Figure 1 on page 3).

PR #1043, opened 2017-11-22
Add igraph_vertex_coloring_greedy()

szhorvat:
This pull request adds a basic vertex colouring algorithm per #1042 

Feedback needed on:

 - function naming
 - which section to put documentation in

Still needs to be finished:

 - [x] integrate documentation
 - [x] add test / benchmark
 - [ ] add more heuristics for greedy colouring

I'll wait for feedback before continuing with these tasks.

I did not check that certain `long` or `double` values fit into `igraph_integer_t` because this is exceedingly unlikely to happen.


src/coloring.c
```diff
@@ -0,0 +1,107 @@
...
+    while (VECTOR(cn)[vertex] >= 0) {
+        igraph_vector_int_t *neighbors = igraph_adjlist_get(&adjlist, vertex);
+        long neigh_count = igraph_vector_int_size(neighbors);
+
+        /* colour current vertex */
+        {
+            igraph_integer_t col;
+            igraph_vector_int_t neigh_colors;
+
+            IGRAPH_CHECK(igraph_vector_int_init(&neigh_colors, neigh_count));
```
ntamas: It is probably unfortunate to re-allocate `neigh_colors` in every iteration of the loop. I would move the allocation outside the `while` loop and just use `igraph_vector_int_resize()` here.

Note that if the allocation is moved outside the `while` loop, you'll also need to guard it with `IGRAPH_FINALLY(igraph_vector_int_destroy, &neigh_colors)`.

src/coloring.c
```diff
@@ -0,0 +1,107 @@
...
+ * This function assigns a "colour"---represented as a positive integer---to
+ * each vertex of the graph in such a way that neighbouring vertices never have
+ * the same colour. The obtained colouring is not necessarily minimal.
+ *
+ * \param graph The input graph.
+ * \param colors Pointer to an initialized integer vector. The colours will be stored here.
+ * \param heuristic The vertex ordering heuristic to use during greedy colouring.
+ *
+ */
+int igraph_vertex_coloring_greedy(const igraph_t *graph, igraph_vector_int_t *colors, igraph_coloring_greedy_t heuristic) {
```
ntamas: Return type should be `igraph_error_t`.

src/coloring.c
```diff
@@ -0,0 +1,107 @@
+
+#include "igraph_coloring.h"
+#include "igraph_interface.h"
+#include "igraph_adjlist.h"
+#include "igraph_interrupt_internal.h"
+#include <assert.h>
+
+
+int igraph_i_vertex_coloring_greedy_cn(const igraph_t *graph, igraph_vector_int_t *colors) {
```
ntamas: Return type should be `igraph_error_t`.

src/coloring.c
```diff
@@ -0,0 +1,107 @@
...
+ * \param graph The input graph.
+ * \param colors Pointer to an initialized integer vector. The colours will be stored here.
+ * \param heuristic The vertex ordering heuristic to use during greedy colouring.
+ *
+ */
+int igraph_vertex_coloring_greedy(const igraph_t *graph, igraph_vector_int_t *colors, igraph_coloring_greedy_t heuristic) {
+    switch (heuristic) {
+    case IGRAPH_COLORING_GREEDY_CN:
+        return igraph_i_vertex_coloring_greedy_cn(graph, colors);
+    }
```
ntamas: `default` branch missing from `switch()` so we may end up returning something undefined if the heuristic is invalid. We need to throw an `IGRAPH_EINVAL` error instead.

src/coloring.c
```diff
@@ -0,0 +1,107 @@
...
+ * \function igraph_vertex_coloring_greedy
+ * \brief Computes a vertex colouring using a greedy algorithm.
+ *
+ * </para><para>
+ * This function assigns a "colour"---represented as a positive integer---to
+ * each vertex of the graph in such a way that neighbouring vertices never have
+ * the same colour. The obtained colouring is not necessarily minimal.
+ *
+ * \param graph The input graph.
+ * \param colors Pointer to an initialized integer vector. The colours will be stored here.
```
ntamas: It has to be documented how the colors are represented; currently they seem to be 1-based integers. I think it would be better if we declared that the colors are 0-based integers, but then we need to subtract 1 from the result of the greedy algorithm.

ntamas:
Function naming is okay. The documentation should probably be in a completely separate chapter, e.g., after "Graph Isomorphism". Right now the chapter wouldn't be too long, but it can change in the future.

Looking forward to the tests as well.

src/coloring.c
szhorvat: Most existing functions use `int` as the return type. `igraph_error_type_t` is an enum, and my compiler represents it with an unsigned value. This then causes a large number of warnings due to signedness conversion in `IGRAPH_CHECK`.

Question: Do any of the functions that return an error code ever return a negative value? (Of course, that wouldn't correspond to any existing enum value, but that doesn't mean that it's never used...)  If not, it would be nice to change `IGRAPH_CHECK` to use `igraph_error_type_t` internally, and eliminate some of the warnings.

src/coloring.c
szhorvat: Fixed, though it didn't result in an appreciable speedup.

szhorvat:
I changed the algorithm to use an indexed heap. I used `igraph_2wheap_t`.  I am not 100% confident about using this. Could you please look over the code again?

In particular, it wasn't clear what the `has_active` and `deactivate` functions do, so I ignored them.

src/coloring.c
szhorvat: It was documented, but very subtly: "represented as positive integers". I made it explicit now (and use non-negative integers instead).

szhorvat:
I added the missing pieces (tests, documentation), so I think this is basically ready for merging (unless you find problems).

More vertex ordering methods can be added later.

I didn't implement the bitfield thing, as it didn't seem worth the trouble. Does igraph has a data structure for this already, or would I need to implement it from scratch?

src/coloring.c
ntamas: Okay, just stick to `int` for the time being - I have forgotten that we weren't using the enum for the return value, sorry.

szhorvat:
Forgot to say, I can squash the commits into a single one before merging, to keep the mess to a minimum. Just let me know.

ntamas:
That bitfield thing was only an idea and it's probably not worth the fuss at this stage. I'll merge the PR as is, thanks!

Issue #981, opened 2017-03-27
Wishlist: Random walk with weights


szhorvat:
There is already `igraph_random_walk`, but it does not support weights.

It would be nice to have a similar function (or an extension to this function) which chooses an out-edge to take a step along with probability proportional to its weight.

I might do this in the future if @gaborcsardi or @ntamas approve and will let me know if a separate function or an extension to `igraph_random_walk` is preferable.

ntamas:
Sounds like a good idea and I would prefer a separate function for this.

I will probably not have time to review the PR this week but I'll try to do it during the weekend or next week.

szhorvat:
OK. I'll just go ahead and add weight support for the original function as well, so you can review everything at the same time (there's no rush).  It will be a minor change.

ncdingari:
Did you implement random_walk based on Weights of Edges and Nodes?

szhorvat:
Yes, but I didn't send it as a PR as it was basically done in terms of `random_edge_walk` as a quick-fix for my own application.

It's something like this:

```c
igraph_random_edge_walk(&graph, &weights, &edgeWalk, start, IGRAPH_OUT, steps-1, stuck)

long len = igraph_vector_size(&edgeWalk) + 1;
long last = start;
vertexWalk[0] = last;
for (long i=1; i < len; ++i) {
    last = IGRAPH_OTHER(&graph, edgeWalk[i-1], last);
    vertexWalk[i] = last;
}
```

(This is simplified code and I almost certainly introduces mistakes, but you get the idea.)

An implementation that uses edge weights would be almost a copy-paste of what `random_edge_walk` does anyway. However, it would avoid having to keep *two* walk vectors in memory.

----

As for "based on weights of nodes", it is not clear to me what you mean.

ncdingari:
Thanks for quick reply. Do you have similar function `random_edge_walk` in R. I can't find it in `igraph `library

szhorvat:
My contribution was to the C core. Perhaps the next version of the R API will expose it.

jefferis:
Did `random_edge_walk` ever get added to R igraph?

vtraag:
@jefferis, no it is not yet implemented in the R interface. We are currently working on updating the R interface to the recently released C core, version 0.8.0, see https://github.com/igraph/rigraph/pull/372. If that is done, the new contributions could start to be added.

vtraag:
[...]

PR #1046, opened 2017-11-27
Add igraph_random_edge_walk()

szhorvat:
This pull request adds the function `igraph_random_edge_walk()`.

It is analogous to `igraph_random_walk()` with the following differences:

 - it returns a list of edge indices
 - it supports edge weights

Related bug: #981 

There I asked if `random_walk()` should be changed to return an edge list instead of a vertex list. It is probably best to have both variants. Obtaining an edge list from a vertex list is not possible when multiple edges are present between the same vertices.  Obtaining a vertex list from an edge list is possible, but it is inconvenient for undirected graphs (because it is not clear in which direction the edge was traversed).

Still missing:

 - [x] Unit tests
 - [ ] Update `igraph_random_walk()` to also handle weights

I will do these after the review.

### DIFF
--- a/src/random_walk.c
+++ b/src/random_walk.c
@@ -26,6 +26,7 @@
 #include "igraph_interface.h"
 #include "igraph_random.h"
 #include "igraph_memory.h"
+#include "igraph_interrupt_internal.h"
 
 /**
  * \function igraph_random_walk
@@ -268,6 +269,8 @@ int igraph_random_edge_walk(const igraph_t *graph,
             }
             break;
         }
+
+        IGRAPH_ALLOW_INTERRUPTION();
     }
 
     RNG_END();
--- a/src/random_walk.c
+++ b/src/random_walk.c
@@ -230,6 +230,8 @@ int igraph_random_edge_walk(const igraph_t *graph,
                 long j;
 
                 *cd = igraph_malloc(sizeof(igraph_vector_t));
+                if (*cd == NULL)
+                    IGRAPH_ERROR("random edge walk failed", IGRAPH_ENOMEM);
                 IGRAPH_CHECK(igraph_vector_init(*cd, degree));
 
                 IGRAPH_CHECK(igraph_vector_resize(&weight_temp, degree));
--- a/src/random_walk.c
+++ b/src/random_walk.c
@@ -111,3 +111,118 @@ int igraph_random_walk(const igraph_t *graph, igraph_vector_t *walk,
 
   return 0;
 }
+
+
+/**
+ * \function igraph_random_edge_walk
+ * \brief Perform a random walk on a graph and return the traversed edges
+ *
+ * Performs a random walk with a given length on a graph, from the given
+ * start vertex. Edge directions are (potentially) considered, depending on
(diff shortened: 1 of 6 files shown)


szhorvat:
Changes in the last two commits:

 - fix a bug (wrong result when `mode != IGRAPH_OUT` or the graph is undirected)
 - add benchmark
 - attempt to improve performance

About the performance optimization:

In the case of a weighted graph, we need to select an edge to traverse next with a probability proportional to its weight. This is done by constructing a cumulative distribution vector for the incident edges.  In the previous iteration, this was done in every single step of the algorithm.  This new version will cache and re-use the result. This takes an extra O(|E|) memory.

However, the performance improvement isn't clear:

 - on a small graph, the timing improved from 1.5 s to 1.0 s.
 - on a big graph, the timing increased from 5.4 to 6.0 s.

I suspect that this is due to caching effect.

A further possible improvement would be using a lazy incidence list instead of an incidence list. In practice, this slows down the program (5.4 s -> 7.2 s with original algorithm), possibly because the lazy incidence list uses `double` vectors instead of integer ones.

Finally, due to many conditionals in the main loops, branch prediction failures may have a performance impact too. I am not sufficiently familiar with these effects to be able to guess at performance without testing, so I am going to stop here.

src/random_walk.c
```diff
@@ -196,31 +219,61 @@ int igraph_random_edge_walk(const igraph_t *graph,
...
 
-            IGRAPH_CHECK(igraph_vector_resize(&ws, degree));
-            for (j=0; j < degree; ++j)
-                VECTOR(ws)[j] = VECTOR(*weights)[ VECTOR(*edges)[j] ];
+            /* compute out-edge cdf for this node if not already done */
+            if (IGRAPH_UNLIKELY(! *cd)) {
+                long j;
 
-            IGRAPH_CHECK(igraph_vector_cumsum(&cs, &ws));
+                *cd = igraph_malloc(sizeof(igraph_vector_t));
```
ntamas: You need to check whether `igraph_malloc` returned null and return with `IGRAPH_ENOMEM` if so.

szhorvat:
Would a similar implementation be okay for the weighted case of `igraph_random_walk()`?  I was thinking of leaving the current implementation of `igraph_random_walk()` mostly alone (other than adding a weights argument), and just adding a second (non-public) function that will be called for the weighted case.

src/random_walk.c
szhorvat: Sorry about the slow response.  Would this look reasonable (message phrasing, etc.)?

                if (*cd == NULL)
                    IGRAPH_ERROR("random edge walk failed", IGRAPH_ENOMEM);

ntamas: Sure, that's fine.

ntamas:
Re your last comment above (which I have missed somehow): yes, that's okay, just add another non-public function for the weighted case.

ntamas:
Shall I merge the current PR or will you do it in the same PR?

szhorvat:
I'll do it in a separate PR.

PR #1047, opened 2017-12-04
Fix compile error with -Werror-format-security.

shlomif:
This patch was originally written for the Mageia Linux package, where
the GCC flag is enforced.

### DIFF
--- a/src/f2c/arithchk.c
+++ b/src/f2c/arithchk.c
@@ -110,7 +110,8 @@ icheck(void)
 	return 0;
 	}
 
-char *emptyfmt = "";	/* avoid possible warning message with printf("") */
+/* avoid possible warning message with printf("") */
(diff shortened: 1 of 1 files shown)


ntamas:
Makes sense, thanks.

shlomif:
On Mon, 04 Dec 2017 03:14:08 -0800
Tamás Nepusz <notifications@github.com> wrote:

> Makes sense, thanks.
> 

You're welcome and thanks! That was also very quick. 

-- 
-----------------------------------------------------------------
Shlomi Fish       http://www.shlomifish.org/
http://www.shlomifish.org/humour/bits/facts/Summer-Glau/

The one and only, 100% original, real, actual, and unmatched, Slimy Fish™!

Please reply to list if it's a mailing list post - http://shlom.in/reply .

PR #1065, opened 2018-02-23
Add Gexf Writer based on Graphml writer.

greebie:
This PR provides a writer for GEXF format (1.2 draft) following the code in foreign-graphml.c.

A few notes:

- **Gexf supports visualization outputs.**  
- `<viz:position>`. look for attributes x, y & z for values.  If there is x & y but no z, z will be 0.0.
- `<viz:size>` - looks for "size" attribute and uses that.
- `<viz:color>` currently looks for r, g, & b values.  For consideration: could include alpha 'a' (transparency) with rgb and/or look for hex if there is no rgb.

- **Gexf 1.2 does NOT support graph attributes in the schema**
- I have included graph attributes, even though its not supported in the schema. 
- I tested outputs in sigmaJS with no troubles. In gephi you get some errors that are ignored.  Graph attributes do not seem to be included in 1.3 so far.
- I can easily comment out the graph attribute sections for later support if desired.

- ** This currently has no unit tests **
- I will add these, but wanted to get some advice on desired features above before I lock it in.
- I had an odd test failure for Adjacency Spectral Embedding, which has nothing to do with my code (I don't think).

- ** I still need to examine places where I can free up memory **

- ** having issues compiling in Mac, so moved to a Debian VM **

### DIFF
--- a/src/foreign-gexf.h
+++ b/src/foreign-gexf.h
@@ -1,0 +1,433 @@
+/* -*- mode: C -*-  */
+/*
+   IGraph R package.
+   Copyright (C) 2006-2012  Gabor Csardi <csardi.gabor@gmail.com>
+   334 Harvard street, Cambridge, MA 02139 USA
+
+   This program is free software; you can redistribute it and/or modify
+   it under the terms of the GNU General Public License as published by
+   the Free Software Foundation; either version 2 of the License, or
+   (at your option) any later version.
+
+   This program is distributed in the hope that it will be useful,
+   but WITHOUT ANY WARRANTY; without even the implied warranty of
+   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+   GNU General Public License for more details.
+
+   You should have received a copy of the GNU General Public License
+   along with this program; if not, write to the Free Software
+   Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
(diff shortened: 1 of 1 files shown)


ntamas:
(First of all, apologies for the delay; I know that I'm quite late to the party here, but sadly I hardly have any time to work on igraph any more).

Thanks for your contribution! Seems like a good starting point, but the entire code is in a _header_ file. Header files should not contain code; the code itself should go in a `.c` file instead.

Further comments:

* The code of `igraph_i_xml_escape` seems to be duplicated from the GraphML reader; instead of that, one should simply use the existing implementation.

* There are no test cases that validate whether the GEXF writer works, and there is no documentation for the implemented function.

* `GEXF_NAMESPACE_URI` is not referenced anywhere (although it should be as its value is clearly being used).

* `igraph_cattribute_has_attr()` and related functions assume that the graph uses the C attribute handler; this approach will fail if igraph is being used from R or Python because those higher-level interfaces use a different attribute handler. The same applies to the macros `VANV`, `VASV` and so on. You need to use the `igraph_i_attribute_...` functions just like they are being used from the GraphML writer because they call the right functions from the attribute handler table currently attached to igraph instead of assuming that the C attribute handler is being used.

szhorvat:
@greebie Are you still planning to work on this? It would be great to finish this PR.

PR #1068, opened 2018-02-27
Compiles with musl libc

### LINKED ISSUE #1066: Does not build with musl
<details>

Make sure that these boxes are checked before submitting your issue -- thank you!

- [x] This issue is for the C core of igraph.

</details>

here you define `OFF_T` as `__off64_t` when building under linux, but when building with musl, it doesn’t exist:

https://github.com/igraph/igraph/blob/02005680aebb4281f2964c6d2d8d98851fce63b9/src/f2c/sysdep1.h#L16-L18

musl [defines](https://github.com/runtimejs/musl-libc/blob/master/include/sys/types.h#L83) `off64_t` instead of `__off64_t`, so the check needs to be different than `__linux__` or `__sun__`.

flying-sheep:
Fixes #1066

I’m not sure which of the branches is hit, but I assume the fallback branch (empty ieee0 function).

This fits with the [musl FAQ entry](https://wiki.musl-libc.org/functional-differences-from-glibc.html#Floating-point-exceptions) explaining that there is no standards compliant behavior that enables floating point exceptions.

If we really need the behavior specified [here](https://github.com/igraph/igraph/blob/02005680aebb4281f2964c6d2d8d98851fce63b9/src/f2c/uninit.c#L298-L299) and feel adventurous, we could try the assembler code [here](https://github.com/iXit/Mesa-3D/pull/214/files) on x86 and x86_64.

### DIFF
--- a/src/f2c/sysdep1.h
+++ b/src/f2c/sysdep1.h
@@ -15,8 +15,13 @@
 
 #ifdef __linux__
 #define USE_LARGEFILE
+
+#ifdef __GLIBC__
 #define OFF_T __off64_t
-#endif
+#else
+#define OFF_T off64_t
+#endif /* __GLIBC__ */
+#endif /* __linux__ */
 
 #ifdef _AIX43
 #define _LARGE_FILES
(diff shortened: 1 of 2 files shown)


ntamas:
Thanks!

PR #1077, opened 2018-03-21
igraph_realize_degree_sequence

szhorvat:
This is an implementation of a number of variations on the Havel-Hakimi algorithm.

There are multiple similar versions of functions in the code. Before it's merged, I would appreciate some help on reducing code duplication without affecting performance.


szhorvat:
The documentation now builds properly, but I'm not sure how to get rid of the `extern "C"` that shows there without an extra wrapper function.

ntamas:
Seems good so far but there are conflicts that have to be resolved before merging. Are you planning to add anything else to this PR?

szhorvat:
Sorry about being slow with some of the PRs. I'll try to clean them up next week. Please don't merge any of my stuff until then.  I'll let you know.

ntamas:
No need to rush -- I have quite a long backlog as well :)

src/degree_sequence.cpp
```diff
@@ -0,0 +1,490 @@
...
+ * \function igraph_realize_degree_sequence
+ * \brief Generates a graph with the given degree sequence
+ *
+ * This function constructs a simple graph that realizes the given degree sequence
+ * using the Havel-Hakimi algorithm, or the given (directed) out- and in-degree
+ * sequences using the related Kleitman-Wang algorithm.
+ *
+ * The algorithms work by choosing an arbitrary vertex and connecting all its stubs
+ * to other vertices of highest degree.  In the directed case, the "highest" (in, out) degree
+ * pairs are determined based on lexigographic ordering.
```
ntamas:
```suggestion
 * pairs are determined based on lexicographic ordering.
```

ntamas:
Also, would it be possible to add some basic test cases? After all, these are deterministic algorithms so it should be easy to write a deterministic test case for them.

Otherwise I'm happy with the proposed changes so I'll merge them once these points are addressed.

szhorvat:
> After all, these are deterministic algorithms so it should be easy to write a deterministic test case for them.

It turns out that this is a bit trickier than I expected.

The algorithms rely on `std::sort`, which may give different results on different platforms when there are ties.

To put it simply, when two nodes have the same degree, the order in which they are connected up effectively becomes platform dependent.

I think I should switch to `std::stable_sort` for the sake of a consistent output on all platforms, even if this comes with a small performance degradation.  Do you agree?

szhorvat:
> Would it be possible to merge igraph_i_havel_hakimi_smallest and igraph_i_havel_hakimi_largest?

Actually, I merged an unmerged this multiple times during development.

Now they're merged again.

Unless I missed something, I think this PR is ready to merge now.

ntamas:
Can you rebase this to the current master? I'll merge it once it's done.

szhorvat:
By rebase, do you mean squashing all commits into one?

szhorvat:
I'm confused. This is what I see:

image

"This branch has no conflicts with the base branch"

I have also merged the master branch into the realize_degseq branch and there were no conflicts.

szhorvat:
Do you have the realize_degseq branch locally on your computer?

There were indeed conflicts, see your May 11 comment above.

While cleaning those up, I changed the history in this branch (with rebase) in an attempt to make it "clean" again.  If you previously had this branch on your own computer, then the equivalence with the GitHub copy broke at that point.  Maybe I should not have done that.

ntamas:
This is what I see on my machine (Github web UI, I don't have this PR checked out on my machine):

screenshot 2018-12-02 at 20 33 31

So, this has nothing to do with what I have on my computer as it is the Github web UI that claims that I cannot merge this PR right now.

szhorvat:
Do you have any idea what I can do to fix this (since on my end it shows no conflicts)? I'm thinking of just opening a new PR instead with the same changes.

ntamas:
I'll try a manual merge from the CLI and see how Github reacts.

szhorvat:
Please wait, I'll send a new PR. SOmething's really wrong with this branch.

ntamas:
Too late :) Manual merge seems to have worked, thanks!

PR #1079, opened 2018-04-05
Prüfer sequences and random trees

szhorvat:
This pull request includes two new functions:

 - `igraph_from_prufer` creates a tree from a [Prüfer sequence](https://en.wikipedia.org/wiki/Pr%C3%BCfer_sequence)

 - `igraph_tree_game` will sample uniformly from the set of labelled trees

The tree sampler works by generating random Prüfer sequences and then converting them to trees.

There are other functions that can create random trees, e.g. `igraph_growing_game` with `m=1` and `citation=true`. However, these don't sample uniformly.  (Specifically, `growing_game` does generate all its possible results with equal probability, but it cannot generate *all* labelled trees.)


szhorvat:
Potential improvements before merging:

 - Support directed (i.e. rooted) trees. This can be done by generating an undirected tree, picking a root at random, then orienting all edges away from it.

 - Add other generating methods. (However, I was unable to find a simple, direct construction method that results in uniform sampling. What I did find was blog posts and papers claiming to have such a method, but not sampling uniformly in reality.)

szhorvat:
I added a second method for random tree generation based on loop-erased random walks. Since it was easier this time, I also added support for directed trees.

Both methods generate all labelled trees with equal probability.

szhorvat:
I should note that I did numerically test uniform sampling for both methods, to verify the implementations.  This is not something that can be easily made into an automated test, so it's not added as such.

image

ntamas:
Seems good to me, thanks! Please add some test cases nevertheless because I don't have the capacity to review everything thoroughly (including memory access issues and so on), and it improves the process greatly if there are some tests that I could run through Valgrind.

In this particular case, `igraph_from_prufer()` is deterministic so it should be easy to create some tests for it (I would create a few example graphs, print their edge lists to stdout, and then compare stdout with an expected version coming from a file, like we do it in many other test cases). `igraph_tree_game()` is stochastic but you can initialize the igraph RNG with a predetermined seed to make the test deterministic.

ntamas:
Thanks a lot!

szhorvat:
Sorry, this was not quite done yet!  I'll send the remaining test in an additional pull request later.

ntamas:
Okay, thanks and no problem.

Pascal-Ortiz:
> This pull request includes two new functions:
> 
> * `igraph_from_prufer` creates a tree from a [Prüfer sequence](https://en.wikipedia.org/wiki/Pr%C3%BCfer_sequence)


> This is an implementation of the feature requested in #1084.
> 
Hi Szabolcs!

Good work! there are not so many libraries implementing Prufer conversions. 

The algorithm has quadratic time complexity so the implentation was not able to build in a reasonable time a tree having one million of nodes from a Prufer sequence.
There exists linear time algorithms for encoding a Prufer sequence into a tree, for instance this [paper](https://pdfs.semanticscholar.org/f6a7/dffda435d1f732d220312fb667e1ba4d41a5.pdf).

PR #1081, opened 2018-04-09
igraph_bridges

szhorvat:
A function to find bridges in the graph. An edge is a bridge if its removal disconnects the graph.

Based on https://www.geeksforgeeks.org/bridge-in-a-graph/


ntamas:
Seems good -- please add some tests and then I'll be happy to merge it.

szhorvat:
Changes:

 - more cross references in the documentation
 - added a simple test

ntamas:
`src/cs/cs.h` has a `CS_MAX` macro. igraph itself does not have its own variant. If you need one, feel free to add it; `src/igraph_math.h` seems to be a good candidate as it is not part of the public API so we don't provide yet-another-max-macro needlessly.

ntamas:
Can I merge this or do you plan to add something else?

szhorvat:
You can merge it.

ntamas:
Thanks!

PR #1083, opened 2018-04-10
igraph_is_tree

### LINKED ISSUE #1082: Wishlist: is_tree
There should be a function to check if a graph is a tree. It should work for directed trees too. There are two cases: all edges point away from the root, or all point towards the root.

`igraph_tree_game` (still in pull request phase) generates tree where the nodes point away from the root.

Suggested prototype:

```
int igraph_is_tree(const igraph_t *graph, igraph_neimode_t mode);
```

`mode` is ignored for undirected.

For directed,

 - `IGRAPH_ALL` ignores directions
 - `IGRAPH_OUT` gives true if all edges point away from the root
 - `IGRAPH_IN` gives true if all edges point towards the root

An alternative is to not have `mode` and just behave as `IGRAPH_OUT` for the directed case.

Is there really no such function that I missed? It would also be useful for implementing other functions and for writing tests.

szhorvat:
A function to test if a graph is a tree.  Fixes #1082. Tests are forthcoming, but comments are welcome before then.  I'd like to merge this before my other pull requests.


ntamas:
I have reviewed the code briefly and I am a bit worried about the DFS recursion. Would this work even with an undirected path graph with, say, ten million vertices? Won’t we run out of stack space earlier than we could recurse all the way down?

szhorvat:
I changed it to a recursion-free implementation.

src/structural_properties.c
```diff
@@ -6897,3 +6897,185 @@ int igraph_i_is_graphical_degree_sequence_directed(
...
+    IGRAPH_CHECK(igraph_vector_bool_init(&visited, igraph_adjlist_size(al)));
+    IGRAPH_FINALLY(igraph_vector_bool_destroy, &visited);
+
+    IGRAPH_CHECK(igraph_stack_int_init(&stack, 0));
+    IGRAPH_FINALLY(igraph_stack_int_destroy, &stack);
+
+    *visited_count = 0;
+
+    /* push the root into the stack */
+    igraph_stack_int_push(&stack, root);
```
ntamas: Pushing to the stack should be protected with `IGRAPH_CHECK()` as we may run out of memory here.

src/structural_properties.c
```diff
@@ -6897,3 +6897,185 @@ int igraph_i_is_graphical_degree_sequence_directed(
...
+        VECTOR(visited)[u] = 1;
+        *visited_count += 1;
+
+        /* register all its yet-unvisited neighbours for future processing */
+        neighbors = igraph_adjlist_get(al, u);
+        ncount = igraph_vector_int_size(neighbors);
+        for (i=0; i < ncount; ++i) {
+            igraph_integer_t v = VECTOR(*neighbors)[i];
+            if (! VECTOR(visited)[v])
+                igraph_stack_int_push(&stack, v);
```
ntamas: Pushing to the stack should be protected with `IGRAPH_CHECK()`

ntamas:
Thanks! Let me know if you are okay with merging this (after addressing my two comments above).

szhorvat:
It can be merged. Now all commits are squashed together. Once this is merged, I'll continue with the rest.

szhorvat:
One possible change could be using `igraph_tree_mode_t` instead of `igraph_neimode_t`. Let me know which one you prefer.

PR #1100, opened 2018-05-24
A portability fix for configure.ac

msk:
The "`test`" command, as well as the "`[`" command, are not required to
know the "`==`" operator. Only a few implementations like bash and some
versions of ksh support it.

When you run "`test foo == foo`" on a platform that does not support the
"`==`" operator, the result will be `false` instead of `true`. This can
lead to unexpected behavior.


gaborcsardi:
Thanks!

PR #1117, opened 2018-10-06
Fix isomorphic_bliss verification of coloring

frederik-h:
If igraph_isomorphic_bliss is called with colorings, it tries to verify that the isomorphism constructed by bliss does indeed preserve the colors.
The current check applies the isomorphism
in the wrong direction. That is, although the isomorphism map12 is a mapping
from the first to the second graph, it is checked that the coloring of 
each vertex in the second graph is equal to the coloring of its image
in the first graph. This leads to igraph_isomorphic_bliss rejecting
isomorphic colored graphs. The problem can be fixed by exchanging the roles
of the two graphs in the check.
An example which demonstrates the problem has been added as a test. (The last test in igraph_isomorphic_bliss.c; the other tests have been adapted from igraph_isomorphic_vf2.c since there seemed to be no dedicated tests covering the behavior of bliss for colored graphs).
The graph of this example has been adapted from the IsomorphismTests.testIsomorphic test of python-igraph.

### DIFF
--- a/examples/simple/igraph_isomorphic_bliss.c
+++ b/examples/simple/igraph_isomorphic_bliss.c
@@ -1,0 +1,164 @@
+/* -*- mode: C -*-  */
+/* 
+   IGraph library.
+   Copyright (C) 2009-2012  Gabor Csardi <csardi.gabor@gmail.com>
+   334 Harvard st, Cambridge MA, 02139 USA
+   
+   This program is free software; you can redistribute it and/or modify
+   it under the terms of the GNU General Public License as published by
+   the Free Software Foundation; either version 2 of the License, or
+   (at your option) any later version.
+   
+   This program is distributed in the hope that it will be useful,
+   but WITHOUT ANY WARRANTY; without even the implied warranty of
+   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+   GNU General Public License for more details.
+   
+   You should have received a copy of the GNU General Public License
+   along with this program; if not, write to the Free Software
+   Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 
+   02110-1301 USA
+
+*/
+
+#include <igraph.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <time.h>
+
+int random_permutation(igraph_vector_t *vec) {
+  /* We just do size(vec) * 2 swaps */
+  long int one, two, tmp, i, n=igraph_vector_size(vec);
+  for (i=0; i<2*n; i++) {
+    one= (double)rand() / RAND_MAX * n;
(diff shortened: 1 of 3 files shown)


szhorvat:
Thanks @frederik-h ! As I remember, that was my code.  You are right that the colour-vector permutation has to go the other way.

ntamas:
Thank you!

szhorvat:
It's been already merged, but I would like to make a few comments anyway, partly for the sake of keeping the tests also useful as an example of good igraph programming.

Line 47 of `igraph_isomorphic_bliss.c`:

Tests should be deterministic. If they fail, they should always fail. If they pass, they should always pass. Use a fixed seed for the RNG.

Use `igraph_vector_shuffle` to create a random permutation. Additional advantage: it uses igraph's RNG so it behaves the same on all systems.

Line 101:

Variable declarations must go to the start of the function in standard C89 (which igraph uses). Many compilers accept this, but it is not standard.

Line 103:

Do not use `igraph_empty_attrs` if not working with attributes. Use `igraph_empty`.

However, instead of creating an empty graph and adding edges to it, use `igraph_small` here. Adding edges to a graph is much slower than creating a graph directly from a list of edges (either using `igraph_small` for conveniently making small graphs, or `igraph_create` to use a stored edge list).

Line 132:

Again, declarations go to the front. However, why not re-use existing vectors `color1` and `color2` (after an appropriate `igraph_vector_resize`)?

Line 137:

`igraph_vector_init` already sets the elements to zero. No need for `vector_fill` or `vector_null`.

ntamas:
Thanks @szhorvat , I should have probably checked the PR more thoroughly. I'll take care of these tomorrow.

frederik-h:
At least the first two issues mentioned by @szhorvat were already present in the vf2 tests from which the new file was derived. They should probably also be fixed there. By the way, I think that it is fine to use a RNG in tests as long as you record the seed to enable repeating the test. 

Regarding the use of RNGs in tests and the version of the C-standard used be igraph: perhaps this could be clarified, for instance, in the CONTRIBUTING.md? 

@ntamas I can also make the changes and send a PR, if that's easier for you.

ntamas:
@frederik-h Thanks, it's not needed, I have done the changes yesterday already, I was just too tired to wait for the tests to compile. It's committed now.

I have also fixed the VF2 tests and added a note in `CONTRIBUTING.md` about test case determinism.

frederik-h:
> 
> 
> @frederik-h
> 
> > By the way, I think that it is fine to use a RNG in tests as long as you record the seed to enable repeating the test.
> 
> See here why it wasn't fine in this specific case: #1139
> 
> (It turns out that seeding was done incorrectly even after changes to this PR.)

Thanks for the pointer.

PR #1120, opened 2018-10-09
Update auto NCV selection for ARPACK

### LINKED ISSUE #1119: PageRank calculation with Arnoldi method does not work on 3-vertex graphs
Since the recent ARPACK-related changes, `igraph_pagerank` fails on graphs with 3 vertices.

The error message is:

> Error at arpack.c:1234 :ARPACK error, NCV must be greater than NEV and less than or equal to N (and for the non-symmetric solver NCV-NEV >=2 must also hold)

Here is a minimal example:

```c
#include <igraph.h>

int main() {
    igraph_t graph;
    igraph_vector_t result;
    double value;
    igraph_arpack_options_t arpack_options;

    igraph_small(&graph, 3, 0 /* directed=false */, 0,1, 1,2, -1);
    igraph_vector_init(&result, 0);

    igraph_arpack_options_init(&arpack_options);

    igraph_pagerank(
            &graph, IGRAPH_PAGERANK_ALGO_ARPACK,
            &result, &value,
            igraph_vss_all(), 0 /* directed=false */,
            0.85 /* damping */,
            NULL /* no weights */,
            &arpack_options);

    igraph_vector_destroy(&result);
    igraph_destroy(&graph);

    return 0;
}
```

szhorvat:
Maximum NCV should be N and not N-1, otherwise it may happen that NCV - NEV < 2, which causes ARPACK to fail.

Fixes #1119

I am unsure why the cap on NCV was N-1 previously. I am going by the error message issued by ARPACK, 

> NCV must be greater than NEV and less than or equal to N (and for the non-symmetric solver NCV-NEV >=2 must also hold)

### DIFF
--- a/src/arpack.c
+++ b/src/arpack.c
@@ -794,9 +794,9 @@ void igraph_i_arpack_auto_ncv(igraph_arpack_options_t* options) {
   if (options->ncv < min_ncv) {
     options->ncv = min_ncv;
   }
-  /* ...but at most n-1 */
-  if (options->ncv > options->n - 1) {
-    options->ncv = options->n - 1;
+  /* ...but at most n */
+  if (options->ncv > options->n) {
+    options->ncv = options->n;
   }
 }
 


ntamas:
Thanks!

Issue #1089, opened 2018-04-12
Wishlist: k_regular_game should sample uniformly


szhorvat:
Currently, `igraph_k_regular_game()` uses `igraph_degree_sequence_game()` with the `SIMPLE_NO_MULTIPLE` method. This method does not sample uniformly.

There should be a method implemented that samples uniformly.

The simplest way is to use the configuration model and reject non-simple graphs (#1072 is tracking this for `degree_sequence_game`). This method is, however, not efficient at all due to the potentially high likelihood of rejection.

Are there better methods?  References and implementations are most welcome (as I only have so much time).

One paper on the topic is:
 * [B. McKay, N. Wormald: Uniform Generation of Random Regular Graphs of Moderate Degree](http://users.cecs.anu.edu.au/~bdm/papers/RandRegGen.pdf).

A more immediate task is to update the documentation and comment on the uniformity of sampling.

szhorvat:
These seem promising:

 - Gao and Wormald, https://arxiv.org/abs/1511.01175 (peer review version here: https://doi.org/10.1137/15M1052779)
 - http://users.monash.edu/~gfarr/research/slides/Gao-dREG.pdf (presentation about the same)

szhorvat:
> A more immediate task is to update the documentation and comment on the uniformity of sampling.

The documentation update should be done before 0.8.0.

szhorvat:
I changed the milestone to 0.8 just so that we don' forget to update the docs. The new method absolutely should not be implemented for 0.8.

vtraag:
@szhorvat, you specifically mean the documentation for `igraph_k_regular_game`, right? I believe the documentation for `igraph_degree_sequence_game` already mentions the non-uniform sampling. Will you implement this?

szhorvat:
Yes.

ntamas:
Can we move this back to 0.9.0?

szhorvat:
Anything of this sort is still marked as 0.8 only by accident ...

Issue #1072, opened 2018-03-08
degree_sequence_game: what does SIMPLE_NO_MULTIPLE do?


szhorvat:
[The documentation](http://igraph.org/c/doc/igraph-Generators.html#igraph_degree_sequence_game) clearly explains what the `SIMPLE` method of `degree_sequence_game` does.  However, it does not explain the algorithm behind `SIMPLE_NO_MULTIPLE`.  My impression from reading the documentation was that `SIMPLE_NO_MULTIPLE` does the same thing as `SIMPLE`, except that when a loop or multi-edge is generated, it throws away the result and starts again from scratch.

> This method is similar to `IGRAPH_DEGSEQ_SIMPLE` but tries to avoid multiple and loop edges and restarts the generation from scratch if it gets stuck.

However, if it did this, then I believe it *would* sample uniformly, which it doesn't (as the documentation states, and as I found by experiment). Also, while I did not take the time to fully understand the code, it seems clear that [it won't throw away the result if it finds a loop or multi-edge](https://github.com/igraph/igraph/blob/bbb00935104bda9e76da8a288e4fc56379c7bc65/src/games.c#L996).

Thus my questions are:

 - What does `SIMPLE_NO_MULTIPLE` do, in simple terms?  This should be documented. Otherwise, it should be made clear that this method gives absolutely no guarantees and the user shouldn't care about the algorithm. This is already suggested, but I found it confusing ...

 - Why isn't there a method that does the same as `SIMPLE`, but keeps retrying until it gets a result that is a simple graph?  This would be useful because the sampling would be uniform (consider this a feature request).  While the user can filter non-simple results manually after they have been fully generated, a function that does this internally could in principle be faster.

ntamas:
Not sure if this is a simple explanation, but here it is. Basically, the `SIMPLE_NO_MULTIPLE` method works as follows:

1. Generate a long vector `v` where vertex _i_ appears as many times as its desired degree.
2. Shuffle the vector `v` and create an empty vector `w`.
3. Start creating edges between `v[0]` and `v[1]`, `v[2]` and `v[3], `v[4]` and `v[5]`, and so on. Whenever an edge would be created between vertices that are already connected, put the involved vertices aside in another vector `w`.
4. If there were no edges created in the previous step, it means that we've reached a dead end. Start again from scratch.
5. Otherwise, if `w` is empty, we are done.
6. Otherwise, let `v` equal to `w` and go back to step 2.

[...]

PR #1144, opened 2018-12-07
degree_sequence_game: add uniform sampling

### LINKED ISSUE #1091: igraph_set_contains is buggy
`igraph_set_contains` is broken when used on empty sets.

https://github.com/igraph/igraph/blob/02005680aebb4281f2964c6d2d8d98851fce63b9/src/igraph_set.c#L257

It has `right = igraph_set_size(set)-1;`. If the set is empty, then `right == -1`. Despite this, it uses `right` for indexing (out-of-bounds error).

In my situation this led to `igraph_set_contains` claiming that an empty set contains `0`. 

What is the status of the `igraph_set_t` functions?  Are they safe to use?  They seem to get very limited use at the moment.

szhorvat:
This PR is another one back from April that I did not send in. It adds uniform sampling for *simple* graphs to `degree_sequence_game` based on the configuration model. Please do review the code before merging.

Related: #1089 

Related: #1072 

It also fixes #1091


szhorvat:
Details:

Currently in `degree_sequence_game` we have:

`IGRAPH_DEGSEQ_SIMPLE`, which generates multigraphs.  Here, the documentation is improved, explaining that this is the configuration model, and providing specific details on sampling probabilities.

There is one point that I notice I did not write down: the probabilities are given in terms of the adjacency matrix. For undirected graphs, this is the kind of adjacency matrix that has a 2 (not a 1) on the diagonal for loop edges.  There are two conventions about loop edges and adjacency matrices: put a 1 in the adjacency matrix (because there is one loop edge) or put a 2 (because both ends of that loop edge connect to the relevant vertex). Maybe I should not this explicitly.

`IGRAPH_DEGSEQ_SIMPLE_NO_MULTIPLE`, which does not generate multigraphs, but does not sample uniformly either.

This PR adds another method, which does sample uniformly.

There is also `IGRAPH_DEGSEQ_VL`, but that is a very different method that generates only connected graphs, and does not sample *exactly* uniformly.

----

About the naming of enum value:

I find the current naming confusing because the method that does generate multigraphs is called "SIMPLE". I did not rename them because that would break compatibility, and it seemed like a minor thing.  Users of the C interface should read the documentation carefully. Renaming can be done in the high-level interfaces, if deemed necessary.

ntamas:
Please let me know whether you'd like to try implementing the check-for-loop-edges-first idea outlined above; if not, I might try it on my own after this one is merged.

szhorvat:
Sorry for the late response ... I will be away for a while, then work will be busy. So I might not work on this at all until late January.  Feel free to implement this!  I will let you know when I come back to this, so we don't duplicate work.

ntamas:
Okay, I've merged the PR for the time being and we can return to this later as time allows.

szhorvat:
> Please let me know whether you'd like to try implementing the check-for-loop-edges-first idea outlined above; if not, I might try it on my own after this one is merged.

If you have not yet done this, I'll do it around next weekend. Let me know.

szhorvat:
Note to self: 

Another improvement that could be made: if the graph is dense (> 50%), then generate the complement instead, then convert it back to the desired graph. The configuration model's algorithm will fail on even tiny almost-complete graphs.

Counter-argument for doing this: we may not be able to judge in the best way when it is worth complementing the sequence. Thus doing this is better left to the user than trying to make our algorithm too smart.

@ntamas  What do you think?

ntamas:
I would resist the tempation to guess; just document the caveat and then let the user try it if needed.

PR #1152, opened 2019-02-12
igraph_vector_init_copy: Add const to the source pointer

heavywatal:
The copy source should be passed to `igraph_vector_init_copy()` via `const` pointer.


ntamas:
Seems good, thank you!

Issue #1084, opened 2018-04-11
Wishlist: Find Prüfer sequence of trees


szhorvat:
After #1082 is fixed, a function should be added to find the Prüfer sequence of trees.

szhorvat:
PR is already merged. Closing.

PR #1154, opened 2019-02-17
Add function for converting a tree to its prufer sequence

frederik-h:
This is an implementation of the feature requested in #1084.

Regarding licensing: I'm fine with GPL 2 or later and FDL.

### DIFF
--- a/src/conversion.c
+++ b/src/conversion.c
@@ -844,3 +844,93 @@ int igraph_get_stochastic_sparsemat(const igraph_t *graph,
   
   return 0;
 }
+
+
+/**
+ * \ingroup conversion
+ * \function igraph_to_prufer
+ * \brief Converts a tree to its Pr&uuml;fer sequence
+ *
+ * A Pr&uuml;fer sequence is a unique sequence of integers associated
+ * with a labelled tree. A tree on n >= 2 vertices can be represented by a
+ * sequence of n-2 integers, each between 0 and n-1 (inclusive).
+ *
+ * \param graph Pointer to an initialized graph object which
+          must be a tree on n >= 2 vertices.
+ * \param prufer A pointer to the integer vector that should hold the Pr&uuml;fer sequence;
+                 the vector must be initialized and will be resized to n - 2.
+ * \return Error code:
+ *          \clist
+ *          \cli IGRAPH_ENOMEM
+ *             there is not enough memory to perform the operation.
+ *          \cli IGRAPH_EINVAL
+ *             the graph has less than 2 vertices.
+ *          \endclist
+ *
+ * \sa \ref igraph_tree(), \ref igraph_from_prufer()
+ *
+ */
+int igraph_to_prufer(const igraph_t *graph, igraph_vector_int_t* prufer) {
+  /* For generating the Prüfer sequence, we enumerate the vertices u of the tree.
+     We keep track of the degrees of all vertices, treating vertices
+     of degree 0 as removed. We maintain the invariant that all leafs
+     that are still contained in the tree are >= u.
+     If u is a leaf, we remove it and add its unique neighbor to the prüfer
+     sequence. If the removal of u turns the neighbor into a leaf which is < u,
+     we repeat the procedure for the new leaf and so on. */
+  igraph_integer_t u;
+  igraph_vector_t degrees, neighbors;
+  igraph_integer_t prufer_index = 0;
+  igraph_integer_t n = igraph_vcount(graph);
+
+  if(n < 2)
+    return IGRAPH_EINVAL;
+
+  IGRAPH_CHECK(igraph_vector_int_resize(prufer, n-2));
+  IGRAPH_CHECK(igraph_vector_init(&degrees, n));
+  IGRAPH_CHECK(igraph_vector_init(&neighbors, 1));
+
+  IGRAPH_FINALLY(&degrees, &igraph_vector_destroy);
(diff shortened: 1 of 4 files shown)


src/conversion.c
```diff
@@ -844,3 +844,93 @@ int igraph_get_stochastic_sparsemat(const igraph_t *graph,
...
+     sequence. If the removal of u turns the neighbor into a leaf which is < u,
+     we repeat the procedure for the new leaf and so on. */
+  igraph_integer_t u;
+  igraph_vector_t degrees, neighbors;
+  igraph_integer_t prufer_index = 0;
+  igraph_integer_t n = igraph_vcount(graph);
+
+  if(n < 2)
+    return IGRAPH_EINVAL;
+
```
frederik-h: Should we also check that the graph is a tree?

szhorvat: I think that would be useful.

src/conversion.c
```diff
@@ -844,3 +844,93 @@ int igraph_get_stochastic_sparsemat(const igraph_t *graph,
...
+     If u is a leaf, we remove it and add its unique neighbor to the prüfer
+     sequence. If the removal of u turns the neighbor into a leaf which is < u,
+     we repeat the procedure for the new leaf and so on. */
+  igraph_integer_t u;
+  igraph_vector_t degrees, neighbors;
+  igraph_integer_t prufer_index = 0;
+  igraph_integer_t n = igraph_vcount(graph);
+
+  if(n < 2)
+    return IGRAPH_EINVAL;
```
szhorvat: On error, do not simply return, use `IGRAPH_ERROR` instead and report a user-readable error. `IGRAPH_ERROR` also ensures that the error handler will be called and any allocated resources can be freed (should a future version of this function allocate resources before this line).

src/conversion.c
```diff
@@ -844,3 +844,93 @@ int igraph_get_stochastic_sparsemat(const igraph_t *graph,
...
+
+  IGRAPH_CHECK(igraph_vector_int_resize(prufer, n-2));
+  IGRAPH_CHECK(igraph_vector_init(&degrees, n));
+  IGRAPH_CHECK(igraph_vector_init(&neighbors, 1));
+
+  IGRAPH_FINALLY(&degrees, &igraph_vector_destroy);
```
szhorvat: The `IGRAPH_FINALLY` should follow the `igraph_vector_init` that it is paired with immediately. Consider that the *next* `igraph_vector_init` may fail too.

You can consider using `IGRAPH_VECTOR_INIT_FINALLY` to do both initialization and `IGRAPH_FINALLY` in a single line.

src/conversion.c
```diff
@@ -844,3 +844,93 @@ int igraph_get_stochastic_sparsemat(const igraph_t *graph,
...
+    while(degree == 1 && leaf <= u) {
+      igraph_integer_t i;
+      igraph_integer_t neighbor = 0;
+
+      VECTOR(degrees)[leaf] = 0; /* mark leaf v as deleted */
+
+      IGRAPH_CHECK(igraph_neighbors(graph, &neighbors, leaf, IGRAPH_ALL));
+
+      /* Find the unique remaining neighbor of the leaf */
+      for(i = 0; i < igraph_vector_size(&neighbors); i++) {
```
szhorvat: I would call `igraph_vector_size` *once* before the loop and store the vector size (as function calls can be slow when inlining is prevented, like in this case).

frederik-h:
> I am not a maintainer of igraph, just trying to help with the review. 
> I do suggest adding a check for whether the input graph is a tree. (...)
> I did test the code a bit by integrating it into the Mathematica interface.

Thanks! I have added the check.

szhorvat:
Thank you for the contribution. I wanted to do this for a while myself, but I just couldn't get around to it.

ntamas:
Thanks for the contribution guys. I'm on holiday in Cyprus till the end of the week, I'll try to get around to reviewing this next week when I am back.

gaborcsardi:
Thanks both!

Pascal-Ortiz:
> This is an implementation of the feature requested in #1084.
> 
Hi Frederik!

Good work! there are not so many libraries implementing Prufer conversions. 

Looking at the source code, I thought your algorithm has linear complexity. To clarify, I have done some testing and it appears that execution seems to be quadratic (if you double the data size then time execution is multiplied by a factor 4) and the implentation was not able to build in a reasonable time a tree having one million of nodes from a prufer sequence.

szhorvat:
@Pascal-Ortiz 

> and the implentation was not able to build in a reasonable time a tree having one million of nodes from a prufer sequence.

This PR was for a function that takes a tree and creates a Prüfer sequence. You seem to be talking about taking a Prüfer sequence and building a tree. Can you clarify which one you are referring to?

Pascal-Ortiz:
I'm talking about the `igraph_from_prufer` function.

Pascal-Ortiz:
> This PR was for a function that takes a tree and creates a Prüfer sequence.

Sorry for the confusion. PR #1079 is the appropriate one.

szhorvat:
@Pascal-Ortiz Can you please open a new issue and describe the problem in detail, along with a link to the linear-time one? (So that I wouldn't forget about this—I did a quick test and indeed it's much slower than it should be.)

szhorvat:
Also a note:

If you want to use this for generating random trees, there's a different function for that (`igraph_tree_game`) which has a method that will be (much) faster for large trees.

PR #1160, opened 2019-04-14
Fix bug in igraph_is_tree

szhorvat:
There was a bug in `igraph_is_tree`.  See here: https://github.com/szhorvat/IGraphM/issues/90

I contributed this code originally.

The `igraph_i_is_tree_visitor` function is effectively measuring the size of a connected component that contains `root`.  It worked like this: it pushed neighbours of the current version onto a stack. It popped off vertices from the stack and pushed their neighbours, etc.

A counter was incremented every time a vertex was popped. I did not think about the case when a vertex may be pushed more than once before it gets visited. This is exactly what happened here.

Now we only increment the counter if the popped vertex has no yet been visited.

### DIFF
--- a/examples/simple/igraph_is_tree.c
+++ b/examples/simple/igraph_is_tree.c
@@ -109,5 +109,17 @@ int main() {
 
     igraph_destroy(&g);
 
+    /* Regression test, see:
+     * https://github.com/szhorvat/IGraphM/issues/90
+     * https://github.com/igraph/igraph/pull/1160
+     */
+    igraph_small(&g, 5, 0,
+                 0,3, 0,4, 1,3, 1,4, -1);
+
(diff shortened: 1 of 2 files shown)


ntamas:
Seems good! Can you add a regression test to `examples/simple/igraph_is_tree.c` in a way that it fails without the patch but passes _with_ the patch? This should allow us to ensure that the bug does not resurface in later versions.

szhorvat:
Done.

I also added an `IGRAPH_LIKELY`.

`__builtin_expect` appears to work with clang as well, and the `IGRAPH_LIKELY` macro could be updated at one point. https://llvm.org/docs/BranchWeightMetadata.html  Not sure if it's worth the effort ... I won't do it now.

ntamas:
Thanks, it works now!

PR #1164, opened 2019-04-20
Lad crash fix

### LINKED ISSUE #927: LAD crashes with certain inputs
Here's the R version of the crash demo:

```
g<-make_lattice(c(4,8))
patt<-make_lattice(vcount(g))
subgraph_isomorphisms(patt,g,"lad")
```

This code attempts to find all Hamiltonian paths in a (4,8) grid graph.  For (4,7) it doesn't crash.  For (4,8) it does.

Mathematica version:

```
g = GridGraph[{4, 8}];
IGLADSubisomorphismCount[PathGraph@Range@VertexCount[g], g]
```

The crash is present in the current master branch version.

It would be reasonable to think that this is an out-of-memory crash.  But the (4,5), (4,6) and (4,7) grids have 2012, 7220 and 24020 Hamiltonian paths, respectively.  Thus it seems unlikely that the (4,8) one could run out of memory.

Do you have any hints on what may be going on before I test the original (unmodified) LAD code?

szhorvat:
Fixes #927 (finally!)

See there for comments.

The two other commits:
 - change all tabs to spaces (it was really annoying to have to deal with tab sizes while debugging this)
 - remove some unnecessary code which was also removed from the original LAD sources

### DIFF
--- a/src/lad.c
+++ b/src/lad.c
@@ -106,24 +106,24 @@ int igraph_i_lad_createGraph(const igraph_t *igraph, Tgraph* graph) {
   
   IGRAPH_VECTOR_INIT_FINALLY(&graph->nbSucc, no_of_nodes);
   IGRAPH_CHECK(igraph_degree(igraph, &graph->nbSucc, igraph_vss_all(), 
-			     IGRAPH_OUT, IGRAPH_LOOPS));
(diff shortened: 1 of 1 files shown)


ntamas:
Seems good to me, thanks!

Issue #1166, opened 2019-04-21
VF2 wrong result for graph with self-loop


szhorvat:
VF2 returns a wrong result for a graph with a self-loop.

```
> patt<-make_graph(c(1,1),dir=F)
> targ<-make_graph(c(1,2, 2,3), dir=F)
> subgraph_isomorphic(patt, targ, "vf2")
[1] TRUE
```

Need to clarify if VF2 is meant to support such graphs at all, and if not, add a check (just like for multigraphs).

Possible workaround: colour each node based on the number of self-loops it has and use the coloured graph isomorphism facilities.

szhorvat:
`igraph_subisomorphic` is currently just a wrapper around `igraph_subisomorphic_vf2`, thus it is also affected.

This is a more serious issue, and should be fixed.  A fix along the lines of PR #1168 should be possible.

ntamas:
Actually, I'm not sure how much effort one should invest in VF2 in general; we could just put a warning in front of the function call if the input graph has multiple edges or self-loops. The old VF2 site seems to propose [VF3](https://mivia.unisa.it/datasets/graph-database/vf3-library/) now, there's also [VF2++](https://web.cs.elte.hu/egres/tr/egres-18-03.pdf) and there could be other new algorithms out there that would serve the purpose better.

szhorvat:
Closing, as it is now documented that VF2 only supports simple graphs. General non-simple-graph isomorphism should go in another issue.

Issue #815, opened 2015-02-12
vf2 says two graphs are isomorphic when they are not


benruijl:
The following example shows two non-simple graphs which are not isomorphic, yet vf2 says they are.

```
c = Graph(n=10)
c.add_edges([(0, 1), (1, 2), (1, 2), (1, 3), (2, 3), (2, 9), (2, 4), (3, 4), (3, 4)])
h = Graph(n=10)
h.add_edges([(0, 1), (1, 3), (1, 4), (1, 6), (1, 6), (3, 4), (3, 4), (3, 6), (4, 9)])
m = c.isomorphic_vf2(h, None, None, None, None, return_mapping_12=True)
print(m)
```

ntamas:
I strongly suspect that VF2 does not support non-simple graphs while BLISS does, although I cannot confirm it because I did not implement either of them in igraph. (Maybe @gaborcsardi can chime in). But anyway, it's definitely a bug because `isomorphic_vf2` should throw an exception if it does not support non-simple graphs and you give it one.

ntamas:
Just checked the VF2 code and indeed it seems to simplify the graph first; see the code around https://github.com/igraph/igraph/blob/master/src/topology.c#L1178 . Not sure whether this is an inherent limitation of the algorithm, though.

benruijl:
Thanks. I don't think it is a limitation of the algorithm, since boost seems to support non-simple graphs with vf2: http://www.boost.org/doc/libs/1_57_0/libs/graph/doc/vf2_sub_graph_iso.html

gaborcsardi:
I am fairly sure that the original VF2 did not handle multi-graphs. Anyway, one simple way to handle them is to use the multiplicity of the edges as labels or colors that are required to match. Or, to label the multi-edges individually and require them to match. This is somewhat less efficient, and seemingly this is what boost is doing.

A quick "fix" from our part would be to warn if VF2 is called with a multigraph. I guess that's not quite a solution, but a proper fix is slightly more difficult and I don't want to get into it right now.

szhorvat:
VF2 not supporting non-simple graphs is now documented.

`igraph_simplify_and_colorize` makes it easy to implement a workaround.

Issue #1097, opened 2018-04-26
igraph_tree example/test is empty


szhorvat:
This is empty:

https://github.com/igraph/igraph/blob/master/examples/simple/igraph_tree.c

Did it get lost on the way or was it never written?

ntamas:
It was never written (although it should have been):

https://github.com/igraph/igraph/commits/master/examples/simple/igraph_tree.c

szhorvat:
OK, I'll handle this one.

PR #1168, opened 2019-04-21
igraph_colored_simple_graph

szhorvat:
Currently, no included isomorphism algorithm supports multigraphs.

VF2 does not seem to support either self loops (#1166, please verify) or multigraphs (#815)

We can still theoretically deal with multigraphs like so:

 - simplify the graph (no self-loops / multi-edges)
 - colour edges based on their original multiplicity (simple edges get 1, edges merged from two parallel ones get 2, etc.)
 - colour vertices based on how many incident self-loops they had

Now we can use coloured-graph isomorphism (using VF2) to handle the multigraph.

This PR adds a function to do this graph simplification while counting self-loops and multi-edges.

It also rearranges the Isomorphism documentation sections a bit.

Please verify before merging!  It closely follows `igraph_simplify` in its implementation.

Question: In undirected graphs, do edges always appear sorted, i.e. always 1-2 and not 2-1?  `igraph_simplify` seems to rely on this where it checks `to == pto && from == pfrom` but not `to == pfrom && from == pto`.  I did the same in this function

### DIFF
--- a/doc/isomorphism.xxml
+++ b/doc/isomorphism.xxml
@@ -49,7 +49,7 @@
 
 <section><title>Utility functions</title>
 <!-- doxrox-include igraph_permute_vertices -->
-<!-- doxrox-include igraph_colored_simple_graph -->
+<!-- doxrox-include igraph_simplify_and_colorize -->
 </section>
 
 </chapter>
--- a/doc/isomorphism.xxml
+++ b/doc/isomorphism.xxml
@@ -9,7 +9,6 @@
 
 <section><title>The simple interface</title>
 <!-- doxrox-include about_graph_isomorphism -->
-<!-- doxrox-include igraph_permute_vertices -->
 <!-- doxrox-include igraph_isomorphic -->
 <!-- doxrox-include igraph_subisomorphic -->
 </section>
@@ -46,6 +45,11 @@
 <!-- doxrox-include igraph_isoclass -->
 <!-- doxrox-include igraph_isoclass_subgraph -->
 <!-- doxrox-include igraph_isoclass_create -->
+</section>
+
+<section><title>Utility functions</title>
+<!-- doxrox-include igraph_permute_vertices -->
+<!-- doxrox-include igraph_colored_simple_graph -->
 </section>
 
 </chapter>
--- a/examples/tests/simplify_and_colorize.c
+++ b/examples/tests/simplify_and_colorize.c
@@ -1,0 +1,58 @@
+
+#include <igraph.h>
+
+#include "test_utilities.inc"
+
+#define SIMPLIFY_PRINT_DESTROY(name) \
+    printf(name "\n"); \
+    igraph_simplify_and_colorize(&graph, &res, &vcol, &ecol); \
+    print_graph(&res, stdout); \
+    print_vector_int(&vcol, stdout); \
+    print_vector_int(&ecol, stdout); \
+    printf("\n"); \
+    igraph_destroy(&graph);
+
+int main() {
+    igraph_t graph, res;
+    igraph_vector_int_t vcol, ecol;
+
+    igraph_vector_int_init(&vcol, 0);
+    igraph_vector_int_init(&ecol, 0);
+
+    /* null graph */
+    igraph_empty(&graph, 0, 0);
+    SIMPLIFY_PRINT_DESTROY("K0");
+
+    /* singleton graph */
+    igraph_empty(&graph, 1, 0);
+    SIMPLIFY_PRINT_DESTROY("K1");
+
+    /* 4-cycle-graph */   
+    igraph_ring(&graph, 4, 0, 0, 1);
+    SIMPLIFY_PRINT_DESTROY("C4");
+
+    /* both multi-edges and self loops */
+    igraph_small(&graph, 2, 0,
+                 0,1, 0,1, 1,1, -1);
+    SIMPLIFY_PRINT_DESTROY("Undirected graph 1");
+
+    /* parallel edges specified with different vertex orderings */   
+    igraph_small(&graph, 3, 0,
+                 0,1, 1,2, 2,0, 2,2, 2,2, 2,1, -1);
+    SIMPLIFY_PRINT_DESTROY("Undirected graph 2");
+
+    /* directed version of the same as above */
+    igraph_small(&graph, 3, 1,
+                 0,1, 1,2, 2,0, 2,2, 2,2, 2,1, -1);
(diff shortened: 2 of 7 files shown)


szhorvat:
This should probably be renamed to something better: any suggestions?

I implemented multigraph subgraph matching in IGraph/M based on this. It would be nice to include that in the igraph C core eventually. It's not as trivial as using colours because a smaller multiplicity edge in the pattern graph should match the higher multiplicity edge in the target graph.  Thus we need a custom edge comparator function, which is not available in the high-level interfaces. Even if it were, constant callbacks to the high-level interface would kill performance.

szhorvat:
Related (Tamás's answer): 

 - https://stackoverflow.com/questions/33980048/checking-for-multigraph-isomorphism-in-python-igraph

ntamas:
>  In undirected graphs, do edges always appear sorted, i.e. always 1-2 and not 2-1?

Right now this is indeed the case.

> This should probably be renamed to something better: any suggestions?

How about `igraph_simplify_and_colorize()`?

Otherwise the implementation seems okay to be, but as always, there should be some tests in `examples/simple` that verify the implementation. The ones for `igraph_simplify()` could probably be re-used.

szhorvat:
Since the motivation for adding this function is to handle isomorphism for non-simple graphs, shall we do that at the same time, and merge everything together? Some issues might appear along the way.

If yes, would it be better to extend `igraph_isomorphic` and `igraph_subisomorphic` to handle multigraphs, or create entirely new functions?  I think that extending them are a good idea since these are "generic" functions that dispatch to the "best" implementation for each specific case.

`igraph_isomorphic` and `igraph_subisomorphic` only return a yes/no answer but they do not return an actual mapping. Eventually, it may be a good idea to also add generic functions which return one or more mapping. In particular for the subgraph case, one would typically want to know _where_ a specific subgraph appears.  It's not clear to me what's the best approach here: generic function that returns _one_ mapping, generic function that returns multiple mappings (if requested), or specific functions (i.e. `vf2` in the name) that return multiple mappings.  The main motivation for adding such a function is that it is currently not easily implementable for subgraphs/multigraphs in the high-level interfaces.

For the subgraph case, in principle this is implementable both for VF2 and LAD. I would not bother with LAD for now because later versions of LAD handle colours differently than LADv1 (which is in igraph). If LAD ever gets updated, this would have to change as well.

The state-of-the-art algorithms that could be used for this are: [RI](https://github.com/InfOmics/RI), [VF3](https://github.com/vincarlet/vf3lib/) and [LADv3](https://perso.liris.cnrs.fr/christine.solnon/LAD).  

I experimented a bit with VF3 and the naive integration is often slower than the existing VF2, so I put it aside for now. I suspect that this either has to do with converting the graph format (I did this for `vf3lib` but igraph's VF2 is well-integrated and _maybe_ it doesn't do a conversion), or more likely with the pre-processing step (it's not clear to me when that's actually necessary).

szhorvat:
Should the `res` argument come before or after the `graph` argument? igraph is not entirely consistent in this and has functions of both kinds.

szhorvat:
>> In undirected graphs, do edges always appear sorted, i.e. always 1-2 and not 2-1?
>
> Right now this is indeed the case.

This should be documented as an invariant, just in case someone contributes a function that messed with the internal graph structure directly ...

On a related note, how do I build the documentation so that it includes `\ingroup internal` things, such as `igraph_t` itself?

ntamas:
> If yes, would it be better to extend igraph_isomorphic and igraph_subisomorphic to handle multigraphs, or create entirely new functions? I think that extending them is a good idea since these are "generic" functions that dispatch to the "best" implementation for each specific case.

I would vote for extending `igraph_isomorphic` and `igraph_subisomorphic`.

> Eventually, it may be a good idea to also add generic functions which return one or more mappings.

Yes, indeed.

> Should the `res` argument come before or after the `graph` argument? igraph is not entirely consistent in this and has functions of both kinds.

I know that we haven't been consistent with the API in several cases. Right now I would say that I would follow the convention that the first argument is always the "this" argument (i.e. what would be "this" if we were programming in C++), just like the explicit `self` argument in Python. The second one could be the output argument (if there is one), and the input arguments could come after that.

ntamas:
> On a related note, how do I build the documentation so that it includes \ingroup internal things, such as igraph_t itself?

Sorry, I have no idea :( The whole documentation build system was brewed in-house and I don't know exactly how it works by now. I know that a long while ago we started with Doxygen, but then it turned out that it did not really satisfy our needs (we wanted to generate documentation in HTML, PDF and INFO formats), so we wrote a quick regexp-based converter (`doc/doxrox.py`) that parses the C source code and converts everything into Docbook format (this is done by the rules defined in `doc/c-docbook.re`). Some of the Doxygen-related tags (such as `\ingroup`) were not needed in this new setup so we simply instructed `doxrox.py` to ignore these tags. I don't think that the `\ingroup` tags are actually used for anything by now - the documentation's XXML files basically specify explicitly what needs to be included where.

szhorvat:
> > Should the `res` argument come before or after the `graph` argument? igraph is not entirely consistent in this and has functions of both kinds.
> 
> I know that we haven't been consistent with the API in several cases. Right now I would say that I would follow the convention that the first argument is always the "this" argument (i.e. what would be "this" if we were programming in C++), just like the explicit `self` argument in Python. The second one could be the output argument (if there is one), and the input arguments could come after that.

That's how I think of it too, but then the question can be rephrased:

Is this function a constructor (that takes a multigraph as input)? Or a method of `igraph_t` that outputs another graph?

It makes sense to say that it's a method that outputs a graph because it _also_ outputs the colours (multiplicities).  Then we keep it as it currently it (input graph is the first, output graph is the second).

ntamas:
In the face of ambiguity, I'd say let's keep stuff as it is now :)

szhorvat:
I forgot about attribute combiners, but given that this is specifically meant as a helper function to deal with multigraph isomorphism, I think it's fine not to include that. Agreed?

ntamas:
Yes.

szhorvat:
Added tests. Also added a file with utility functions for testing.  Once merged, I'll send a PR which adds tests for `igraph_tree` (see #1097). I have them but they use the new helper functions.

szhorvat:
An explanation for why the vector printing and the graph printing are so fancy in the helper functions:

This is so that it can be clear when a vector is empty or a graph has no edges. It would be confusing if a test failed because the number of empty lines didn't match with the expected output.

ntamas:
Seems good to me!

Issue #548, opened 2013-11-26
Fix C test cases that fail on some platforms


gaborcsardi:
A bunch of tests are skipped non-OSX platforms, using code like

```
#ifdef __APPLE__
  return 0;
#else
  return 77;
#endif
```

These should be fixed to be platform independent, as they indicate that igraph itself is platform-dependent, even when using the same random seed.

szhorvat:
The following files have this in `examples/simple`:


```
igraph_community_infomap.c
igraph_community_multilevel.c
igraph_hrg2.c
igraph_hrg3.c
igraph_layout_merge2.c
igraph_scg_grouping3.c
igraph_scg_grouping4.c
igraph_scg_semiprojectors2.c
igraph_scg_semiprojectors3.c
scg2.c
scg3.c
```

szhorvat:
A return code of 77 signals a skipped test.

From here:

https://www.gnu.org/software/automake/manual/html_node/Scripts_002dbased-Testsuites.html

> When no test protocol is in use, an exit status of 0 from a test script will denote a success, an exit status of 77 a skipped test, an exit status of 99 an hard error, and any other exit status will denote a failure.

PR #1184, opened 2019-05-16
Reenable tests

szhorvat:
This PR does two things:

 - Adds a dead simple test to make sure that the default RNG produces the same sequence of numbers on all platforms. There are other tests which would fail if this were not the case, but I wanted a test that does this and nothing else. As discussed, this went into `examples/tests` instead of `examples/simple`. The idea is that `examples/simple` are supposed to be educational (and some of them included in the docs), while `examples/tests` have one purpose: testing.

 - Don't disable tests on non-Mac platforms, see #548

I ran the tests on Mac and x86_64 Linux only so far, but I do not expect problems on MinGW. I suggest merging, and if something breaks on Windows, consider that a separate issue.


ntamas:
Let's see how it goes. Thanks!

Issue #1050, opened 2017-12-31
igraph_betweenness_estimate() returns all-zero result with cutoff=0


szhorvat:
The documentation of `igraph_betweenness_estimate()` says,

```
 * \param cutoff The maximal length of paths that will be considered.
 *        If zero or negative, the exact betweenness will be calculated
 *        (no upper limit on path lengths).
```

However, with `cutoff=0`, it returns an all-zero vector.  This is true both for weighted and unweighted calculations.

Negative cutoff values do calculate the exact betweenness, as they should.

`igraph_edge_betweenness_estimate()` suffers from the same problem, but only for weighted calculations.

This is a report for the C core (latest git), but here's a demo with R:

```
> g<-erdos.renyi.game(10,20,'gnm')
> betweenness.estimate(g,cutoff=0)
 [1] 0 0 0 0 0 0 0 0 0 0
> betweenness.estimate(g,cutoff=-1)
 [1] 6.0833333 2.9166667 1.2500000 1.9166667 0.7916667 1.9166667 3.2500000 6.8333333 1.2500000
[10] 0.7916667
```

szhorvat:
One of the things I would like to know is if this will be fixed by changing the function to return exact results for `cutoff=0` or changing the documentation to require `cutoff < 0` instead.

ntamas:
`cutoff=0` does not really make sense to me (I mean, why would anyone run the function with `cutoff=0`?), so I think it's okay if we fix it by bending the implementation to the documentation (and not vice versa). Any objections?

szhorvat:
Fixed by #1187

Issue #56, opened 2013-05-04
shortest path outputs not properly


gaborcsardi:
Hi, 

sp function outputs paths of node 3 - 1- 7 with weight of 16
but
path 3 - 4 - 7 has only  weight of 5

is there any problem in my code ? 

-------------code-------------------------------------------------------------------------------------------------------------

library(igraph)
# edge list

d <- data.frame(matrix(c( "2","1", "1","2", "2","0", "0","2", "3","2", "2","3", "3","1", "1","3", "4","3", "3","4", "4","0", "0","4", "5","0", "0","5", "5","3", "3","5", "6","5", "5","6", "6","1", "1","6", "7","1", "1","7", "7","4", "4","7", "8","7", "7","8", "8","1", "1","8", "9","1", "1","9", "9","2", "2","9", "0","7", "7","0", "0","1", "1","0", "1","5", "5","1"),nc=2,byrow=TRUE),stringsAsFactors=FALSE)
colnames(d) <- c("from","to")
# node list

vers <- data.frame(c("0","1","2","3","4","5","6","7","8","9"),stringsAsFactors=FALSE)
# create graph

g <- graph.data.frame(d,directed=TRUE,vertices=vers)
# weight

w1 <- c(3, 1, 8, 6, 2, 7, 8, 3, 3, 3, 7, 8, 1, 4, 3, 3, 1, 4, 8, 3, 1, 8, 6, 2, 7, 8, 3, 3, 3, 7, 8, 1, 4, 3, 3, 1, 4, 8)
E(g)$weight <- w1
# shortest path

sp <- get.shortest.paths(g,from=3,to=7)
sp
 c(c(d[[1]][7], d[[2]][7]),c(d[[1]][22],c(d[[2]][22])))
 E(g)$weight[[7]]+E(g)$weight[[22]]
 c(c(d[[1]][10], d[[2]][10]),c(d[[1]][24],c(d[[2]][24])))
 E(g)$weight[[10]]+E(g)$weight[[24]]

---

Imported from Launchpad using lp2gh.
- date created: 2009-11-30T01:15:29Z
- owner: hashimoto-ie
- assignee: gabor.csardi
- the launchpad url was https://bugs.launchpad.net/bugs/490166

PR #1187, opened 2019-05-21
Start working on betweenness cutoff=0 #1050

iosonofabio:
Changing an `>= 0` into `> 0` should work, as the doc says. Of course, as already noted in #1050, there's no good reason to set `cutoff = 0` except to get exact betweenness.

### DIFF
--- a/examples/simple/igraph_betweenness.c
+++ b/examples/simple/igraph_betweenness.c
@@ -135,6 +135,67 @@ int main() {
   igraph_vector_destroy(&bet);
   igraph_destroy(&g);
 
+
+  /* test corner case of cutoff = 0 */
+  igraph_tree(&g, 20, 3, IGRAPH_TREE_UNDIRECTED);
+
+  /* unweighted */
+  igraph_vector_init(&bet, 0);
+  igraph_betweenness_estimate(/* graph=     */ &g,
+			      /* res=       */ &bet,
+			      /* vids=      */ igraph_vss_all(),
+			      /* directed = */ 0,
+			      /* cutoff=    */ 0,
+			      /* weights=   */ 0, 
+			      /* nobigint=  */ 1);
+
+  igraph_vector_init(&bet2, 0);
+  igraph_betweenness_estimate(/* graph=     */ &g,
+			      /* res=       */ &bet2,
+			      /* vids=      */ igraph_vss_all(),
+			      /* directed = */ 0,
+			      /* cutoff=    */ -1,
+			      /* weights=   */ 0, 
+			      /* nobigint=  */ 1);
+
+  if (!igraph_vector_all_e(&bet, &bet2)) {
+    return 1;
+  }
+
+  igraph_vector_destroy(&bet);
+  igraph_vector_destroy(&bet2);
+
+  /* weighted */
+  igraph_vector_init(&weights, igraph_ecount(&g));
+  igraph_vector_fill(&weights, 2.0);
+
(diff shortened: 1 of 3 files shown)


iosonofabio:
I think this is done. I understand this is planned for 0.9 but since it's done we could include it already

szhorvat:
Please also change the weighted case, as well as the edge betweenness case (only the weighted calculation is affected for that one).

Have you tested the change already?

iosonofabio:
Thank you @szhorvat, I can write a test for CI.

Are those other things mentioned in other issues on GitHub? If so it'd be great to collect them all here so I can understand the situation better and close them at once.

szhorvat:
I mentioned them in #1050:

 - unweighted and weighted vertex betweenness is affected
 - weighted edge betweenness is affected
 - unweighted edge betweenness is not affected

The same fix you applied should work for the missing two as well.

szhorvat:
I just verified that closeness calculation is not affected (but found another problem there ... will open an issue).

iosonofabio:
You mean issue #56?

szhorvat:
Sorry, I'm not sure I understand. Are you referring to the closeness issue I mentioned? It's #1197

iosonofabio:
My fault, wrong link 😋

The closeness is a separate regression bug, the weighted version I'm not sure but I don't have my computer now, let me check and get back to you 😊

szhorvat:
Yes, the closeness is independent and unrelated. I simply confirmed that there is no issue with the cutoff parameter and closeness calculations. The only problems are with 3 varieties of betweenness, see my message above. https://github.com/igraph/igraph/pull/1187#issuecomment-496248605

iosonofabio:
Good, I fixed the two remaining betwennness measures.

Anything else I should do about this?

Shall we just wait until the checks pass on CI or shall I make a test against this corner case?

szhorvat:
In my personal experience, it is very important to test (= try out) every change one makes.  I do not mean adding a test case, but did you actually try it out? Often the easiest way is to build a high-level interface with the change and try it from that.

iosonofabio:
Agreed. Then I would need to compile python-igraph against the locally built C core and then write a short Python script against that I guess. Is there any info on how to do that?

iosonofabio:
Ok I added tests for the vertex and edge betweenness, both for weighted and unweighted. They work fine, so the PR is good to go.

ntamas:
> I would need to compile python-igraph against the locally built C core

Personally, I do it like this:

1. Install `pkg-config` and add `$HOME/lib/pkgconfig` to the environment variable named `PKG_CONFIG_PATH`. (I have this in my shell profile).
2. Add `$HOME/lib` to `LD_LIBRARY_PATH` (or to `DYLD_LIBRARY_PATH` on macOS). This is to ensure that libraries installed in `$HOME/lib` are picked up by the linker.
3. Build and install igraph locally with `mkdir build && cd build && ../configure --prefix=$HOME && make && make install`. This installs igraph in `$HOME/include/igraph/...` and `$HOME/lib`, and also puts a file in `$HOME/lib/pkgconfig` that helps `pkg-config` tell what the igraph compiler and linker flags are. (You can validate this with `pkg-config --cflags --libs igraph`).
4. Check out `python-igraph`'s source code somewhere and then run `python setup.py install`. (Preferably in a virtualenv to avoid messing around with the system Python).

Let me know if you need more help with this.

iosonofabio:
Thank you, we should write this up in some docs...

Issue #969, opened 2016-11-02
Clear instructions for compiling on Windows


szhorvat:
I think there is a need for clear, step-by-step instructions for compiling igraph-c on Windows. I propose a [wiki article](https://github.com/igraph/igraph/wiki) because that will hopefully be easier to keep up to date than a file in the git repo.  This should be specifically for compiling the development version.

There should be separate instructions for msys/MinGW and for different versions of MSVC.  Perhaps @janschulz can help with this.

I can contribute to the msys2/mingw guide, as that's how I compile igraph on Windows now.

Currently I find it quite painful to compile igraph on Windows.  I never managed to compile it with MSVC, and when I go back to try it with msys2 and MinGW I often find that something got broken (perhaps due to an msys2 update) and have to spend a lot of time to figure out how to get it working again.  Having a community guide to which everyone can contribute would save time and effort for all of us.

janschulz:
IMO the best way to document the build of it is by building a appveyor config and then keep tat green: https://github.com/igraph/igraph/blob/master/appveyor.yml

E.g. in this case: simply add another build configuration for mingw alone.

janschulz:
Ok, I just saw that igraph already has a mys2 build (and the missleaing comment should eb removed :-) ):

```
   - TARGET_ARCH: "x64"
      PYTHON_VERSION: "NONE"
      CONDA_INSTALL_LOCN: "C:\\Miniconda-x64"
      MSYSTEM: MINGW64
      PATH: C:\msys64\usr\bin;C:\msys64\mingw64\bin;C:\Windows\System32;C:\Windows;%PATH%
[...]
install:
[...]
    - cmd: if %PYTHON_VERSION%==NONE  make
[...]
test_script:
[...]
    - cmd: if [%PYTHON_VERSION%]==[NONE] make check
```

Compiles with mys2 alone, not the msvc target and does the make test

janschulz:
Unfortunatelly, it doesn't comüile right now due to the mentioned changes in msys2: 
https://github.com/igraph/igraph/commit/49c2429a02e4dfb3045d09461e3281117e45fa9b#commitcomment-19609938
https://ci.appveyor.com/project/ntamas/igraph/build/1.0.64

szhorvat:
[...]

PR #1189, opened 2019-05-21
Compilation instructions Windows

vtraag:
Updated the instructions for compilation on Windows slightly. This essentially reflects what @jankatins wrote in https://github.com/igraph/igraph/issues/969#issuecomment-258247860.

Removed the instructions for compiling `python-igraph`. Those instructions should probably be better included in that repository (and will need to be updated anyway). 

This (at least partially) addresses #969. You indicated you had problems @szhorvat compiling `igraph` using MSVC on Windows. Do you have specific issues you run into that aren't in these (admittedly succinct) instructions?


szhorvat:
I'll give it a go on the weekend.

ntamas:
I believe this is now superseded by #1279, right?

vtraag:
Yes, indeed, it can be closed!

Issue #899, opened 2015-11-26
igraph_bipartite_game_gnm overflow problem


ntamas:
`igraph_bipartite_game_gnm` cannot generate a bipartite graph with 2 x 1 million vertices and 8 million edges as the part that checks whether the edge count is feasible tries to multiply 2 times 1 million times 1 million so it overflows.

iosonofabio:
closed via #1190 1190

PR #1190, opened 2019-05-21
Fix for #899: Prevent overflow of maxedges by early conversion to double

iosonofabio:
Try and convert `n1` and `n2` to double before multiplying to avoid overflowing integers


ntamas:
Seems good to me!

Issue #1056, opened 2018-01-27
error reading boolean attributes in igraph_cattribute_GAB


robgjansen:
Make sure that these boxes are checked before submitting your issue -- thank you!

- [x] This issue is for the C core of igraph.
- [x] This issue is a bug report or a feature request, not a support question.

I believe that there is an error reading boolean graph attributes in `igraph_cattribute_GAB`. I'm not sure if this also affects boolean vertex and edge attributes.

Reproduce the error as follows. First, create a simple test graphml file using python:
```
import networkx as nx
from cStringIO import StringIO

G = nx.Graph(mybool=True)

s = StringIO()
nx.write_graphml(G, s)

with open('mybool.graphml.xml', 'w') as outf:
    print >>outf, s.getvalue(),

```

the output file produced by the above script is:
```
<?xml version='1.0' encoding='utf-8'?>
<graphml xmlns="http://graphml.graphdrawing.org/xmlns" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://graphml.graphdrawing.org/xmlns http://graphml.graphdrawing.org/xmlns/1.0/graphml.xsd">
  <key attr.name="mybool" attr.type="boolean" for="graph" id="d0" />
  <graph edgedefault="undirected">
    <data key="d0">True</data>
  </graph>
</graphml>
```

Then load the graph using the c core, and try to print the boolean attribute value.
```
// file: testbool.c
// build like: gcc -g -o testbool testbool.c -ligraph

#include <stdio.h>
#include <igraph/igraph.h>

#define FILENAME "mybool.graphml.xml"

igraph_t graph;

int main() {
    igraph_attribute_table_t* oldHandler = igraph_i_set_attribute_table(&igraph_cattribute_table);

    FILE* graphFile = fopen(FILENAME, "r");
    igraph_read_graph_graphml(&graph, graphFile, 0);
    fclose(graphFile);

    if (igraph_cattribute_has_attr(&graph, IGRAPH_ATTRIBUTE_GRAPH, "mybool")) {
        igraph_bool_t value = igraph_cattribute_GAB(&graph, "mybool");
        printf("found boolean value %d\n", value);
    }

    igraph_destroy(&graph);
    return 0;
}
```

The above can be built with `gcc -g -o testbool testbool.c -ligraph`. Running in GDB produces the following output and backtrace:
```
$ gdb testbool
[...]

PR #1191, opened 2019-05-21
Check for #1056: Test for boolean graph attribute

iosonofabio:
Added two files inside `examples/simple' for now:
- a C file with the example code
- a graphml file with the example graph

Not sure what else needs to be done to include this in the test suite. Also, do we need an XML file in the same folder?


ntamas:
Thanks!

PR #1192, opened 2019-05-22
modified igraph_clique_number to use the new maximal clique implementation

vtraag:
This continues PR #824 for which the original fork was removed, so we cannot continue to work on it. In this PR we can continue the work. This should address issue #804.

At the moment further work is required, at the moment I only integrated PR #824 with the current master, and corrected something based on the earlier comments by @szhorvat.


szhorvat:
Just a warning, so we don't duplicate work (also mentioned in the chatroom). I have existing working meant to replace PR #824 , but not complete. I was meaning to send the completed parts soon. We should coordinate on this.

vtraag:
OK, good. I'll let you submit your work first then and won't touch it further. Feel free to either use your fork or merge your changes into this PR.

szhorvat:
Was it ever verified that the new maximal clique finder performs better even for finding maximal independent vertex sets than the currently used independent vertex set finder? This is the premise of #824.

Equivalently, was it verified that this new max clique finder is faster even for dense graphs?

@gaborcsardi, since you implemented these, do you have any comment on this?

gaborcsardi:
sry, I don't remember. I would think that it is much faster at least for sparse graphs.

szhorvat:
I'll do some benchmarks before swapping out the independent vertex set finder (which is effectively the dense graph case).

ntamas:
I doubt that the independent vertex set based finder would perform better. The key problem was that in order to find maximal cliques in a graph using independent vertex sets, you need to construct the complement graph first, which is prohibitively expensive (both memory- and time-wise) for large sparse networks as the complement would be an almost full graph. (I'm talking about the scale of protein-protein interaction networks here with vertices in the order of thousands).

szhorvat:
@ntamas Sorry, I misremembered what #824 was about. No problems with the proposals in that PR. I thought it also proposed transitioning the independent vertex set functions to use the new clique finders. But it did not.

As you say, computing the complement is not nice ... would take up a lot of memory. Still, _eventually_ it will be good to investigate if that could help.  We now have fast clique finding implementations for both _all cliques_ (Cliquer) and _maximal cliques_ (the subject of this PR). Maybe they can be modified to work without having to store the complement graph. (I _think_—not sure— that Cliquer stores the graph as a bit-based adjacency matrix so it wouldn't be a problem to do the complement.)

Anyway, the independent vertex set thing is a project for much later.

I'll send a PR with the few changes I already have towards this before 0.8 and the rest can go into 0.9.

szhorvat:
@ntamas Before I send that PR, there's another thing to discuss.

I have a version that can find a limited number of cliques. Just like the VF2 functions, it take a "clique handler function" as the argument. This function can signal to stop the search.

In order to stop the search, it uses a special, undocumented return value.

https://github.com/szhorvat/igraph/blob/IGraphM-040/src/maximal_cliques_template.h#L118

https://github.com/szhorvat/igraph/blob/IGraphM-040/include/igraph_error.h#L393

I am not completely comfortable with "abusing" error codes like this ... but I don't have a much better _simple_ solution. I think you suggested this method originally, but I can't find the discussion.

What do you think about this? Shall we go on with this method and add the `IGRAPH_SEARCH_STOPPED` return code / "error code"?

ntamas:
I think it's fine; I would probably use a shorter name like `IGRAPH_STOP`, but that's just a matter of taste.

One thing though: if you extend the error enum, please also add a human-readable description of the error in `igraph_i_error_strings[]` in `src/igraph_error.c`. I think we also forgot to do this for the `IGRAPH_ERWSTUCK` error code. `igraph_strerror()` could return an invalid pointer if we don't do this.

szhorvat:
@ntamas I'll add the missing error strings. This "error code" will be a bit special because it should never propagate to the top and should never be visible to the user. I'll note this in the error string. If anyone sees this message, it indicates a bug, and they should know that it needs to be reported.

ntamas:
Yes, that’s exactly how I was thinking about it.

T.

PR #1194, opened 2019-05-22
Removed membership argument of modularity_matrix.

### LINKED ISSUE #1035: modularity_matrix() asks for membership vector but never uses it(?)
The function `igraph_modularity_matrix()` demands a vector of memberships be specified. It is not clear why; the definition of the `modularity_matrix()` is given in igraph's R documentation as:

> M[i,j] = A[i,j] – d[i]d[j] / (2m)
> where A[i,j] is the (possibly weighted) adjacency matrix, d[i] is the degree of vertex i and m is the number of edges (or the total weights in the graph, if it is weighed *[sic]*

This definition does not use community memberships, because it corresponds to the *initial* modularity matrix, at the start of most modularity optimisation algorithms when every node is in its own community.
I tried running `modularity_matrix` (via R) with different membership vectors and always get the same result.

It seems to me that the documentation is unclear or the `membership` argument is unnecessary.

I'm not much of a C programmer, but in the source code (lines 960–1001 of `community.c`) it looks like the only time `membership` is used when throwing an error to say the user hasn't supplied it. Maybe I am missing something, however?

vtraag:
Fixes #1035.


vtraag:
Ah yeah, of course, sorry. Will do so Friday (won't have time tomorrow).

Issue #1145, opened 2018-12-23
cited_type_game does not check its arguments


szhorvat:
`igraph_cited_type_game` does not check that its arguments are valid.

Unless I misunderstood something, the following fixes are needed:

 - Check that the length of `types` is `nodes`
 - Check if the node count is zero and if yes, return an empty graph right away (do not index into `types`!)
 - Before indexing into `prefs`, check that the index is within the bounds (`prefs` may not be long enough to have values for all types, or a type value may be invalid).
 - Before using any values from `prefs`, check that it is not negative
 - Before `vector_binsearch` check that `cumsum` is not all-zero (i.e. that `sum` is not zero)

ntamas:
The above is probably correct. Note that similar problems are probably also present in `igraph_citing_cited_type_game()`.

szhorvat:
Yes, they are (I checked since opening this issue).

iosonofabio:
I think I have added all these fixes to both functions, anyone can have a look?

ntamas:
Guess I can close this as the PR is merged, right? Please reopen if needed.

PR #1198, opened 2019-06-07
First fixes to #1145, citation games

iosonofabio:
DO NOT MERGE YET!

Started implemented suggested fixes to both citation games in #1145 , seems quite smooth up to now.

Will let you know when this is likely merge-able.

### DIFF
--- a/src/games.c
+++ b/src/games.c
@@ -3509,7 +3509,20 @@ int igraph_cited_type_game(igraph_t *graph, igraph_integer_t nodes,
   igraph_real_t sum;
   long int i,j;
   
-  IGRAPH_VECTOR_INIT_FINALLY(&edges, 0);
+  if (igraph_vector_size(types) != nodes) {
+      IGRAPH_ERROR("Invalid size of types", IGRAPH_EINVAL);
+  }
+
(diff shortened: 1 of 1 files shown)


iosonofabio:
This only fails on windows (AppVeyor) and only with Python2.7, whatever that means since it's a C library. The error message is:

`.\src\games.c(3532) : error C2143: syntax error : missing ';' before 'type'`

could anyone who knows AppVeyor or Microsoft quirks please have a look?

szhorvat:
ANSI C requires declarations to come at the beginning of blocks.

```c
{
    doSomething();
    long x = 1; /* invalid */
    doSomethingElse(x);
}
```

```c
{
    long x; /* valid */
    doSomething();
    x = 1; /* valid */
    doSomethingElse(x);
}
```

With newer MSVC it does not fail because of some support for C99 (enabled by default), which allows declarations elsewhere.

With gcc it does not fail because it allows this as a GNU extension. The default is `-std=gnu89`. With `-std=c89` it would fail too.

You can fix such problems either by moving the declaration to the beginning of the block, or by adding a new block starting at the point where the declaration was.

iosonofabio:
Ok, this is ready for merging

Issue #1009, opened 2017-05-31
R session Aborted when computing betweenness


mpoirel:
I attached a network for which I cannot compute the betweenness as it aborts the R session. I can't see what's wrong with it, do you?

I use R studio, R version 3.3.2 and igraph 1.0.1.

When I use the NetSwan::swan_combinatory function, which relies on and recomputes betweenness, I get the same result: R session aborted.

Thanks for your help,

[Network_pbl_betweenness.zip](https://github.com/igraph/igraph/files/1041854/Network_pbl_betweenness.zip)

mpoirel:
I extracted a small part of the network which cause the disruption

[NetworkSmall_pbl_betweenness.zip](https://github.com/igraph/igraph/files/1041899/NetworkSmall_pbl_betweenness.zip)

ntamas:
Also happens in the Python interface, and it is caused by the weight of the last edge (`1e-23`). Unweighted betweenness centrality works correctly for the graph; also, even the weighted centrality works correctly if the weight of the last edge is replaced with something larger (e.g., `1e-5`).

szhorvat:
Reproduced with the Mathematica interface (based on a recent dev version).

ntamas:
#591 seems related. In one of the comments for #591, @gaborcsardi says:

> The problem is that you have edge weights that are smaller than the machine precision, and that leads to some numerical instability and violates the assumptions the igraph code has about the shortest paths. If you remove edges with weight less than 1e-10, then it works fine, but I guess this is not an option for you.
> 
> igraph should not crash, of course, but I am not sure what we can do to actually calculate the betweenness for these weights, probably nothing.

He is probably correct about the numerical instability issue, although in this particular simple case, it probably wouldn't matter. I'll check whether there is a simple way to prevent the crash, though.

ntamas:
#820 is also related; although it is about edge betweenness, the stack trace where the crash happens is almost identical.

mpoirel:
Thank you very much for your prompt answer and your explenation!

It indeed works when I put bigger figure for the too small weights.

I'll have to figure out a proper way of converting the edge weights but that's a lesser issue. Thanks!

ntamas:
My advice is: decide in advance what precision you will need (e.g., 10 significant digits), then multiply every weight by 10^10, and round them to the nearest integers. You will not have this problem then because integers are not prone to the accumulation of numeric errors.

pkreissel:
[...]

Issue #820, opened 2015-04-10
R: Segfault when trying to run edge.betweenness.community on weighted network


khughitt:
I have recently been trying out some of the community detection methods available in igraph and ran into segfault issues when I tried to use the `edge.betweenness.community` on an undirected weighted network.

I tried to reproduce the issue on some random datasets, but have not been able to do so, so instead I produced a much smaller (21 node) version of the original graph which still produces the bug for me.

Nothing obvious stands out to me as to what could be producing the issue. The adjacency matrix is pretty uninteresting:

![adjmatrix](http://i.imgur.com/sH0Btjs.png)

Any ideas?

Adjacency matrix CSV: https://gist.github.com/khughitt/3b72574b8a00320ae4cf

**code to reproduce the issue**

```
#!/bin/env Rscript
library('igraph')
mat = as.matrix(read.csv('igraph_edgebetweenness_bug_adjmat.csv',
                header=1, row.names=1))
g = graph.adjacency(mat, mode='undirected', weighted=TRUE, diag=FALSE)
x = edge.betweenness.community(g, directed=FALSE)
```

**sessionInfo()**

```
R version 3.1.3 (2015-03-09)
Platform: x86_64-unknown-linux-gnu (64-bit)
Running under: Arch Linux

locale:
 [1] LC_CTYPE=en_US.UTF-8       LC_NUMERIC=C               LC_TIME=en_US.UTF-8            LC_COLLATE=en_US.UTF-8     LC_MONETARY=en_US.UTF-8    LC_MESSAGES=en_US.UTF-8   
 [7] LC_PAPER=en_US.UTF-8       LC_NAME=C                  LC_ADDRESS=C               LC_TELEPHONE=C             LC_MEASUREMENT=en_US.UTF-8 LC_IDENTIFICATION=C       

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

other attached packages:
[1] igraph_0.7.1   setwidth_1.0-3 colorout_1.0-3
```

khughitt:
p.s. One thing I noticed is that any amount of rounding I applied (even up to eight or nine digits)  alleviated the issue.

ntamas:
Crashes the Python interface as well so it's likely a bug in the C core.

ntamas:
Simply calculating the edge betweenness scores also crashes the C core.

leifeld:
[...]

PR #1199, opened 2019-06-07
Improved numerical stability of betweenness

### LINKED ISSUE #591: betweennes seg-faults in R
Hi, I have been seeing segrmentation faults when I try to run betweennes, and my graph has weighted edges.
I currently use the development branch version, but I have seen the segfault in versions since versions starting in 6.5 (master branch).

 **\* caught segfault ***
address 0xc69ec70, cause 'memory not mapped'

Traceback:
 1: base::.Call(.NAME, ...)
 2: .Call("R_igraph_betweenness", graph, v - 1, as.logical(directed),     weights, as.logical(nobigint), PACKAGE = "igraph")
 3: betweenness(theGraph)

For the environment, here is some info:

sessionInfo> sessionInfo()
R version 3.0.2 (2013-09-25)
Platform: x86_64-redhat-linux-gnu (64-bit)

locale:
 [1] LC_CTYPE=en_US.UTF-8       LC_NUMERIC=C  
 [3] LC_TIME=en_US.UTF-8        LC_COLLATE=en_US.UTF-8  
 [5] LC_MONETARY=en_US.UTF-8    LC_MESSAGES=en_US.UTF-8  
 [7] LC_PAPER=en_US.UTF-8       LC_NAME=C  
 [9] LC_ADDRESS=C               LC_TELEPHONE=C  
[11] LC_MEASUREMENT=en_US.UTF-8 LC_IDENTIFICATION=C       

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base

I am using fedora 20 x86_64, with :
glibc-2.18-12.fc20.x86_64
gcc-4.8.2-7.fc20.x86_64

If I run R with valgrind, this is what I get : 
==11891== Invalid write of size 8
==11891==    at 0x1472D5F9: igraph_2wheap_modify (heap.c:1030)
==11891==    by 0x14610E0F: igraph_i_betweenness_estimate_weighted (centrality.c:1669)
==11891==    by 0x14611453: igraph_betweenness_estimate (centrality.c:1818)
==11891==    by 0x147C064F: R_igraph_betweenness (rinterface.c:3668)
==11891==    by 0x4EBCED8: ??? (in /usr/lib64/R/lib/libR.so)
==11891==    by 0x4EFA114: Rf_eval (in /usr/lib64/R/lib/libR.so)
==11891==    by 0x4EFBD39: ??? (in /usr/lib64/R/lib/libR.so)
==11891==    by 0x4EF9F61: Rf_eval (in /usr/lib64/R/lib/libR.so)
==11891==    by 0x4EFAEDE: Rf_applyClosure (in /usr/lib64/R/lib/libR.so)
==11891==    by 0x4EF9DB4: Rf_eval (in /usr/lib64/R/lib/libR.so)
==11891==    by 0x4EFCB97: ??? (in /usr/lib64/R/lib/libR.so)
==11891==    by 0x4EF9F61: Rf_eval (in /usr/lib64/R/lib/libR.so)
==11891==  Address 0x15c70630 is 16 bytes before a block of size 8,192 alloc'd
==11891==    at 0x4C293AA: realloc (in /usr/lib64/valgrind/vgpreload_memcheck-amd64-linux.so)
==11891==    by 0x1482949E: igraph_vector_reserve (vector.pmt:437)
==11891==    by 0x14829551: igraph_vector_push_back (vector.pmt:536)
==11891==    by 0x1472D2D2: igraph_2wheap_push_with_index (heap.c:944)
==11891==    by 0x14610CB5: igraph_i_betweenness_estimate_weighted (centrality.c:1660)
==11891==    by 0x14611453: igraph_betweenness_estimate (centrality.c:1818)
==11891==    by 0x147C064F: R_igraph_betweenness (rinterface.c:3668)
==11891==    by 0x4EBCED8: ??? (in /usr/lib64/R/lib/libR.so)
==11891==    by 0x4EFA114: Rf_eval (in /usr/lib64/R/lib/libR.so)
==11891==    by 0x4EFBD39: ??? (in /usr/lib64/R/lib/libR.so)
==11891==    by 0x4EF9F61: Rf_eval (in /usr/lib64/R/lib/libR.so)
==11891==    by 0x4EFAEDE: Rf_applyClosure (in /usr/lib64/R/lib/libR.so)
 ... repeats several times, then
==11891== Use of uninitialised value of size 8
==11891==    at 0x1472D0D2: igraph_i_2wheap_switch (heap.c:866)
==11891==    by 0x1472D3A7: igraph_i_2wheap_sink (heap.c:898)
==11891==    by 0x1472D605: igraph_2wheap_modify (heap.c:1031)
==11891==    by 0x14610E0F: igraph_i_betweenness_estimate_weighted (centrality.c:1669)
==11891==    by 0x14611453: igraph_betweenness_estimate (centrality.c:1818)
==11891==    by 0x147C064F: R_igraph_betweenness (rinterface.c:3668)
==11891==    by 0x4EBCED8: ??? (in /usr/lib64/R/lib/libR.so)
==11891==    by 0x4EFA114: Rf_eval (in /usr/lib64/R/lib/libR.so)
==11891==    by 0x4EFBD39: ??? (in /usr/lib64/R/lib/libR.so)
==11891==    by 0x4EF9F61: Rf_eval (in /usr/lib64/R/lib/libR.so)
==11891==    by 0x4EFAEDE: Rf_applyClosure (in /usr/lib64/R/lib/libR.so)
==11891==    by 0x4EF9DB4: Rf_eval (in /usr/lib64/R/lib/libR.so)
==11891== 
==11891== Invalid write of size 8
==11891==    at 0x1472D0D2: igraph_i_2wheap_switch (heap.c:866)
==11891==    by 0x1472D3A7: igraph_i_2wheap_sink (heap.c:898)
==11891==    by 0x1472D605: igraph_2wheap_modify (heap.c:1031)
==11891==    by 0x14610E0F: igraph_i_betweenness_estimate_weighted (centrality.c:1669)
==11891==    by 0x14611453: igraph_betweenness_estimate (centrality.c:1818)
==11891==    by 0x147C064F: R_igraph_betweenness (rinterface.c:3668)
==11891==    by 0x4EBCED8: ??? (in /usr/lib64/R/lib/libR.so)
==11891==    by 0x4EFA114: Rf_eval (in /usr/lib64/R/lib/libR.so)
==11891==    by 0x4EFBD39: ??? (in /usr/lib64/R/lib/libR.so)
==11891==    by 0x4EF9F61: Rf_eval (in /usr/lib64/R/lib/libR.so)
==11891==    by 0x4EFAEDE: Rf_applyClosure (in /usr/lib64/R/lib/libR.so)
==11891==    by 0x4EF9DB4: Rf_eval (in /usr/lib64/R/lib/libR.so)
==11891==  Address 0xfb1da635458e34a0 is not stack'd, malloc'd or (recently) free'd

Finally, reproducing the error while running R under gdb, I get:

Program received signal SIGSEGV, Segmentation fault.
igraph_i_2wheap_switch (h=h@entry=0x7fffffffaea0, e1=-2, e2=e2@entry=-3) at heap.c:865
(in this case, this would be the  version igraph_0.7.999-66  of heap.c, and line 865 is inside function "igraph_i_2wheap_switch")

So I was wondering if my problem had anything to do with the 2wheap issue that was fixed before, but that somehow got back (for my specific architecture).

vtraag:
This should fix #591. I've tested it against the provided problem case, and it does solve that issue, and should be more numerically stable in general, I believe. It now also gives a warning in the case of small weights, that there may be numerical issues.

The reason the problem emerged at all is because of some numerical inaccuracies. This resulted in a node being popped already from the heap using `igraph_2wheap_delete_max`, but it was still updated afterwards using `igraph_2wheap_modify`. This is not possible, but this error is not explicitly caught by the `igraph_2wheap` interface. 

Doing explicit range checking etc on these type of things will make it easier to do bughunting, but it will of course also slow it down. Perhaps we could use range checking on some of the data structures when compiled with `--enable-debug`? Anyway, just a thought, this is not for now.

Unfortunately, my editor also removed the trailing white space everywhere, so that the diff is a bit more difficult to read I now see. I can correct that at some later point if you want.


vtraag:
FYI: relevant changes start at line 1576.

ntamas:
> Perhaps we could use range checking on some of the data structures when compiled with --enable-debug?

We could rely on the `NDEBUG` macro in C and put range checks and stuff like that in `#if !defined(NDEBUG) ... #endif`. As far as I know `NDEBUG` is standard and it is defined when the app is compiled in non-debug mode. (Needs someone to confirm that).

ntamas:
Tests 90 and 91 seem to fail; is this also a case on your machine?

ntamas:
> Unfortunately, my editor also removed the trailing white space everywhere, so that the diff is a bit more difficult to read I now see.

You can ask Github to ignore whitespace when diffing so it is actually pretty readable on Github.

vtraag:
Yes, I just noticed that these tests fail, I'll investigate, I didn't run them locally.

vtraag:
I've changed it around so that the tests now also work correctly. Probably a similar problem plays in `igraph_i_edge_betweenness_estimate_weighted`, I'm looking into correcting that as well. So don't merg yet, I'll look into that further.

Do we want to include a test for this case as well by the way?

vtraag:
Hmm, compilation fails for old MSVC that compiles using C89 rules that forbid mixed declaration and code. We said we would drop support for python2.7 anyhow, so should we then also disable that build and start allowing C99 from now on?

szhorvat:
Mixed declarations are very easy to fix, and perhaps not a good enough reason to go to C99 ... 

I sometimes write blocks in C:

```c
void fun() {
    int i, j;

    {
        int a;
        doThing(a);
    }

    {
        int b;
        doThing(b);
    }
}
```

Do you know which is the first MSVC version that supports C99 reasonably fully?

szhorvat:
It would seem that Visual Studio 2017 still doesn't support VLAs ... https://stackoverflow.com/a/48615664/695132 Which was one of the major issues with porting LADv3.

Microsoft doesn't even seem to document C standards conformance (they have detailed documentation about C++ standards conformance).

vtraag:
Yes, I believe the support is always partial. But at least the mixed declaration and code is supported, I think from 2014 onwards?

Anyway, I'll change it for now, it's indeed straightforward to fix.

ntamas:
Seems good to me - once the declaration is moved I'm happy to merge this.

vtraag:
Thanks! I'll first take a look at `igraph_i_edge_betweenness_estimate_weighted` before merging. I'll let you know when it's good to go.

vtraag:
I've corrected the mixed declaration and initialization. I've also improved the numerical stability for `igraph_i_edge_betweenness_estimate_weighted`. 

However, it now runs into a problem with test 92. The problem is that test 92 uses weights of 0. Is there any specific reason for that @iosonofabio ? It is currently allowed to use zero weights for `igraph_i_edge_betweenness_estimate_weighted`, whereas for `igraph_i_betweenness_estimate_weighted` zero weights are not allowed. Is there any specific reason for that? To me that doesn't make sense, and I would adjust `igraph_i_edge_betweenness_estimate_weighted` to only allow strictly positive weights. That then also requires that test 92 is updated to use strictly positive weights. Is there any reason not to do so @iosonofabio ?

vtraag:
The same numerical problem also appeared in `igraph_i_closeness_estimate_weighted`. Again, zero weights are allowed, and I don't understand why, and I would suggest restricting it to strictly positive weights.

vtraag:
This PR also fixes #820.

vtraag:
This PR also fixes #1009.

ntamas:
> Again, zero weights are allowed, and I don't understand why

I agree that we should simply exclude edges with zero weight. It does not really make sense for algorithms where the weights correspond to distances; an edge with zero weight basically means tha the two endpoints are "infinitely close", and I think that these measures are not prepared for that. (However, we can allow edges with zero weights in algorithms such as PageRank where the weight means "closeness"; in that case, an edge with zero weight is essentially a nonexistent edge).

I'm happy to merge this if you add a check that prevents edges with zero weights in these functions.

vtraag:
OK, I'll update the code and the tests.

ntamas:
Thanks!

szhorvat:
@vtraag 

The IGraph/M test suite reveals some serious looking breakage in the closeness calculations after this ... I was hoping that I could tell you that everything passes, and be done with it, but apparently it'll take a bit of work to investigate. So this is just an advance notice. I'll come back to it later.

szhorvat:
In case someone else would investigate before I can get to it:

image

vtraag:
Hmm, that is a pity! Thanks for checking it carefully though, it seems I was too quick. I'll try to take a look as well, but if you make any progress, let me know!

vtraag:
I can't reproduce your closeness measures for `igraph` in C (assuming these are returned by `IGCloseness`). I'm running this:

```
  igraph_i_set_attribute_table(&igraph_cattribute_table);
  
  igraph_t graph;
    
  igraph_vector_t edges;
  igraph_vector_init(&edges, 0);
  igraph_vector_push_back(&edges, 0); igraph_vector_push_back(&edges, 1);
  igraph_vector_push_back(&edges, 1); igraph_vector_push_back(&edges, 2);
  igraph_vector_push_back(&edges, 2); igraph_vector_push_back(&edges, 0);
  igraph_create(&graph, &edges, 3, 0);

  printf("Graph has %i nodes and %i edges.\n", igraph_vcount(&graph), igraph_ecount(&graph));
  
  igraph_vector_t weights;
  igraph_vector_init(&weights, 3);
  VECTOR(weights)[0] = 1;
  VECTOR(weights)[1] = 2;
  VECTOR(weights)[2] = 3;

  igraph_vector_t c;
  igraph_vector_init(&c, igraph_vcount(&graph));  
  igraph_closeness(&graph, &c, igraph_vss_all(), IGRAPH_ALL, &weights, 1);
  for (int i = 0; i < 3; i++)
  {
        printf("Closeness node %i: %e\n", i, VECTOR(c)[i]);
  }
  igraph_vector_destroy(&c);
      
  igraph_vector_destroy(&weights);
  igraph_destroy(&graph);
```

This yields 

```
Graph has 3 nodes and 3 edges.
Closeness node 0: 5.000000e-01
Closeness node 1: 6.666667e-01
Closeness node 2: 4.000000e-01
```

which I believe is in line with the closeness from (I guess) Mathematica?

Am I misinterpreting the Mathematica code in any way?

szhorvat:
> which I believe is in line with the closeness from (I guess) Mathematica?

Yes.

> Am I misinterpreting the Mathematica code in any way?

No, instead you don't need to set up an attribute handler. Passing a `weights` vector is sufficient.

I'll take a closer look later.

szhorvat:
Are you certain you are building with the correct version of igraph?

With your code, I get:

```
Graph has 3 nodes and 3 edges.
Closeness node 0: 2.857143e-01
Closeness node 1: 3.333333e-01
Closeness node 2: 2.500000e-01
```

This is with latest `master`.

Here's a simpler way to write the same:

```
#include <igraph.h>

int main() {
    igraph_t g;
    igraph_vector_t weights, res;

    igraph_small(&g, 3, 0, 0,1, 1,2, 0,2, -1);

    igraph_vector_init(&weights, 3);
    VECTOR(weights)[0] = 1;
    VECTOR(weights)[1] = 2;
    VECTOR(weights)[2] = 4;

    igraph_vector_init(&res, 0);

    igraph_closeness(&g, &res, igraph_vss_all(), IGRAPH_ALL, &weights, 1);

    igraph_vector_print(&res);

    igraph_vector_destroy(&res);
    igraph_vector_destroy(&weights);

    igraph_destroy(&g);

    return 0;
}
```

vtraag:
Sorry, yes, I just forgot that I actually checked out code to double check the result before the PR. I'll take a closer look at what's going wrong later today...

ntamas:
FYI, the Python tests are also breaking: https://travis-ci.org/igraph/python-igraph/jobs/544192558

(It's a pity that some of the igraph functions are actually tested indirectly by higher-level interfaces)

Issue #930, opened 2016-03-02
igraph_write_graph_graphml() is locale dependent


Ariki:
The function `igraph_write_graph_graphml()` writes floating point attribute values using locale dependent decimal separator. In locales that use comma as separator, this causes generation of locale dependent XML which can't be read back properly, even on the same system.The numbers containing a comma instead of a period are truncated when read back.

Here is a code example demonstrating the issue. This assumes that the locale using decimal comma (`"ru_RU.UTF-8"` in this example) presents in the system.

``` c
#include <locale.h>
#include <igraph.h>

int main(void) {
    igraph_t graph;
    igraph_i_set_attribute_table(&igraph_cattribute_table);
    igraph_full(&graph, 1, 0, 0);
    SETGAN(&graph, "myattr", 3.14);
    setlocale(LC_NUMERIC, "ru_RU.UTF-8"); // This locale uses decimal comma
    igraph_write_graph_graphml(&graph, stdout, 0); // The output contains "3,14"
    setlocale(LC_NUMERIC, "C");
    igraph_write_graph_graphml(&graph, stdout, 0); // The output contains "3.14"
    igraph_destroy(&graph);
    return 0;
}
```

ntamas:
Note to self (for future testing and reproduction): ANSI C says that all programs start by default in the standard `C` locale, so this becomes a problem only if the user sets the locale using `setlocale()` first; it is not enough to set the `LANG`, `LC_ALL` or `LC_NUMERIC` environment variables.

Ariki:
Yes, you are right. I've discovered the issue when using Python bindings from some exotic environment (ArcGIS ArcToolbox) which seems to set locale for me.

ntamas:
Okay, you can then probably work around this from Python as follows:

```python
import locale
locale.setlocale(LC_NUMERIC, "C")
```

I'll commit a fix soon(ish).

Ariki:
Yes, it works. Thank you.

iosonofabio:
@ntamas I started PR #1203 for this. Not sure whether it was what you had in mind, so some advice would help.

szhorvat:
Note: This has been reverted because it turns out that the fix was problematic.

 - It would leave the locale changed on failure.
 - `setlocale()` cannot be safely used in multithreaded programs (such as ArcGis).

It is now left to the user to ensure that the locale uses a decimal point (not a decimal comma) before calling any foreign format readers / writer.

[...]

PR #1203, opened 2019-06-09
GRAPHML export and locale, aka #930

iosonofabio:
This PR tries to address issue #930 

For now no error handling so the locale might be messed up if anything happens during the XML export.

### DIFF
--- a/src/foreign-graphml.c
+++ b/src/foreign-graphml.c
@@ -1439,7 +1439,7 @@ int igraph_write_graph_graphml(const igraph_t *graph, FILE *outstream,
 
   /* set standard C locale lest we sometimes get commas instead of dots */
   char *saved_locale = strdup(setlocale(LC_NUMERIC, NULL));
-  setlocale(LC_NUMERIC, "C")
+  setlocale(LC_NUMERIC, "C");
   
   ret=fprintf(outstream, "<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n");
   if (ret<0) IGRAPH_ERROR("Write failed", IGRAPH_EFILE);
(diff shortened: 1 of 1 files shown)


ntamas:
Seems good as a starting point, but we should take care of error handling. In particular:

* if `strdup()` returns `NULL`, return `IGRAPH_ENOMEM` and bail out
* otherwise, add `IGRAPH_FINALLY(igraph_free, saved_locale)` right after `strdup()` so `saved_locale` gets freed properly when any of the functions we call return an igraph error code.
* finally, at the end, call `igraph_free()` on `saved_locale` yourself and bump the counter in `IGRAPH_FINALLY_FREE()` by one to clean up the "finally" stack properly

ntamas:
Not quite there yet; `IGRAPH_FINALLY(free, saved_locale)` basically makes a note that if any subsequent `IGRAPH_CHECK()` call returns with an error code, igraph should call `free(saved_locale)` automatically. These entries are stored in a stack which we call the "finally stack". More about this [here](https://igraph.org/c/doc/igraph-Error.html#IGRAPH_FINALLY).

Note that the finally stack is not cleaned if you return from the current function with `return`, so you still need `free(saved_locale)` at the end. For normal return conditions, the idea is that you free everything that you have put in the finally stack on your own, and then you call `IGRAPH_FINALLY_FREE()` and tell igraph how many items to _discard_ from the finally stack (the ones that you have taken care of yourself). If you add a new memory allocation somewhere in a function, you need to:

* add a corresponding `IGRAPH_FINALLY()` call to take care of freeing the memory in case of an error (some macros like `IGRAPH_VECTOR_INIT_FINALLY()` unify the initialization and the handling of the finally stack in a single call).
* free the memory on your own upon normal exit
* find the `IGRAPH_FINALLY_FREE()` call at the end of the normal exit route and bump the counter in it by one

iosonofabio:
This should be good now as soon as tests pass

ntamas:
Seems good to me as well, thanks!

Issue #831, opened 2015-06-12
community_leading_eigenvector() returns incorrect results (sort of)


Pelonza:
I'm running igraph version 0.7.1 on SageMathCloud.

When I ran the community detection algorithm using Newman's leading eigenvector method on the Political books network by Krebs, it produced a different result than the answer Newman published.

Figure 3 in "Modularity and community structure in networks" by Newman, 2006 in PNAS,
http://www.pnas.org/content/103/23/8577.full

This reports 4 communities (which the algorithm finds), but according to our checking, seems to report the wrong communities, and has a much lower modularity score than the paper reports.

A little more investigation finds that igraph has implemented Newman's "simple" algorithm (correct, based on the citation given in the documentation), however, in the citation above, he introduces a further refinement which improves modularity further (based on the Kernighan-Lin agorithm, possibly implemented elsewhere?). 

It may be a speed/complexity choice not to introduce the extra refinement method, however, it should probably be noted in the documentation that it does not include that refinement, so results may vary.

I checked this by running on the karate network. Newman reports that without his refinement, the leading eigenvector method reports Q=0.393, but with the refinement Q=0.419  (igraph's method on the karate indeed reports Q=0.393)

gaborcsardi:
Sure, we can add to the docs, that this is not implemented.

PR #1207, opened 2019-06-11
Corrected doc: no refinement in leading eigenvector

### LINKED ISSUE #830: Error: GraphBase does not have attribute 'community_leading_eigenvector_naive'
In the igraph documentation, there is listed the method "community_leading_eigenvector_naive" ... as a graph attribute, which can be called. 

While the attribute is present on a graph instance, if you look at the GraphBase object, it does NOT occur.

I ran into this error in SageMathCloud, so it _might_ be localized to their instance of igraph.
The reported version (via print igraph.**version**) is 0.7.1

I checked the "attributes" via a call to dir() on both igraph.Graph and igraph.GraphBase

If in-fact, the naive version is implemented, then the GraphBase attributes need to be properly updated. If, on the other hand, it's not implemented, the documentation should reflect this.

vtraag:
Fixes #830.


vtraag:
Sorry, fixes #831, not #830.

PR #1209, opened 2019-06-11
Correction of improved numerical stability for closeness

vtraag:
This is the correction for the closeness that was found to be incorrect in the earlier fix in PR
#1199.

The provided example now provides the expected results. 

@ntamas, I tried to run the `python-igraph` tests locally, but I run into an unrelated error
```
testRandomWalkStuck (igraph.test.walks.RandomWalkTests) ... Segmentation fault (core dumped)
```
I'm investigating this currently. Perhaps you can run the python tests for this PR to check the closeness calculations?


szhorvat:
Thanks for the quick fix Vincent! The Mathematica interface tests all pass now, but I have not yet had time to understand that changed in this PR (or the original one for the same issue).

vtraag:
The original problem was the following. The problem is similar for `igraph_i_betweenness_estimate_weighted`, `igraph_i_edge_betweenness_estimate_weighted` and `igraph_i_closeness_estimate_weighted`, but I'll explain what happened in `igraph_i_betweenness_estimate_weighted`. The `igraph_2wheap_t Q` contains a (two way) heap, which has at its top the node at the shortest distance of `source`. This heap was maintained at a zero-based distance, i.e. the source was pushed to the heap at distance 0:
https://github.com/igraph/igraph/blob/9bb8167ad6d6e5ae3a8359610a85bca57916e4b8/src/centrality.c#L1632
At the same time, the distance to all nodes is maintained in the vector `dist`, which used a one-based distance
https://github.com/igraph/igraph/blob/9bb8167ad6d6e5ae3a8359610a85bca57916e4b8/src/centrality.c#L1633
In other words, the distance maintained in the heap `Q` and the vector `dist` always differed 1.0. They were therefore compared as follows:
https://github.com/igraph/igraph/blob/9bb8167ad6d6e5ae3a8359610a85bca57916e4b8/src/centrality.c#L1659
where `altdist` is heap-based (i.e. starting at 0) and `curdist` is vector based (i.e. starting at 1). This is used to decide whether the current distance (`curdist`) is the shortest or whether the alternative distance (`altdist`) is actually shorter.

The problem now is that the relative ordering of the heap compares floats around `0`, while the decision of whether the current distance (`curdist`) is the shortest, only happens after adding and extracting `1`. In general `(x + 1) - 1 != x` due to limited floating point accuracy, and so this leads to numerical problems, where the relative ordering in the heap is inconsistent with the comparison in 
https://github.com/igraph/igraph/blob/9bb8167ad6d6e5ae3a8359610a85bca57916e4b8/src/centrality.c#L1659
Finally, this resulted in an attempt to update the heap in
https://github.com/igraph/igraph/blob/9bb8167ad6d6e5ae3a8359610a85bca57916e4b8/src/centrality.c#L1679
while it was actually already popped from the heap in 
https://github.com/igraph/igraph/blob/9bb8167ad6d6e5ae3a8359610a85bca57916e4b8/src/centrality.c#L1638
My solution is to maintain the heap also at a one-based distance, which solves the numerical comparisons. Of course, the accuracy is still limited, but the relative ordering of distances in the heap and the vector is identical.

Finally, the problem with closeness was quite simple. It actually used the zero-based distance from the heap for its calculations
https://github.com/igraph/igraph/blob/9bb8167ad6d6e5ae3a8359610a85bca57916e4b8/src/centrality.c#L2581
and also for doing the cutoff comparison
https://github.com/igraph/igraph/blob/9bb8167ad6d6e5ae3a8359610a85bca57916e4b8/src/centrality.c#L2584
These were handled differently in `igraph_i_betweenness_estimate_weighted` and `igraph_i_edge_betweenness_estimate_weighted`, so I didn't check it. This caused the problem you noticed.

In fact, the code can be cleaned up a little bit more, but I thought it better to keep the edits to a minimum.

ntamas:
Python tests are OK as well.

Issue #1051, opened 2018-01-06
It would be desirable that igraph_avg_nearest_neighbor_degree() implemented Barrat's knn measure


yoelcabo:
As explained in #987 by @szhorvat, `igraph_avg_nearest_neighbor_degree()` for weighted graphs does not compute the measure as specified in the referenced paper by Barrat et al. This has been clarified in the documentation by https://github.com/igraph/igraph/commit/08c04de141e13ef07e5079ea934ceeff973cbd62, closing #987.

However, I believe this is still a bug, although unfortunately I lack knowledge about the project and time to submit a PR. So I think it is good to continue having an issue opened for this topic. 

Even if that was the intended behaviour, I believe it would be good to have the option to compute Barrat's `knn`. For instance adding a type field like in: http://igraph.org/r/doc/transitivity.html.

PR #1214, opened 2019-07-10
Corrected average nearest neighbour (direction and weights)

### LINKED ISSUE #643: Consider edge directions in graph.knn
BEcause right now they are ignored.

### LINKED ISSUE #987: igraph_avg_nearest_neighbor_degree() weighted result dubious
`igraph_avg_nearest_neighbor_degree()` computes the mean degree the vertices' nearest neighbours.  To find out what it does for edge-weighted graphs, I looked at the source code.

For weighted graphs, for a given node `i` it will add up the strengths of `i`'s neighbours, then divide the result by `i`'s strength.

This is not clear from the documentation, and is inconsistent with the reference given there.

http://igraph.org/r/doc/knn.html

The documentation says that for weighted graphs, "vertex strength (see strength) is used instead of vertex degree". From this, I would have assumed that it computes the mean neighbour strength, but this is not the case.  In fact, what is being computed cannot even be called an average.

The reference being given in the documentation is http://dx.doi.org/10.1073/pnas.0400087101

This paper has a similar "knn" measure defined in equation (6).  But again, it does not agree with what igraph computes.  Barrat's "knn" is a weighted average of degrees (not strengths). The weighting is done according to edge weights.

The documentation suggests that this function was meant to compute what is in the paper, but it doesn't. I assume that this is a bug.

vtraag:
This should fix #643 about the directionality and fix #987 and #1051 about the weights. 

You can now specify the direction (`IGRAPH_IN`, `IGRAPH_OUT`, `IGRAPH_ALL`) for which neighbours to consider and the direction for which degree of the neighbours to average over.

The weighted version now respects the definition provided by [Barrat et. al](https://dx.doi.org/10.1073/pnas.0400087101), Eq. (6).

### DIFF
--- a/src/structural_properties.c
+++ b/src/structural_properties.c
@@ -5772,23 +5772,28 @@ int igraph_i_avg_nearest_neighbor_degree_weighted(const igraph_t *graph,
     IGRAPH_CHECK(igraph_vector_resize(knn, no_vids));
   }
 
+  // Get degree of neighbours
   IGRAPH_VECTOR_INIT_FINALLY(&deg, no_of_nodes);
   IGRAPH_CHECK(igraph_degree(graph, &deg, igraph_vss_all(),
 			       neighbor_degree_mode, IGRAPH_LOOPS));
   IGRAPH_VECTOR_INIT_FINALLY(&strength, no_of_nodes);
+
+  // Get strength of all nodes
   IGRAPH_CHECK(igraph_strength(graph, &strength, igraph_vss_all(),
 			       mode, IGRAPH_LOOPS, weights));
+
+  // Get maximum degree for initialization
   IGRAPH_CHECK(igraph_maxdegree(graph, &maxdeg, igraph_vss_all(),
 				mode, IGRAPH_LOOPS));
-  IGRAPH_VECTOR_INIT_FINALLY(&neis, maxdeg);
-  IGRAPH_VECTOR_INIT_FINALLY(&edge_neis, maxdeg);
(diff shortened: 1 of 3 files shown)


src/structural_properties.c
```diff
@@ -5885,53 +5908,52 @@ int igraph_avg_nearest_neighbor_degree(const igraph_t *graph,
...
 
   for (i=0; !IGRAPH_VIT_END(vit); IGRAPH_VIT_NEXT(vit), i++) {
     igraph_real_t sum=0.0;
     long int v=IGRAPH_VIT_GET(vit);
-    long int nv=(long int) VECTOR(deg)[v];
-    IGRAPH_CHECK(igraph_neighbors(graph, &neis, (igraph_integer_t) v,
-				  IGRAPH_ALL));
-    for (j=0; j<nv; j++) { 
+    IGRAPH_CHECK(igraph_neighbors(graph, &neis, (igraph_integer_t) v, mode));
+    long int nv=igraph_vector_size(&neis);
```
ntamas: The declaration of `nv` should be moved to the top of the block; otherwise the CI build will fail on Windows with Python 2.7.

vtraag: A recurrent mistake of mine :)

Is there any way to force `gcc` to conform to this declaration standard, without forcing also additional `C89` features that are compatible with `VC9++` to compile for Python 2.7 (e.g. `//` comment style)?

ntamas: I don't think so; let me know if you find a way ;)

ntamas:
Thanks!

Issue #670, opened 2014-08-20
layout.reingold.tilford does not work with trees mode="in" 


tonigi:
As per subject:

```
plot(graph.tree(20),layout=layout.reingold.tilford)
```

works, while...

```
plot(graph.tree(20,mode="in"),layout=layout.reingold.tilford)
```

...doesn't.

szhorvat:
Potentially related: https://github.com/igraph/igraph/issues/879

iosonofabio:
should this bug be moved to the R repo? It seems like the C interface for `mode` is well documented in `layout.c`, line 941:
```
 * \param mode Specifies which edges to consider when building the tree.
 *   If it is \c IGRAPH_OUT then only the outgoing, if it is \c IGRAPH_IN
 *   then only the incoming edges of a parent are considered. If it is 
 *   \c IGRAPH_ALL then all edges are used (this was the behavior in 
 *   igraph 0.5 and before). This parameter also influences how the root 
 *   vertices are calculated, if they are not given. See the \p roots parameter.
```

vtraag:
The bug itself concerns some problems when there are no incoming/outgoing edges I believe.

On a related note, the documentation for `layout.reingold.tilford` in R may be be incomplete. @ntamas, perhaps we should open a specific issue in `rigraph` for the documentation.

ntamas:
> The bug itself concerns some problems when there are no incoming/outgoing edges I believe.

Yes; you can probably still reproduce it in C by creating a tree graph with edges pointing _towards_ the root, and then trying to construct a layout with `layout_reingold_tilford()` such that it is configured to follow the _outbound_ edges only. Of course there are no such edges from the root so the layout should not produce anything meaningful, but at least it should not put NaNs in the coordinate matrix.

@vtraag Please do if the documentation is indeed incomplete.

iosonofabio:
I'm looking at this, but I'm a bit confused as of what the expectation for this corner case would be (tree "in" but layout "out").

First, the function states it makes the input graph connected by adding (directed?) edges, but that part remains quite obscure to me. Second, the internal function `igraph_i_layout_reingold_tilford` initializes the output matrix to be of size `n_nodes x 2` and then goes on a breath-first search to set the Y coordinates. However, in the example above it seems like all those are `NaN`s, indicating this search fails to traverse the whole tree and leaves chunks of the matrix uninitialized.

[...]

PR #1215, opened 2019-07-15
Reingold Tilford layout and directed graphs

iosonofabio:
DO NOT MERGE YET!

Trying to address #670 for layouting tree-style directed graphs. The current implementation, which needs a bit of testing, addresses the issue of unreachable nodes within weakly connected components by adding directed nodes from each root to each unreachable node within that component. Better than nothing for 0.8 - and we can brainstorm a bit to improve it in 0.9 if we have time.

Will let you know when this needs serious assessment for merging.


src/layout.c
```diff
@@ -1115,21 +1116,158 @@ int igraph_layout_reingold_tilford(const igraph_t *graph,
...
+  /* the graph is disconnected and directed, add edges within
+     weakly connected components from the root to all nodes that
+     cannot be reached
+     
+     TODO: a better strategy would be nice here
+     */
+  if (igraph_vector_size(proots) != 1 && mode != IGRAPH_ALL) {
+
+    igraph_vector_t newedges;
+    long int no_of_newedges, root;
```
vtraag: `no_of_newedges` is never initialized, and should be set to 0 at some point. This yields a (possible) error later on with memory allocation when initializing the vector.

vtraag:
Sorry, I don't have further time to look at this now, but I just noticed one problem. I'll try to take another look at some later point.

src/layout.c
```diff
@@ -686,6 +687,81 @@ int igraph_layout_lgl(const igraph_t *graph, igraph_matrix_t *res,
...
+        j++;
+      }
+    }
+  }
+
+  igraph_dqueue_destroy(&q);
+  igraph_adjlist_destroy(&allneis);
+  igraph_vector_destroy(&visited);
+  IGRAPH_FINALLY_CLEAN(3);
+}
```
vtraag: There is now no proper return value, you have to return `IGRAPH_SUCCESS`. Doing so solves the out of memory error I believe (this is just a "random" error, depending on what value was returned).

iosonofabio:
thank you Vincent, good catch. I think we can merge now

PR #1225, opened 2019-08-07
Corrected all s-t cuts

### LINKED ISSUE #1102: Failure in igraph_all_st_cuts
I'm not quite sure what is going on here.  Here's an example, demonstrated using the R interface:

```
g<-make_graph(c(1, 2, 1, 5, 1, 7, 1, 8, 2, 1, 2, 4, 3, 10, 4, 2, 4, 6, 4, 8, 5, 1,  5, 2, 5, 6, 5, 10, 6, 8, 7, 1, 7, 6, 9, 3, 9, 5, 9, 10))

stCuts(g,1,6)
```

It fails with:

```
Error in stCuts(g, 1, 6) : 
  At st-cuts.c:415 : Invalid root vertex id for dominator tree, Invalid value
```

vtraag:
This should fix #1102. The problem was that some vertices were not relabeled while working on a subgraph. I've also added tests to demonstrate that it functions correctly, I believe.

Potentially a similar problem plays also in some of the other `_pivot` functions, I'll have to check. Please wait until I've done so. I just wanted to share this, so that you can check whether this solution works.

### DIFF
--- a/examples/simple/igraph_all_st_cuts.c
+++ b/examples/simple/igraph_all_st_cuts.c
@@ -32,6 +32,37 @@ int igraph_i_all_st_cuts_pivot(const igraph_t *graph,
 			       long int target,
 			       long int *v,
 			       igraph_vector_t *Isv);
+int test_all_st_cuts(const igraph_t *graph,
+             long int source,
+             long int target)
+{
+  igraph_vector_ptr_t cuts, partition1s;
+  long int n, i;
+
+  igraph_vector_ptr_init(&cuts, 0);
+  igraph_vector_ptr_init(&partition1s, 0);
+  igraph_all_st_cuts(graph, &cuts, &partition1s,
+                     source, target);
+
+  n=igraph_vector_ptr_size(&partition1s);
+  printf("Partitions and cuts:\n");
+  for (i=0; i<n; i++) {
+    igraph_vector_t *v=VECTOR(partition1s)[i];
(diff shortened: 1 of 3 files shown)


vtraag:
OK, I believe the same problem doesn't play a role in `igraph_i_all_st_mincuts_pivot`, it was only a bug in `igraph_i_all_st_cuts_pivot`. As far as I'm concerned we can merge it, but perhaps @szhorvat wants to weight in still?

szhorvat:
I absolutely do. It's been an extremely busy time. Please give me a day or too and I'll give feedback.

szhorvat:
I tried to test it extensively and find the corner cases, and it works well.

There is one minor thing that's unclear (and unrelated to the changes you made):

When the two given vertices are already disconnected, sometimes the function returns zero cuts, and sometimes it returns one empty cut. It's not clear to me why this is happening.

Take e.g. this graph:

image

For `s=2, t=4` I get `{{}}` as a result, i.e. one cut of size zero.  For `s=3, t=1` I get `{}`, i.e. no cuts.  I don't see why the result shouldn't be the same in both cases.

szhorvat:
My suggestion is that we merge this now and I open a new issue for the small problem I described above. @vtraag do you agree or will you look into this problem before merging?

vtraag:
Thanks for the careful check! I agree that is is best to merge this and open another issue for the other problem you identified.

Issue #1188, opened 2019-05-21
Docs back to doxygen


iosonofabio:
We currently generate documentation from C source code using doxygen + Docbook intermixed throughout the code. That relies on a hack script, `doxrox.py`, which is brittle and does not like Python 3 (#1186).

We should try and switch back to pure doxygen throughout the codebase and use other tools, e.g. Sphinx + [breathe](https://github.com/michaeljones/breathe) to expand the list of doc formats beyond what doxygen can do.

This issue superseeds #1186.

iosonofabio:
Alright, closing this one. There are a number of improvements @ntamas is working on re docs and cmake, which improve the state of affairs in a similar way as using doxygen.

PR #1231, opened 2019-08-14
Versioning documentation

### LINKED ISSUE #279: Add igraph version to documentation
I mean in the C manual.

---

- date created: 2010-08-13T19:09:44Z
- owner: gabor.csardi
- the launchpad url was https://bugs.launchpad.net/bugs/617548

vtraag:
This is my attempt to fix #279. It correctly generates html using `make html` inside the `doc` directory.

However, there seem to be some problems with some character encodings. In particular, during `make info`, there is a problem with the `ü` from `Prüfer` and the `å` from `Östergård`. I managed to fix one error I got (during the parsing of the XML by `tools/removeexamples.py`) by adding appropriate entity definitions in `igraph-docs.xxml`. It then still runs into an error when running `db2x_texixml`. Unfortunately, I don't know what the problem exactly is.

Suggestions are welcome!


ntamas:
Can you post the error here? I don't have the full DocBook toolchain on my machine yet but it would be nice to see it.

vtraag:
Here is the output of `make info`:
```
$ make info
if [ "x.." != "x.." ]; then cp ../doc/fdl.xml ../doc/gpl.xml ../doc/installation.xml ../doc/introduction.xml ../doc/tutorial.xml ../doc/licenses.xml ../doc/pmt.xml . ; fi && \
cat tutorial.xml | sed '/<link/{N;s/<link\n*[^>]*>//g;}' | \
  sed 's/<\/link>//g' >tutorial-info.xml \
&& sed 's/tutorial\.xml/tutorial-info.xml/' igraph-docs.xml > igraph-docs-info.xml \
&& xmllint --xinclude -o igraph-docs-info2.xml igraph-docs-info.xml \
&& ../tools/removeexamples.py igraph-docs-info2.xml igraph-docs-info3.xml \
&& sed '/<book/q' igraph-docs-info2.xml | grep -v '<book' >igraph-docs-info4.xml\
&& cat igraph-docs-info3.xml >> igraph-docs-info4.xml \
&& db2x_xsltproc --xinclude -s texi igraph-docs-info4.xml -o igraph.txml \
&& db2x_texixml igraph.txml \
&& mv igraph_reference_manual.texi igraph.texi \
&& makeinfo --no-split --number-sections igraph.texi \
&& mv igraph_reference_manual.info igraph.info \
&& patch <../doc/igraph.info.diff
docbook2texi://book[@id='index']: no description for directory entry
/usr/bin/iconv: illegal input sequence at position 445009
/usr/bin/db2x_texixml: program in pipeline exited with an error
make: *** [Makefile:683: igraph.info] Error 1
```
The problem is in `db2x_texixml igraph.txml`. The problematic line is
```
Patric R. J. Östergård, <uref url="http://users.aalto.fi/~pat/cliquer.html">http://users.aalto.fi/~pat/cliquer.html</uref>
```
and the problematic character is `å` of this line. Changing that manually to an ordinary `a` everywhere solves the immediate issue, but the problem is of course how to deal with this correctly. In `igraph-docs-info4.xml` this line reads

```
Patric R. J. &#214;sterg&#229;rd, <ulink url="http://users.aalto.fi/~pat/cliquer.html">http://users.aalto.fi/~pat/cliquer.html</ulink>
```
so the HTML character codes are translated to utf8 by `db2x_xsltproc`. However, I don't understand why `Ö` is not problematic, while `å` is.  Both are proper utf8 characters I believe.

ntamas:
Okay, I think the problem is not that `db2x_texixml` does not understand what the input is; it knows the input perfectly fine. The problem is that `db2x_texixml` tries to convert the input to `us-ascii` in order to produce the _output_ file; see the documentation [here](http://docbook2x.sourceforge.net/latest/doc/db2x_texixml.html), in particular the default value of the `--encoding` switch, which is `us-ascii`. We could try `--encoding=utf-8`, or `--encoding=us-ascii//TRANSLIT` and see how it goes (accepting the fact that `utf-8` encoded Texinfo documents might not be readable on systems without proper UTF-8 support).

vtraag:
That indeed solves the issue. I still don't fully understand why it only complains about some utf8 characters, and not all. Anyway, it should be solved now, thanks!

I had to change a bit the setup, because the version would actually appear in the filename, while making the `info` document, which is inconvenient. I also had to change some other things for the inclusion of the version number in the `info` document.

ntamas:
Okay, great! Is this ready to merge now?

vtraag:
Good question. I think it should be fine for now. I believe the idea is to change the documentation system anyhow a bit (#1188), but perhaps @iosonofabio wants to weigh in on this?

szhorvat:
Right now it looks like this:

image

Two questions:

 - There's a version in between the author list and legal notice, and one within the legal notice. Do we need to include the version note specifically in the legal notice? If yes, perhaps it should be in its own paragraph. It is written on a separate line, but HTML doesn't care about lines.

 - The version info between the author list and the legal notice is not aligned correctly. If this is where we will put it, an extra entry is needed in `style.css`.

vtraag:
I left the version both on its own and inside the legal notice. It is now on separate lines though. That way, the version is separately identified as `releaseinfo` of the docbook (which may be useful in some regard), but it is also explicitly in the release info.

ntamas:
> That indeed solves the issue. I still don't fully understand why it only complains about some utf8 characters, and not all.

Finally tracked this down -- it is because `doc/igraph-docs.xml` lists a few entities (that we have used in the past) in the header of the document so `docbook` "knows" these. If you start using other HTML entities, then you need to add them to the top of `doc/igraph-docs.xml` and it will be okay.

PR #1232, opened 2019-08-16
Limit all simple paths to a certain cutoff.

### LINKED ISSUE #774: Limit for all possible paths between two nodes [Feature Request]
Add command to return all possible path between two nodes (with an optional argument to set a limit for the maximum path length)

Related request here: http://lists.nongnu.org/archive/html/igraph-help/2013-11/msg00164.html

vtraag:
Fixes #774. Changes the public API by introducing additional cutoff argument. Providing a negative cutoff generates all simple paths.

### DIFF
--- a/examples/simple/igraph_get_all_simple_paths.c
+++ b/examples/simple/igraph_get_all_simple_paths.c
@@ -1,0 +1,63 @@
+/* -*- mode: C -*-  */
+/* 
+   IGraph library.
+   Copyright (C) 2010-2012  Gabor Csardi <csardi.gabor@gmail.com>
+   334 Harvard street, Cambridge, MA 02139 USA
+
+   This program is free software; you can redistribute it and/or modify
+   it under the terms of the GNU General Public License as published by
+   the Free Software Foundation; either version 2 of the License, or
+   (at your option) any later version.
+   
+   This program is distributed in the hope that it will be useful,
+   but WITHOUT ANY WARRANTY; without even the implied warranty of
+   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+   GNU General Public License for more details.
+   
+   You should have received a copy of the GNU General Public License
+   along with this program; if not, write to the Free Software
+   Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 
+   02110-1301 USA
+
+*/
+
+#include <igraph.h>
+
+int main() {
+
+  igraph_t g;
+  igraph_small(&g, 6, IGRAPH_UNDIRECTED,
+    0,1, 1,2, 2,5,
+    0,3, 3,4, 4,5,
+    3,2, 3, 5,
+    -1);
+  igraph_vector_int_t res;
+  igraph_vector_int_init(&res, 0);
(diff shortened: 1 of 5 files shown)


ntamas:
Thanks!

szhorvat:
Due to an API change, the Python interface needs an update too now (ran into this when installing python-igraph for someone today).

vtraag:
Yes, this is also why we should move to using the C core as a submodule in python-igraph.

ntamas:
Hmmm, I think we use submodules now in `python-igraph` so it should be pinned at the revision that was the head of `master` when I switched to submodules. This commit should not be part of the build.

vtraag:
Yes, you are right of course. Perhaps @szhorvat still had an old commit from the `python-igraph` master branch?

ntamas:
...or he did not run `git submodule update --init`, in which case we probably still fall back to the old methods of finding the C core of igraph (`pkg-config` et al).

vtraag:
Ah yes, good point, that might also be a possibility. Any reason why you prefer to keep that method of installation in?

ntamas:
Because running `pip install python-igraph` should work even if the user does not have `git` on his system. Right now this would download the source code of `python-igraph` from the Python Package Index and run `python setup.py install` on it, no submodule initialization or anything like that. (We cannot even expect the user to have `git` installed on his system). Yes, this could break if the user _does_ have some kind of an igraph version on his system but it does not match the one required for `python-igraph`; however, in the long run I would like to have pre-built Python wheels for all platforms so `pip install python-igraph` would fall back to running `python setup.py install` on the source code only if the user has some kind of an obscure Python installation, in which case we would be screwed anyway :)

vtraag:
Fair enough!

szhorvat:
Yes, this was my fault, I explicitly downloaded the dev version of the C core using [the instructions here](https://github.com/igraph/python-igraph#compiling-the-development-version). I was under time pressure to install a Python 3 compatible version for a summer school student, and I am not yet used to working with the dev version of python-igraph.

iosonofabio:
hi all,
this PR has been merged, which breaks `python-igraph` because of the API change. I now opened a PR in `python-igraph` to fix this by an optional argument in Python:

https://github.com/igraph/python-igraph/pull/239

ntamas:
That's great, I'll merge it soon. However, this probably means that the build process in Travis is _not_ using the git submodule yet; instead of that, it tries to check out the master revision of the C code and link with that. We should clean this up in the Travis CI spec such that it simply runs `git submodule update --init` instead of getting the C core from elsewhere.

iosonofabio:
I'll go through the CI stuff anyway since we have to adapt it to produce artifacts for conda, pip, etc etc. so I'll clean up this aspect as well

vtraag:
Note that `conda-forge` feedstocks for [python-igraph](https://github.com/conda-forge/python-igraph-feedstock) and [igraph](https://github.com/conda-forge/igraph-feedstock) already exist.

iosonofabio:
hmm ok that looks useful

stevepowell99:
hi, I think this is not implemented in the R package yet, is there a workaround?

PR #1233, opened 2019-08-16
Corrected loop in graphlets

### LINKED ISSUE #869: graphlets() crashes
First I have to say that I am not familiar with graphlets yet, so I am very likely missing some points here (such as what weight values make sense).  I'll be showing the problem with the R interface, but it applies to the C interface as well.

First, `graphlets` will crash with any unweighted graph.  My understanding is that is requires weighted ones, but the R interface (not the C one) should never crash anyway.

Example:

```
g <- erdos.renyi.game(10,20,'gnm')
graphlets(g)
```

With C I get this:

```
Assertion failed: (v != NULL), function igraph_vector_size, file ./vector.pmt, line 481.
```

Second, depending on the weights passed, this will sooner or later crash or hang irrecoverably:

```
g<-erdos.renyi.game(10,20,'gnm')
graphlets(g, weights=runif(ecount(g)))
```

When this happens, and I'm calling the C interface from Mathematica, the Mathematica kernel prints this error, which might be helpful:

```
MathKernel(92230,0x7fff74b41300) malloc: *** error for object 0x7fcd43c2e328: incorrect checksum for freed object - object was probably modified after being freed.
```

vtraag:
This fixes #869 I believe. The error seemed to have been simply a for loop over the wrong number of items (which may have lead to some memory issues).

I also added `IGRAPH_CHECK` calls.

Finally, I had some trouble understanding how to properly destroy the returned cliques. It turns out that you have to call

```
  for (i = 0; i < igraph_vector_ptr_size(&cliques); i++)
  {
    igraph_vector_t* c = VECTOR(cliques)[i];
    igraph_vector_destroy(c);
    igraph_free(c);
  }
```
so, both `igraph_vector_destroy` and `igraph_free`. Perhaps this is already documented somewhere else, but I failed to find it. I therefore also made this more explicit in the documentation.

### DIFF
--- a/src/glet.c
+++ b/src/glet.c
@@ -498,7 +498,7 @@ int igraph_i_graphlets_filter(igraph_vector_ptr_t *cliques,
  * \param weights Weights of the edges, a vector.
  * \param cliques An initialized vector of pointers. 
  *        The graphlet basis is stored here. Each element of the pointer 
- *        vector will be a vector of vertex ids. 
+ *        vector will be a vector of vertex ids.
  * \param thresholds An initialized vector, the (highest possible)
  *        weight thresholds for finding the basis subgraphs are stored
  *        here.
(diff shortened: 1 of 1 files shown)


iosonofabio:
Hi guys, this is a stopper for 0.8, but it seems like Vincent already make a working PR... any comments?

szhorvat:
I commented in chat. There's an issue with the graphlet projection function. I get crashes due to double free. I have this on my TODO list and will send the feedback once I figured out what's going on.

vtraag:
Yes, I believe that the error was that a vector was being written to out of bounds, which lead to the memory problem. This is addressed in this PR. You found another problem @szhorvat ?

szhorvat:
Yes, I did. Please give me some time.

vtraag:
OK, sure!

szhorvat:
I am sorry, the problem was indeed with my code. This is good to go.

PR #1241, opened 2019-08-20
Correcting make dist

### LINKED ISSUE #1240: make dist not working
It seems that `make dist` is not working properly. After successfully building the package using `make`, it errors out on `make dist` with
```
make[4]: *** No rule to make target 'bliss_bignum.hh', needed by 'distdir-am'.  Stop.
```
I can't locate `bliss_bignum.hh` in the tree either.

vtraag:
This fixes #1240, that is, the second problem that was noticed, the missing `structural_properties_internal.h` file when running `make dist`.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -36,7 +36,7 @@ install-info:
 dist-hook:
 	rm -rf `find $(distdir)/examples -type d -name .arch-ids`
 
-MAINTAINERCLEANFILES = Makefile.in
+MAINTAINERCLEANFILES = Makefile.in aclocal.m4
 
 ## to make sure make deb will generate Debian packages
 .PHONY: framework msvc parsersources
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -281,7 +281,8 @@ HEADERS_PRIVATE = atlas-edges.h \
 		scg_headers.h igraph_hacks_internal.h triangles_template.h \
 		triangles_template1.h maximal_cliques_template.h prpack.h \
 		igraph_cliquer.h cliquer/graph.h cliquer/cliquer.h cliquer/misc.h \
-		cliquer/cliquerconf.h cliquer/reorder.h cliquer/set.h
+		cliquer/cliquerconf.h cliquer/reorder.h cliquer/set.h \
+		structural_properties_internal.h
 
 HEADERS_PUBLIC =../include/igraph.h 		../include/igraph_memory.h    \
 		../include/igraph_random.h 	../include/igraph_types.h     \
--- a/tests/Makefile.am
+++ b/tests/Makefile.am
@@ -23,7 +23,7 @@ TESTSUITE_AT = \
 	other.at foreign.at conversion.at flow.at community.at eigen.at \
 	cliques.at attributes.at arpack.at bipartite.at centralization.at \
 	version.at separators.at hrg.at microscopic.at mt.at random.at scg.at \
-	matching.at qsort.at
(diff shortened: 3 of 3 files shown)


vtraag:
Let me first check if everything is now correctly built.

ntamas:
Okay - feel free to rebase and merge when the builds pass.

vtraag:
I believe it now correctly generates a source distributions by running `make dist`. The resulting distribution correctly builds using `make` and successfully completes `make check`. 

I ran into a problem with generated expected output by some tests because there were some versioning conflicts. Running `make distclean` did not solve this. I believe this problem is now solved by commit https://github.com/igraph/igraph/pull/1241/commits/f777ea7b320faefccda2c3cc703cdbe25bfe80c1. Any comments or thoughts on this @ntamas?

ntamas:
Seems good to me!

ntamas:
Footnote: running `make maintainer-clean` seems to delete one of the essential files in `test/` that prevents one from running `make check` without regenerating everything with `autoconf` and `automake`. Not sure if it's a problem, and I don't even know whether this was introduced with this patch or not (I did not even know about `make maintainer-clean` before looking at this patch and trying to figure out what `MAINTAINERCLEANFILES` does).

vtraag:
The problem I had was that somehow some old version information was retained. The expected output of `dot.out`, `cattributes2.out` and `gml.out` still had an old version number, whereas the compiled version would have the most recent version number. Running `make check` then encounters errors because of the differences in version numbers.

Running `make clean` or `make distclean` and then running `./configure && make && make check` did not resolve that situation. Even removing the `configure` script and rerunning `./bootstrap.sh` did not correct this.

Perhaps I should have run `make maintainer-clean` instead and then restart from `./bootstrap.sh`, making commit https://github.com/igraph/igraph/commit/f777ea7b320faefccda2c3cc703cdbe25bfe80c1  unnecessary? I'll try it out.

ntamas:
Well, I don't know how it works now, but ideally, the test outputs that include the version number should be generated or updated when the user runs `make dist`. This way we could ensure that the distributed tarballs always contain the up-to-date test outputs that match the version number of the tarball. If the user checks out the repository, then the test outputs should either be generated by `./bootstrap.sh`, or when the user runs `make check`.

There should also be a way for regenerating these test outputs from the command line because it could be an annoyance during development to update the test outputs manually all the time when the git commit hash changes.

Or, even better, we should get rid of the version numbers in the outputs completely or make `autotest` ignore them somehow :)

vtraag:
At the moment, the configure script does not seem be regenerated with a new version number if you don't run `make maintainer-clean`. This has also consequences for making the tarball during `make dist` because the tarball is named using the old version, while the new version is actually compiled. It then includes the test files that use the old version, but since the new version is actually compiled it gives a conflict when running `make check` from the source code in the generated tarball.

It seems that this is impossible to resolve without removing all generated files, since the version is only correctly passed on when rerunning from an entirely clean tree from scratch starting with `./bootstrap.sh`. I think that running `make maintainer-clean` should be sufficient to clean the tree, but I am still investigating.

szhorvat:
> At the moment, the configure script does not seem be regenerated with a new version number

You know, that's been bugging me forever, but I couldn't figure out how to fix it (other than with a clean start, as even a second run of `./bootstrap.sh` isn't sufficient). Thanks for sorting this out!

vtraag:
I sorted it out, see #1243. It turns out that you need to call `make maintainer-clean` and then start from scratch using `./bootstrap.sh`.

Issue #1242, opened 2019-08-20
Undefined reference to 'igraph_sparsemat'


advaitb:
Hi,

1) I've installed the igraph c library successfully, however while running the sparsemat functions I run into ld error caused by undefined reference. I don't seem to have this problem when accessing other functions like vector_t or matrix_t. I had a look at Issue #988 and made the respective changes but I'm not sure how to resolve this.Any idea how I can use sparsemat? 

`/tmp/ccklFFmC.o: In function `main':
test.cpp:(.text+0xce): undefined reference to `igraph_sparsemat_init(igraph_sparsemat_t*, int, int, int)'
test.cpp:(.text+0x174): undefined reference to `igraph_sparsemat_entry(igraph_sparsemat_t*, int, int, double)'
test.cpp:(.text+0x1ec): undefined 
reference to `igraph_sparsemat(igraph_s*, igraph_sparsemat_t const*, int)'
test.cpp:(.text+0x24b): undefined reference to `igraph_sparsemat_destroy(igraph_sparsemat_t*)'
collect2: error: ld returned 1 exit status`

2) Also, is there anyway to convert spmatrix into a graph?

Thanks!

Make sure that these boxes are checked before submitting your issue -- thank you!

- [x] This issue is for the C core of igraph.
- [ ] This issue is a bug report or a feature request, not a support question.

vtraag:
Could you share some minimal reproducible code that shows the exact mentioned problem? What version of `igraph` did you install? 

To answer your second question, yes you can use [`igraph_sparsemat`](https://igraph.org/c/doc/ch07.html#igraph_sparsemat).

advaitb:
Hi @vtraag , thanks for your prompt response. I'm using v0.7.1. Here is the code I'm trying to run:
```
#include <igraph.h>
#include <cstdlib>
#include <vector>
#include <iostream>
#include <ctime>
#include <string>
#include <igraph_sparsemat.h>


#define SIZE 900000
using namespace std;


void printConsole(string printThis){
    time_t result = time(nullptr);
    cout<<asctime(localtime(&result))<<" "<<printThis<<endl;
}


int main(){

        igraph_t graph;
        igraph_sparsemat_t mat;
        igraph_vector_t cores;
        igraph_sparsemat_init(&mat, SIZE, SIZE, 805000000000);
        long i;
        printConsole("Start populating adjmat");
        for (i=0;i<SIZE-700000;i++){
[...]

PR #1244, opened 2019-08-21
Corrected include files

vtraag:
This PR changes `igraph.h` so that it can be included directly in `C++` without necessitating `extern "C"` (see #1242). 

I have also changed the example/test files to ensure that only `igraph.h` is included instead of any separate headers such as `igraph_sparsemat.h` since they are already included in `igraph.h`. I also changed any non-public headers (e.g. `igraph_types_internal.h`) to be included using `"..."` instead of `<...>` to clarify the difference.

### DIFF
--- a/examples/simple/2wheap.c
+++ b/examples/simple/2wheap.c
@@ -22,7 +22,7 @@
 */
 
 #include <igraph.h>
-#include <igraph_types_internal.h>
+#include "igraph_types_internal.h"
 #include <time.h>
 #include <stdlib.h>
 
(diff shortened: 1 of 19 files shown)


szhorvat:
All the other headers used the constructs `__BEGIN_DECLS` and `__END_DECLS`. Should we use these in the main header as well, for consistency?

vtraag:
That definition is incorrect, since it defines the `extern C` at compile time. In other words, when built as a C library `extern C` isn't included, so you still get problems when including it in a C++ project. I tried to those definitons completely, but that generates some problems elsewhere.  I therefore thought to simple leave it in, rather than spending time tracking that down.

PR #1248, opened 2019-08-27
Corrected all s-t cuts for disconnected source-target pairs

### LINKED ISSUE #1236: all_st_cuts: inconsistent behaviour when s and t are not connected
`igraph_all_st_cuts` behaves in an inconsistent way when `s` and `t` are not connected.

Sometimes it returns no cuts at all, and sometimes it returns a single cut of zero size.

Here's a small example using the (unfinished) Mathematica interface:

image

The edge list is:

```
1,2
2,3
1,4
4,3
```

For `s=3, t=1` we get no cuts at all.

For `s=2, t=4` we get a single cut of zero size.

Either would be an acceptable result (it's up to interpretation), but the result should be the same in both cases.

vtraag:
This fixes #1236, where incorrect results were obtained for disconnected source-target pairs.

The problem was that `GammaS` contained the source node by default if `S` is empty. However, in the paper it is (tacitly) assumed that the source and target nodes are connected, in the sense that all nodes in `GammaS` are contained in the nodes in the dominator tree. This created an inconsistency, which is resolved by ensuring that nodes that are not in the dominator tree are not part of `GammaS`.

### DIFF
--- a/src/st-cuts.c
+++ b/src/st-cuts.c
@@ -742,12 +742,6 @@ int igraph_i_all_st_cuts_pivot(const igraph_t *graph,
 				     &leftout, IGRAPH_IN));
   IGRAPH_FINALLY(igraph_destroy, &domtree);
 
-  /* Relabel left out vertices (set K in Provan & Shier) to
-     correspond to node labelling of graph instead of SBar. */
-  n = igraph_vector_size(&leftout);
-  for (i = 0; i < n; i++)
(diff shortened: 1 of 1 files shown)


szhorvat:
I didn't look at the algorithm, but I did check carefully that the results are correct. I think it's good to go.

One weird thing: running my tests for this with AddressSanitizer _through the Mathematica interface_ leaks memory badly. Without AddressSanitizer, it's fine. It may even be some bad interaction between AddressSanitizer and Mathematica—I didn't check.

PR #1253, opened 2019-09-03
Corrected all s-t mincuts

### LINKED ISSUE #1235: all_st_mincuts returns empty cuts
For many different inputs, `all_st_mincuts` returns _some_ empty results, in addition to the correct results.

Here's a small example in R:

```R
g<-make_graph(c(4,3, 3,1, 3,2, 2,1))
plot(g)
cuts<-st_min_cuts(g,3,1)
cuts$cuts
```

The output is

```
[[1]]
+ 2/4 edges from 6cc9c03:
[1] 3->1 3->2

[[2]]
+ 2/4 edges from 6cc9c03:
[1] 3->1 2->1

[[3]]
+ 0/4 edges from 6cc9c03:
```

The graph looks like this:

image

This function should return all minimum cuts (compare to `igraph_all_st_cuts` which returns all minimal cuts, and can thus be used to verify the result from `igraph_all_st_mincuts`).  Therefore, the size of all returned cuts should be the same.

Possibly related:

 - https://github.com/igraph/igraph/issues/204
 - https://github.com/igraph/igraph/issues/628
 - https://lists.nongnu.org/archive/html/igraph-help/2014-06/msg00005.html

vtraag:
This fixes #1235 I believe.

There seemed to have been two problems. In [Provan & Shier (1996)](10.1007/s004539900020) there are two ingredients that seemed to have gone wrong in the implementation.
1. In Eq. (3.8a) effectively the target node `t` is excluded as a pivot element `v`. This was not implemented. I must admit that Eq. (3.8a) is also a bit strangely written, since only pivot elements `v` of indegree 0 are admitted, and so `t` can never reach `v`, hence the latter condition of (3.8a) only holds for the case that `v==t`.
2. It seemed that the following sentence was wrongly interpreted: "After this preprocessing has been done, the set M0 can be found by successively removing vertices of indegree 0 in G3" (p. 365). I believe the "succesive" here refers to the fact that we repeatedly need to find the set M0. It was now programmed to succesively remove vertices of indegree 0 for simply finding M0. I have corrected this.

Finally, it seemed that for one example included in the test, the results were actually already incorrect. Because the expected output was incorrect, this was of course not noted previously. I believe this is now also corrected.

### DIFF
--- a/src/st-cuts.c
+++ b/src/st-cuts.c
@@ -1185,7 +1185,7 @@ int igraph_i_all_st_mincuts_pivot(const igraph_t *graph,
   const igraph_vector_bool_t *active=data->active;
 
   long int no_of_nodes=igraph_vcount(graph);
-  long int i;
+  long int i,j;
   igraph_vector_t Sbar_map, Sbar_invmap;
   igraph_vector_t keep;
   igraph_t Sbar;
@@ -1239,21 +1239,21 @@ int igraph_i_all_st_mincuts_pivot(const igraph_t *graph,
     /* OK, we found a pivot element. I(S,v) contains all elements
        that can reach the pivot element */
     igraph_vector_t Isv_min;
-    long int isvlen;
     IGRAPH_VECTOR_INIT_FINALLY(&Isv_min, 0);
     *v=(long int) VECTOR(Sbar_invmap)[ (long int) VECTOR(M)[i] ];
     /* TODO: restricted == keep ? */
     IGRAPH_CHECK(igraph_bfs(graph, /*root=*/ (igraph_integer_t) *v,/*roots=*/ 0,
-			    /*mode=*/ IGRAPH_IN, /*unreachable=*/ 0,
-			    /*restricted=*/ &keep, /*order=*/ &Isv_min,
-			    /*rank=*/ 0, /*father=*/ 0, /*pred=*/ 0,
-			    /*succ=*/ 0, /*dist=*/ 0, /*callback=*/ 0,
-			    /*extra=*/ 0));
-    for (isvlen=0; isvlen<no_of_nodes; isvlen++) {
-      if (!IGRAPH_FINITE(VECTOR(Isv_min)[isvlen])) { break; }
+          /*mode=*/ IGRAPH_IN, /*unreachable=*/ 0,
+          /*restricted=*/ &keep, /*order=*/ &Isv_min,
+          /*rank=*/ 0, /*father=*/ 0, /*pred=*/ 0,
+          /*succ=*/ 0, /*dist=*/ 0, /*callback=*/ 0,
+          /*extra=*/ 0));
+    for (j=0; j<no_of_nodes; j++) {
+      igraph_real_t u = VECTOR(Isv_min)[j];
+      if (!IGRAPH_FINITE(u)) { break; }
+      if (!igraph_estack_iselement(T, u))
+        igraph_vector_push_back(Isv, u);
     }
-    igraph_vector_resize(&Isv_min, isvlen);
-    igraph_vector_update(Isv, &Isv_min);
     igraph_vector_destroy(&Isv_min);
     IGRAPH_FINALLY_CLEAN(1);
   }
--- a/src/st-cuts.c
+++ b/src/st-cuts.c
@@ -1232,7 +1232,8 @@ int igraph_i_all_st_mincuts_pivot(const igraph_t *graph,
   nomin=igraph_vector_size(&M);
   for (i=0; i<nomin; i++) {
     long int min=(long int) VECTOR(Sbar_invmap)[ (long int) VECTOR(M)[i] ];
-    if (!igraph_estack_iselement(T, min)) { break; }
+    if (min != target)
+      if (!igraph_estack_iselement(T, min)) { break; }
   }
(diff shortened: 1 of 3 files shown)


szhorvat:
I did some randomized tests on this and found a graph where the result looks strange.

Here's the edge list:

```
[[0, 10], [0, 17], [1, 16], [3, 18], [4, 8], [5, 2], [5, 9], [6, 1], [8, 12], [8, 19], [9, 8], [9, 16], [10, 0], [10, 6], [10, 15], [11, 8], [11, 17], [12, 13], [12, 14], [13, 9], [14, 2], [14, 18], [14, 19], [15, 0], [15, 8], [15, 18], [16, 17], [17, 19], [18, 2], [18, 15]]
```

Vertices are 0, 1, ..., 19.

Note that there's an isolated vertex, and that the graph is directed.

This is the result I get:

image

Python format:

```
[[[0, 10], [0, 17]], [[10, 15], [17, 19]], [[15, 8], [17, 19]], [[0, 10], [17, 19]], [[10, 15], [17, 19]], [[15, 8], [17, 19]]]
```

All returned results are correct, but two of them are returned twice, which is strange.

szhorvat:
Smaller example:

image

Result for cuts from 0 -> 8 is:

image

Notice the many duplicates.

Edge list of this graph:

```
[[0, 4], [0, 7], [1, 6], [2, 1], [3, 8], [4, 0], [4, 2], [4, 5], [5, 0], [5, 3], [6, 7], [7, 8]]
```

vtraag:
I've simplified the problematic case to this
original
with edge list
```
[(0,2),  (0,4),  (1,4),  (2,1),  (2,3),  (3,0),  (3,5),  (4,5)]
```
It yields the cuts for source `0` and target `5`:
```
Found 6 cuts.
Cut 0:
0 - 2
0 - 4
Partition 0: 0

Cut 1:
2 - 3
4 - 5
Partition 1: 0 2 1 4

Cut 2:
3 - 5
4 - 5
Partition 2: 0 2 1 4 3

Cut 3:
0 - 2
4 - 5
Partition 3: 0 4

Cut 4:
2 - 3
4 - 5
Partition 4: 0 4 2 1

Cut 5:
6 7
3 - 5
4 - 5
Partition 5: 0 4 2 1 3
```
The last two are duplicates. There should be six cuts, but the last two cuts should include edge `0-4` instead of `4-5`.

I have not yet found anything that is actually going wrong, I'm investigating.

vtraag:
I've simplified it even further to
original
With source `0` and target `4` this yields the cuts:
```
Found 4 cuts.
Cut 0:
0 - 2
0 - 3
Partition 0: 0

Cut 1:
2 - 4
3 - 4
Partition 1: 0 2 1 3

Cut 2:
0 - 2
3 - 4
Partition 2: 0 3

Cut 3:
2 - 4
3 - 4
Partition 3: 0 3 2 1
```
The last cut should include edge `0-3` instead of `3-4`.

vtraag:
I've managed to figure it out. Again, this is not really explicit in [Provan & Shier (1996)](https://github.com/igraph/igraph/pull/10.1007/10.1007/s004539900020), but the sets `S` and `T` should be disjoint (see Lemma 2.1). Because Eq. (3.8b) simply states that `Isv` should include all nodes that can reach `v` in `Sbar`, this was not properly considered. Nodes that were already included in `T` could in this way end up in `S`.

Another consideration is that the actual partitions and the edges that are cut should be considered only for edges that have positive flow. This is not made explicit in Provan & Shier (1996), but otherwise you end up with edges that are cut which are not necessary. Note that when viewed as a partition, this works slightly differently. For example, in the example provided in https://github.com/igraph/igraph/pull/1253#issuecomment-527820920, the partition `0, 2` will generate the cut edges `0 -> 3` and `2 -> 4`, which are correct, but it also includes the cut edge `2 -> 1`, which should not be included because it has no positive flow (i.e. it cannot reach the target). This is now correctly handled.

Finally, the error was actually already in the test case again! The fourth test case listed the same cut twice. This is now corrected, and the cuts are printed more explicitly.

src/st-cuts.c
```diff
@@ -1240,27 +1232,28 @@ int igraph_i_all_st_mincuts_pivot(const igraph_t *graph,
...
+          /*mode=*/ IGRAPH_IN, /*unreachable=*/ 0,
+          /*restricted=*/ &keep, /*order=*/ &Isv_min,
+          /*rank=*/ 0, /*father=*/ 0, /*pred=*/ 0,
+          /*succ=*/ 0, /*dist=*/ 0, /*callback=*/ 0,
+          /*extra=*/ 0));
+    for (j=0; j<no_of_nodes; j++) {
+      igraph_real_t u = VECTOR(Isv_min)[j];
+      if (!IGRAPH_FINITE(u)) { break; }
+      if (!igraph_estack_iselement(T, u))
+        igraph_vector_push_back(Isv, u);
```
szhorvat: I don't quite follow whre `Isv` originally came from, but unless it is guaranteed that this `push_back` will never need to allocate more storage for it, there should be an `IGRAPH_CHECK` on this.

vtraag: Thanks!

ntamas:
Good to go then! Thanks a lot!

PR #1256, opened 2019-09-08
Documentation: Spell check intro

szhorvat:
Fixed some language/spelling in the introduction to the documentation.

Removed mention of the Ruby interface, as it seems dead.

Added mention of the Mathematica interface.


vtraag:
One thing by the way is that the documentation is not compiled on the CI. Should we set this up perhaps, to ensure that documentation builds correctly?

PR #1257, opened 2019-09-09
Cycle basis

iosonofabio:
DO NOT MERGE YET!

Trying to implement functions for finding a cycle basis:
- starting from undirected, unweighted graphs
- undirected, weighted graphs in progress

**Final TODO before merging**:
- [ ] finish implementation
- [ ] rebase onto most recent `develop` branch
- [ ] check for memory leaks
- [ ] special checks for `igraph_vector_ptr` since current convention is shaky
- [ ] convert tabs to spaces
- [ ] check docs
- [ ] check tests
- [ ] add 1-2 examples


szhorvat:
Just some very quick notes without looking through it fully:

 - This function should return a `ptr_vector_t`, each element of which is a basis element, i.e. a `vector_t` of edge indices.
 - It must work for multigraphs, e.g. `1-2, 1-2, 2-3` has one cycle.
 - It should work on directed graphs and simply ignore edge directions. One may want to use the result to create an edge-cycle matrix where 0, 1 and -1 are used to indicate whether an edge is no in a cycle, in the cycle pointing in the cycle direction, or in the cycle pointing in the reverse direction, respectively. This is a common thing to do with cycle bases.

You seem to be using some raw `char *` arrays for boolean vectors. We also have `igraph_vector_bool_t`. I did notice that an `igraph_bool_t` is an `int`, i.e. typically 32-bit instead of 8-bit. Perhaps this is why you went with the raw `char` array? @ntamas Can you comment on why `igraph_bool_t` is `int` and not, say, and `unsigned char`?

ntamas:
I have no idea; this has always been this way before I joined the development of igraph. I think we could change this easily. You could try changing the typedef of `igraph_bool_t` to `unsigned char` in the source and see if anything breaks (it shouldn't - if it does, then the point where it breaks is probably an interesting place to look at as it could uncover places where we are (ab)using an `igraph_bool_t` to do something else).

src/cycle_basis.c
```diff
@@ -0,0 +1,217 @@
...
+ * \return Error code.
+ * 
+ * Time complexity: ???.
+ * 
+ * \sa \ref igraph_maximum_cardinality_search().
+ */
+int igraph_cycle_basis_unweighted_undirected(const igraph_t* graph,
+    igraph_vector_t* cycles) {
+
+    // FIXME: deal with unconnected graphs
```
ntamas: Check whether MSVC in C89 mode is okay with comments starting with double slashes. I have always tried to avoid this because I wasn't sure whether MSVC likes it or not and it's too complicated for me to get hold of a Windows machine :)

szhorvat: MSVC 2015 and later do support `//` comments as "a Microsoft extension", unless standards mode is triggered. https://docs.microsoft.com/en-us/cpp/c-language/c-comments?view=vs-2015 This can also be tested e.g. at godbolt.org

I _think_ earlier versions support it too, but I do not have proof.

I always went with `/* ... */` _just in case_. You never know what compilers people use on what weird platforms that we don't have access to. Maybe someone uses an IBM compiler on a Power architecture machine. I'd be in favour of sticking with the standard (even though there are already several `//` comments in the igraph code base—my IDE keeps flagging them actually).

gaborcsardi: `//` is C99 I believe. Older MSVC compilers do no support it. I don't know what Python 2.x uses now to compile its extensions, but R is C99, so if you want to use them I don't mind.

gaborcsardi:
> I have no idea; this has always been this way before I joined the development of igraph. I think we could change this easily. You could try changing the typedef of igraph_bool_t to unsigned char in the source and see if anything breaks

This breaks the R interface because is there a logical is an `int`.

szhorvat:
Some more comments:

I think the following would be reasonable names / prototypes:

```c
int igraph_fundamental_cycle_basis(const igraph_t *graph, igraph_vector_ptr_t *basis);
int igraph_minimum_cycle_basis(const igraph_t *graph, igraph_vector_ptr_t *basis, const igraph_vector_t *weights);
```

Alternatively, `minimum_weight_cycle_basis` would be more precise, but longer.

----

I do not think we need to include "undirected" in the name, as this is understood. In many applications, people would run this on directed (i.e. oriented) graphs and also want to know if each edge points along the cycle direction or opposite to it. This information could even be optionally returned by the function in an additional vector (however, it is easy enough to compute, so maybe we don't need to include it).

I do not think we need to include "unweighted" in the name, as weights are not relevant for bases unless we look for the minimum weight one (which is a special application). However, it is useful to indicate that it returns a fundamental basis.

src/cycle_basis.c
szhorvat: There are two separate issues:

 1. Is it okay to use arbitrary C99 features? No, because even MSVC 2019 doesn't support them fully. We should not use e.g. variable-length arrays for this reason.

 2. Is it okay to use `//` comments? Compilers typically support `//` as an extension, even if they do not fully support C99. So it might not be a problem in practice. It is claimed that [MSVC 2010 already supported `//` comments](https://softwareengineering.stackexchange.com/q/224747/29979).

vtraag:
I believe the implementation can be made more efficient in the following way.

While doing a BFS, you maintain a vector `father` that denotes for each node the node through which it was first visited (i.e. this corresponds to the spanning tree). Let us initially set `father` to `-1` for all nodes, so that if `father[i] < 0` it means that node `i` was not yet visited. In addition, we keep a vector `dist` which is the distance from the root vertex. Whenever you encounter a node `i` that was already visited (`father[i] > 0`) when traversing the edge `j -> i`, you backtrack the fathers of `i` and `j` (i.e. ancestors) until `dist[i] == dist[j]` and then keep iterating until the ancestors are identical. In other words:
```C
int ancestor_i = i;
int ancestor_j = j;
while (ancestor_i != ancestor_j)
{
  if (dist[ancestor_i] > dist[ancestor_j])
    ancestor_i = father[ancestor_i];
  else if (dist[ancestor_i] < dist[ancestor_j])
    ancestor_j = father[ancestor_j];
  else
  {
    ancestor_i = father[ancestor_i];
    ancestor_j = father[ancestor_j];
  }
}
```
By maintaining the traversed list of ancestors, you have the list of vertices in the cycle. If `ancestor_i == j` you know that the cycle was directed and otherwise it is an undirected cycle (i.e. a cycle ignoring the edge directions). If you want a vector of edges instead of vertices (which is probably more useful in a multigraph) you can use edge ids instead of vertex ids to identify the father (i.e. the edge that leads to the father). Finally, for multiple components you should include a `root` vector that identifies from which root the node was reached. If `root[i] != root[j]` no cycle can be identified.

iosonofabio:
Thank you Vincent, you mean the BFS happens instead of my DFS in the code? Or outside of the for loop? Or we build the tree as we backtrack already?

vtraag:
I believe that the initial BFS that you perform is sufficient for immediately identifying the cycles, while creating the spanning tree. In other words, you identify the cycles as you are building the spanning tree in one go. You now repeatedly perform a DFS on the resulting spanning tree, but you will then repeatedly backtrack the same paths I believe (i.e. the `parent` variable remains essentially identical across `igraph_dfs` runs).

vtraag:
And probably you should indeed call it `parent` instead of `father` :)

iosonofabio:
Sounds good, I'll get rid of those DFS loops :-)

I knew it was optimizable, but wanted to have a rough version out there to start the discussion

ntamas:
@iosonofabio If you are still working on this, feel free to remove the `wontfix` label. Stalebot will also stop bothering this PR if you add the `pinned` label to it.

iosonofabio:
I restarted working on this one, don't worry about the massive merge commits for now, I can squash-rebase at the end.

vtraag:
Great! I noticed you merged in `develop` so perhaps we should discuss whether this should go into `master` or `develop`. If my understanding is correct, this is an entirely new function, right? That means it can go into `master`, because it is backwards compatible. But that means that we should not bring in all the changes from the `develop` branch. What do you think?

iosonofabio:
Good point. I'm working on the weighted graph function which is significantly more messy than the unweighted one, so I'm unclear on the time frame - not so familiar with Horton cycles I admit. There's also no urgency I believe. So maybe develop is fine? But I'm also fine rebasing against master

ntamas:
I'd say keep it on `develop`. As for me, I tend to do all development on `develop` now.

iosonofabio:
Ok, I'm sharpening the tools for this one. Some advice would be appreciated, but in terms of general approach and specific code. Two points in particular:

1. It seems reasonable to me to implement this algorithm: https://dl.acm.org/doi/abs/10.1145/1644015.1644023, see fig 2. Anyone objects or has a better suggestion?

2. I starting shaping up the code for that and it needs a bunch of auxiliary functions that might be useful beyond this particular PR, e.g. an approximate calculation of a feedback vertex set. If anyone could have a quick look at my `igraph_minimum_cycle_basis` (only a stub) and comment on what functions exist already and whether any of the private ones should instead be lifted to be part to the API, I'be be grateful.

Thanks!
Fabio

vtraag:
> 1. 
> It seems reasonable to me to implement this algorithm: https://dl.acm.org/doi/abs/10.1145/1644015.1644023, see fig 2. Anyone objects or has a better suggestion?

I actually ran into the same publication recently. It sounded like an interesting option, although I have not yet explored the details.

include/igraph_vector.h
```diff
@@ -136,6 +136,11 @@ __BEGIN_DECLS
     do { IGRAPH_CHECK(igraph_vector_bool_init(v, size)); \
         IGRAPH_FINALLY(igraph_vector_bool_destroy, v); } while (0)
 #endif
+#ifndef IGRAPH_VECTOR_CHAR_INIT_FINALLY
```
ntamas, line 139: Definitely useful - let's keep this.

src/cycle_basis.c
```diff
@@ -0,0 +1,462 @@
...
+	igraph_i_shortest_nonorthogonal_cycle(
+			candidate_cycles, &candidate_weights, &nonorth_cycles, basis);
+
+    }
+
+    return IGRAPH_SUCCESS;
+}
+
+
+int igraph_i_feedback_vertex_set_approx(const igraph_t *graph,
```
ntamas, line 379: `igraph_feedback_arc_set()` already contains an `algorithm` argument that opens up the possibility of implementing different approximation heuristics right within `igraph_feedback_arc_set()`. I propose adding whatever you plan to implement to `igraph_feedback_arc_set()` instead.

ntamas:
@iosonofabio `igraph_feedback_arc_set()` already has an `algo` argument that allows us to implement various heuristics for the feedback arc set problem. I think that you should add the particular approximation that you want to use there.

szhorvat:
This feedback vertex set is a bit different from what we have because it concerns undirected graphs. What we have is for directed graphs. Specifically, in undirected graphs, the feedback edge set problem is trivial (same as spanning tree) while the feedback vertex set problem is NP-complete.

Our feedback arc set functionality could use some expansion, and I do have [some stuff](https://arxiv.org/abs/2001.01440) coming for that, if I ever get to it ...

In particular, we should add a mapping between the feedback edge set (FES) and feedback vertex set (FVS) problems in directed graphs.  Solve FEP based on FVP is easy (take the line graph), the other direction not so much.

ntamas:
Ah, sorry, my bad, I mistook feedback _vertex_ sets for feedback _edge_ sets.

iosonofabio:
Ok guys, tea time again, thank you if you have 5 minutes to answer the following:

1. calculating a shortest path/minimum spanning tree **rooted** at a specified vertex?
2. argsort? sorry I'm more of a Pythonista and there we just have a util function do the thing for us...

Thanks!
Fabio

szhorvat:
> not so familiar with Horton cycles I admit

Do you need a short summary of Horton's algorithm (which is relatively simple, and not inefficient), or do you want to do the one the paper you linked?

iosonofabio:
I can read the paper, that's not the problem. How much difference do you think it will make in practice versus the increased complexity?

include/igraph_vector.h
iosonofabio: Would you like me to make a separate PR onto `develop` just with this? Happy to do it!

vtraag:
>     2\. argsort?

If you do need this is practice, I think it would make a great contribution to the `igraph_vector*` types. It would probably be good to make this a public implementation then.

src/cycle_basis.c
```diff
@@ -0,0 +1,660 @@
...
+
+		/* find the neighbour identity */
+                igraph_integer_t from, to;
+                igraph_edge(graph, (igraph_integer_t) edge, &from, &to);
+		
+		/* we only know that either one is actnode, swap if needed */
+                if (to==actnode) {
+		  to = from;
+		  from = to;
+		}
```
szhorvat, line 132: Noting in case you were not aware:

We have `IGRAPH_TO` and `IGRAPH_FROM`, which are both simpler to use and more efficient than `igraph_edge` (they avoid any function call).

There is also `IGRAPH_OTHER`. If an undirected `edge` is `u-v` then `IGRAPH_OTHER(graph, edge, v)` gives `u`. Did you need `IGRAPH_OTHER(graph, edge, actnode)` here?

src/cycle_basis.c
```diff
@@ -0,0 +1,660 @@
...
+          /* current node, incoming edge, and distance from tree root */
+          long int actnode=(long int) igraph_dqueue_pop(&q);
+	  long int actedge=(long int) igraph_dqueue_pop(&q);
+          long int actdist=(long int) igraph_dqueue_pop(&q);
+
+	  /* check all edges connected to this vertex */
+          IGRAPH_CHECK(igraph_incident(graph, &inc_edges, (igraph_integer_t) actnode,
+				       IGRAPH_ALL));
+
+          for (j=0; j<igraph_vector_size(&inc_edges); j++) {
```
szhorvat, line 118: Please avoid using `vector_size` in `for` loops like this. Instead, get the size first, store it in a variable, then use that variable instead.

Here it would not matter that much because there are already several other function calls in the loop, but in general it's good to get into the habit of storing the size in a variable. In a tight loop, that function call would bring a noticeable performance hit.

src/cycle_basis.c
```diff
@@ -0,0 +1,660 @@
...
+        added_nodes[i] = 1;
+
+	/* The queue stores the next vertex, the edge to it, and its distance to root */
+        IGRAPH_CHECK(igraph_dqueue_push(&q, i));
+        IGRAPH_CHECK(igraph_dqueue_push(&q, -1));
+        IGRAPH_CHECK(igraph_dqueue_push(&q, 0));
+        while (! igraph_dqueue_empty(&q)) {
+          /* current node, incoming edge, and distance from tree root */
+          long int actnode=(long int) igraph_dqueue_pop(&q);
+	  long int actedge=(long int) igraph_dqueue_pop(&q);
```
szhorvat, line 111: Let's not use tab characters for indentation.

src/cycle_basis.c
```diff
@@ -0,0 +1,660 @@
...
+ *    direction is ignored.
+ * \param basis A pointer to a vector of vectors. Each fundamental cycle
+ *    will be a member as a list of edge ids.
+ * \return Error code.
+ * 
+ * Time complexity: O(|V| + |E|) (spanning tree) + ?? (backtracking).
+ * 
+ */
+int igraph_fundamental_cycle_basis(const igraph_t *graph,
+    igraph_vector_ptr_t *basis) {
```
szhorvat, line 57: Don't forget that the user may pass in a non-empty `vector_ptr`. Clear the `vector_ptr` when you receive it (an analogous issue had to be corrected in Yasir's PR as well).

src/cycle_basis.c
```diff
@@ -0,0 +1,660 @@
...
+	IGRAPH_CHECK(igraph_i_shortest_nonorthogonal_cycle(
+			candidate_cycles, &candidate_weights, &nonorth_cycles, basis));
+
+    }
+
+
+    /* Clean */
+    igraph_vector_destroy(&fvs);
+    igraph_vector_ptr_destroy_all(trees);
+    IGRAPH_FINALLY_CLEAN(1 + n + 1); // FIXME: Check this number
```
szhorvat, line 383: This may be a useful read:

https://igraph.discourse.group/t/who-should-clean-up-vector-ptr-out-argument-in-case-of-interruption-or-error/153

Unfortunately, the handling of vector_ptrs in an error conditions is not completely standardized yet.

I am not aware of any existing case where all vector_ptr elements are put on the finally stack, and maybe it would be better not to start doing this. See the above thread for how this is usually handled.

Something to pay attention to:

Should the function exit due to an error condition *or* interruption, it should be safe for the caller to destroy/free any non-NULL elements of the ptr_vector. Thus, anything you destroy must be set to NULL. Alternatively, if everything was destroyed, the vector_ptr can be cleared.

As far as I can tell, existing functions do not leave any elements undestroyed or unfreed in an error condition.

Tip: `igraph_Free(ptr)` will set `ptr` to `NULL`, but `igraph_free` will not (cannot).

src/feedback_vertex_set.c
```diff
@@ -0,0 +1,90 @@
...
+/**
+ * \ingroup structural
+ * \function igraph_feedback_vertex_set
+ * \brief Calculates a feedback vertex set of the graph.
+ *
+ * </para><para>
+ * A feedback vertex set is a set of vertices whose removal makes the graph acyclic.
+ *
+ * </para><para>
+ * The problem is NP complete except in undirected graphs of maximum degree three.
```
szhorvat, line 47: The *minimum* feedback vertex set is NP-complete. Finding an arbitrary feedback vertex set (FVS) is not.

I assume this function will use a heuristic to find a "reasonably small" FVS, but not a minimum one because it would be counterproductive to use an exponential complexity algorithm as a subroutine in a polynomial-time one. 

Can you please clarify these in the description?

src/cycle_basis.c
szhorvat: We need to do this https://github.com/igraph/igraph/issues/1296

iosonofabio:
Thanks @szhorvat . At this stage I'm not fixing memory leaks and similar stuff, but thanks.

I haven't seen any comments on the general strategy, I'm just pushing ahead then. I'll come back to the nitty gritty details once the core of the algorithm is implemented :+1:

szhorvat:
@iosonofabio Is the fundamental cycles part considered to be ready now?

For the minimum cycle basis, I am familiar with Horton's algorithm, which is simple:

 - find fundamental cycles corresponding to a BFS tree starting from each vertex
 - order the entire group of cycles by weight
 - starting from the smallest, use Gaussian elimination to find a linearly independent set

I am not familiar with the paper you linked. I'd need to look at it in detail to comment, which would take a few hours. I need to force myself off igraph for now, so I can make progress on other projects ...

iosonofabio:
Fantastic @szhorvat , thanks for the recap. The paper is an incremental improvement over that. Don't worry too much about this one, it'll take me a few more days ;-)

src/cycle_basis.c
ntamas: Speaking about tabs vs spaces: I think we have an `.editorconfig` at the root, but not all editors pick up the settings from `.editorconfig` by default, so we need another safety mechanism to prevent misformatted code from slipping into the codebase. Initially I wanted to propose using [pre-commit](https://pre-commit.com/) for validating the codebase with pre-commit hooks, but then I realized that third-party contributors would probably not have `pre-commit` installed and then we would be back to square one. Do you guys know about any Github plugin that could be used to validate code style in the PRs, or do we have to create a CI test for that?

src/cycle_basis.c
ntamas: > I am not aware of any existing case where all vector_ptr elements are put on the finally stack

The size of the finally stack is limited so probably not all elements would fit there anyway. So far I was following the principle that the vector_ptr takes "ownership" of any pointer that you put into it, so you only need to put a cleanup function on the finally stack that iterates over the elements of the vector_ptr and destroys and frees the non-null ones. This function can then be popped off the finally stack when the vector_ptr is returned to the user. The "item destructor" pattern outlined in the Discourse group is supposed to help with the implementation.

include/igraph_vector.h
iosonofabio: PR done and merged

src/feedback_vertex_set.c
iosonofabio: I changed the name appending `_approx` and clarified a little in the function doc. Implementation of a decent algorithm is still missing, returning all vertices for now.

src/cycle_basis.c
iosonofabio: This is copycat of the exact same pattern in `spanning_trees.c`, e.g. line 321. I think the compiler will put the `igraph_vector_size(&inc_edges)` into a variable anyway, so I'm not sure a human should deal with these kinds of optimizations.

So if this is a performance problem because the compiler does not do its job, I can open a separate PR to fix it everywhere else too. @ntamas @vtraag ?

src/cycle_basis.c
iosonofabio: This is also a copycat from `spanning_trees.c`. Shall I make a PR to fix that?

src/cycle_basis.c
ntamas: I have been surprised several times in the past by the complexity of optimizations compilers can do automatically, but I'm still unsure :)

On one hand, yes, it should be the compiler's job to optimize this and leave it to us mere mortals to deal with the more interesting parts of programming. On the other hand, in order to do the optimization here in this particular case, the compiler (let it be `gcc`, `clang`, MSVC or something else) must be able to infer that 1) `igraph_vector_size()` is side-effect-free (okay, that's actually quite easy), and that 2) the thing that `igraph_vector_size()` calculates is an invariant of the loop so it can be lifted to an external variable (which might even end up being in a register only). If the compiler can infer this, that's great.

As for me, I routinely lift the size of a vector into a variable on its own and then use that variable in the loop -- compilers were not that good in the past and old habits die hard. Also, it gives me peace of mind to do so because then I can say that it probably works even on a crappy compiler. However, if someone leaves the function call in the loop, that's also okay if I know it's a simple function call that does not do anything complex.

src/cycle_basis.c
ntamas: I think it could be fixed in `spanning_trees.c` as well. That file is a very old part of the codebase, probably well before the time we've added `IGRAPH_TO` and `IGRAPH_FROM`. It probably does not make any difference performance-wise, but it helps people to pick up best practices if we have less places in the codebase where we don't use the shortcuts we've implemented to make our life easier in the long run.

src/cycle_basis.c
szhorvat: The compiler cannot infer that `igraph_vector_size` is side-effect free because it does not have the definition of that function. It is in a different translation unit.

Link-time optimization might be able to mitigate this to some extent, but it's not guaranteed, and LTO does not even work on all platforms. We should not rely on it.

szhorvat: @iosonofabio In C++ one can just do `for (int i=0; i < vec.size(); ++i)` because `vec.size()` will be an inline function that lives in the header. The compiler can analyse it and figure out that it does not need to call it for every loop iteration. In C, this is not possible because there is no general support for inline functions and the definition will simply not be visible to the compiler.

It is good practice to make a habit of avoiding this pattern. It's very easy to do so and then you don't have to keep thinking about how big the performance impact is anyway.

src/cycle_basis.c
szhorvat: > It probably does not make any difference performance-wise

It depends. If this is *the only* function call in a loop, eliminating that call may very well make a noticeable difference. Function calls are expensive.

vtraag:
FYI, I just came across `igraph_vector_order` (and it siblings `igraph_vector_order1` and `igraph_vector_order1_int`), which essentially provide an argsort of a vector. This really shows that we should make these functions public and provide decent documentation for them.

iosonofabio:
This one is also not urgent, esp. since @szhorvat has implemented a similar algorithm. Closing for now.

PR #1259, opened 2019-09-17
Added basic implementation of Leiden algorithm.

vtraag:
This is a basic implementation of the [Leiden algorithm](https://dx.doi.org/10.1038/s41598-019-41695-z) for community detection. It closely follows the [Java implementation](https://github.com/cwtsleiden/networkanalysis) of the algorithm.

Any comment regarding coding style, memory management or `igraph` coding standards (i.e. how to best do things) is welcome.

Some things remain to be done; I believe I have included all of them below. If some todo item is missing, let me know, so that I can add it to the list. I just wanted to already share this with you so that you can already take a closer look.

Todo list:
- [x] Add test cases.
- [x] Add documentation.
- [x] Add checks for correct graph types (e.g. undirected, etc...)
- [x] Add additional interface function to use modularity / CPM more easily.
- [x] Double check memory cleaning (also in case of `IGRAPH_FINAL`).
- [x] Add `IGRAPH_ALLOW_INTERRUPTION`.

Please only merge after all todo items are finished.


src/community.c
```diff
@@ -3567,3 +3567,878 @@ int igraph_i_compare_communities_rand(const igraph_vector_t *v1,
...
+  igraph_vector_bool_destroy(&neighbor_cluster_added);
+  igraph_vector_destroy(&edge_weights_per_cluster);
+  igraph_stack_destroy(&empty_clusters);
+  igraph_vector_int_destroy(&nb_nodes_per_cluster);
+  igraph_vector_destroy(&cluster_weights);
+  igraph_vector_destroy(&node_order);
+  igraph_dqueue_destroy(&unstable_nodes);
+  igraph_vector_bool_destroy(&node_is_stable);
+
+  IGRAPH_FINALLY_CLEAN(8);
```
vtraag: This should be `IGRAPH_FINALLY_CLEAN(9)` (noted by @szhorvat ).

src/community.c
```diff
@@ -3567,3 +3567,878 @@ int igraph_i_compare_communities_rand(const igraph_vector_t *v1,
...
+    /* Add neighbours to unstable_nodes  */
+    if (best_cluster != current_cluster)
+    {
+        VECTOR(*membership)[v] = best_cluster;
+
+        for (i = 0; i < degree; i++)
+        {
+          long int e = VECTOR(*edges)[i];
+          long int u = (long int)IGRAPH_OTHER(graph, e, v);
+          if (!VECTOR(node_is_stable)[u] && VECTOR(*membership)[u] != best_cluster)
```
vtraag: Quite crucially, this should be `if(VECTOR(node_is_stable)[u] ...` (i.e. not the negation).

vtraag:
I believe I covered everything that I wanted to cover. 

I decided not to implement a "simpler" interface, this could be easily done in any interface implementing the Leiden algorithm. The `node_weight` should be set to the degree to optimize modularity (or the strenght in the case of weights), and the `resolution_parameter` divided by 2m. This is also explained in the documentation.

@szhorvat, feel free to do some further checking.

vtraag:
Sorry, one more thing to do, I still have to ensure that the documentation is actually compiled (in `community.xxml`), doing that now.

src/community.c
```diff
@@ -3567,3 +3567,1094 @@ int igraph_i_compare_communities_rand(const igraph_vector_t *v1,
...
+int igraph_community_leiden(const igraph_t *graph,
+  const igraph_vector_t *edge_weights, const igraph_vector_t *node_weights,
+  const igraph_real_t resolution_parameter, const igraph_real_t beta,
+  igraph_vector_t *membership, igraph_integer_t *nb_clusters, igraph_real_t *quality) 
+{
+  igraph_vector_t *i_edge_weights, *i_node_weights;
+  int ret;
+
+  if (igraph_vector_size(membership) != igraph_vcount(graph)) 
+    IGRAPH_ERROR("Membership length does not equal the number of vertices", IGRAPH_EINVAL);
```
szhorvat: Other community detection functions only require the membership vector to be initialized, but not to have a certain size. It will be resized automatically to the correct size.

szhorvat:
A general comment: 

This contribution is huge. It would be better to put it in its own file to keep things more manageable.

src/community.c
vtraag: Yes, that is true, but this function starts optimisation from a given membership vector, which will be updated inplace. It would be possible to check for a membership vector of an incorrect length and then initialize it with a singleton partition. However, that may only lead to confusion if people pass a membership vector of the correct length, without intending to. This is why I opted for this solution.

vtraag:
> It would be better to put it in its own file to keep things more manageable.

Yes, that may be a good idea. In general, `community.c` is indeed quite bloated because of the large number of functions. I have no objection against including a separate source file, but it feels somewhat strange to only do this for `igraph_community_leiden`.

src/community.c
```diff
@@ -3567,3 +3567,1094 @@ int igraph_i_compare_communities_rand(const igraph_vector_t *v1,
...
+  /* Check edge weights to possibly use default */
+  if (!edge_weights)
+  {
+      i_edge_weights = igraph_Calloc(1, igraph_vector_t);
+      if (i_edge_weights == 0)
+        IGRAPH_ERROR("Leiden algorithm failed, could not allocate memory for edge weights", IGRAPH_ENOMEM);    
+      IGRAPH_CHECK(igraph_vector_init(i_edge_weights, igraph_ecount(graph)));
+      IGRAPH_FINALLY(free, i_edge_weights);
+      IGRAPH_FINALLY(igraph_vector_destroy, i_edge_weights);
+      igraph_vector_fill(i_edge_weights, 1);
```
szhorvat: Minor comment: The `igraph_vector_t` `struct` could have been allocated on the stack for simplicity, saving some `free`/`calloc`. (Of course, the actual vector storage still needs to be allocated/freed conditionally on whether the user supplied any weights).

vtraag: That might be possible. However, I am not sure how I then do
```
i_edge_weights = edge_weights;
```
in case that `i_edge_weights` is allocated on the stack?

src/community.c
```diff
@@ -3567,3 +3567,1094 @@ int igraph_i_compare_communities_rand(const igraph_vector_t *v1,
...
+  /* Perform actual Leiden algorithm */
+  ret = igraph_i_community_leiden(graph, i_edge_weights, i_node_weights,
+                                      resolution_parameter, beta,
+                                      membership, nb_clusters, quality);
+
+  if (!edge_weights)
+  {
+    igraph_vector_destroy(i_edge_weights);
+    igraph_Free(i_edge_weights);
+    IGRAPH_FINALLY_CLEAN(1);
```
szhorvat: I think this should be `2` and not `1` (one for the `vector_destroy`, one for the `free`).

src/community.c
```diff
@@ -3567,3 +3567,1094 @@ int igraph_i_compare_communities_rand(const igraph_vector_t *v1,
...
+  }
+
+  if (!node_weights)
+  {
+    igraph_vector_destroy(i_node_weights);
+    igraph_Free(i_node_weights);
+    IGRAPH_FINALLY_CLEAN(1);
```
szhorvat: Same as above, should be `2`.

src/community.c
vtraag: Something going wrong with GitHub. My comment triggered a new review by me, which I then deleted, but it also deleted the original comment.

It might be possible to use the stack instead of the heap, but I am not sure how to do
```
i_edge_weights = edge_weights;
```
in that case.

src/community.c
szhorvat: E.g. `igraph_community_leading_eigenvector` has a separate boolean `start` argument, which controls whether the user-supplied `membership` vector should be used as a starting point, or whether the membership vector should be initialized automatically.

szhorvat: Here's how it does the checks: https://github.com/igraph/igraph/blob/master/src/community.c#L1591

vtraag: OK, good point. I will also use that.

vtraag: N.B. `nb_clusters` should also be checked or set correctly (in the case `start=TRUE`).

src/community.c
szhorvat: We could do

```
igraph_vector_t i_edge_weights_vec;
igraph_vector_t *i_edge_weights;
```

Then set `i_edge_weights = &i_edge_weights_vec;` or `i_edge_weights = edge_weights;` as needed.

But again, it's not a big deal.

szhorvat:
> but it feels somewhat strange to only do this for `igraph_community_leiden`.

InfoMAP is also in a separate file (possibly only because it's C++, not C).

src/community.c
vtraag: Thanks, that would be a possibility indeed. But I think I then prefer the current solution.

vtraag:
OK, I will create a new file `community_leiden.c` then and move all relevant functions there.

ntamas:
In the long run it would definitely be a good idea to separate the various community detection methods into their own files.

To be honest, I would even start structuring the `src` folder a bit based on "topics"; e.g., all community detection functions could be placed in `src/communities`, all flow-related functions could be placed in `src/flow`, all shortest path-related functions could be placed in `src/paths` and so on. The smaller files we have, the better it is, and probably it also makes recompilations faster - a single change in one of the community detection algorithms won't recompile all the other algorithms.

One thing that we have to be careful about is that the filenames should be unique even if they are in different folders; I think that [Visual Studio does not like having multiple files with the same name in different folders](https://stackoverflow.com/questions/14364362/visualstudio-project-with-multiple-sourcefiles-of-the-same-name).

src/community.c
```diff
@@ -3567,3 +3567,1100 @@ int igraph_i_compare_communities_rand(const igraph_vector_t *v1,
...
   return IGRAPH_SUCCESS;
 }
+
+/* Clean membership.
+ *
+ * This function ensures that the membership vector is renumbered in consecutive
+ * numbers 0, ..., q, where q is the number of distinct clusters. The number of
+ * distinct clusters is also returned in \c nb_clusters.
+ */
+int igraph_i_community_leiden_clean_membership(igraph_vector_t *membership, igraph_integer_t* nb_clusters)
```
ntamas: I think we already have this functionality in `igraph_reindex_membership()`.

vtraag: Yes, true.

I though the existing method was a bit inefficient because of the unnecessary sorting that is done.

But if you think it is better to use the already existing functions I can do so. From a maintenance point of view you would be right I think.

I could also make the method more efficient (but use the same call signature of course). But my current implementation makes some assumptions, which may not be warranted in all function uses. That is why I didn't touch it originally. I could take a closer look.

vtraag:
Good point about updating the overall structure! Perhaps it would be best to open a separate issue for that?

I could still move everything for this contribution to `community_leiden.c` but leave the rest as is.

src/community.c
ntamas: I'd rather have a bit more inefficient implementation than the duplication of functionality across the codebase. It could also be an option to make `igraph_reindex_membership()` more efficient by replacing its implementation with yours.

ntamas:
> Perhaps it would be best to open a separate issue for that?

I think so.

vtraag:
`RNG_BEGIN` and `RNG_END` should be added to the relevant parts.

src/community.c
vtraag: `nb_clusters` does not need to be checked, it is set when cleaning the membership.

vtraag:
> `RNG_BEGIN` and `RNG_END` should be added to the relevant parts.

Done

szhorvat:
Test suite fails on macOS because one of the expected output lines is

    Leiden found 10 clusters using modularity, quality is -nan.

while the actual output is

    Leiden found 10 clusters using modularity, quality is nan.

Maybe it's best to avoid printing `nan`.

ntamas:
...or simply use `isnan()` in the test case to test for nan-ness.

vtraag:
OK, I did not know that it failed on MacOS, thanks for letting me know. We don't have a CI setup for MacOS apparently?

I will correct the test.

szhorvat:
> We don't have a CI setup for MacOS apparently?

Yes, maybe we should add one on Travis, if possible.

ntamas:
Travis now builds igraph on macOS as well. (Unfortunately I could not get it to rebuild this PR -- @vtraag, maybe you need to rebase your PR on master to make it run on macOS as well).

vtraag:
Thanks @ntamas. I've cherry picked the commit instead, rebasing would require me to force push to the PR invalidating earlier comments.

szhorvat:
@vtraag Do you have recommendations for reasonable default for the parameters, to be used by a high-level interface?

For beta, you recommend 0.01 as a starting point.

There was the case of using the modularity as the quality function, as described in the docs: vertex weights are chosen to be vertex strengths and `resolution = 1/(2*edge_count)`.

Do you have any other tips/comments on automatically choosing reasonable values?

szhorvat:
Your Python package is actually a good starting point: https://leidenalg.readthedocs.io/en/latest/reference.html#modularityvertexpartition

szhorvat:
As far as I'm concerned, this is good to go. @ntamas if you have no further comments either, please merge.

vtraag:
Thank you for the feedback @szhorvat ! 

The reasonable default for modularity is a resolution parameter of 1, since it then corresponds to the original definition. 

For CPM, I prefer not to give any default resolution parameter, as it forces people to make an active choice. A reasonable range is somewhere around the average (weighted) density of the network. Whether that is most suitable for any particular network really depends on the needs and interpretations of the results by the user. The property of being a resolution-limit-free method depends on not having such a default resolution parameter (see http://arxiv.org/abs/1104.3083). This is a fine thin line, but I think it is worthwhile to stress this by not having such a default parameter in practice.

iosonofabio:
I've been using cpm for biology data a lot and I agree with Vincent, it really depends on the application

ntamas:
Thanks a lot!

szhorvat:
I posted in the chat, but not sure @vtraag has seen it. 

I believe that on this line:

https://github.com/igraph/igraph/blob/master/examples/tests/igraph_community_leiden.c#L61

`gamma` should be set to `1.0 / (2*sum_of_weights)` rather than `1.0 / (2*igraph_ecount(graph))`.

Of course, it's just a small test case, but it also serves as an example for people who want to use the function. It's better to have it be conceptually correct.

PR #1265, opened 2019-11-18
Motifs: check size of cut probability vector

szhorvat:
This will check that the size of the cut probability vector is not smaller than the motif size (which, as far as I could tell, would result in an out-of-bounds access even though it never crashed for me).

For convenience reason, and also so that we do not have to modify the tests, I still allow longer `cut_prob` vectors than the motif size.

I opted not to check that the probabilities are all `0 <= p <= 1`. Values < 0 or > 1 are effectively treated as 0 and 1, but I think they won't cause trouble.


ntamas:
Thanks!

PR #1273, opened 2019-12-20
Perform range checking when reindexing membership.

vtraag:
This addresses the issue that the build of `python-igraph` now [fails](https://travis-ci.org/igraph/python-igraph/builds/626633793). This is because a cluster index is supplied that is outside the range of 0, ..., n - 1, where n is the number of nodes. This is now explicitly checked and an error is thrown.

This does not yet resolve the failing build of `python-igraph`, which I need to address in a separate PR for `python-igraph`.


ntamas:
Thanks!

PR #1279, opened 2020-01-16
Updated instructions for compilation on Windows.

vtraag:
This addresses #969. Please do not merge yet however. I will check whether the instructions work correctly on my Windows machine tomorrow. There are also (very short) instructions on the website for installation on Windows, we probably also need to update those. I will take a look at that tomorrow.


ntamas:
Thanks, this will be very useful - also for me to get things up and running with Appveyor in the `python-igraph` repo!

ntamas:
One comment: in Appveyor, `devenv.exe` was not on the PATH for some reason so I had to add its full path manually. Not sure if this is the case with an ordinary Visual Studio installation.

vtraag:
Thanks for the heads up. I'll look into it.

I'm already running into some other issues before that step. For example, http://mingw.org only provides a 32-bits version, it seems that https://www.msys2.org provides the easiest way to to have a 64-bits version. I'll make adjustments accordingly.

szhorvat:
I believe MinGW and MinGW-w64 are separate projects. You should only use the latter.

I linked this before, but will put it here again: https://github.com/szhorvat/IGraphM/blob/master/Development.md#windows This is (mostly) how I compile igraph on Windows with msys2.

I'll try to test your instructions, but I won't get to it before the freeze date (19th).

vtraag:
Yes, they indeed are separate projects, from what I understand. MinGW will not provide any 64-bits support, so indeed `msys2` is the way forward. Unfortunately I am running into some issues when installing `msys2`, presumably related to this issue https://github.com/msys2/MSYS2-packages/issues/1532. Unfortunately I am not in a position to update the BIOS (which may be the cause of the issue), so I'm now trying to take a detour, by simply using the tarball from http://repo.msys2.org/distrib/x86_64/ instead of the `.exe`.

vtraag:
I can confirm that [this](https://github.com/msys2/MSYS2-packages/issues/1532#issuecomment-514589920) solution indeed provides a workaround for me with the problem I encountered. I am now proceeding through the remaining steps to see what other problems I encounter.

vtraag:
I run into a problem while running `./configure` in `msys2`:

```
Something went wrong bootstrapping makefile fragments
for automatic dependency tracking. Try re-running configure with the
'--disable-dependency-tracking' option to at least be able to build
the package (albeit without support for automatic dependency tracking).
```

I have now run `./configure` with `--disable-dependency-tracking` as suggested. This then runs fine, but I run into another problem with `make`, where it cannot find `sys/times.h`. I am not sure whether that is related, I'll look into that later. I just wanted to report this already here. It seems that this may be a problem of `automake`, I'm running version `1.16.1`. There are some changes to the dependency tracking in the `1.16` version of `automake`, see [release notes](https://lists.gnu.org/archive/html/info-gnu/2018-02/msg00008.html). I'm not sure if that causes the problem.

I have already pushed some changes to the instructions, so that if anyone wants to test this further, they at least already have these improved instructions. I will continue to look into this over the weekend (if I can find some time).

szhorvat:
> Something went wrong bootstrapping makefile fragments

I have never seen this, even after building igraph with msys2 on more than one computer.

> where it cannot find `sys/times.h`

You need to add `-DMSDOS` to `CPPFLAGS` before running `./configure`. This was required by one of the dependencies (f2c? I don't remember which one).

szhorvat:
Here's what I usually do, in as much detail as I can remember:

 - Install msys2, update components. See https://wiki.qt.io/MSYS2
 - Install needed packages using `pacman -S`. We need at least `mingw-w64-x86_64-toolchain`. I do not remember the msys2-specific names of the other needed packages, but try at least `automake`, `autoconf`, `bison`, `flex`, `libtool`, `git`.
 - We get up to three terminals to work with, "msys", "mingw 32-bit" and "mingw 64-bit". Use the last one to compile igraph.
 - I build igraph with gmp and without graphml support. I'll go on with this assumption to keep things consistent, even though you'd want to *not* disable graphml support (which requires additional dependencies).
 - Build GMP from source and install it. I usually install into `$HOME/local`, meaning that I pass `--prefix=$HOME/local` to `./configure`. From some reason, igraph won't link properly unless I remove the file `libgmp.la` from `PREFIX/lib`. Thus, remove it. I have not yet been able to understand the details of this.
 - Clone the igraph repo with git
 - `./bootstrap.sh`
 - Assuming GMP is in `$HOME/local`, first set `export CPPFLAGS="-I$HOME/local/include -DMSDOS" LDFLAGS=-L$HOME/local/lib`  The `-DMSDOS` part is necessary.
 - `./configure --prefix=$HOME/local --disable-graphml`
 - `make -j4` (or however many cores you have)
 - `make install`
 - Now add `PREFIX/bin` to your `PATH` environment variable. The igraph `.dll` needs to be in the path before we can run the tests
 - `make check`

These are the steps I actually follow. There are parts that I do not fully understand, and there are things that can probably be done better. Thus I would not publish these steps are they are. But they worked without issue on several Windows machines, so they are a good starting point for polishing things.
 -

vtraag:
Thanks for reporting the steps you are taking @szhorvat! That should prove useful. I have been using `cygwin` earlier, which yielded fewer problems it seems. Could you check whether you also encounter the error with the dependency tracking in the most recent installation of `msys2` (provided you can install it—otherwise use the tarball)?

szhorvat:
I regularly update msys2 with `pacman -Syuu`, and it does work fine.

I can try to uninstall and re-install it and see what happens.

iosonofabio:
Oh another Archer!!

vtraag:
You seem to found out the hint to pass the `MSDOS` flag @szhorvat [here](https://lists.nongnu.org/archive/html/igraph-help/2015-09/msg00057.html), for `f2c` indeed, in these files:
https://github.com/igraph/igraph/blob/c517ad26e537bbdf363a1f6f9cabd66040664fdd/src/f2c/dtime_.c#L3-L6
https://github.com/igraph/igraph/blob/c517ad26e537bbdf363a1f6f9cabd66040664fdd/src/f2c/etime_.c#L3-L6

Unfortunately, I still have not been able to run `igraph` successfully using `msys2`, also not with the hints provided by @szhorvat [above](https://github.com/igraph/igraph/pull/1279#issuecomment-575879441). At this point, it does seem that `igraph` is compiled, but there are some linking issues. I will look into that later (no more time today I'm afraid).

Is there any reason why it is necessary to compile `igraph` on `msys2`? Frankly, the process seems quite tedious and fragile to me, at least in comparison to the `cygwin` route. If my understanding is correct, the `msys2` route produces a DLL that can be used by native Windows libraries that people may want to develop, while `cygwin` does not, but I am not 100% sure if that is the case (?). But if a DLL is required in development work, it is probably just as easy (if not easier) to simply compile using MSVC. @szhorvat , do you need `msys2` to support the Mathematica interface on Windows? Or would it also be possible to use a DLL that is compiled using MSVC?

Finally, I ran into one point: the `patch` done in `bootstrap.sh` is not run because it cannot find the `patch` program. Perhaps it can be installed using `pacman`, I have not yet looked into it, but just wanted to raise another point. To be clear, it concerns this part:
https://github.com/igraph/igraph/blob/c517ad26e537bbdf363a1f6f9cabd66040664fdd/bootstrap.sh#L23-L26
I am wondering whether it is actually still necessary to do this, maybe libtool has been upgraded already indeed? Do you know @ntamas ? This patching seems a bit hacky, but perhaps there are good reasons to keep this, I am not sure.

ntamas:
Re the patch: it all depends on the version of `libtool` that you have on your computer when you invoke `bootstrap.sh`. If `libtool` is too old (not sure _how_ old, though - our patch is from 2017 so it was definitely a problem in 2017 even though [there is a patch from 2015](https://savannah.gnu.org/patch/?8775) that should have fixed this in `libtool`), the generated `ltmain.sh` script does not let the `-fsanitize=...` flags get passed through to the linker, which makes it impossible to test igraph with Clang's AddressSanitizer.

All in all, _not_ patching `ltmain.sh` is not a big deal; it breaks a few stuff that probably only the developers care about. Feel free to update the `bootstrap.sh` script to skip patching if `patch` is not installed.

vtraag:
Thanks @ntamas, perhaps good to check if patch is installed or not. Otherwise we might get question about that later. I'll do that later.

The thing I don't understand is that from `appveyor.yml` the compilation using `msys2` seems straightforward. Why I then encounter so many difficulties in compilation I don't understand. I'm comparing the output from the AppVeyor build to my local results to learn more.

szhorvat:
> Is there any reason why it is necessary to compile igraph on msys2? 

One might also ask: is there any reason not to (since it seemed to work for me on multiple machines)? I'm sure we'll be able to sort out the problems.

MinGW-w64 is much more useful than Cygwin (since Cygwin is not very interoperable).

I believe the R interface is compiled with MinGW-w64, though I'm not familiar with the details. The standard CRAN toolchain uses MinGW-w64.

The Mathematica interface is currently compiled with msys2 because I have not yet been able to get everything working with MSVC.  

I compile the Mathematica interface with GMP, which is essential for me. In fact, I would like to propose making GMP an essential part of igraph because several of the feature I plan to contribute require performant bigint arithmetic. GMP does not support MSVC at all because the developers are against it. This is solvable: MPIR is an API-compatible replacement, but using iy *only on Windows* is a pain and some extra work. Now that we have much more manpower we can handle it, but I haven't been able to find the time to do it alone for the Mathematica interface.

I also compile the Mathematica interface with an external GLPK because the newest version of GLPK performs better (I needed this for my work). This is again a bit problematic but not unsolveable with MSVC (the solution: WInGLPK).

If all issues get sorted out with MSVC, I would prefer switching the Mathematica interface over to MSVC, as that's Mathematica's preferred toolchain on Windows. However, as the above two examples (GMP and GLPK) illustrate, going for MSVC would have been *a lot more work*, while msys2, which can run autotools, was quite straightforward. Therefore I think there is great value in supporting msys2, and I am strongly in favour of it.

----

Regarding the problems you were having with msys2: I strongly suspect that this might be an exception rather than the rule. I compiled igraph on msys2 without problems on three different Windows machine. One of them was a Windows 7 VM set up just for this purpose, i.e. a clean slate. Therefore I'm confident that we will be able to debug the issues.

szhorvat:
@vtraag Since we're short on time, I propose that I review the msys2 parts of the instructions and make sure that they work on the two Windows systems I have access to. Let's just go with these imperfect instructions (we know they don't work on your system) for 0.8, as they're still much better than no instructions at all. 

Then we'll sort out the rest for 0.9.

vtraag:
> Regarding the problems you were having with msys2: I strongly suspect that this might be an exception rather than the rule. I compiled igraph on msys2 without problems on three different Windows machine.

Fair enough!

I would prefer to get it to work on my machine as well. It seems I'm making progress, and most tests now seem to work. There are still some things to look at (e.g. compiling with `libxml2`), which I think should be properly explained. If you can give me until Tuesday, I hope to have found enough time to have sorted it all out.

szhorvat:
> If you can give me until Tuesday

I thought the freeze today was meant to cover PRs adding new features. We were still going to work on fixes, documentation, installation instructions etc. this week, right?

ntamas:
Yes; PRs regarding documentation, homepage updates, bugfixes, code formatting etc are still okay during next week.

vtraag:
Yes, in theory I would have until the end of the week (or even until 28 January). But at some point I should also cut my losses, so if I'm still not there by Tuesday, we'll have to make do with what we have. There are still other things to be updated as well, and other non-igraph things, which also need to be done before the 28th. 

But it seems I'm making some progress though, so I'm still hopeful! :) I believe I now have a correctly compiled version of `igraph` in `msys2`, including GMP and XML support. I will now restart from scratch again to recheck my instructions. If all goes well, I will push new instructions for that part, and then continue with the MSVC compilation. Explaining compiling with an external GLPK library should not be necessary I think?

szhorvat:
> Explaining compiling with an external GLPK library should not be necessary I think?

Actually, it would be useful to note that GLPK needs to be compiled with `--disable-reentrant` when using MinGW, as this is a non-obvious issue. Unfortunately, I do not remember how I found the solution to this.  https://github.com/szhorvat/IGraphM/blob/master/Development.md#glpk-1

szhorvat:
I just compiled with full GraphML support in msys2 and ran the test suite. Things work, but there are two failures.

The Leiden tet failed because once the quality is computed to be 0.2813 but 0.2812 is expected. Clearly some numerical issue that's not a true failure.

SCG of a graph, stochastic matrix fails because complex numbers are printed differently:

`-3.14295e-16+0i` vs `-3.14295e-016+0i`

Again, not a true failure.

Things like this keep popping up from time to time as I update MinGW ... (I was aware of one of these "failures" before.)

vtraag:
For future reference, this is the problem I encounter when running the `msys2-x86_64-20190524.exe` installer:
image
Perhaps it is due to this issue https://bugreports.qt.io/browse/QTIFW-1110. I'm not sure how to resolve it though. This issue https://github.com/msys2/MSYS2-packages/issues/1532 suggests it may be a problem related to the BIOS (which I am in no position to update), although the reported error is somewhat different. Again, for future reference: I'm running Windows 10, version 1809, build 17763.973.

I will not explain this in the instructions, and simply assume it is a (hopefully) incidental glitch.

vtraag:
Thanks for testing @szhorvat ! These were indeed the two tests that still failed me, which are clearly not critical indeed.

vtraag:
It seems I have now a working set of instructions to compile on `msys2`. Some of my earlier problems were probably due to incorrect permissions when unpacking the tarball to `c:\msys64`. I removed all references to administrator rights, since that is probably a result of that faulty installation process.

I think it is possible to simply compile with the provided GMP version for MinGW `mingw-w64-x86_64-gmp`. This is automatically installed when installing `mingw-w64-x86_64-toolchain`. Similarly, I think it is better to compile with `mingw-w64-x86_64-libxml2` instead of `libxml2`. The latter gave some issues (some include files of `icu` were incorrectly not found), while the former worked flawlessly. Perhaps that should also be changed in `AppVeyor.yml`?

I would propose by the way to include the setting of `-DMSDOS` in `configure.ac`, so that this is automatically defined when compiling using MinGW. Do you agree? I actually still don't understand exactly how this is not a problem for the AppVeyor configuration, apparently `sys/times.h` is included from one path or another, but I'm not sure how.

szhorvat:
> When cloning from GitHub, please ensure that you run `git config --global core.autocrlf false` before cloning the repository. Otherwise, you might run into problems with line endings.

I never do this and have not had problems so far. Did you experience problems?

More importantly, will turning this off risk introducing CRLF (instead of LF) endings into the source files by a future contributor who uses Windows? <-- This is really my main concern and I do not know the answer.

szhorvat:
I checked the MSYS2 installation instructions on a borrowed clean-slate system that didn't have MSYS2 installed before. They work.

While testing them, I made a few edit. These are all minor things, and none are strictly necessary. @vtraag Please review and feel free to revert anything. There was no problem in the instructions themselves.

Minor nitpick:

MinGW and MinGW-w64 are separate projects. I believe the former is no longer maintained. MinGW-w64 is not simply a 64-bit version of MinGW. It is a different project that provides both 32-bit and 64-bit tools (and I *think*, though I'm not sure, that the 32-bit tools from MinGW-w64 are not quite compatible with the original MinGW).

I think the `patch` package is not needed to I removed it.

Mentioned that GMP can be installed to enable additional igraph functionality, though in practice GMP will already be installed as a dependency of some other package.

Corrected `--prefix [DIRECTORY]` to `--prefix=[DIRECTORY]`

Mentioned how to run the tests. We should look into fixing those failures.

"If you dispose of the full Microsoft Visual Studio" <-- I think the usual meaning of "dispose of" is "to get rid of".

Changes XXX to the actual https://igraph.discourse.group/ URL (which will be made public soon).

vtraag:
I have now checked the instructions for Windows, and I have updated them according to some issues that I ran into.

The compilations works fine, but there is one thing that should be clarified (which took me some time to figure out). By default the `igraph` project targets a static library (`.lib`), but the exports are suitable for a dynamic library (`.dll`) by using the `__declspec(dllexport)` keyword (upon compilation) and the `__declspec(dllimport)` keyword (upon inclusion in another project). When including the static library, one should then include `#define IGRAPH_NO_IMPORTS 1` to make sure that the static library names are properly used. If this is not properly done, you will run into linker issues that `__imp_*` external symbols are unresolved. However, I believe this is nowehere (yet) documented (this is also a recent addition https://github.com/igraph/igraph/commit/47ee0ce45902193ec56f58bed999368804c11de2). I have now clarified this in the instructions. I am not exactly sure how the static library was linked to previously actually. Perhaps that never worked properly before?

Regarding this aspect, there are two things that I would like to suggest. First of all, I think we should rename `IGRAPH_NO_IMPORTS` to `IGRAPH_STATIC` to clarify its intention (this is for example also how it is named in [`libssh`](https://vovkos.github.io/doxyrest/samples/libssh/page_libssh_linking.html)). People might get confused because they *do* want to import function definitions. Secondly, we need to make sure that the `igraphtest` that is now included can be properly compiled and runs correctly. We therefore need to add `#define IGRAPH_NO_IMPORTS 1` to `igraphtest.cpp`. In the long run, it might be good to include all tests from the test suite also in an MSVC project.

ntamas:
> However, I believe this is nowehere (yet) documented (this is also a recent addition 47ee0ce). I have now clarified this in the instructions. I am not exactly sure how the static library was linked to previously actually. Perhaps that never worked properly before?

Most likely that never worked properly before. I have figured this out the hard way when trying to make `python-igraph` compile in Appveyor. That's why it's also not documented - I came up with it only last week.

ntamas:
> First of all, I think we should rename IGRAPH_NO_IMPORTS to IGRAPH_STATIC to clarify its intention (this is for example also how it is named in libssh)

Sure, go ahead and send a PR. We also need to update `setup.py` in the Python interface to reflect the change, but it should be okay. I did not know that there are other examples of this in other projects.

vtraag:
Right now the igraph [website](https://igraph.org/c/) briefly explains how to install/compile on Windows. Putting the entire explanation in this PR on the website will be too much I feel. Shall I simply provide a link to these instructions instead? Or perhaps we can include these instructions as a separate page on the website, instead of as a file in the `igraph` GitHub repository?

ntamas:
`IGRAPH_NO_IMPORTS` is now renamed to `IGRAPH_STATIC`

ntamas:
I would create a separate page on the website. If you can wrap these instructions up nicely in a Markdown document, I can add it to the website. We also need to get rid of any references to Cygwin - if I understand correctly, Cygwin is outdated and no one tests it regularly from our end.

vtraag:
> `IGRAPH_NO_IMPORTS` is now renamed to `IGRAPH_STATIC`

Great! I now see that `libxml` actually also follows the same pattern (see [Notes](https://www.zlatkovic.com/projects/libxml/index.html)), so good to be consistent there.

I am actually trying to compile with GraphML support by using the provided Windows binaries for `libxml`. However, the `winclude` directory (copied from `msvc/include`) contains include files from `libxml`, but apparently outdated ones. That is to say, one particular symbol `XML_PARSE_HUGE` seems to be missing from the `parse.h` header, which is present in the include files provided by the Windows binaries from [Zlatkovic](https://www.zlatkovic.com/pub/libxml/). What is the reason to include those header files in `winclude`?

vtraag:
> I would create a separate page on the website. If you can wrap these instructions up nicely in a Markdown document, I can add it to the website. We also need to get rid of any references to Cygwin - if I understand correctly, Cygwin is outdated and no one tests it regularly from our end.

OK, I will create a separate page for that. I believe compilation on `cygwin` works, but the `msys2` instructions now seem to work flawlessly, including support of GMP and `libxml`. In the interest of keeping it simple, I agree that it would be better to focus on MinGW/`msys2` instead of `cygwin`.

ntamas:
> What is the reason to include those header files in winclude?

I don't remember any more - is there a standard mechanism in Windows that allows the discovery of these header files without hardcoding some path in the project file?

vtraag:
Since we moved the installation instructions to the website (see https://github.com/igraph/igraph.org/pull/17), I removed the instructions from the repository.

szhorvat:
Just an opinion: I think there's value in having the instructions directly in the repo as many people will look for them there, but won't look on the website. You decide what to do :-)

vtraag:
I prefer not to maintain two different places. We also don't provide installation instructions for Linux or macOS. Perhaps we can include a general INSTALL file that points to the website though.

szhorvat:
> Perhaps we can include a general INSTALL file that points to the website though.

Let's do that.

vtraag:
As far as I am concerned, we are good to go.

PR #1280, opened 2020-01-19
Create .zenodo.json

szhorvat:
Zenodo automatically extracts metadata about repositories from GitHub APIs. For example, the authors are determined from the repository's contributor statistics. This automatic extraction is solely a best guess. Therefore it is useful to set certain values explicitly in a `.zenodo.json` files.

This is in preparation for requesting a DOI from Zenodo, which should be set up *before* we create a release. It is creating a release that triggers auto-archiving on Zenodo. At that point Zenodo will take some of the data from this file.

### DIFF
--- a/.zenodo.json
+++ b/.zenodo.json
@@ -1,0 +1,13 @@
+{
+    "title": "igraph",
+    "upload_type": "software",
+    "keywords": [
+        "graph theory",
+        "network analysis"
+    ],
+    "creators": [
+        {
+            "name": "The igraph Development Team"
+        }
+    ]
+}


szhorvat:
We need to decide about how to present the authorship. For now, I put "The igraph Development Team". The main purpose of this file is not to let Zenodo just fetch the user accounts that contributed because that tends to end up containing things that don't belong (I ended up with the [Gitter Badger](https://github.com/gitter-badger) ...)

ntamas:
Having "The igraph development team" in Zenodo is okay for me. I'll let @vtraag and @gaborcsardi  chime in in the next few days before I merge this.

vtraag:
I'm fine with "The igraph Development Team" in Zenodo, sounds good!

gaborcsardi:
Fine with me as well.

PR #1281, opened 2020-01-19
Add more maximal clique functions

szhorvat:
This PR:

 - adds a callback-type function for maximal cliques
 - adds a function that computes a histogram of maximal clique sizes
 - fixes minor issues in error codes (e.g. missing strings) and adds comments

I want to get this in for 0.8 so that I can start building the Mathematica interface from the main igraph (instead of my fork)

Tests are forthcoming, but I believe this to be correct as I have been using it for a while.


szhorvat:
The function naming and the argument ordering is perhaps not ideal, but neither is it bad. I would keep it as is for 0.8, then maybe improve it for 0.9 when we'll have several other breaking changes anyway.

The function naming and argument order takes some work to change because of the templating mechanism that is being used.

szhorvat:
There was an issue with stopping the search with the callback function. It's fixed now.

I added tests (which discovered the above issue).

Ready to review/merge.

PR #1283, opened 2020-01-20
Add missing includes and other small fixes

szhorvat:
This PR adds some missing includes to get rid of `implicit-function-declaration` warnings and to make IDEs be happier when editing igraph source files.

There are also a few other minor changes:

 - re-encode some files to UTF-8 from ISO-Latin-1
 - change `LAPACK_H` header guard to `IGRAPH_LAPACK_H` (as `LAPACK_H` is too likely to be encountered in other projects)
 - Add some missing `__BEGIN_DECL` or `__END_DECL`
 - Add missing entries for `igraph_rng_type_t` structs

### DIFF
--- a/include/igraph_lapack.h
+++ b/include/igraph_lapack.h
@@ -21,8 +21,8 @@
 
 */
 
-#ifndef LAPACK_H
-#define LAPACK_H
+#ifndef IGRAPH_LAPACK_H
+#define IGRAPH_LAPACK_H
 
 #include "igraph_vector.h"
 #include "igraph_matrix.h"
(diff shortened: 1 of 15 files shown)


szhorvat:
One thing that bothers me and is unfixed in this PR is the use of `igraph_i_vector_binsearch_slice` in `community_leiden.h` without declaration. There should be an internal header for every internal function that we might want to use in other files.

ntamas:
> re-encode some files to UTF-8 from ISO-Latin-1

Can't we do it the other way round and re-encode everything in UTF-8 instead? After all, it's 2020 now :) Are there any (serious) IDEs out there that cannot cope with UTF-8?

szhorvat:
> Can't we do it the other way round and re-encode everything in UTF-8 instead? After all, it's 2020 now :) Are there any (serious) IDEs out there that cannot cope with UTF-8?

My phrasing was awkward. That's exactly what I did. It was Latin-1 and now it is UTF-8.

ntamas:
Thanks!

PR #1288, opened 2020-01-23
Correct compilation on Windows, including test

### LINKED ISSUE #1287: Compilation on Windows
Right now, the `igraphtest` solution does not immediately build correctly because of two things.

1. We need to add `#define IGRAPH_NO_IMPORTS 1` to `igraphtest.cpp`, as explained [here](https://github.com/igraph/igraph/pull/1279#issuecomment-577080247). As discussed in that thread, we will change the name of the symbol to `IGRAPH_STATIC`.

2. We need to correct the include/library directory. Right now, all include/lib directories refer to the `igraph-0.7-msvc` directory. There are two possible solutions for this problem:
  a. We generate the `igraphtest.vcproj` dynamically.
  b. We simply use the directory name `igraph-msvc` instead of `igraph-[DIRECTORY]-msvc`.

I would prefer option (b). When making the MSVC zip file for distributing the source when running `make msvc` we could still include the version number though.

vtraag:
This is intended to fix #1287. This corrects some aspects related to the `make msvc` command, which failed if there were no files yet in the `igraph-*-msvc` directory.

I still intend to look at compiling with `libxml` support, so please do not merge yet.

Another issue that I preferably would like to correct in this PR is to automatically include the `-DMSDOS` flag when compiling using MinGW.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -65,9 +65,9 @@ msvc: parsersources
 	cp -r $(top_srcdir)/include/*.h $(top_builddir)/igraph-$(VERSION)-msvc/include
 	if [ "x$(top_srcdir)" != "x$(top_builddir)" ]; then cp -r $(top_builddir)/include $(top_builddir)/src $(top_builddir)/igraph-$(VERSION)-msvc; fi
 	cp -r $(top_srcdir)/msvc/include $(top_builddir)/igraph-$(VERSION)-msvc/winclude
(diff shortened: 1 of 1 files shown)


ntamas:
> Another issue that I preferably would like to correct in this PR is to automatically include the `-DMSDOS` flag when compiling using MinGW.

Sorry, I missed this part and merged the PR without it. Feel free to open a new one.

PR #1289, opened 2020-01-23
Compilation on Windows, part 2

vtraag:
This is the continuation of #1288, fixing #1287. 

PLEASE DO NOT MERGE YET

I have now set -DMSDOS for compilation when sys/times.h is missing automatically in `configure.ac`.

Next point to address will be to compile with `libxml2` support.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -41,8 +41,10 @@ AC_LANG_POP([C++])
 AM_PROG_LEX
 AC_PROG_YACC
 
-AC_CHECK_HEADER([sys/times.h], 
-      [AC_DEFINE([HAVE_TIMES_H], [1], [Define to 1 if you have the sys/times.h header])])
+AC_CHECK_HEADER([sys/times.h],
+      [AC_DEFINE([HAVE_TIMES_H], [1], [Define to 1 if you have the sys/times.h header])],
+      [CPPFLAGS="$CPPFLAGS -DMSDOS"],
+      )
 
 AC_LIBTOOL_WIN32_DLL
 AC_LIBTOOL_DLOPEN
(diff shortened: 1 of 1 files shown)


szhorvat:
Would it not be better to try to make the test more specific?

I wonder if there is a good way to detect an msys2 environment.

ntamas:
The `MSYSTEM` environment variable is always defined in an MSYS2 environment.

szhorvat:
One additional issue when building on msys2 is that one must explicitly add the location of the igraph DLL to the `PATH` before `make check` will work. It would be nice (but not a 0.8-blocker) to do this automatically on that platform.

vtraag:
> more specific

I actually thought this would be the most specific. In the AppVeyor build environment apparently the `sys/times.h` is provided (but I'm not sure from where). Hence, testing for `msys2`, or more precise, for MinGW, would be less specific, not more specific. Of course we could also test for the combination: both `sys/times.h` is missing *and* we are compiling using MinGW? I believe we can test that using something like [`AC_CANONICAL_HOST`](https://www.gnu.org/software/autoconf/manual/autoconf-2.69/html_node/Canonicalizing.html)?

vtraag:
I'm looking into compilation with `libxml`. This might be best supported using `vcpkg`, which seems to work quite nicely. Do you have any experience with this? I'm not sure if I will finish that in time, so perhaps that will have to wait.

vtraag:
I'm not sure about how to solve the `make check` thing, suggestions are welcome.

vtraag:
As far as I am concerned, we can merge this. I don't have further time to complete compilation with `libxml` from [Zlatkovic](https://www.zlatkovic.com/projects/libxml/index.html). I believe the detection of `sys/times.h` works satisfactorily, for now at least.

Issue #1208, opened 2019-06-11
Add tests for igraph_closeness


szhorvat:
There are no tests for `igraph_closeness`. We need to add tests both for the weighted and unweighted cases.

A reminder about the guidelines: https://github.com/igraph/igraph/wiki/Checklist-for-new-(and-old)-functions

YasirKusay:
Hello, I would like to be assigned to this issue please.
I have already begun working on some of the tests for igraph_closeness_estimate.

iosonofabio:
Hi Yasir,

It'd be best to proceed as we agreed, which is first open a PR so I/we can see your proposed changes and then we'll assign the issue.

It'll get easier as you get credibility but for now maybe that works best.

iosonofabio:
Assigned to Yasir, as planned. Chances are progress on this will happen on PR #1290 and will be merged, if all goes well, right after 0.8 is released (assuming the code is working).

Issue #1053, opened 2018-01-12
Closeness centrality and disconnected graphs


szhorvat:
Currently, this is what `igraph_closeness` does for disconnected graphs:

> If the graph is not connected, and there is no path between two vertices, the number of vertices is used instead the length of the geodesic. This is always longer than the longest possible geodesic.

Whether this is useful is debatable. It does this also for weighted graphs, where the reasoning that "This is always longer than the longest possible geodesic" does not apply.

Arguably, *closeness* just doesn't make much sense for disconnected graphs. If someone is calculating it for a disconnected graph, it may be unintentional (they didn't notice it's disconnected) or due to a lack of proper understanding of the concept.

Therefore it may be useful to issue a warning in this situation.  Consider this a feature suggestion.

----

This is what some other systems do:

 - Mathematica computes it per connected component
 - networkx seems to compute it per connected component, then weight it by the component size (unless I misunderstood—I haven't tested) https://networkx.github.io/documentation/stable/reference/algorithms/generated/networkx.algorithms.centrality.closeness_centrality.html

vtraag:
Reopened, because this issue also touches on the estimation of closeness centrality for large graphs (cf. #1197).

Silmathoron:
To give some update on this issue, I'll summarize here [a discussion on discourse](https://igraph.discourse.group/t/user-defined-distance-for-unconnected-node-in-closeness-and-harmonic-centrality/183) on this issue.

The arithmetic definition of the closeness centrality leads to significant "awkwardness" regarding how to compute it for graphs where the in- or out-component of some nodes does not contain all other nodes in the graph.

One way to alleviate the issue is related to the convention used for the distance between nodes ``i`` and ``j`` when there is no path linking ``i`` to ``j``.
Currently, igraph uses the convention ``d_{ij} = N``, with N the number of nodes in the graph.
Other graph libraries (graph-tool and networkx) use the convention ``d_{ij} = 0`` so that the distance does not appear in the sum.

To the best of my knowledge, this second convention is the least problematic because it can be combined with another convention for the normalization, which is to use ``n - 1``, with ``n`` the number of nodes in the in- or out-component of node ``i``.
This can be further improved using the Wasserman and Faust formula (by default in networkx).

[...]

PR #1290, opened 2020-01-24
Tests for igraph_closeness_estimate

YasirKusay:
I've implemented a few tests for igraph_closeness_estimate for as many cases that I can think of. I will admit that the code may be unnecessarily complicated to read and that some of the tests are too simple. Furthermore, I have yet to implement tests for specific cutoff values, rather than just calculating the true closeness centrality as I am confused as to how it works. From what I understand, if the path from one node to another exceeds the cutoff value, the cutoff value is used instead but the output I am getting is not consistent with this. Feel free to give me advice as how to improve the tests and the code, and if I need to change my code in order to conform to your test protocols.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@
 
 ### Changed
 
+ - Improved interruptability: `igraph_degree_sequence_game()`
+ - Improved argument checking: `igraph_forest_fire_game()`
+
 ### Fixed
 
  - `igraph_community_edge_betweenness()`: fix for graphs with no edges (PR #1312)


szhorvat:
Thanks for all your work on this @YasirKusay !

Since you are still working on it, I think it's best to merge this only *after* the 0.8 release. However, if you discover any bugs while creating the tests, please let us know ASAP.

szhorvat:
In case you have not yet seen this wiki page, here are some tips on what to verify: https://github.com/igraph/igraph/wiki/Checklist-for-new-(and-old)-functions

vtraag:
Yes, indeed thanks for you work @YasirKusay ! As @szhorvat indicates, we should only merge this after the upcoming release.

One other thing: it is useful to indicate in your PR which issue you intend to fix (if any). This will both link the PR to the issue, and indicate in the issue which PR fixes it. In this case, you intend to address issue #1208. If you indicate something like: "this should fix #1208", GitHub even automatically picks up on this and closes the issue if we merge the PR. You only need to write the issue number (prepended with `#`), GitHub will pick up on this and link to the appropriate issue.

YasirKusay:
Thank you all very much for your advice. I will continue to work on it for now. These tests were intended to address issue #1208 (add tests for igraph_closeness). One more thing, can you explain to me how the cutoff parameter is meant to work? From what I understand, if the path from one node to another exceeds the cutoff value, the cutoff value is used instead but the output I am getting is not consistent with this.

szhorvat:
Just a note that I haven't forgotten about this and will get back to you in a couple of days. Thank again for doing all this work!

szhorvat:
Can you please separate changes related to different issues? This PR was started for adding tests for closeness calculations. Things related to bridges should be kept separate.

iosonofabio:
Hi Yasir,

Thank you for your commitment! The usual way to submit several PRs at once is to have separate branches (both locally on your laptop and remote branches on your GitHub) and then each remote branch is used for one PR. Git will take care of merging them all.

szhorvat:
> One more thing, can you explain to me how the cutoff parameter is meant to work? From what I understand, if the path from one node to another exceeds the cutoff value, the cutoff value is used instead but the output I am getting is not consistent with this.

This related to issue https://github.com/igraph/igraph/issues/1053 : igraph's current handling of non-connected graphs is rather questionable in my opinion ... We proceeding, we need to think about this carefully, and choose how to handle disconnected graphs, or cases with look "disconnected" with a low cutoff.

YasirKusay:
With regards to issue #1053, I definitely agree! When I was writing tests for a weighted graph, I found it unusual that I was using the number of vertices as the default distance for vertices in separated subgraphs (or even in connected, directed graphs where there was no possible way to reach a vertex from another vertex, e.g. in 1->2->3, where you cannot reach node 2 from node 3). Furthermore, say we have a small graph: 1-2-3, where the distance from 1-2 is 10 and 2-3 is 20, therefore disproving that the number of vertices is " is always longer than the longest possible geodesic". I actually did a bit of research into this and I found some very interesting ways of working out closeness centrality, particularly this one: 

https://toreopsahl.com/2010/03/20/closeness-centrality-in-networks-with-disconnected-components/

To summarise what that website stated, since the distance between 2 disconnected nodes is infinite, the reciprocal of that distance will be 0 and therefore useless when you are finding the reciprocal of the sum of all distances, as centrality will be 0. Instead, it proposes that you find the distance between 2 nodes and get the reciprocal of that, (doing the same from that node to all of the nodes in the graph, even if they are disconnected) and then sum up these reciprocals. I acknowledge that this will result in a different closeness centrality for graphs that are connected, but perhaps it can be implemented for disconnected graphs and directed graphs where there is no possible path from one node to another (whilst triggering a warning to the user).

YasirKusay:
@szhorvat, @iosonofabio 

Thank you for your advice and I will do that from now on. I am a bit new to github and didnt realise that commits are connected to the same pull request.

vtraag:
@YasirKusay, you may want to read the general [contributing](https://github.com/igraph/igraph/blob/master/CONTRIBUTING.md) guidelines. That also contains some explanations of how to best prepare your PR (it also took me some time to learn it).

PR #1295, opened 2020-01-26
Draft changelog for 0.8

szhorvat:
This is a draft changelog for 0.8.

Feel free to edit (edits from maintainers are enabled in this PR).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,16 +27,15 @@
  - `igraph_weighted_clique_number()` computes the weighted clique number
  - `igraph_largest_weighted_cliques()` finds the largest weighted cliques
 
- #### Games
+#### Graph generators
+
  - `igraph_hsbm_game()` for a hierarchical stochastic block model
  - `igraph_hsbm_list_game()` for a more general hierarchical stochastic block model
  - `igraph_correlated_game()` generates pairs of correlated random graphs by perturbing existing adjacency matrix
  - `igraph_correlated_pair_game()` generates pairs of correlated random graphs
  - `igraph_tree_game()` samples uniformly from the set of labelled trees
  - `igraph_dot_product_game()` generates a random dot product graph
- - `igraph_sample_sphere_surface()` samples points uniformly from the surface of a sphere
- - `igraph_sample_sphere_volume()` samples points uniformly from the volume of a sphere
- - `igraph_sample_dirichlet()` samples points from a Dirichlet distribution
+ - `igraph_realize_degree_sequence()` creates a single graph with a given degree sequence (Havel-Hakimi algorithm)
 
 #### Graph embeddings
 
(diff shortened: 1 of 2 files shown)


szhorvat:
I just noticed that the `ChangeLog` file I renamed to `CHANGELOG.md` had this commit note:

> Readded (empty) ChangeLog file, needed for automake.

Despite that, `,/configure` gave me no complaints. Does anyone remember why `ChangeLog` was needed? Should I re-add an empty one or rename `CHANGELOG.md` back?

vtraag:
I've compare the `include` directory with the `0.7.1` version, and added some additional changes I noticed. Regarding the need of `ChangeLog`, I don't know, sorry.

ntamas:
`ChangeLog` is mandated by the GNU coding standards, so `automake` complains if you don't have one. (It probably does not fail the build, though). Even the format of the `ChangeLog` file is strictly specified so you cannot symlink a generic Markdown file as `ChangeLog` to fool `automake`. *sigh*

szhorvat:
> `ChangeLog` is mandated by the GNU coding standards

Do we care about GNU standards?

Should I leave it as is (markdown is nice!) or reformat to GNU standards? I *roughly* followed https://keepachangelog.com/ combined with a bit of common sense to improve readability.

vtraag:
Note that I didn't follow a chronological order per se.

szhorvat:
`igraph_solve_lsap()` is not documented. I never knew if it was ready and tested. I believe it was added to eventually replace the R implementation of graph matching. 

If we include it in the changelog, we should add documentation.

@gaborcsardi, do you know what status of this function is? Ready to use?

gaborcsardi:
> @gaborcsardi, do you know what status of this function is? Ready to use?

IDK, just ignore it, I would say.

szhorvat:
Removed `igraph_solve_lsap` from the changelog per Gábor's comment. We can test it later and perhaps make it public in 0.9

ntamas:
> Do we care about GNU standards?

I guess not so much any more, at least compared to how much we used to care about them several years ago :) I'd prefer to have the changelog in Markdown as well - Github formats it nicely, and later on we could even switch to [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0-beta.4/), which would allow us to generate changelogs from the commit messages instead of writing them manually. However, in order to silence `automake` for the time being, we should have a regular `ChangeLog` file that just points the user to the "real" changelog. We can remove it once we switch the build system from Autotools to CMake in the future.

szhorvat:
Re-added `ChangeLog` file and kept `CHANGELOG.md` as-is.

Please everyone have a read-though of the changelog. If there are no concerns by the end of the day, we should merge this.

ntamas:
Seems OK to me so far! I'll merge around 10pm CET if there are no objections.

szhorvat:
I just noticed that there's already a `NEWS` file with changes up to 0.6.5 but not 0.7. What do we do with that? Transfer its contents?

Unrelated: There is also an ancient `TODO` file which should probably be removed.

ntamas:
Both of these are obsolete and can be removed. The NEWS file was also mandated by the GNU coding standards, so we should probably keep a placeholder in it. 

T.

ntamas:
Old news file renamed from `NEWS` to `ONEWS` (as suggested by the GNU coding standards). We can remove this file in 0.9 or so. Added new `NEWS` file that simply refers the user to `CHANGELOG.md`. These are already in master.

PR #1300, opened 2020-01-26
Added missing DECLDIR

vtraag:
Some existing functions did not yet have a proper `DECLDIR` to ensure proper export under Windows.

### DIFF
--- a/include/igraph_visitor.h
+++ b/include/igraph_visitor.h
@@ -75,7 +75,7 @@ typedef igraph_bool_t igraph_bfshandler_t(const igraph_t *graph,
         igraph_integer_t dist,
         void *extra);
 
-int igraph_bfs(const igraph_t *graph,
+DECLDIR int igraph_bfs(const igraph_t *graph,
                igraph_integer_t root, const igraph_vector_t *roots,
                igraph_neimode_t mode, igraph_bool_t unreachable,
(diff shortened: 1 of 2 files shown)


include/igraph_visitor.h
```diff
@@ -85,7 +85,7 @@ int igraph_bfs(const igraph_t *graph,
                igraph_vector_t *dist, igraph_bfshandler_t *callback,
                void *extra);
 
-int igraph_i_bfs(igraph_t *graph, igraph_integer_t vid, igraph_neimode_t mode,
+DECLDIR int igraph_i_bfs(igraph_t *graph, igraph_integer_t vid, igraph_neimode_t mode,
```
ntamas: Hmmm, I'm a bit unsure about this one. As its name implies, this is an internal function. I'd rather not have it in the header file at all. I think the only reason it's there is that we can test it in `examples/simple/igraph_bfs.c`.

What about removing this declaration from the header and copying it into `examlpes/simple/igraph_bfs.c` instead?

vtraag: Yes, it wasn't my intention to add it to the `igraph_i_*` declarations. There are quite a number of other such declarations across the public interface. We should probably move all of them, but let's postpone that until 0.9.0. For now, I'll remove the `DECLDIR` on this one.

PR #1305, opened 2020-01-28
Updated the central mail address to igraph@igraph.org.

vtraag:
I only updated the central mail address for the package, as used in `configure.ac`. I left all other mail address (e.g. in copyright statements) intact.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,5 @@
+{
+  "files.associations": {
+    "igraph_community.h": "c"
+  }
+}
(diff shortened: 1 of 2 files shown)


szhorvat:
@vtraag Was adding the VS Code setting accidental?

vtraag:
> @vtraag Was adding the VS Code setting accidental?

Sorry, no that passed right by me! Probably better to delete it. But it might be nice to add `.vscode` to the `.gitignore`.

PR #1307, opened 2020-01-29
Add link to Mathematica interface from README

szhorvat:
Hope this is OK


ntamas:
Absolutely. If you find any more places where the Mathematica interface is not mentioned, feel free to add it on your own.

PR #1309, opened 2020-01-29
Update local scan docs, remove unused chapters

szhorvat:
I just noticed that some of the missing functions that I added to the documentation recently already had their own chapter, those chapters were just not included. 

This PR incorporates the additional info from those chapters and removes the chapters.

Epidemics had its own (unincluded) chapter. Now it is in "Games on graphs", as it is just another process that happens on graphs.

Local scan statistics had its own chapter. Now it is in "Structural properties", as it seems to fit there.

I think that this PR is safe to merge at the last minute for 0.8, but it is not essential. We can also include it in 0.8.1. The only user-visible change is a minor reorganization of the local scan statistics functions.


ntamas:
I'm still messing with a few minor compilation issues with `f2c` on Windows, so I'll let the current CI tests finish and then I'll merge this before tagging the release.

vtraag:
Are there any issues with `f2c` in Windows that we should report in the installation/compilation instructions? If not before the release, we can do so afterwards still?

ntamas:
No, nothing to worry about. I made a last-minute change in the compilation of the C core when I realized that `src/f2c/arith.h` should actually be _generated_ on the machine where igraph is being compiled, but this broke the MSVC build because we cannot generate this file on MSVC (this is basically a pre-build step that I'm not familiar how to hack into the build process). So I ended up providing a pre-generated copy of `arith.h` for MSVC, but at least it is now generated properly in POSIX environments.

PR #1312, opened 2020-01-31
Fix #1311 by checking if number of edges is zero

### LINKED ISSUE #1311: igraph_community_edge_betweenness segfaults on graph with zero edges
If the function `igraph_community_edge_betweenness` is called on a graph with zero edges, it crashes when attempting to `free` and `invalid pointer`. Here is a reproducible example:

```
#include <igraph.h>

int main(void)
{
    igraph_t g;
    igraph_vector_t eb;
    igraph_vector_t res;

    igraph_full(&g, 1, 0, 0);
    igraph_vector_init(&res, igraph_ecount(&g));
    igraph_vector_init(&eb, igraph_ecount(&g));

    igraph_community_edge_betweenness(&g, 
        &res, // result
        &eb, // edge_betweenness result
        NULL, // merges result
        NULL, // bridges
        NULL, // modularity
        NULL, // membership
        IGRAPH_UNDIRECTED, // directed
        NULL // weights
        );    

    igraph_vector_destroy(&eb); // crashes here
    igraph_vector_destroy(&res);
    igraph_destroy(&g);

}
```

I believe that memory corruption is caused by indexing an array at -1, which leads to invalidating a pointer. This will cause a segfault when `free` is called on the invalid pointer. 

I believe this is the ultimate cause of  igraph/rigraph#336 segfault in the R interface.

nograpes:
In igraph_community_edge_betweenness, if the number of edges in the
graph is zero, and an initialized pointer is passed for the edge
betweenness variable, then the array will be indexed at -1, which
causes memory corruption in the chunk header, making the pointer
invalid. This will segfault when free is called on the pointer.

I have added a test as well, but you may not find that necessary.

I think this directly fix igraph/igraph#1311, which will indirectly fix
igraph/rigraph#336.


ntamas:
Tests are always nice to have :) Thanks a lot!

PR #1319, opened 2020-02-04
Configuration model: check for loop edges in an initial pass

szhorvat:
This implements Tamás's idea from here:

https://github.com/igraph/igraph/pull/1144#issuecomment-446934997

The speedup is very minor in practice. It was only noticeable for heavy-tail degree distributions. An example of the biggest speedups I noticed: 7.7 s -> 5.9 s.

This is because the set data structure is quite efficient, and because most of the time is spent in the random number generator anyway.

### DIFF
--- a/src/games.c
+++ b/src/games.c
@@ -1257,6 +1257,7 @@ int igraph_degree_sequence_game_no_multiple_undirected_uniform(igraph_t *graph, 
         igraph_bool_t success = 1;
         IGRAPH_CHECK(igraph_vector_int_shuffle(&stubs));
 
+        /* optimization: we do an initial pass looking for self-loops */
         for (i = 0; i < ecount; ++i) {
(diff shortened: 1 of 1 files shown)


szhorvat:
@ntamas I think what we (or I) should try is to not do a complete shuffle of the vector before rejecting a trial. 

What we do now: 1) do complete shuffle 2) check for self-loops 3) check for multi-edges. Originally, 2) and 3) were combined. This PR separates them. Since 3) works with sets, it was potentially slow, and this PR did indeed lead to some speedup. But I believe most time is spent in calling the RNG (i.e. shuffling), and not in checking whether the result was a multigraph. We could try to combine steps 1) and 2): do not complete the Fisher-Yates shuffle once we detect that a self-loop would arise. Just short-circuit and restart.  This will save a lot of calls to the RNG.

Issue #1180, opened 2019-05-12
igraph_bridges and multigraphs


szhorvat:
`igraph_bridges` appears to ignore multi-edges.  Verify that this is the only limitation with multigraphs, and document it (or generalize the function).

YasirKusay:
This is a very interesting issue that I found and decided to run a small program to investigate. I created an undirected graph: 0=1-2 (i.e, two edges between 0 and 1, indexed as edges 0 and 1 respectively and 1 edge between 1 and 2, indexed as edge 2). When I ran the program, I assumed that the multi-edge edges will not be considered as a bridge and that igraph_bridges will only return edge id 2 but surprisingly, edge id 1 and 2 were returned after using igraph_bridges. I removed one of the edges from 0-1 and igraph_bridges  returned correctly (edge id 1 and 2). Is this what the implementation of igraph_bridges intended?

ntamas:
As stated above in the comment of @szhorvat, `igraph_bridges` appears to ignore multi-edges in the current implementation. My conjecture is that essentially it works with _vertex pairs_ instead, so basically for each vertex pair it investigates whether disconnecting the _vertex pair_ would disconnect the graph. If it does, then it returns _all_ edges between that vertex pair as bridges. So, in some sense it may have been better for `igraph_bridges` to return vertex pairs instead of edge indices. On the other hand, most people don't work with multi-edges and this way we spare them an extra edge index lookup in the result vector.

We are open to suggestions about how to generalize `igraph_bridges` in the next release so it works nicely with multigraphs.

szhorvat:
When I implemented `igraph_bridges`, I followed the guide here: https://www.geeksforgeeks.org/bridge-in-a-graph/  I simply forgot about multi-edges, so the code wasn't originally *meant* to handle them one way or another. 

If you'd like to look into this, perhaps the link above will be helpful

YasirKusay:
From my observation of the function, if there is a multi-edge, it is treated as a single edge and if it is a bridge, then the final edge from the multi-edge group is returned. Personally, I think the user should have a choice (perhaps including an extra function parameter) to treat multi-edges as just one edge, such that removing all the edges in that multi-edge group may disconnect the graph and therefore all edges are considered to be a bridge, and returned to the user. In my opinion, that does not require a complete overhaul of how the function is implemented, but rather some minor checks (check to see if the edge pair are multiedges, etc).  If you agree, I am more than happy to give it a go.

szhorvat:
[...]

PR #1326, opened 2020-02-10
Check in igraph_bridges if graph has loops and edges

YasirKusay:
This is a pull request intended to solve issue #1180 
This will check if the current graph has multi-edges and loops, raising a warning if there is. It does this at runtime and therefore, is more efficient than calling a function to check if there are bridges or loops.


iosonofabio:
Hi Yasir,

Great to see you are starting to delve into PR plumbing! It's best if we sit down and go through some details together before you proceed.

Let me write you a separate email about possible dates.

szhorvat:
Closing, since this was split into separate PRs for bridges (which is resolved) and closeness (which is not yet resolved).

Issue #678, opened 2014-08-28
igraph_arpack_rnsolve.c test case sometimes fails when an external ARPACK library is used


ntamas:
It is not really a failure; some ARPACK implementations (such as the one used in Debian Linux when compiling the igraph package from source) do not respect the `nev` setting; they will return _at least_ `nev` eigenvalues but they always seem to return both eigenvalues from a conjugate complex pair, so one may end up with more eigenvalues than what was requested. The test case does not account for this. (For what it's worth, the expected output of the test case anticipates that `nev` is respected for the number of eigenvalues but not for the number of eigenvectors).

There are two possible fixes that we could apply here: either we make `igraph_arpack_rnsolve` resize the result vector (and the matrix of eigenvectors as well) such that igraph does not return more than `nev` eigenvalues and/or eigenvectors even if ARPACK returned more, or we let this quirk of ARPACK leak through the abstraction and fix the test case instead.

Comments are appreciated (ping @gaborcsardi).

orlitzky:
We're hitting this on Gentoo, too, where users can choose to run the test suite before installing a package. Currently we've "fixed" it by disabling the test suite entirely which works but is not optimal =)

nullr0ute:
We see this on non x86 architectures on Fedora too (aarch64/ppc64/ppc64le/s390x)

rdauria:
Hello, I am not sure if this is related but on a CentOS 6.10 I get:

```
Failed tests:
igraph 0.7.1 test suite: igraph test groups:

 NUM: FILE-NAME:LINE     TEST-GROUP-NAME
      KEYWORDS

  26: types.at:151       Sparse matrix, ARPACK eigensolver (igraph_sparsemat_t): 
      sparse matrix igraph_sparsemat_t arpack
 195: arpack.at:45       Solving linear systems with LU factorization (igraph_lapack_dgesv):
      lapack dgesv solve lu factorization
 198: arpack.at:63       Non-symmetric ARPACK solver (igraph_arpack_rnsolve):
      arpack eigenvalue eigenvector eigen eigenproblem non-symmetric
```
I have configure igraph with:

```.
[...]

PR #1330, opened 2020-02-13
Make all tests deterministic

szhorvat:
This PR tries to make all tests deterministic by setting a fixed random seed. Some tests use igraph's own RNG, some use the one from the C standard library.

This was motivated by this spurious failure which may have happened because of numerical precision issues:

https://travis-ci.org/igraph/igraph/jobs/648828440

I am not sure how to get the full build state from Travis (if it's even possible).

### DIFF
--- a/examples/simple/2wheap.c
+++ b/examples/simple/2wheap.c
@@ -23,7 +23,6 @@
 
 #include <igraph.h>
 #include "igraph_types_internal.h"
-#include <time.h>
 #include <stdlib.h>
 
 int main() {
@@ -33,7 +32,7 @@ int main() {
     long int i;
     igraph_real_t prev = IGRAPH_INFINITY;
 
-    srand(time(0));
+    srand(42); /* make tests deterministic */
 
     igraph_vector_init(&elems, 100);
     for (i = 0; i < igraph_vector_size(&elems); i++) {
--- a/examples/simple/igraph_are_connected.c
+++ b/examples/simple/igraph_are_connected.c
@@ -21,7 +21,6 @@
 
 #include <igraph.h>
 #include <stdio.h>
-#include <time.h>
 
 #define R_INTEGER(a,b) (igraph_rng_get_integer(igraph_rng_default(), (a), (b)))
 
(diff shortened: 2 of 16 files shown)


szhorvat:
What is happening? Here's another failure, and this test was already deterministic. 

https://travis-ci.org/igraph/igraph/jobs/649885327?utm_medium=notification&utm_source=github_status

vtraag:
I don't think that test was deterministic to start with? That is, as far as I know 
https://github.com/igraph/igraph/blob/e4851413e08ebfd3419cbbb290519529747682ed/examples/simple/igraph_community_label_propagation.c#L54
is not a deterministic function, and so the RNG should be seeded for this. I hadn't seen the error before, but perhaps we just got lucky earlier?

szhorvat:
Of course! The method is not determinisitc. 

@vtraag  I am not familiar with the details of this method. Is there any reason to expect it not to return more than 4 clusters for this graph?

BTW the probability of this test failing is about 0.1%:

```wl
In[116]:= g = ExampleData[{"NetworkGraph", "ZacharyKarateClub"}];

In[129]:= Table[
     Length[IGCommunitiesLabelPropagation[g]["Communities"]],
     {10000}
     ] // Counts // KeySort // N // Normalize[#, Total] &

Out[129]= <|1 -> 0.0351, 2 -> 0.5074, 3 -> 0.4066, 4 -> 0.0493, 5 -> 0.0016|>
```

The output is the probabilities of getting a certain number of clusters.

szhorvat:
It is still good to have a test that exercises the function even if it does not explicitly check any condition. It helps prevent crashes and it's useful with AddressSanitizer. But the test should at least check the return value of the function (`IGRAPH_SUCCESS`).

vtraag:
No, as far as I know, label propagation does not provide any guarantee on the number of clusters. Well, that is perhaps too strong: I assume that the number of clusters will always be strictly lower than the number of nodes. But there's no guarantee that it will be not more than 4 clusters.

We can either make the test deterministic by seeding the RNG. Or we can simply not check the condition, and only check the return value, as you suggest.

By the way: a similar problem of not checking values that may be under-determined (and can hence change in certain conditions) appears in issue #678. Perhaps better to address that separately, but I just wanted to mention it here.

szhorvat:
> and only check the return value, as you suggest.

I think it is actually not necessary to check the return value of functions in tests. If the return value is not zero, the function will print an error or warning message. It should be sufficient to let the test frame work check the output of the program.

PR #1331, opened 2020-02-13
degree_sequence_game: make DEGSEQ_SIMPLE_NO_MULTIPLE interruptible

szhorvat:
`degree_sequence_game` with `DEGSEQ_SIMPLE_NO_MULTIPLE` does not do well with some degree sequence (e.g. power-law distributed) and will in practice keep running forever.

This makes it interruptible.

### DIFF
--- a/src/games.c
+++ b/src/games.c
@@ -965,6 +965,8 @@ int igraph_degree_sequence_game_no_multiple_undirected(
      * until it finally succeeds. */
     finished = 0;
     while (!finished) {
+        IGRAPH_ALLOW_INTERRUPTION();
+
         /* Be optimistic :) */
         failed = 0;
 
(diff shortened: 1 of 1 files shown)


szhorvat:
As far as I can tell, there are some unnecessary `IGRAPH_CHECK`'s, such as this:

https://github.com/igraph/igraph/pull/1331/files#diff-734e1282742f6a82c91db2984452de0dL975

That vector seems to be already initialized to the correct size, so the `_update` function shouldn't fail due to out-of-memory.

I didn't remove them as it's not harmful and it won't be what slows things down. I didn't want to mess up anything by accident.

ntamas:
That particular `IGRAPH_CHECK()` is probably not necessary with the current implementation (and it's unlikely that we'll ever implement another version of `igraph_vector_update()` that might return an error code if no memory allocation is necessary for the update). However, I'd rather stick to the practice of putting `IGRAPH_CHECK()` macros everywhere where the wrapped function returns an `igraph_error_t` so if we ever decide to return an error code from a function that did not do so in certain conditions before, we won't break existing code with it. The overhead of the `IGRAPH_CHECK()` macro is probably minimal; it might even be the case that the compiler optimizes it away if it detects that the only valid return code from a function is `IGRAPH_SUCCESS`.

PR #1333, opened 2020-02-17
Tests for closeness centrality calculations

YasirKusay:
This is a new PR rebased on an updated master aimed at adding tests for `igraph_closeness_estimate` issue #1208.

DO NOT MERGE YET, FOR REVIEW ONLY (fixing the cutoff values).

### DIFF
--- a/examples/tests/igraph_closeness.c
+++ b/examples/tests/igraph_closeness.c
@@ -1,0 +1,190 @@
+#include <igraph.h>
+#include <stdio.h>
+#include "test_utilities.inc"
+
+
+void simple_test_case_no_weights_undirected() {
+
+    igraph_t g;
+    igraph_vector_t vector_actual_results;
+
+    printf("Simple test case, no weights, undirected\n");
+
+    igraph_vector_init(&vector_actual_results, 0);
+
+    igraph_small(&g, 0, IGRAPH_DIRECTED, 0,1 , 1,2, -1);
+
+    /* NOT NORMALISED TEST BELOW */
+
+    igraph_closeness_estimate(&g, &vector_actual_results /*store results here*/, 
+                             /*calculating for all vectors in the graph*/ igraph_vss_all(), 
+                             IGRAPH_ALL  /*graph is "undirected"*/, /*cutoff*/ 0 /*calculating exact centrality*/, 
+                             NULL /*unweighted*/, /*not normalised*/ 0);
+
+    printf("Non normalised results below\n");
+    print_vector(&vector_actual_results, stdout);
+
+    /* NORMALISED TEST BELOW */
+
+    igraph_closeness_estimate(&g, &vector_actual_results /*store results here*/, 
+                             /*calculating for all vectors in the graph*/ igraph_vss_all(), 
+                             IGRAPH_ALL  /*graph is "undirected"*/, /*cutoff*/ 0 /*calculating exact centrality*/, 
+                             NULL, /*normalised*/ 1);
+
+    printf("\nNormalised results below\n");
+    print_vector(&vector_actual_results, stdout); 
+
+    igraph_vector_destroy(&vector_actual_results);
+    igraph_destroy(&g);
+}
+
+void simple_test_case_with_weights_undirected() {
+
+    igraph_t g;
+    igraph_vector_t vector_edges, vector_weights, vector_actual_results;
+
+    igraph_real_t real_edges[] = {0,1 , 1,2};
+    igraph_real_t real_weights[] = {3, 5};
+
+    printf("\nSimple test case, with weights, undirected\n");
+
+    igraph_vector_init(&vector_actual_results, 0);
+    igraph_vector_view(&vector_edges, real_edges, sizeof(real_edges)/sizeof(igraph_real_t));
+    igraph_create(&g, &vector_edges, /*number of vertices*/ 2, IGRAPH_DIRECTED);
+    
+    igraph_vector_view(&vector_weights, real_weights, 
+                       sizeof(real_weights)/sizeof(igraph_real_t));
+
+    /* NOT NORMALISED TEST BELOW */
+
+    igraph_closeness_estimate(&g, &vector_actual_results /*store results here*/, 
+                             /*calculating for all vectors in the graph*/ igraph_vss_all(), 
+                             IGRAPH_ALL  /*graph is "undirected"*/, /*cutoff*/ 0 /*calculating exact centrality*/, 
+                             &vector_weights, /*not normalised*/ 0);
+ 
+    printf("Non normalised test below\n"); 
+    
+    print_vector(&vector_actual_results, stdout);
+
+    /* NORMALISED TEST BELOW */
+
+    printf("\nNormalised test below\n");
+
+    igraph_closeness_estimate(&g, &vector_actual_results /*store results here*/, 
+                             /*calculating for all vectors in the graph*/ igraph_vss_all(), 
+                             IGRAPH_ALL  /*graph is "undirected"*/, /*cutoff*/ 0 /*calculating exact centrality*/, 
+                             &vector_weights, /*normalised*/ 1);
+
+    print_vector(&vector_actual_results, stdout);
+
+    igraph_vector_destroy(&vector_actual_results);
+    igraph_destroy(&g);
+}
+
+void advanced_test_case_no_weights_undirected() {
+
+    igraph_t g;
+    igraph_vector_t vector_actual_results;
+
+    /* note, denominatory calculated as (shortest dist)*(n-1) for no normalisation
+    normalisation excludes n-1 as part of the denominator */
+
+    printf("\nAdvanced test case, no weights, undirected\n");
+
+    igraph_vector_init(&vector_actual_results, 0);   
+
+    igraph_small(&g, 0, IGRAPH_DIRECTED, 1,0 , 0,5 , 5,6 , 5,4, 4,1 , 1,2 , 2,4 , 4,6 ,
+                                 2,3 , 3,7 , 7,6 , 2,6 , -1);
+
+    /* NOT NORMALISED TEST BELOW*/
+
+    printf("Non normalised test below\n");
+
+    igraph_closeness_estimate(&g, &vector_actual_results /*store results here*/, 
+                             /*calculating for all vectors in the graph*/ igraph_vss_all(), 
+                             IGRAPH_ALL  /*graph is "undirected"*/, /*cutoff*/ 0 /*calculating exact centrality*/, 
+                             NULL, /*not normalised*/ 0);
+
+    print_vector(&vector_actual_results, stdout);
+ 
+    /* NORMALISED TEST BELOW*/
+
+    printf("\nNormalised test below\n");
+
+    igraph_closeness_estimate(&g, &vector_actual_results /*store results here*/, 
+                             /*calculating for all vectors in the graph*/ igraph_vss_all(), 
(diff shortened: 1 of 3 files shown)


ntamas:
FYI, CI tests were failing on Windows with MSVC. This is because some variable declarations were in the middle of their blocks (not at the top). This is not permitted in C89 but is totally fine with any modern C compiler. Unfortunately we still need to support C89 because igraph's Python interface for Python 2.7 is compiled with MSVC and MSVC is very picky about this.

(It is probably a good practice anyway to move all variable declarations to the top of the corresponding blocks because otherwise one would have to scan the entire block to figure out which variables are declared in the block).

ntamas:
> DO NOT MERGE YET, FOR REVIEW ONLY

There is a "draft PR" feature in Github - you can mark a PR as "draft" to indicate that it's still work in progress. I think it's not possible to change this after the PR has been created, but you can use it next time.

szhorvat:
I meant to comment on this a few days ago. Unfortunately a bad flu prevented me from doing so.

Do I understand it correctly that this is basically complete, except for testing cutoffs?  There are a lot of tests here, so I suggest we finalize and merge these as soon as possible. The cutoff case is problematic, and requires a complete re-think of how closeness estimation should work, and whether estimation with a cutoff makes sense at all. It will almost certainly require changes to the closeness function. That should be a separate project, and I suggest we leave it for another time.

Some comments on the current tests:

When a test fails, one would want to determine (with as little effort as possible) which part of the test failed, and how it failed exactly. Therefore it is preferable to print the results and let the test framework compare the printed output to the expected one. Is there are reason you think the direct comparison was necessary (instead of printing results)?

I notice you implemented a function to compare with tolerance. If we print the output, a small difference can be allowed by printing with a limited number of digits. This is, in fact, the default behaviour of `printf` with `%f`, so usually nothing special needs to be done.

To make it easier to print vectors, and to reduce duplication of code in test programs, you can `#include` the file [`test_utilities.inc`](https://github.com/igraph/igraph/blob/master/examples/tests/test_utilities.inc) and use the printing functions there. 

Some smaller nitpicks:

We aim for C89 compatibility in igraph, and also try to maintain a consistent style. Therefore please change `//`-comments to `/* */`-comments and `main(void)` to `main()`. `main()` (i.e. no prototype) will work with any old C compiler, and I believe some MSVC versions may choke on `main(void)`.

Don't forget to move the declarations to the beginning of blocks, as Tamás asked.

Thanks for all your work on this so far! Is there any other tests you wanted to add (other than the cutoff ones), or do you agree with merging after the above points have been addressed?

iosonofabio:
Agree with Szabolcs,

@YasirKusay let's meet one of these days and fix these few things?

YasirKusay:
Hello @szhorvat, thank you for your feedback. With regards to your statement concerning printing the output, I did do that originally when I was first writing the function. Whenever I had a mismatch, I would call the print function that I wrote to see the output but when I saw that they were matching, I decided to remove it to make my output cleaner as a function . However, although I have confirmed that all the tests have passed, you are correct in pointing out that it may be hard for someone else to determine if I passed or failed some of the tests, and I can add some print statements in the main() function to address that, e.g. "test 1 passed", "test 1 failed", "all tests passed", etc. Although none of the tests fail here, I can add a feature to print the expected results and actual results side by side, in case someone wants to play around with my tests. What do you think?

szhorvat:
> Whenever I had a mismatch, I would call the print function

You do not need to check the result in your program at all. Leave this to the testing framework. Simply print all results, then create a file that contains the expected results and add it to the test. Take a look at the other tests which have two items in `AT_COMPILE_CHECK` to see examples.

> and I can add some print statements in the main() function to address that, e.g. "test 1 passed", "test 1 failed"

Yes, it's a good idea to print some information about what is being tested, so when the test fails, we can quickly jump to the part of the test program which failed.

But I suggest you leave the comparison to the testing framework. Thus, do not print "failed" or "passed". Simply print the title of the sub-test, then the result of the closeness calculation. Verify manually that the results are correct. You have already done this, so no need to do it again. When you are sure the result is correct, then create the expected output file:

```
./myprog > igraph_closeness.out
```

Then add the test program and the expected output to the `.at` file, and we're good to go.

----
One

szhorvat:
@YasirKusay Looking forward to seeing this completed. Let us know if you have questions/comments.

YasirKusay:
Passes CI on all 3 platforms. Ready to merge.

examples/tests/igraph_closeness.c
```diff
@@ -0,0 +1,287 @@
...
+    igraph_t g;
+    igraph_vector_t vector_edges, vector_expected_results, vector_actual_results, vector_expected_results_normalised;
+
+    igraph_real_t real_edges[] = {0,1 , 1,2};
+    igraph_real_t real_expected_results[] = {0.333333, 0.5, 0.333333};
+    igraph_real_t real_expected_results_normalised[] = {0.666666, 1, 0.666666};
+
+    printf("Simple test case, no weights, undirected\n");
+
+    igraph_vector_init(&vector_actual_results, 3);
```
szhorvat: It is not necessary to initialize result vectors to their final size. Functions like `igraph_closeness` will always resize them as needed.

examples/tests/igraph_closeness.c
```diff
@@ -0,0 +1,287 @@
...
+
+/* helper function, useful for viewing output */
+void igraph_vector_t_print(igraph_vector_t *exp, igraph_vector_t *actual, int size) {
+    for (int i = 0; i < size; i++) {
+        double res = VECTOR(*exp)[i];
+        double res_2 = VECTOR(*exp)[i];
+        printf("%lf %lf\n", res, res_2);
+    }
+    printf("\n");
+}
```
szhorvat: Please do not include the expected results in the code.

The expected results should to in the `.out` file. The program prints the actual results. Leave it to the test framework to compare them, and do not print superfluous information.

You can use the printing function from `test_utilities.inc` to simplify the test program and reduce code duplication

szhorvat: (Note that there was also a bug in this: you print `exp` twice instead of printing `exp` and `actual`. But let's just get rid of this function and go with the method in my comment above.)

szhorvat: Yet another issue here was that 
```
for (int i=0; i < size; i++)
```
is not valid C89. The variable declaration cannot be inside of the `for`. Declarations must always go to the beginning of the block.

examples/tests/igraph_closeness.c
```diff
@@ -0,0 +1,287 @@
+#include <igraph.h>
+#include <stdio.h>
+#include <math.h>
```
szhorvat: It seems that the `math.h` header is not needed.

examples/tests/igraph_closeness.c
```diff
@@ -0,0 +1,287 @@
...
+            100000 == 99999 */
+            if (v2+1 != v1) {
+                return 0;
+            }
+        }
+    }
+    return 1;
+}
+
+int simple_test_case_no_weights_undirected() {
```
szhorvat: Since the test functions don't actually return anything (the return value, which is always `0`, is never used), make them `void`.

iosonofabio:
The CI tests do not catch any of these things, in particular not the C89 comment. It's ok that you look at every PR by hand for now, but that's obviously not ideal. Any suggestions for checking these things at CI?

szhorvat:
It is possible to catch these using the `-std=c89` compiler flag (supported by gcc and clang). The problem is that this is much too strict and disables stuff we rely on and is not easy to replace (such as `isnan()`). It also disables extensions that are sometimes used in igraph and which may potentially improve performance (`inline`).

What might work better is handpicking a number of relevant compiler flags to enable warnings and turn warnings into errors.

I am in favour of relaxing the standards in some reasonable way (I have asked for this in the past). But until that is done, and until we come to an agreement about which constructs are important to avoid (for compatibility without various older systems) and which are safe to use, I think it is better to be conservative and stick to the standards that we were using before.

szhorvat:
BTW a good IDE that understands code and flags issues early can be really helpful. I use Qt Creator.

ntamas:
> I am in favour of relaxing the standards in some reasonable way (I have asked for this in the past)

We can switch to C99 once we drop support for Python 2.7 in `python-igraph` (which we did, in theory, after the release of 0.8.0). However, I'd still prefer to have variable declarations at the top of code blocks and not at random places throughout the code - it makes it easier to see what variables a particular code block relies on.

One thing that we need to confirm is that recent versions of Microsoft's C compiler are okay with a more modern dialect of C. It has been notoriously lagging behind in the past.

szhorvat:
I don't think any version of MSVC has full C99 support, and it is not a bad idea to keep supporting somewhat older versions too. Older MSVC versions definitely require declarations to be at the start of the block. It's actually really weird that they didn't throw an error about the `for (int i=0; ...)`

At this point, C99 would mostly give us _convenience_ features. There is nothing in it that we _need_ at this point. I do not want to risk compatibility (or creating extra work in the future to fix issues with an older system or older compiler) just to get some minor convenience.  The one exception is maybe `//`-comments which are very unlikely to cause problems with any compiler (and there are already multiple instances in the codebase).

When I said that I am in favour of relaxing standards, I was mainly thinking of C++11. We might want to integrate code which requires C++11. Porting C++11 to C++98 is often not at all easy (unlike porting C99 to C89).

----

@YasirKusay, for now, please try to stick to C89 and address the comments I made.

szhorvat:
@YasirKusay Thanks for the changes. Is there any reason you insist on using a custom printing function instead using what was already provided? 

If you really want to use this printing function, e.g. to print as columns, then please remove the `size` argument. The size is encoded in the `igraph_vector_t` and should be retrieved from there instead of hard-coding it.

YasirKusay:
When I started work on this, I wrote the functions that I needed as I did not have as much knowledge of the functions back then. Is there a function that will enable me to print values from an igraph_vector_t type as I couldn't find one.

szhorvat:
There is a function for this purpose in `examples/test/test_utilities.inc`. This file is meant to be included as 

```
#include "test_utilities.inc"
```

in test programs, immediately after the usual header `#include`s. It is not part of igraph.

If you judge that this function is not suitable for this application, or not the best one, then we should think about whether an improved version could be useful for other (future) tests as well. If yes, we should include it in `test_utilities.inc` so people wouldn't have to re-invert their own helper functions in the future. If not, it's fine to create a one-time helper function just for this test.

For an example use of these utilities, see `tree.c` or `simplify_and_colorize.c` in `examples/tests`.

Let me know if you need any help with this! :-)

iosonofabio:
Let's keep the discussion on C89/99 for another day. Also for that day is the unrelated topic of using CI for what it's designed to do, which is catch these problems that might be hard to spot by the human eye.

@YasirKusay: feel free to drop in today and we can look into the print function and hopefully get this merged ;-)

szhorvat:
@YasirKusay If you have any questions or concerns, do not hesitate to ask. If you disagree with the changes I asked for, feel free to say so and explain why. That's how we find the best solution. 

Let's merge this ASAP, then you can focus on more exciting things such as #1346

YasirKusay:
Oh no I had no issues with it. I did not get the chance to work on it until just now and it has been finalised. Sorry for any inconveniences.

szhorvat:
No inconvenience caused. Thanks for all your work on this!  Once the CI tests pass, I will squash the commits and merge this.

szhorvat:
Oops, I didn't realize that there was a "squash and merge" option and I squashed the commits right here on this branch. Sorry about that Yasir. We're all learning!

PR #1334, opened 2020-02-17
Issue #1180, new clean PR

YasirKusay:
This PR addresses issue #1180, which checks if there are loops or multi-edges in both directed/undirected graphs on runtime in the function, igraph_bridges.

DO NOT MERGE YET, WORK IN PROGRESS

### DIFF
--- a/src/components.c
+++ b/src/components.c
@@ -1166,6 +1166,29 @@ static int igraph_i_bridges_rec(const igraph_t *graph, const igraph_inclist_t *i
         long edge = (long) VECTOR(*incedges)[i];
         igraph_integer_t v = IGRAPH_TO(graph, edge) == u ? IGRAPH_FROM(graph, edge) : IGRAPH_TO(graph, edge);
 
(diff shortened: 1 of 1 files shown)


szhorvat:
Hi Yasir,

I looked at the bridges algorithm again, and the fix to allow multigraphs turned out to be relatively simple. The algorithm kept track of the parent of each vertex in the DFS tree so that it wouldn't try to "go back the way it came from" during the DFS. In a multigraph, we need to keep track of the incoming edge instead. Please take a look at #1335 

The problem with checking for multi-edges the way it is done here is that it worsens the complexity of the algorithm, due to the nested for loop, as you note.

Some tips for future PRs:

Please use `/* ... */` comments only, as `//` is not valid in C89.

As Tamas noted, please put variable declarations to the start of a block.

ntamas:
PR obsoleted by #1335 by @szhorvat -- thanks for your work on this @YasirKusay!

PR #1335, opened 2020-02-17
igraph_bridges: support multigraphs

### LINKED ISSUE #1180: igraph_bridges and multigraphs

szhorvat:
This PR fixes igraph_bridges so it supports multigraphs. Fixes #1180

After reviewing the algorithm, the fix turned out to be very simple: instead of keeping track of parents of vertices in the DFS tree, we keep track of the incoming edges. This is important only so that we don't try to "go back the same way we came from" during DFS.

There is still room for improvement here, in particular, we should get rid of the recursion to prevent any stack overflow. But that's for later.


ntamas:
Seems good to me! Is this ready to be merged? Would this affect #1334?

szhorvat:
Yes, it's ready.

It obsoletes #1334 as that PR was about adding a check for multigraphs. We no longer need that since now the algorithm works correctly for multigraphs. Originally I did not realize that the fix would be so easy.

PR #1339, opened 2020-02-25
Some cleaning up of memory was not done properly.

vtraag:
Some vectors were not properly destroyed when calculating the weighted average nearest neighbour degree.

### DIFF
--- a/src/structural_properties.c
+++ b/src/structural_properties.c
@@ -5968,6 +5968,10 @@ int igraph_i_avg_nearest_neighbor_degree_weighted(const igraph_t *graph,
         }
     }
 
+    igraph_vector_destroy(&edge_neis);
+    igraph_vector_destroy(&neis);
+    IGRAPH_FINALLY_CLEAN(2);
+
     if (knnk) {
         for (i = 0; i < maxdeg; i++) {
             igraph_real_t dh = VECTOR(deghist)[i];
@@ -5982,7 +5986,7 @@ int igraph_i_avg_nearest_neighbor_degree_weighted(const igraph_t *graph,
         IGRAPH_FINALLY_CLEAN(1);
     }
 
-    igraph_vector_destroy(&neis);
+    igraph_vector_destroy(&strength);
     igraph_vector_destroy(&deg);
     IGRAPH_FINALLY_CLEAN(2);
 


szhorvat:
I would like to suggest bringing a bit more order to commit messages. I would not go as far as to make strict rules, but we could come up with some guidelines which will make it easier to read commits.

One of these could be to mention which function is being modified.  For example,

```
avg_nearest_neighbor_degree: fix memory leak
```

or similar.

It's just a suggestion. It will also help in preparing changelogs.

ntamas:
How about [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/)?

szhorvat:
I have no objection to that, but I must say that it looks a bit complicated. In the interest of not putting too many obstacles in front of external contributions, I would write a shorter summary with many igraph-specific examples.

So what would be the appropriate message for this change?

```
fix(avg_nearest_neighbor_degree): memory leak
```

The easiest way forward would be to just try to use it for a while, see if everyone likes it, and let igraph-specific conventions develop naturally.

If there are no objections, I will try to follow the Conventional Commits standard from now on.

vtraag:
@szhorvat, you mean for commits or for PRs? I'm fine with doing this for PRs. Do you want me to update my [initial PR comment](https://github.com/igraph/igraph/pull/1339#issue-379513234)? For commits it looks actually a bit overkill to me.

vtraag:
An alternative could be to also update the `CHANGELOG.md` with every PR?

szhorvat:
My original suggestion was to bring some (soft) order to commit message formats so that it is easier for all of us to keep track of what is happening (not specifically to construct a changelog, although it would make that easier too).

Mentioning `avg_nearest_neighbor_degree` in the commit message makes it easier to understand what was changed, should one ever have to skim through commits in the future.

Whether that is done with Conventional Commits or otherwise, I do not mind. Originally I was imagining something laid back that does not feel like a burden to anyone, but I can certainly follow the Conventional Commits standard too.

My comment referred to commit messages specifically, not to PRs (which are not part of the git repo).

vtraag:
As far as I am concerned, following Conventional Commits will prove impractical. In practice, this will require a lot of rewriting commits when merging PRs (or asking people submitting a PR to properly rewrite their commits). Especially when responding to reviews of PRs this will involve commits of corrections (or fixes) of new features, which would be better rewritten as a single commit (as which point you lose the PR reviews). This can become quite burdensome I believe.

The question is what the goal is of following such a Conventional Commit type of style. If the goal is to simply have a more extensive description of a commit, as you suggest, that is fine. I'm used to making one-liner commits, but having a somewhat more extensive description could be fine. Having said that, I think the context (or scope) is quite clear from the commit itself. The file and function are clearly identifiable, and to my mind it would be doing double work to then state also explicitly in the commit message what part is affected. Isn't it immediately clear that when looking at the commit that is concerns `igraph_i_avg_nearest_neighbor_degree_weighted`?

Screenshot from 2020-02-25 20-26-31

If the goal is to help with maintaining the CHANGELOG, or making sure Semantic Versioning is followed, I think this will not be achieved in this way. To use it to keep track of (or even generate) a CHANGELOG we would need such a strict adherence to this that it becomes impractical, and a burden, as you say. If we want to maintain the CHANGELOG, I think it would be more doable to simply request to update the CHANGELOG with every PR. I think the PR is the more reasonable unit of work. It can consist of multiple interdependent commits which form a coherent whole, while each individual commit may only be adjusting relatively minor elements. Updating the CHANGELOG then after each PR (if necessary) sounds quite reasonable, and can also always be requested during the review of a PR.

In summary, I would suggest that, as long as the commit message is sufficiently descriptive, the exact format shouldn't matter. So, I don't object against using Conventional Commits, I just think we shouldn't require it of anybody in particular.

ntamas:
Okay, let's keep it simple and forget about Conventional Commits for now.

Back to this PR: CI checks seem to have completed without problems - can I merge this?

szhorvat:
> If the goal is to simply have a more extensive description of a commit, as you suggest, that is fine.

Yes, that was my goal.

> Isn't it immediately clear that when looking at the commit that is concerns `igraph_i_avg_nearest_neighbor_degree_weighted`?

Well, not always, at least not to me. The screenshot shows GitHub's special feature where it extracts the function name. The Git GUI I use is not that smart :-)

> If the goal is to help with maintaining the CHANGELOG, or making sure Semantic Versioning is followed

No, that was not why I suggested it.

> I think it would be more doable to simply request to update the CHANGELOG with every PR

I've been keeping an eye on the changelog since 0.8.0 and periodically verified that it was up to date. I would not enforce a changelog update in every PR because it will lead to merge conflicts.

> In summary, I would suggest that, as long as the commit message is sufficiently descriptive, the exact format shouldn't matter.

I did not feel the need to make commits machine-readable, but I think it would be beneficial to improve human-readability. As time passes, I tend to get more and more (not less!) disciplined with documenting what I do. I have had too many bad experiences where something that seemed completely obvious made zero sense to me a few years later. Then I had to spend a couple of hours figuring out what I was originally thinking.

This applies much more to commenting code than to commit messages, but the two aren't entirely independent. Sometimes the only documentation of a change is the commit message.

Suppose we have some problem with the knn function in the future. If the commit message includes the function name, it'll be quicker to find relevant recent changes. Of course, it's still *possible* to find them (git's "blame") but with clearer messages it's even quicker (for a human).

So I think I'll try to keep adding a bit more context and will try to maintain clarity in commit messages, but I won't push others to do it too :-)  Here's what I did in the last few:

image

vtraag:
> If the commit message includes the function name, it'll be quicker to find relevant recent changes. 

OK, fair enough. I have indeed been using `git blame` or the interactive blame on GitHub to identify this. I have just force pushed the commit with a more appropriate commit message.

> Back to this PR: CI checks seem to have completed without problems - can I merge this?

Yes, you can merge it. Probably best to await the CI to pass for the force pushed commit, but since there's nothing new, there shouldn't be any surprises.

vtraag:
@ntamas , thanks for merging. @szhorvat , do you want me to update the CHANGELOG?

szhorvat:
> do you want me to update the CHANGELOG?

Sure, go ahead.

PR #1340, opened 2020-02-27
Ensure that all licenses are noted and included.

szhorvat:
This PR notes the licenses of all included code and all linked-to libraries.

The `ACKNOWLEDGEMENTS.md` files has all copyright lines, and either includes the full license text or references licenses. Referenced licences are in `doc/licenses`. I believe that shipping all these (`ACKNOWLEDGEMENTS.md` and `doc/licenses`) with binary distributions will be sufficient.

plfit is now marked a GPLv2+. We can correct this depending on the decision Tamás makes.

The only other change is the addition of a `Copyright.txt` file in `msvc/include/libxml`.

As far as I could tell, all other code has the license noted. It is either included in a separate file in the directory where the code lives, or it is noted in the source files (when the source has only a reference, the full license text is found in `doc/licenses`).

### DIFF
--- a/ACKNOWLEDGEMENTS.md
+++ b/ACKNOWLEDGEMENTS.md
@@ -1,0 +1,272 @@
+# Acknowledgements
+
+[igraph](https://igraph.org) includes or links to code from the following sources.
+
+
+#### [bliss 0.73](http://www.tcs.hut.fi/Software/bliss/)
+
+Copyright (c) 2003-2015 Tommi Junttila.
+
+License: [GNU LGPLv3][lgpl3]
+
+
+#### [Cliquer 1.21](https://users.aalto.fi/~pat/cliquer.html)
+
+Copyright (C) 2002 Sampo Niskanen, Patric Östergård.
+
+License: [GNU GPLv2][gpl2] or later
+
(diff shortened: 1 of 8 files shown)


ntamas:
`plfit` will be licensed as GPLv2+ from version 0.8; I'm in the process of incorporating it into igraph but it takes more time than expected because the computation of p-values changed between `plfit` 0.6 and 0.7 (plus there are other minor discrepancies in the results). But anyway, it's coming soon.

Re the included licenses: the license files for GPLv2 and GPLv3 contain an additional section about "How to Apply These Terms to Your New Programs". I think these sections should be removed from the license files that we include.

szhorvat:
> but it takes more time than expected

I think we're not in a rush.

> "How to Apply These Terms to Your New Programs". I think these sections should be removed from the license files that we include.

Done. 

Keep the comments coming. Once everyone is happy, I will squash all the commits to keep things tidy. Please do not merge until then.

ntamas:
There were no comments in the last five days; can this be merged now?

szhorvat:
As far as I'm concerned, it is good.

Issue #1219, opened 2019-07-16
Checking graphs in arguments and caching


vtraag:
Several functions only work on specific graphs, e.g. only (un)directed graphs or only simple graphs. Checking directionality is easy, because it is cheap, but checking whether a graph is simple is more involved. This issue is meant to track our discussion on a proper solution for addressing this problem. There a few possible alternatives.

1. Do no checks and simply state the requirements in the documentation.
2. Do all necessary checks, thereby enforcing correct input.

The advantage of (1) is that it is very cheap, and it does not cost anything. A disadvantage is that it may lead to bugs that are somewhat more difficult to track down. Similarly, it may not be immediately obvious to a user that a certain function is not suitable for a certain type of graph, and the result may be incorrect, without the user being aware of it.

The advantage of (2) is that it is always clear that the input is correct. It would then not be possible to use a certain function with an incorrect input, which makes it more clear to a user. The disadvantage is that these checks may be costly, and may be run repeatedly (e.g. where one function would call another function). This would therefore require a type of caching, which may be somewhat involved to setup. Such a caching system would probably work by caching the result of a function call (e.g. `igraph_is_simple`) inside `igraph_t`, and marking the cached result as dirty whenever the graph was changed. Any new function call would then simply return the cached result, or recalculate the result if it was marked dirty.

The third possible option would then be

3. Do all necessary checks only when compiled in debug mode.

This has the same advantages as (1), and some of the advantages of (2), without requiring a caching system. By introducing checks only if igraph is compiled in debug mode, it doesn't incur any costs when used by a user. It still has the disadvantage that some of the requirements of a certain function are not immediately obvious to an end user. The question is whether the C library has many actual end users, and whether such checking would not be more appropriate in the higher level interfaces. On the other hand, the enforced requirements by checking input may then become inconsistent between higher level interfaces, and may still leave it unclear to users whether a function is really suitable for a certain type of input. It also requires us to maintain these checks in multiple places, thereby increasing the maintenance costs.

Possibly there may be other considerations still. Let us discuss and settle on some system to how to deal with this. Personally, my preference is option (2), but perhaps there are good reasons to choose (1) or (3), or perhaps there is still some other option.

ntamas:
I have no strong opinions on this; let's pick one, document the underlying "philosophy" and then stick to it.

[...]

PR #1341, opened 2020-02-28
Made vector_binsearch_slice public

vtraag:
As discussed in https://github.com/igraph/igraph/issues/981#issuecomment-592440175, we decided to make `igraph_vector_binsearch_slice` properly public, because the `igraph_community_leiden` was improperly relying on the `igraph_i_vector_binsearch_slice` function.

In accordance with our earlier discussion in https://github.com/igraph/igraph/issues/1219#issuecomment-512514619, I made an internal version, `igraph_i_vector_binsearch_slice` (which is used by `igraph_vector_intersect_sorted`, for which it is properly defined) which does not perform any argument checking, and a public version `igraph_vector_binsearch_slice`, which checks arguments. From the perspective of `igraph` as a whole, `igraph_community_leiden` should be allowed to use "internal" functions that perform no argument checking. This then in turn begs the question of how to deal exactly with these type of internal functions that should be included in a proper (internal) header. For now `igraph_community_leiden` simply uses the public `igraph_vector_binsearch_slice` to keep it simple.

### DIFF
--- a/doc/vector.xxml
+++ b/doc/vector.xxml
@@ -1,5 +1,5 @@
 <?xml version="1.0"?>
-<!DOCTYPE section PUBLIC "-//OASIS//DTD DocBook XML V4.3//EN" 
+<!DOCTYPE section PUBLIC "-//OASIS//DTD DocBook XML V4.3//EN"
                "http://www.oasis-open.org/docbook/xml/4.3/docbookx.dtd" [
 <!ENTITY igraph "igraph">
 ]>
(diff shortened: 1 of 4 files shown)


szhorvat:
So should this be included in 0.8.1 even though it's adding to the API? Personally I am not against including it in 0.8.1, but it's also very easy to merge only after 0.8.1 is out.

vtraag:
I'm fine with merging this after 0.8.1.

ntamas:
Let's postpone this to 0.9 to be consistent with semantic versioning.

vtraag:
Actually, now that I think of it: we *can* actually merge this in `master`, given our policy of that it has to be backwards-compatible with previous minor releases. This PR only adds to the API. Do you agree?

ntamas:
Sure, I'm okay with that.

szhorvat:
@ntamas It seems you made essentially the same change as this PR here:

https://github.com/igraph/igraph/commit/f525ef36bdc84e9ee9f6d9a8211ed43b37541238

We should merge this ASAP.

PR #1342, opened 2020-02-28
plfit: ensure math.h defines are present with MSVC

szhorvat:
Hopefully this will fix the missing `M_LN2`. See this test failure:

https://ci.appveyor.com/project/ntamas/igraph/builds/31127044/job/rpktetnh87thu4x3#L4306

### DIFF
--- a/src/plfit/hzeta.c
+++ b/src/plfit/hzeta.c
@@ -42,8 +42,13 @@
 
 /* Author:  Jerome G. Benoit < jgmbenoit _at_ rezozer _dot_ net > */
 
+#ifdef _MSC_VER
+#define _USE_MATH_DEFINES
+#endif
+
 #include <math.h>
 #include <stdio.h>
+#include "hzeta.h"
 #include "error.h"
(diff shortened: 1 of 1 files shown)


szhorvat:
@ntamas I made a mistake and forgot to push a small modification (which I thought I pushed). I guess it makes more sense to put this in `platform.h` (then change the order of includes so `math.h` is included through `platform.h`).

Take another look at let me know which version you want. Then I'll update the PR and squash the commits.

ntamas:
Seems OK to me. Do you mind if I also include this in ntamas/plfit?

szhorvat:
It's good to merge now. 

I went back to the original change because the header reordering caused another failure that I couldn't fix (or didn't have the patience for).

https://ci.appveyor.com/project/ntamas/igraph/builds/31131963/job/nu8srmk8x061el43#L4310

The only difference from the change you made in the plfit repo is that I also added the `hzeta.h` header (my IDE was complaining about missing declarations).

PR #1344, opened 2020-02-29
Implement global/local efficiency and weighted average shortest paths

### LINKED ISSUE #1343: average_path_length should have a weighted counterpart
`igraph_average_path_length` does not consider weights. This function should have a weighted counterpart (or it should be extended to optionally handle weights).

szhorvat:
This PR implements global and local efficiency.

Global: https://dx.doi.org/10.1103/PhysRevLett.87.198701
Local: http://dx.doi.org/10.1103/PhysRevE.71.036122 (this is a modification/improvement of the "local efficiency" from the previous paper)

These measures are based on averaging not shortest path lengths, but their inverses. This makes it easy to handle non-connected vertex pairs, as their distance can be taken to be infinity, and inverse distance 0.

Since it's closely related, this PR also adds _weighted_ average shortest path computation, and modifies the unweighted one too to reduce code duplication.

Still, there is a lot of code duplication. Much of this is copy/paste from other shortest path implementations with modifications on top. Ideally, eventually we would somehow unify the various implementations without a performance hit. Doing this in C is not easy (C++ would make it easier with templates). I am thinking of something like the preprocessor-based templating used in the maximal clique implementation. Unfortunately, that is ugly, very hard to work with (my IDE can't handle it), and quite a bit of work to get right. So for now, I just went with the copy/paste + modify approach. I needed this functionality right now (with good performance).

Naturally, this is for 0.9. Marked as draft, but it is ready for a basic review. Please comment.

Fixes #1343

----

#### Added

 - `igraph_average_path_length_dijkstra()` (#1343)
 - `igraph_global_efficiency()`
 - `igraph_local_efficiency()`
 - `igraph_average_local_efficiency()`

#### Changed

 - Create a file `shortest_paths.c` and moved over these functions from `structural_properties.c`: `average_path_length()`, `diameter()`, `diameter_dijkstra()`
 - `igraph_average_path_length()` now returns `IGRAPH_INFINITY` when `unconn=false` and the graph is not connected (#1345)
 - `igraph_average_path_length()` now also returns the number of unconnected ordered vertex pairs.
 - `igraph_diameter()` now returns an `igraph_real_t` instead of an `igraph_integer_t`
 - `igraph_diameter()` now returns `IGRAPH_INFINITY` when `unconn=false` and the graph is not connected (#1345)

#### Fixed

 - All functions moved to `shortest_paths.c` now use `igraph_free()`/`igraph_Free()` instead of `free()`


szhorvat:
Note that I started a new file, `shortest_paths.c`. `structural_properties.c` is too large (7000+ lines). Eventually, all shortest path stuff can be moved over to this file.

vtraag:
Thanks for this contribution @szhorvat! I will have to review it some later time (but that should not be a problem if it is only for 0.9).

For now, I would just like to comment on the merging of PRs and the maintenance of different versions. If we want to respect Semantic Versioning (which we do), I think we will really need to have separate branches. Right now there are two PRs in the pipeline for 0.9 (this PR and PR #1341), and the longer we wait, the more conflicts we will get when we only start merging those PRs when we want to release 0.9. If, instead, we can simply merge these PRs to a `develop` branch (which may be unstable), other PRs that target 0.9 can simply build on that branch. Otherwise, all new PRs that target 0.9 (i.e. introduce some API change) will always need to be based on the `master` branch, which by definition will not (yet) include those changes intended for 0.9. For example, if you want to now push another PR for shortest paths that builds on this PR, you cannot, unless you also include all of the changes in this PR in the new PR, which the whole workflow very opaque.

Finally, note that some changes are backwards compatible (I believe both this PR and PR #1341 are backwards compatible). Since we are in development versioning still (0.x), an increment of the MINOR version number could be restricted to backwards incompatible changes, while backwards compatible changes might not need to trigger an increment in the MINOR version number. This means that, in principle, we could allow this PR and PR #1341 to merge already in a 0.8.x release. I'm fine with leaving both PRs for 0.9, but I just wanted to raise this point, because at some point we may want to include backwards compatible changes in a 0.8.x release. So perhaps it would be good to already discuss this point now.

szhorvat:
I did not forget that we were going to change the branching structure. It's not a concern. I'll (figure out how to) move the PR over to the develop branch when that is created. For now, it's a draft (and I'm fixing some bugs just now).

vtraag:
OK, good! Yes, it's no problem retargeting the PR to a `develop` branch, whenever we created it.

szhorvat:
Before this can go ahead, we must make a decision about #1345. This PR adds a weighted average shortest path implementation, which creates a similar inconsistency as the diameter calculation.

szhorvat:
@vtraag I need some help. This PR changes the signature of `igraph_average_path_length` and adds the `unconn_pairs` output argument.

https://github.com/igraph/igraph/pull/1344/files#diff-0940a69f1637f0b1b42209a7c34179d1R113

It is okay to just pass NULL for this argument if one does not want the result.

What is the proper way to include this in `functions.def`? I am hoping that it's easy to just set it to NULL there and not expose it in the R interface *for now*. I don't have the expertise to update the R interface.

szhorvat:
@vtraag Just a small reminder about the above question. I am not very confident poking around in the `functions.def` file

vtraag:
@szhorvat , regarding the `functions.def` I believe this is now correct. The changes will require some further changes in the R interface, since `igraph_average_path_length` is called directly in `rinterface.c.in`, so that it doesn't use the generated one from `functions.def` (if I understand the process correctly). But this only needs to be (and can be) changed when the R interface starts using the C core that includes this PR. We need to make sure the `CHANGELOG` is properly updated, so that we keep track of this.

examples/tests/efficiency.c
```diff
@@ -0,0 +1,104 @@
...
+
+#include "test_utilities.inc"
+
+/* Number of edges in the test graph.
+   This value, 'graph', and 'weights_array' must be consistent. */
+#define N_EDGES 24
+
+int main() {
+    igraph_t graph;
+    const igraph_real_t weights_array[] = {4, 4, 4, 3, 1, 5, 1, 2, 4, 5, 3, 5, 5, 4, 1, 1, 5, 4, 1, 1, 2, 1, 3, 5};
```
ntamas, line 12: add `N_EDGES` here between the brackets and then we can be sure that `N_EDGES` is consistent with the length of the array

examples/tests/efficiency.c
```diff
@@ -0,0 +1,104 @@
...
+    const igraph_vector_t weights;
+    igraph_real_t eff;
+    igraph_vector_t eff_vec;
+
+    igraph_small(&graph, 13, /* directed= */ 1,
+                 0,8, 1,3, 1,4, 1,5, 1,8, 1,10, 2,0, 2,1, 2,4, 3,5, 4,2, 4,7,
+                 4,9, 5,3, 5,10, 6,7, 8,2, 8,3, 8,4, 8,9, 9,3, 9,4, 11,9, 11,3,
+                 -1);
+
+    igraph_vector_view(&weights, weights_array, N_EDGES);
```
ntamas, line 22: we can add an assert here (`igraph_ecount(&graph) == N_EDGES`) and bail out early if the edge count is incorrect

ntamas:
Note to self: the `feature/cmake-build` branch needs to be updated separately to make sure that the new test is called with CMake.

ntamas (approved):

ntamas:
Okay, it looks good to me, assuming that we can make the tests pass :)

I assumed that there were no significant changes to the shortest path routines that were moved to `shortest_paths.c` from `structural_properties.c` that would need my review.

src/shortest_paths.c
```diff
@@ -0,0 +1,1164 @@
+/* -*- mode: C -*-  */
+/* vim:set ts=4 sw=4 sts=4 et: */
+/*
+   IGraph library.
+   Copyright (C) 2005-2012  Gabor Csardi <csardi.gabor@gmail.com>
```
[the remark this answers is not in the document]
szhorvat: Fixed.

examples/tests/efficiency.c
szhorvat: Done.

examples/tests/efficiency.c
szhorvat: Done.

PR #1347, opened 2020-03-03
Make functions 'static' when appropriate

szhorvat:
This PR attempts to make functions 'static' (i.e. local to the translation unit) when appropriate.

This is a step towards cleaning up the code, finding unused functions (the compiler can provide better warnings), sorting internal functions from ones that should be public, and making sure that everything used across multiple translation units has a prototype in a header.

This is not completed yet (still some files to go through) but I'm putting it out here in case anyone has comments.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -56,7 +56,8 @@ AC_HEADER_STDC
 AC_CHECK_HEADERS([stdarg.h stdlib.h string.h time.h unistd.h stdint.h sys/int_types.h])
 LIBS_SAVE=$LIBS
 LIBS="$LIBS -lm"
-AC_CHECK_FUNCS([expm1 rint rintf finite log2 snprintf log1p round fabsl fmin strcasecmp isfinite isnan strdup _strdup ftruncate stpcpy])
+AC_CHECK_FUNCS([expm1 rint rintf finite log2 snprintf log1p round fabsl fmin strcasecmp isnan strdup _strdup ftruncate stpcpy])
+AC_CHECK_DECLS(isfinite,,,[#include <math.h>])
 AC_CHECK_DECL([stpcpy],
     [AC_DEFINE([HAVE_STPCPY_SIGNATURE], [1], [Define to 1 if the stpcpy function has a signature])])
 LIBS=$LIBS_SAVE
--- a/src/arpack.c
+++ b/src/arpack.c
@@ -32,7 +32,7 @@
 
 /* The ARPACK example file dssimp.f is used as a template */
 
-int igraph_i_arpack_err_dsaupd(int error) {
+static int igraph_i_arpack_err_dsaupd(int error) {
     switch (error) {
     case  1:      return IGRAPH_ARPACK_MAXIT;
     case  3:      return IGRAPH_ARPACK_NOSHIFT;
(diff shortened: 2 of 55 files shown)


ntamas:
Thanks, this is great, thank you for working on this!

On a somewhat related note, the current codebase throws a bunch of (probably meaningless) warnings on MSVC; at some point we should probably start cleaning that up as well.

ntamas:
By the way, a rule of thumb is that 99% of the functions prefixed with `igraph_i_` can safely be made static (at least that was the original intention).

vtraag:
Good to start doing this @szhorvat !

I think there is a connection here with argument checking, as discussed in issue #1219. 

If all internal (`igraph_i_*`) functions are only visible within their proper translation unit, then they are also not usable from other `igraph` functions. For example, if we do some argument check (e.g. `igraph_has_multiple`) that we perform only in public `igraph` functions (for example, suppose we would do that in `igraph_hub_score`), and we want to use that function elsewhere within `igraph`, we cannot use the internal counterpart that would not do that check (e.g. `igraph_i_hub_score` in this hypothetical example). 

We should hence make clear the intended use of different functions. There seem three possibilities I believe:
- local use, which should not be used by other parts of `igraph`. These should be declared `static`.
- internal use, which could be used by other parts of `igraph`. These should be declared in an internal header.
- external use, which should be used by external consumers of `igraph`. These should be declared in a public header.

Are there other possibilities that I am missing? Or do we not want to (try to) maintain this distinction between internal/external versions of functions that perform argument checking or not as discussed in issue #1219?

szhorvat:
@vtraag Yes, basically this PR does an imperfect pre-sort of functions to separate "local" from "internal".

Let me show an example of why I think this is important. Suppose we have two translation units, A and B, and B uses a function from A, like this:

```
/* A.c */

int foo(int x) { return x; }
```

```
/* B.c */

/* Prototype of function from A.c */
int foo(int x);

int bar(int x) { return foo(x)*foo(x); }
```

There are many instances of this pattern in igraph.

Now suppose someone goes into `A.c` and makes a change to `foo()`, assuming it is not used anywhere else. Now it's `int foo(int x, int y);`. But the file `B.c` has an entirely separate prototype included, which does not know about `A.c` at all. `bar()` calls `foo()` with one argument, but `foo()` wants to read two arguments off the stack: bang! crash, or more likely something worse: `foo()` subtly returns a wrong result without crashing, and we never detect what is going wrong.

This is why both `A.c` and `B.c` should include the _same_ header, which contains the prototype of `foo()`. Prototypes shouldn't be copied around in an ad-hoc way.

----

What I am doing in this PR is that I add `static` to everything that looks local, and try to compile. If the function was used somewhere else I get a linker error. Now we know it's used in multiple places. For now, I simply add a note to the file where the function is defined, and remove `static`.  If linking succeeded with `static`, then we know that the function is not used anywhere else.

----

After this PR is finished, consider anything named as `igraph_i_` without `static` or an explanatory comment as suspect and in need of investigation. There are so many functions that I can't possibly catch everything.

As a next step, we can try to sort out the header/prototype situation.

ntamas:
I think that the policy in the past has been that the headers in `include/` are the public interface, and any other headers in `src/` are cross-module "internal" headers that may be used "inside" igraph but not "outside". We may have been sloppy here and there by not putting a declaration into an "internal" header and just copying it into the translation unit where it was used, and we should fix those of course. Otherwise, I think this pattern is sustainable for the time being.

szhorvat:
@ntamas If the tests pass, this is good to go.  Hope I didn't mess up anything with all the rebasing ...

PR #1355, opened 2020-03-04
avg_nearest_neighbor_degree: improve documentation

szhorvat:
[The docs were a bit messed up](https://igraph.org/c/doc/igraph-Structural.html#igraph_avg_nearest_neighbor_degree), e.g. they claimed that edge directions are ignored. @vtraag please have a look and make sure it looks fine (I'm tired tonight).

### DIFF
--- a/src/structural_properties.c
+++ b/src/structural_properties.c
@@ -6003,10 +6003,10 @@ int igraph_i_avg_nearest_neighbor_degree_weighted(const igraph_t *graph,
 
 /**
  * \function igraph_avg_nearest_neighbor_degree
- * Average nearest neighbor degree.
+ * Average neighbor degree.
  *
- * Calculates the average degree of the neighbors for each vertex, and
- * optionally, the same quantity as a function of the vertex degree.
+ * Calculates the average degree of the neighbors for each vertex (\p knn), and
+ * optionally, the same quantity as a function of the vertex degree (\p knnk).
  *
  * </para><para>
  * For isolated vertices \p knn is set to NaN.
@@ -6016,11 +6016,12 @@ int igraph_i_avg_nearest_neighbor_degree_weighted(const igraph_t *graph,
  * </para><para>
  * The weighted version computes a weighted average of the neighbor degrees as
  *
- * <code>k_nn_u = 1/s_u sum_u w_uv k_v</code>,
+ * <code>k_nn_u = 1/s_u sum_v w_uv k_v</code>,
  *
- * where <code>s_u</code> is the sum of the incident edge weights of vertex \c u,
- * i.e. its strength. The sum runs over the neighbors \c v of vertex \c u
- * as indicated by \p mode (with appropriate weights)
+ * where <code>s_u = sum_v w_uv</code> is the sum of the incident edge weights
+ * of vertex \c u, i.e. its strength.
+ * The sum runs over the neighbors \c v of vertex \c u
+ * as indicated by \p mode. <code>w_uv</code> denotes the weighted adjacency matrix
  * and <code>k_v</code> is the neighbors' degree, specified by \p neighbor_degree_mode.
  *
  * </para><para>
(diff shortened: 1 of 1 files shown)


src/structural_properties.c
```diff
@@ -6006,36 +6006,50 @@ int igraph_i_avg_nearest_neighbor_degree_weighted(const igraph_t *graph,
...
  *
- * \param graph The input graph, it can be directed but the
- *   directedness of the edges is ignored.
+ * </para><para>
+ * The weighted version computes a weighted average of the neighbor degrees as
+ *
+ * <code>k_nn_u = 1/s_u sum_u w_uv k_v</code>,
+ *
+ * where <code>s_u</code> is the sum of the incident edge weights of vertex \c u,
+ * i.e. its strength. The sum runs over the neighbors \c v of vertex \c u
```
vtraag: I would write here, 
`edge weights of vertex \c u, i.e. its strenght, sum_u w_uv`.

src/structural_properties.c
```diff
@@ -6006,36 +6006,50 @@ int igraph_i_avg_nearest_neighbor_degree_weighted(const igraph_t *graph,
...
+ * as indicated by \p mode (with appropriate weights)
```
vtraag: Here I would write 
`as indicated by \p mode, with weights w_uv`

src/structural_properties.c
```diff
@@ -6006,36 +6006,50 @@ int igraph_i_avg_nearest_neighbor_degree_weighted(const igraph_t *graph,
...
+ * The architecture of complex weighted networks,
+ * Proc. Natl. Acad. Sci. USA 101, 3747 (2004).
+ * https://dx.doi.org/10.1073/pnas.0400087101
+ *
+ * \param graph The input graph. It may be directed.
  * \param vids The vertices for which the calculation is performed.
- * \param mode The neighbors over which is averaged.
- * \param neighbor_degree_mode The degree of the neighbors which is
- *   averaged.
+ * \param mode The type of neihbors to consider in directed graphs.
```
vtraag: Small typo: `neihbors` should be `neighbors`

src/structural_properties.c
```diff
@@ -6006,36 +6006,50 @@ int igraph_i_avg_nearest_neighbor_degree_weighted(const igraph_t *graph,
...
+ *   \c IGRAPH_OUT considers out-neighbours, \c IGRAPH_IN in-neighbours
```
vtraag: Here you use UK spelling, I believe we use US spelling throughout, so it should be `neighbors` instead of `neighbours`.

src/structural_properties.c
```diff
@@ -6006,36 +6006,50 @@ int igraph_i_avg_nearest_neighbor_degree_weighted(const igraph_t *graph,
...
+ *   and \c IGRAPH_ALL ignores edge directions.
+ * \param neighbor_degree_mode The type of degree to compute in directed graphs.
```
vtraag: I would write here: `The type of degree in directed graphs to average`

src/structural_properties.c
```diff
@@ -6006,36 +6006,50 @@ int igraph_i_avg_nearest_neighbor_degree_weighted(const igraph_t *graph,
  * Average nearest neighbor degree.
```
vtraag: I'm now actually starting to wonder why this is called `avg_`**`nearest`**`_neighbor_degree`. Shouldn't we actually change this to `avg_neighbor_degree`? "**Nearest** neighbor" seems to imply that we can also do something for neighbors that are further removed (in whatever sense), which we do not.

src/structural_properties.c
```diff
@@ -6006,36 +6006,50 @@ int igraph_i_avg_nearest_neighbor_degree_weighted(const igraph_t *graph,
...
+ *   and \c IGRAPH_ALL ignores edge directions.
+ * \param neighbor_degree_mode The type of degree to compute in directed graphs.
+ *   \c IGRAPH_OUT averages out-degrees, \c IGRAPH_IN averages in-degrees
+ *   and \c IGRAPH_ALL ignores edge directions for the degree calculation.
  * \param vids The vertices for which the calculation is performed.
  * \param knn Pointer to an initialized vector, the result will be
- *   stored here. It will be resized as needed. Supply a NULL pointer
+ *   stored here. It will be resized as needed. Supply a \c NULL pointer
  *   here, if you only want to calculate \c knnk.
  * \param knnk Pointer to an initialized vector, the average nearest
```
vtraag: Here also the "**nearest** neighbor" sounds a bit unclear, as if we can also include less near neighbors.

src/structural_properties.c
szhorvat: Agree to keep US spelling, and I have already been trying to do so. This was a mistake (usually I use UK spelling so I have to make an extra effort here).

szhorvat:
@vtraag I tried to address all your comments, though I did not always literally take the suggestion. I also fixed some incorrect index notation.

About "nearest", I was thinking the same myself. I removed this word from the docs. I believe it is used because the references paper is using the same name. However, igraph itself generally uses "neighbour" to refer to first neighbours. We need to be consistent within igraph.

src/structural_properties.c
vtraag: Yeah, I have to same problem often: I tend to default to UK spelling.

vtraag:
Thanks for the changes, they look good to me! Indeed, the sum had an incorrect index, I did not yet notice.

We can probably best squash these two commits when merging?

> About "nearest", I was thinking the same myself. I removed this word from the docs.

OK, so we may want to update the name in 0.9.0 for example (but keep the old name with a deprecation message)?

szhorvat:
@vtraag  If you're happy with it, please go ahead and "squash and merge" Apparently GitHub has a button for this (which I didn't know before). I don't want more than one commit on master for a simple change like this. If you want to make small changes, feel free to do so on this branch (GitHub's web-based editor is very convenient).

PR #1359, opened 2020-03-08
Various small code cleanups

szhorvat:
This PR contains various small cleanups (with the `#include` cleanup touching lots of files). Please see the commits for details.

----

Any unneeded header includes are removed and headers are included in a consistent order, from most specific to least specific, i.e.:

 - igraph headers
 - `config.h` (unless there's a reason to put it in a different location in the header order)
 - system headers

----

Some always-true conditionals are removed from `sir.c` (if those pointers were NULL, the preceding code would fail too in that function).

----

Consistent header guard naming (I was uncomfortable with names like `ARPACK_H` that lacked the `IGRAPH_` prefix)

----

Re-encode some files are UTF8

----

Comments and minor fixes

### DIFF
--- a/src/cattributes.c
+++ b/src/cattributes.c
@@ -23,10 +23,10 @@
 
 #include "igraph_attributes.h"
 #include "igraph_memory.h"
-#include "config.h"
 #include "igraph_math.h"
 #include "igraph_interface.h"
 #include "igraph_random.h"
+#include "config.h"
 
 #include <string.h>
 
--- a/src/drl_graph.h
+++ b/src/drl_graph.h
@@ -37,7 +37,13 @@
 #include "DensityGrid.h"
 #include "igraph_layout.h"
 
+#include <map>
+#include <vector>
+#include <ctime>
+
 namespace drl {
+
+using namespace std;
 
 // layout schedule information
 struct layout_schedule {
--- a/src/drl_graph.h
+++ b/src/drl_graph.h
@@ -43,8 +43,6 @@
 
 namespace drl {
 
-using namespace std;
-
 // layout schedule information
 struct layout_schedule {
     int iterations;
@@ -87,8 +85,8 @@ private:
     void update_nodes ( );
     float Compute_Node_Energy ( int node_ind );
     void Solve_Analytic ( int node_ind, float &pos_x, float &pos_y );
-    void get_positions ( vector<int> &node_indices, float return_positions[2 * MAX_PROCS] );
-    void update_density ( vector<int> &node_indices,
+    void get_positions ( std::vector<int> &node_indices, float return_positions[2 * MAX_PROCS] );
+    void update_density ( std::vector<int> &node_indices,
                           float old_positions[2 * MAX_PROCS],
                           float new_positions[2 * MAX_PROCS] );
     void update_node_pos ( int node_ind,
@@ -101,11 +99,11 @@ private:
     // graph decomposition information
     int num_nodes;                  // number of nodes in graph
     float highest_sim;              // highest sim for normalization
-    map <int, int> id_catalog;      // id_catalog[file id] = internal id
-    map <int, map <int, float> > neighbors;     // neighbors of nodes on this proc.
+    std::map <int, int> id_catalog;      // id_catalog[file id] = internal id
+    std::map <int, std::map <int, float> > neighbors;     // neighbors of nodes on this proc.
 
     // graph layout information
-    vector<Node> positions;
+    std::vector<Node> positions;
     DensityGrid density_server;
 
     // original VxOrd information
--- a/src/drl_graph_3d.h
+++ b/src/drl_graph_3d.h
@@ -43,8 +43,6 @@
 
 namespace drl3d {
 
-using namespace std;
-
 // layout schedule information
 struct layout_schedule {
     int iterations;
@@ -79,8 +77,8 @@ private:
     void update_nodes ( );
     float Compute_Node_Energy ( int node_ind );
     void Solve_Analytic ( int node_ind, float &pos_x, float &pos_y, float &pos_z );
-    void get_positions ( vector<int> &node_indices, float return_positions[3 * MAX_PROCS] );
-    void update_density ( vector<int> &node_indices,
+    void get_positions ( std::vector<int> &node_indices, float return_positions[3 * MAX_PROCS] );
+    void update_density ( std::vector<int> &node_indices,
                           float old_positions[3 * MAX_PROCS],
                           float new_positions[3 * MAX_PROCS] );
     void update_node_pos ( int node_ind,
@@ -93,11 +91,11 @@ private:
     // graph decomposition information
     int num_nodes;                  // number of nodes in graph
     float highest_sim;              // highest sim for normalization
-    map <int, int> id_catalog;      // id_catalog[file id] = internal id
-    map <int, map <int, float> > neighbors;     // neighbors of nodes on this proc.
+    std::map <int, int> id_catalog;      // id_catalog[file id] = internal id
+    std::map <int, std::map <int, float> > neighbors;     // neighbors of nodes on this proc.
 
     // graph layout information
-    vector<Node> positions;
+    std::vector<Node> positions;
     DensityGrid density_server;
 
     // original VxOrd information
--- a/src/drl_graph_3d.h
+++ b/src/drl_graph_3d.h
@@ -37,7 +37,13 @@
 #include "DensityGrid_3d.h"
 #include "igraph_layout.h"
 
+#include <map>
+#include <vector>
+#include <ctime>
+
 namespace drl3d {
+
+using namespace std;
 
 // layout schedule information
 struct layout_schedule {
--- a/src/drl_parse.h
+++ b/src/drl_parse.h
@@ -41,8 +41,6 @@
 
 namespace drl {
 
-using namespace std;
-
 class parse {
 
 public:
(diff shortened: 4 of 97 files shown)


szhorvat:
@ntamas Is there a better solution here?

https://github.com/igraph/igraph/pull/1359/commits/edcb7d1f0c33aeaf5592aa32752575b9e33d2ff5

See just above [this](https://igraph.org/c/doc/igraph-Error.html#igraph_strerror) why this change was necessary.

ntamas:
>  Is there a better solution here?

Probably not, other than replacing the documentation generator entirely with a solution that does not rely on dumb regular expressions :)

src/cattributes.c
```diff
@@ -23,10 +23,10 @@
 
 #include "igraph_attributes.h"
 #include "igraph_memory.h"
-#include "config.h"
 #include "igraph_math.h"
 #include "igraph_interface.h"
 #include "igraph_random.h"
+#include "config.h"
```
ntamas: Be careful with moving the include of `config.h` around; it might be the case that some of the headers _below_ its original location but _above_ the current one depended on the presence of macros defined in `config.h`. If we decide to move `config.h` around, I'd probably move it _above_ all igraph-related includes.

src/drl_graph.h
```diff
@@ -37,8 +37,14 @@
 #include "DensityGrid.h"
 #include "igraph_layout.h"
 
+#include <map>
+#include <vector>
+#include <ctime>
+
 namespace drl {
 
+using namespace std;
```
ntamas: I'd rather avoid `using namespace` in headers. I know that in this particular place it is in a `namespace {...}` block so the imported namespace should not leak out; however, there are still possibiities for [subtle bugs](https://stackoverflow.com/a/6175850/156771) that you can introduce this way.

src/drl_graph_3d.h
```diff
@@ -37,8 +37,14 @@
 #include "DensityGrid_3d.h"
 #include "igraph_layout.h"
 
+#include <map>
+#include <vector>
+#include <ctime>
+
 namespace drl3d {
 
+using namespace std;
```
ntamas: Same as above regarding the usage of `using namespace` in headers.

src/drl_parse.h
```diff
@@ -37,8 +37,12 @@
     #include <mpi.h>
 #endif
 
+#include <string>
+
 namespace drl {
 
+using namespace std;
```
ntamas: Same as above regarding the usage of `using namespace` in headers.

src/flow.c
```diff
@@ -33,15 +33,12 @@
 #include "igraph_structural.h"
 #include "igraph_components.h"
 #include "igraph_types_internal.h"
-#include "config.h"
 #include "igraph_math.h"
 #include "igraph_dqueue.h"
-#include "igraph_visitor.h"
 #include "igraph_interrupt_internal.h"
 #include "igraph_topology.h"
+#include "config.h"
```
ntamas: Again, watch out for the ordering of imports when moving `config.h` around.

src/hrg_splittree_eq.h
```diff
@@ -62,7 +62,7 @@
 #ifndef IGRAPH_HRG_SPLITTREE
 #define IGRAPH_HRG_SPLITTREE
 
-#include <iostream>
+#include <string>
 
 using namespace std;
```
ntamas: Another example of `using namespace` in headers that should be avoided. You did not introduce this, but if we are fixing headers anyway, we should probably grep the codebase for `using namespace` and get rid of those that are in headers.

src/drl_graph.h
szhorvat: I fully agree with that, but how shall we fix things then? Add the `std::` in the headers where needed?

The `using namespace std;` was already all over the place in the DrL code.  I was trying to clean that up and move it into the namespace (at the same time as cleaning up the includes). What actually happened here was that `drl_graph.h` includes `DensityGrid.h` which has a naked `using namespace std;` (which I forgot to move into the namespace as well).

What I'm trying to say is that I didn't just add a `using namespace` arbitrarily here. I was actually trying to reduce the number of naked `using namespace`s.

What is your preference?

 1. add the `std::` in headers (not that much work)
 2. move the `using namespace std;` into the `drl` namespace everywhere
 3. something else? 

The DrL code is not the only one that has this in headers. The HRG code also does.

ntamas: We should probably add `std::` in the headers where applicable.

src/cattributes.c
szhorvat: My original thought was that in that case the header itself should include `config.h`, thus if moving `config.h` to the end causes errors, then we found a bug that needs to be fixed. This is generally the reasoning behind moving system headers to the end: this practical reveals cases when a system header should have been included in a library header, but it wasn't.

But let me think for a moment because `config.h` is different.  I am going to be thinking aloud a but below. Please correct any mistaken assumptions.

`config.h` does not get installed (`make install`) so no public header should depend on it. Normally, any public header that may depend on thing in `config.h` is generated from an `.in` files (like `igraph_threading.h.in`).

I `grep`ped the `include` directory for `HAVE_` and found just one header: `igraph_vector.h` contains:

```
#ifdef HAVE_STDINT_H
    #include <stdint.h>
#else
    #if defined(HAVE_SYS_INT_TYPES_H) && HAVE_SYS_INT_TYPES_H
        #include <sys/int_types.h>    /* for Solaris */
    #endif
#endif
```

@ntamas Do you know if this is fine? None of these `HAVE_` macros will be defined when using igraph as a library. They would only be defined when compiling igraph itself. Is this intentional? Or is this a bug? It looks like a bug to me.

----

Now let's look at internal headers:

```
grep HAVE_ -R src --include '*.h*'

src/plfit/arithmetic_sse_float.h:#if     HAVE_XMMINTRIN_H
src/plfit/arithmetic_sse_float.h:#endif/*HAVE_XMMINTRIN_H*/
src/plfit/arithmetic_sse_double.h:#if     HAVE_EMMINTRIN_H
src/plfit/arithmetic_sse_double.h:#endif/*HAVE_EMMINTRIN_H*/
src/bliss/defs.hh:#if HAVE_GMP == 1
src/igraph_glpk_support.h:#ifdef HAVE_GLPK
src/igraph_math.h:#ifndef HAVE_LOG2
src/igraph_math.h:#ifndef HAVE_LOG1P
src/igraph_math.h:#ifndef HAVE_FABSL
src/igraph_math.h:#ifndef HAVE_FMIN
src/igraph_math.h:#ifndef HAVE_ROUND
src/igraph_hacks_internal.h:#ifndef HAVE_STRDUP
src/igraph_hacks_internal.h:#ifndef HAVE_STPCPY
src/igraph_hacks_internal.h:    #ifndef HAVE_STPCPY_SIGNATURE
src/igraph_arpack_internal.h:#ifdef HAVE_GFORTRAN
src/bignum.h:#ifdef HAVE_STDINT_H
src/bignum.h:    #ifdef HAVE_SYS_INT_TYPES_H
src/bignum.h:#ifndef HAVE_U64
src/bignum.h:    #define HAVE_U64 1
src/bignum.h:#if HAVE_U64
```

I verified that all of these internal headers have `#include "config.h"` at the beginning.

The one exception is `plfit`. However, those `HAVE_` macros are not in `config.h` and as far as I can tell will never be defined (??).  Is there anything that needs to be done here? As far as I can tell, those intrinsics are being used anyway.

src/drl_graph.h
szhorvat: I pushed a commit that takes care of this in more headers. I had to add `using namespace std;` to the cpp files sometimes.

I did not touch Bliss, I'll take up the issue with the author as he's preparing a new release anyway.

I did not touch simpleraytracer.

```
grep 'using namespace std' -R . --include '*.h*'

./optional/simpleraytracer/Point.h:using namespace std;
./optional/simpleraytracer/Light.h:using namespace std;
./optional/simpleraytracer/Shape.h:using namespace std;
./src/bliss/utils.hh:using namespace std;
```

szhorvat:
I added one more commit where I comment out the prototypes of some functions that were already commented out earlier. This gets rid of a few compiler warnings.

Please check that one as well. 

I responded to your comments in the review. The `using namespace std` is taken care of. Looking at the `config.h` revealed a potential issue in `igraph_vector.h`. I think this should not be there, but it's unclear why it's there in the first place:

```
#ifdef HAVE_STDINT_H
    #include <stdint.h>
#else
    #if defined(HAVE_SYS_INT_TYPES_H) && HAVE_SYS_INT_TYPES_H
        #include <sys/int_types.h>    /* for Solaris */
    #endif
#endif
```

On macOS is compiles fine without that part, and I do not see why `stdint.h` would ever be needed.

See my comment on the review for more details.

szhorvat:
https://github.com/igraph/igraph/commit/6b68ff5f12e700fe0f01ac65567a5b0e589cf626

szhorvat:
The original stdint.h inclusion comes from https://github.com/igraph/igraph/commit/a8092fbb06efefbcd784ef4ac651fa05a79f119d

src/cattributes.c
ntamas: > `config.h` does not get installed (make install) so no public header should depend on it.

That's correct.

> Do you know if this is fine? None of these HAVE_ macros will be defined when using igraph as a library.

This is probably not fine; the question is whether there's anything in `igraph_vector.h` that needs `stdint.h` or `sys/int_types.h`. If there isn't, we can safely remove the whole block. If there is, we should probably find a way to work around it somehow. (But I think that this particular piece of code caters for an obscure use-case that I've found when I tried to get igraph compiled on an old Solaris server several years ago, and it might be the case that it's easier to say that we don't support old Solaris server builds any more).

> The one exception is `plfit`. However, those `HAVE_?MMINTRIN_H` macros are not in `config.h` and as far as I can tell will never be defined (??). 

These macros come from the upstream `plfit` distribution. The L-BFGS optimizer in `plfit` uses SSE2 optimized instructions if `emmintrin.h` is present. If it is not, it simply falls back to "normal" arithmetics. (See the `arithmetic_*.h` headers in `src/plfit`). So, this doesn't hurt, but we could improve things by also testing for `emmintrin.h` in the configure script and then allowing `src/plfit/lbfgs.c` to use SSE2 optimizations.

src/drl_graph.h
ntamas: Okay, that's fine, thanks!

ntamas:
Thanks, merged!

PR #1362, opened 2020-03-24
igraph_sir: interruption support and other fixes

szhorvat:
Please see commit messages for details.

### DIFF
--- a/src/sir.c
+++ b/src/sir.c
@@ -137,7 +137,7 @@ int igraph_sir(const igraph_t *graph, igraph_real_t beta,
         IGRAPH_ERROR("Number of SIR simulations must be positive", IGRAPH_EINVAL);
     }
 
-    igraph_is_simple(graph, &simple);
+    IGRAPH_CHECK(igraph_is_simple(graph, &simple));
     if (!simple) {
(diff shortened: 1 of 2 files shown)


ntamas:
Seems like a tug-of-war between `xz` and `zip` (well, their dependencies). Guess we could remove `xz` from the default installation as we don't need it, but I expect this to be a temporary problem.

By the way, the commits seem okay for me.

szhorvat:
Please wait with merging, there's an occasional crash with the interruption.

szhorvat:
OK, the problem was that `igraph_i_sir_destroy` did not set the destroyed `vector_ptr` elements to `NULL`. Therefore the caller could not detect that some elements should not be destroyed when an interruption occurred.

The fix was a bit ugly because I had to get rid of the `s` variable. Alternatively, I could have made used an `igraph_sir_t **s`, but then the compiler warns about incompatible pointer types (since assigning from a `void **`). I don't see any alternative. @ntamas ?

I also added two other commits.

szhorvat:
Just clearing the `vector_ptr` at the end of `igraph_i_sir_destroy` would have been simpler, I guess ...

PR #1365, opened 2020-03-31
all_st_mincuts: check whether capacities are strictly positive.

### LINKED ISSUE #1363: igraph_all_st_mincuts should not allow zero capacities
`igraph_all_st_mincuts` does not complain about zero capacities, but simply drops zero-capacity edges from the result. It should not allow zero capacities.

Example graph with capacities:

image

Asking for all minimum cuts from 1 to 2:

```
In[188]:= IGFindMinimumCuts[g, 1, 2]
Out[188]= {{1 \[DirectedEdge] 2}}
```

Notice that `1 -> 2` is not a cut. `{1 -> 2, 1 -> 3}` would be a cut. But `1->3` has zero capacity and is dropped from the result.

@vtraag  Can you take a look, since you touched this function last time?

vtraag:
The capacities for `igraph_all_st_mincuts` are asssumed to be strictly positive, but this was not yet checked. This PR also updates the docs to make clear that not all minimal cuts may be found due to numerical precision errors. This fixes #1363.

### DIFF
--- a/src/st-cuts.c
+++ b/src/st-cuts.c
@@ -1284,10 +1284,13 @@ static int igraph_i_all_st_mincuts_pivot(const igraph_t *graph,
  * \function igraph_all_st_mincuts
  * All minimum s-t cuts of a directed graph
  *
- * This function lists all minimum edge cuts between two vertices, in a
- * directed graph. The implemented algorithm
- * is described in JS Provan and DR Shier: A Paradigm for listing
- * (s,t)-cuts in graphs, Algorithmica 15, 351--372, 1996.
+ * This function lists all edge cuts between two vertices, in a directed graph,
+ * with minimum total capacity. Possibly, multiple cuts may have the same total
+ * capacity, although there is often only one minimal cut in weighted graphs.
+ * Due to numerical errors, it is possible that this function will not find all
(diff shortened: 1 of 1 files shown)


szhorvat:
I would like to recommend integer weights in the docs, as these will be free from numerical errors.

Do you want me to go ahead and edit it, or do you want to do it?

vtraag:
Feel free to edit! Does this really hold generally? This means that all integers should be representable as floats. I'm not sure whether that is the case for all architectures and OS's. Probably this will (nearly) never be a problem, but it might be possible, wouldn't it? But still, the general recommendation makes sense.

szhorvat:
I don't mean integers as `int` or `long`. I mean integer values, with no fractional part.

Generally, what is needed is that both the weights and their sums be exactly representable in binary. Integer multiples of `1/2` or `1/4` would also work in practice, but I think that saying so may be too confusing for many users.

I'll edit it a bit in the morning, with a clearer head.

ntamas:
> This means that all integers should be representable as floats. I'm not sure whether that is the case for all architectures and OS's.

Pretty much all architectures use IEEE754 single- or double-precision floats nowadays, where it is guaranteed that integers have an exact representation. [This] is an excellent read about the pitfalls of IEEE754 floating point arithmetic.

Issue #1345, opened 2020-03-01
Shortest-path based measures in disconnected graphs


szhorvat:
Some shortest-path based calculations, such as diameter or average shortest path, have an `unconnected` option.  When it is set to `true`, shortest paths are considered only among pairs of vertices reachable from each other. If it is set to `false`, all pairs are considered, even unconnected ones.

There are two functions for diameter, an unweighted and a weighted one. With `unconnected=false`, they behave differently:

 1. The unweighted one considers the distance between non-connected pairs to be `V`, the number of vertices (the justification being that `V` is greater than any possible shortest path length in the graph)
 2. The weighted one considers the distance between non-connected pairs to be infinity.

However, the weighted function will simply call the unweighted one when no weight vector is present. This means that _usually_ the weighted function treats unweighted graph as if all edge weights were 1, _but not when the graph is disconnected_.

This is inconsistent and very confusing to users. Keep in mind that a high-level interface will typically want to present a unified interface for the diameter calculation and not have two separate functions for weighted/unweighted. Thus this is what happens:

```r
> g<-make_graph(c(1,2, 3,4))
> diameter(g, unconnected = F)
[1] 4
> E(g)$weight <- c(1,1)
> diameter(g, unconnected = F)
[1] Inf
```

----

I think that the behaviour of the unweighted function with disconnected graphs is ad-hoc. There is no reasonable justification for it. I would go as far as to call it incorrect.

I suggest we change it so that even unweighted functions will consider the distance between non-connected pairs to be infinity, and _not_ `V`.

Yes, this will in a way break backwards compatibility, but I think that's okay: if anyone has been taking this value to be `V` just because igraph does this, then perhaps igraph did them a disservice. Lots of people use igraph who are not very experienced with network analysis and cannot always make a good judgement about how to handle these situations.

----

We need to come to a decision about this so that I can proceed with #1343 and #1344. Let me know what you think.  @gaborcsardi as far as I can tell, you implemented these (correct me if I'm wrong). Would you be okay with this change?  Is there any reasonable use case for taking `V` as the distance? It seems to be that this choice may have originally been a hack that stuck around for too long.

vtraag:
I would be in favour of returning infinity always, for both weighted and unweighted versions. This is a clear return value, allows to substitute it for `V` (or any other value you desire), and is consistent with most (or all) definitions of shortest path distance that I know of.

szhorvat:
[...]

PR #1366, opened 2020-04-04
Introduction of develop branching structure

vtraag:
This PR introduces our intended branching structure. 

To recap, we will respect [semantic versioning](https://semver.org/). For the development release (0.x) we follow a specific procedure. Any released minor version (`0.x.z`) will be API backwards-compatible with any previous release of the *same* minor version (`0.x.y`, with `y < z`). If there is an API incompatible change, we will increase the minor version. 

We will have two versions of `igraph`: the most recent released version, and the next upcoming minor release, which is by definition not yet released. The most recent release version is in the `master` branch, while the next upcoming minor release is in the `develop` branch. API-incompatible changes with the current release **must** be merged to the `develop` branch. If changes are API backwards-compatible, it **can** be merged to the `master` branch.

In most cases, changes in the `master` branch should also be merged to the `develop` branch by the maintainers, in order to keep them as much as possible in sync, and make sure that any improvements to the `master` branch will also be included in the next minor release in the `develop` branch.

All of this is explained in the `CONTRIBUTING.md`. I first would like to merge this PR before we setup the new `develop` branch. That way the `CONTRIBUTING.md` is already immediately correct in the `develop` branch as well.

Finally, in order to make the versioning system clear, I propose we switch to a slightly different versioning system (as implemented in `tools/getversion.sh`). All intermediate commits of a release will be versioned as `0.x.y-post+<sha>` in the `master` branch. All commits in the `develop` branch will be versioned as `0.a.b-pre+<sha>`, none of which are released yet of course. In other words, any commits in the `master` branch will be versioned as `0.8.1-post+<sha>` after this PR gets merged. In the `develop` branch, commits should then be versioned as `0.9.0-pre+<sha>`. This will be done automatically by adding back the `NEXT_VERSION` file. If we wouldn't change the versioning system, commits in `master` that end up being part of `0.8.2` would first be labelled `0.9.0-pre+<sha>` in the current versioning system, incorrectly suggesting that it contains API-incompatible changes. I hope you agree that this is more clear.

Let me know if you have any comments or are OK with this.


vtraag:
OK, thanks for merging. I created the `develop` branch. I think we can now go ahead and merge in changes that we want to make for version `0.9.0` (e.g. #1341 and #1344 ) to the `develop` branch. @szhorvat, I will try to review #1344 in not too much time, are you in a hurry?

szhorvat:
No hurry, and it's actually not fully done. I wanted to change the lazy inclist to a normal inclist, since there seems to be no point in having it lazy. Much of the Dijkstra code was copied and adapted from other functions.

I was also going to make changes to diameter calculations, as discussed in #1345.

Other than the inclist thing, please consider the efficiency functions in the PR complete. Feedback is welcome.

PR #1379, opened 2020-04-29
Fix arpack callback for sparse matrixes

khitrin:
Fix wrong vector usage in igraph_i_eigen_matrix_sym_arpack_cb: source vector was completely ignored leading to incorrect multiplications in igraph_eigen_matrix_symmetric for sparse matrix.

Also, there is the duplicate code in sparsemat.c (function igraph_i_sparsemat_arpack_multiply), but I propose to add small patch to next bugfix release without major refactoring.


ntamas:
Thanks a lot!

PR #1384, opened 2020-05-19
Pull in changes from the CRAN branch.

vtraag:
This PR is meant to discuss the changes that were made in the `cran` branch. It seems that there have been some useful changes, but also some changes were superseded by more recent commits in the `master` branch.

I will simply go through all the changes, and indicate which changes I think should be corrected. I believe we will be able to merge to the `master`, if it turns out that we will need to make API breaking changes, we could retarget it to `develop`.

### DIFF
--- a/src/CHOLMOD/Include/cholmod_blas.h
+++ b/src/CHOLMOD/Include/cholmod_blas.h
@@ -106,7 +106,7 @@
 #elif defined (BLAS_NO_UNDERSCORE)
 
 #define BLAS_DTRSV igraphdtrsv
-#define BLAS_DGEMV igraphdgemv
+#define BLAS_DGEMV igraphxdgemv
 #define BLAS_DTRSM igraphdtrsm
 #define BLAS_DGEMM igraphdgemm
 #define BLAS_DSYRK igraphdsyrk
@@ -126,7 +126,7 @@
 #else
 
 #define BLAS_DTRSV igraphdtrsv_
-#define BLAS_DGEMV igraphdgemv_
+#define BLAS_DGEMV igraphxdgemv /* this is a macro */
 #define BLAS_DTRSM igraphdtrsm_
 #define BLAS_DGEMM igraphdgemm_
 #define BLAS_DSYRK igraphdsyrk_
@@ -172,7 +172,7 @@
 /* === BLAS and LAPACK prototypes and macros ================================ */
 /* ========================================================================== */
 
-int BLAS_DGEMV (char *trans, BLAS_INT *m, BLAS_INT *n, double *alpha,
+void BLAS_DGEMV (char *trans, BLAS_INT *m, BLAS_INT *n, double *alpha,
 	double *A, BLAS_INT *lda, double *X, BLAS_INT *incx, double *beta,
 	double *Y, BLAS_INT *incy) ;
 
--- a/src/DensityGrid.cpp
+++ b/src/DensityGrid.cpp
@@ -70,24 +70,10 @@ DensityGrid::~DensityGrid ()
 void DensityGrid::Init() 
 {
   
-  try
-    {
-      Density = new float[GRID_SIZE][GRID_SIZE];
-      fall_off = new float[RADIUS*2+1][RADIUS*2+1];
-      Bins = new deque<Node>[GRID_SIZE*GRID_SIZE];
-    }
-  catch (bad_alloc errora)
-    {
-      // cout << "Error: Out of memory! Program stopped." << endl;
-	  #ifdef MUSE_MPI
-        MPI_Abort ( MPI_COMM_WORLD, 1 );
-	  #else
-	    igraph_error("DrL is out of memory", __FILE__, __LINE__,
-			 IGRAPH_ENOMEM);
-		return;
-	  #endif
-    }
-	
+  Density = new float[GRID_SIZE][GRID_SIZE];
+  fall_off = new float[RADIUS*2+1][RADIUS*2+1];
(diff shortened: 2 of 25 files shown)


vtraag:
Perhaps one immediate question: why were some of the defintions changed from `igraph...` to `igraphx...` @gaborcsardi? Is this necessary for any particular reason?

vtraag:
I would like to raise the following points:
- It seems that most of the explicit `(void*)` casts could be easily integrated.
- There are a couple of return type changes from `int` to `void`. This might reduce some warnings, I will have to see.
- There are a couple of changes from `int` to `long int`, perhaps this is indeed necessary to accommodate some use cases. Perhaps this is also necessary in order to support R (better). Can you comment @gaborcsardi?

ntamas:
It looks like some of the changes should be kept in the `cran` branch unless we find a workaround that makes both us and CRAN happy. For instance, https://github.com/igraph/igraph/pull/1384/commits/df716b975d439477dcc690dd183a83284caab0c9 removes a `try..catch` handler from the C++ code because it makes CRAN unhappy, but it also means that we won't return an `IGRAPH_ENOMEM` when we cannot allocate enough memory.

https://github.com/igraph/igraph/pull/1384/commits/1eee8990081dbd4c5418ba871233e3c0a5c11b35 is legit and it can easily be merged into the master branch.

https://github.com/igraph/igraph/pull/1384/commits/d47463093bf83fa2fa6b01c285b772e5b30e3380 actually seems to be multiple commits in one; some changes (e.g., the ones in the parsers) are unrelated to R and ARPACK. They probably belong to the next commit, but I'm okay with these in general.

The magic with `igraphxdgemv` vs `igraphdgemv` seems to work around some imcompatibility between the `dgemv` function when igraph is built with `gfortran` vs when it is built without it; I cannot comment on why this is necessary.

I don't see why the explicit `(void*)` casts in `realloc()` are necessary; I wonder what sort of warning it triggers on CRAN's side -- isn't it allowed to pass in an arbitrary pointer to a function that expects `void*`?

szhorvat:
> I don't see why the explicit `(void*)` casts in `realloc()` are necessary; I wonder what sort of warning it triggers on CRAN's side -- isn't it allowed to pass in an arbitrary pointer to a function that expects `void*`?

I agree, casting to void should not be necessary in C. In fact, after some searching, it seems to be considered bad practice because it can hide mistakes (because the compiler won't complain about using `void *` where a more specific pointer is required). I also tried if I can trigger warnings, but I cannot, even with `-Wpedantic`. https://godbolt.org/z/GzkXoX

In C++, this cast would be needed, but not in C.

szhorvat:
The Bliss modification, I'll check that in more detail later.

ntamas:
> In C++, this cast would be needed, but not in C.

That could be an issue. I don't know how the CRAN pipeline works, but it could be the case that they compile everything with a C++ compiler by default.

szhorvat:
While we're on the topic of R-specific changes, take a look at this:

https://github.com/igraph/igraph/blob/develop/src/cliquer/misc.h#L38

A Cliquer error would terminate the process, but when `USING_R`, it just call a generic R `error` function.

[This is why I still believe that we need a special `igraph_abort()`](https://igraph.discourse.group/t/c-igraph-0-8-1-release/127/24) that is a replacement for [`abort()`](https://en.cppreference.com/w/c/program/abort) but does not terminate the process.

szhorvat:
> > In C++, this cast would be needed, but not in C.
> 
> That could be an issue. I don't know how the CRAN pipeline works, but it could be the case that they compile everything with a C++ compiler by default.

The Cliquer stuff can't be compiled as C++ at all because it uses `new` as a variable name. So it cannot be this. I know this because I tried to compile it as C++ in order to deal with that `ASSERT` macro I define above. The idea was to use an exception to bail out. (I no longer think that's a good idea.)

gaborcsardi:
I don't think you need to merge this, I can try to use the master branch again for the next R release, and fix there whatever issues come up.

ntamas:
I will go through the changes in this branch and merge some of these changes (the ones that I do understand and that cause no semantic changes) manually into master. Then we can close this PR and try to build igraph's C core from the `master` branch. If the build fails on CRAN, let's try to fix them again based on the `cran` branch to see which one of the changes are still necessary, and merge them back into `master` if they are not too R-specific to avoid having to maintain two separate branches.

ntamas:
@szhorvat 

> While we're on the topic of R-specific changes, take a look at this:

Having an `igraph_abort()` is a good idea and I'm all in favour of it. The R headers do the following to mark a function as not returning:

```
#if defined(__GNUC__) && __GNUC__ >= 3
#define NORET __attribute__((noreturn))
#else
#define NORET
#endif
```

Clang has `_Noreturn` instead; we need to incorporate this as well.

ntamas:
Closing this PR; I have merged a few changes from this PR back to master (and then develop) where it made sense.

As for the remaining hackery with BLAS, LAPACK etc -- I am more and more convinced that we should simply get rid of all the embedded libraries in igraph and just admit it that igraph is a complex library with lots of dependencies that have to be installed first. This should probably be discussed in the forum first, though, so I'm not going to do anything about it right now. Let me know if you have opinions about this.

vtraag:
Thanks for cleaning this up @ntamas ! Shall we remove the `cran` branch to make sure there will be no confusion in the future?

ntamas:
It is @gaborcsardi 's turf; there might be some nightly build servers at CRAN or anywhere else that look for this branch, so let's keep it around until he is okay with the removal.

PR #1386, opened 2020-05-24
Gamma resolution for louvain/multilevel community detection

jmonlong:
I added a the gamma resolution parameter for louvain/multilevel community detection. This stems from a request on the rigraph repo: [https://github.com/igraph/rigraph/issues/239](https://github.com/igraph/rigraph/issues/239).

It compiles and the test from `make check` worked on my machine (ubuntu). I also wrote the rigraph side of it and tested the new parameters on a few examples. It performs as expected so all looks good to me.

I'm not sure what's the next steps to coordinate with the rigraph repo. Should I wait until this is merged here (or in the *cran* branch?) before creating a pull request in the rigraph repo?


szhorvat:
@jmonlong Thanks so much for the PR!

First, a note that this will need to go on the `develop` branch, as it breaks the API (we can deal with this later).

@vtraag What is your opinion on modifying the `igraph_modularity` function directly? I notice that you did not do this when contributing `igraph_community_leiden`, which uses a similarly defined quality parameter. Would it be better to separate this into another function?

If we do change `igraph_modularity` itself, some explanation in the documentation will be necessary (e.g. which included community detection functions does this apply to?), as well as an update of the formula.

vtraag:
> I'm not sure what's the next steps to coordinate with the rigraph repo. Should I wait until this is merged here (or in the cran branch?) before creating a pull request in the rigraph repo?

Because this PR breaks API, as @szhorvat  already noted, we will first have to wait until the next minor version is released (0.9.0). Once that is released, we have to update the C core of `rigraph`, after which we can open a PR to include the updated functionality. We will no longer work with the `cran` branch, `rigraph` should start using the `master` branch again (soon).

szhorvat:
> One other nitpick thing is the order of the arguments. I think it would be good if we default to having `const` parameter first, if possible. That way, any parameters that are used as "output" are at the end of the function. This is not always possible because some arguments function both as input and output, but it might be useful as a principle. That means we should move the `resolution_parameter` before the other arguments. We haven't yet discussed any standard for that @szhorvat , @ntamas, but what do you think?

I think this discussion can be left for later (after this is merged). It's okay to change the API again before the 0.9 release, and this discussion would involve other functions too.

----

> Perhaps we should add a test for modularity with a resolution parameter? Currently only the default is tested, perhaps it would also be good to test with another resolution parameter.

I think this is important. @jmonlong If you have time, and know your way around igraph's testing framework, it would be really great if you could add this.

----

>     1. Probably better to use `resolution_parameter`, to be consistent with the current `leiden` implementation.

Indeed, the best name is an explicitly descriptive name. Most users won't already be familiar with the method and come looking for it. Instead, they'll be browsing the library to see what's available for community detection. For them, the name `gamma` is absolutely uninformative.

If too long, consider `resolution`?

----

@vtraag If you have time, could you fix up the docs a little bit and push to this PR directly, before merging?

In particular, in the docs of `igraph_modularity`:

 - the generalized modularity needs a bit more explanation and the formula needs updating
 - it would be useful to have a reference for this generalized definition
 - explicitly point out that `gamma = 1` needs to be used to get the classical definition
 - link to other igraph functions which use this generalized definition (i.e. `igraph_community_multilevel` at this point)

In the docs of `igraph_community_multilevel`:

 - explicitly point out that the `modularity` output parameter uses the generalized definition (otherwise some will certainly be confused about what the value returned by this function does not match the well-known classic definition).
 - include a reference to the paper where this was originally described

@vtraag If you don't have time now, please help with the paper references and I'll push a draft. Then you can fix it up as the community detection expert.

jmonlong:
No problem, I can do my part today. Even the parameter ordering (`const` first) makes a lot of sense, I added the new thing at the end without thinking.

Just a clarification about having multilevel test with a different resolution. Do you mean adding a small test to prevent future regression or a test to ensure the code works correctly? 

The former is easy, I can add a test checking for the output of the current membership assignment of the multilevel code. The latter is harder because I don't really have an example with a known truth. As a sanity check, I just tested that increasing the resolution increased the number of communities and that they look like communities in the graph.

> > Perhaps we should add a test for modularity with a resolution parameter? Currently only the default is tested, perhaps it would also be good to test with another resolution parameter.
> 
> I think this is important. @jmonlong If you have time, and know your way around igraph's testing framework, it would be really great if you could add this. If you don't know how to add tests, a C program, along with a text file containing its expected output, is fine. We'll take care of the rest.

src/community.c
```diff
@@ -3013,7 +3010,7 @@ static int igraph_i_community_multilevel_step(
         IGRAPH_ERROR("weights must be positive", IGRAPH_EINVAL);
     }
     if (gamma < 0.0) {
```
vtraag, line 3012: I think this should still be changed to `resolution_parameter`, right?

vtraag:
Thanks for the fast response @jmonlong!

Regarding a test, I thought you could do something as follows:

image

This comes from [this article](http://arxiv.org/abs/1103.5569).

szhorvat:
I made some more edits to the docs.

src/community.c
jmonlong: This function was only used internally by the main multilevel function (AFAIK) so I had left the *gamma*. But we could definitely change it to *resolution_parameter* too.

jmonlong:
The good news: I could reproduce the figure from the paper you linked using Rigraph. That's reassuring!

krings-blondel-example-resolution

The bad news: I'm still not sure how I should add a test to igraph. I didn't find a value for the membership/modularity in the paper that I could use as expected value for the test. I think the algorithm also depends on the edges/nodes order in the graph definition anyway, so the same graph would get different results depending on how it was constructed. 

Should I just add whatever output we get now as the expected value for the test? Just as safeguard against future regressions?

vtraag:
Great, good to see it all works as expected!

Adding a test is relatively straightforward. You can simply add it to the existing file [`examples/simple/igraph_community_multilevel.c`](
https://github.com/igraph/igraph/blob/develop/examples/simple/igraph_community_multilevel.c). If you compile and run that C program, you can then store the output in [`examples/simple/igraph_community_multilevel.out`](https://github.com/igraph/igraph/blob/develop/examples/simple/igraph_community_multilevel.out). When running `make check` the testing framework automatically compiles the C program and compares it to the expected output.

If you can't get it to work, no problem, if you can post the code here, I will make sure to add it as a proper test.

ntamas:
Regarding `resolution_parameter` vs `resolution`: I would prefer the latter; `_parameter` seems to be redundant -- in a strict sense, all function arguments are parameters one way or another. We can call it "resolution parameter" in the docs, though, if we want to improve searchability.

szhorvat:
@ntamas Do you have any opinion on the argument reshuffling that happened here?

Personally, I am not sure I am comfortable with the const-first approach applied *in general*, but my question refers only to this single instance. The general discussion should be left for another day.

jmonlong:
I updated the test with the output for a higher resolution and renamed the parameter to *resolution*. Let me know if you want the argument order back like it was before (+ the new parameter at the end).

vtraag:
I'm happy with changing the name to `resolution` instead of `resolution_parameter`. The current order of arguments makes sense to me, we can indeed leave the more general discussion for another time. The test with a higher resolution parameter is also fine with me.

vtraag:
If we all agree that `resolution` is the proper name, and that the arguments are in the right order, then the only thing left to do is to make sure that `resolution` is used everywhere, also for the internal functions.

szhorvat:
> the only thing left to do is to make sure that `resolution` is used everywhere

Done.

Also changed some error messages and added a missing check.

ntamas:
@szhorvat 

> Do you have any opinion on the argument reshuffling that happened here?

What we have here right is perfectly fine for me. The _const-first_ approach is a good rule of thumb, but we can deviate from it if it feels unnatural in some cases. The only rule that I want to stick to at all costs is that the "this" argument (i.e. in most cases the graph that the function operates on) comes first.

szhorvat:
Can be merged as far as I'm concerned.

PR #1391, opened 2020-05-27
topological_sorting: raise error if the graph is not acyclic

### LINKED ISSUE #1380: Confusing results of topological_sorting
I suppose the topological_sorting function will return a list whose length equals the number of vertices in the graph. However, when I run the following codes I get unexpected results. The behavior is rather confusing. I know the topological order is not unique in general, but I can not find out why this function returns these results. Specifically, maybe the list obtained by mode='IN' should be the reverse of the list obtained by mode='OUT'?

```
>>> g=Graph.Erdos_Renyi(nvt, 0.1, directed=True, loops=False)
>>> print(g)
IGRAPH D--- 10 10 --
+ edges:
6->1 5->2 5->3 2->5 4->5 3->6 9->6 9->7 0->8 6->8
>>> g.topological_sorting(mode='IN')
[1, 7, 8, 0, 6, 3, 9]
>>> g.topological_sorting(mode='OUT')
[0, 4, 9, 7]

```

szhorvat:
`topological_sorting` would only show a warning if the graph is not acyclic. Now it raises an error. Fixes #1380.

Additional changes:

 - Error message formatting (see https://igraph.discourse.group/t/standardize-error-and-warning-styles/249)
 - Move tests to examples/tests
 - Retain examples/simple/igraph_topological_sorting.c in a simplified form for the docs
 - Edit the documentation

This goes to develop because it changes the behaviour significantly.


ntamas:
Many thanks!

PR #1392, opened 2020-05-27
Correct cutoff for betweenness.

### LINKED ISSUE #1316: Betweenness with cutoff - wrong output
igraph version 0.7.1

Code:

```
import igraph as ig
g = ig.Graph()
g.add_vertices(4)
g.add_edge(0, 1, weight = 1)
g.add_edge(0, 2, weight = 2)
g.add_edge(1, 3, weight = 1)
g.add_edge(2, 3, weight = 2)

print(g.betweenness(weights='weight', directed=False, cutoff=2))
```

Output:
`[0.25, 1.0, 0.0, 0.25]`

Expected output:
`[0.0, 1.0, 0.0, 0.0]`
because only the length of the shortest path between 0 and 2 (going through node 1) is below the cutoff.

vtraag:
This should fix #1316, where incorrect results for betweenness were reported when a cutoff was being used.

The problem was twofold:
1. Some nodes were incorrectly being seen as too distant for some source node because the distance was not reset if that node was too distant for another source node.
2. Nodes were still being added to the predecessor list, even when they were too distant.

The same problem may also play in other parts:
- `igraph_i_edge_betweenness_estimate_weighted`
- `igraph_i_closeness_estimate_weighted`
- `igraph_closeness_estimate`

These should still be corrected.

I would like to suggest to rewrite significant parts of the code. First of all, it would be a lot clearer if we indeed would use actual distances, instead of using distances + 1. Secondly, some variable names are not as clear as they could be. Finally, some parts seem to use a sort of legacy code, where simply direct arrays are used instead of `igraph_vector_t`. I think it would be good to simply use the types that are supplied by `igraph` as much as possible, instead of arrays. If you agree, I will do so in this PR.

@szhorvat, could you first double check if things are working correctly now? I have a test case, we could also add it to the tests, let me know what you prefer.

### DIFF
--- a/src/centrality.c
+++ b/src/centrality.c
@@ -1726,7 +1726,7 @@ static int igraph_i_betweenness_estimate_weighted(
                     VECTOR(dist)[to] = altdist;
                     IGRAPH_CHECK(igraph_2wheap_modify(&Q, to, -altdist));
                 } else if (cmp_result == 0 &&
-                    (altdist <= cutoff + 1.0 || cutoff <= 0)) {
+                    (altdist <= cutoff + 1.0 || cutoff < 0)) {
                     /* Only add if the node is not more distant than the cutoff */
                     igraph_vector_int_t *v = igraph_adjlist_get(&fathers, to);
                     igraph_vector_int_push_back(v, minnei);
@@ -1826,8 +1826,8 @@ static void igraph_i_destroy_biguints(igraph_biguint_t *p) {
  * \param directed Logical, if true directed paths will be considered
  *        for directed graphs. It is ignored for undirected graphs.
  * \param cutoff The maximal length of paths that will be considered.
- *        If zero or negative, the exact betweenness will be calculated
- *        (no upper limit on path lengths).
+ *        If negative, the exact betweenness will be calculated, and
+ *        there will be no upper limit on path lengths.
  * \param weights An optional vector containing edge weights for
  *        calculating weighted betweenness. Supply a null pointer here
  *        for unweighted betweenness.
@@ -1969,7 +1969,7 @@ int igraph_betweenness_estimate(const igraph_t *graph, igraph_vector_t *res,
             long int actnode = (long int) igraph_dqueue_pop(&q);
 
             /* Ignore vertices that are more distant than the cutoff */
-            if (cutoff > 0 && distance[actnode] > cutoff + 1) {
+            if (cutoff >= 0 && distance[actnode] > cutoff + 1) {
                 /* Reset variables if node is too distant */
                 distance[actnode] = 0;
                 if (nobigint) {
--- a/src/centrality.c
+++ b/src/centrality.c
@@ -1679,10 +1679,17 @@ static int igraph_i_betweenness_estimate_weighted(
             igraph_vector_int_t *neis;
             long int nlen;
 
-            igraph_stack_push(&S, minnei);
-            if (cutoff > 0 && VECTOR(dist)[minnei] >= cutoff + 1.0) {
+            /* Ignore vertices that are more distant than the cutoff */
+            if (cutoff >= 0 && mindist > cutoff + 1.0) {
+                /* Reset variables if node is too distant */
+                VECTOR(tmpscore)[minnei] = 0;
+                VECTOR(dist)[minnei] = 0;
+                VECTOR(nrgeo)[minnei] = 0;
+                igraph_vector_int_clear(igraph_adjlist_get(&fathers, minnei));
                 continue;
             }
+
+            igraph_stack_push(&S, minnei);
 
             /* Now check all neighbors of 'minnei' for a shorter path */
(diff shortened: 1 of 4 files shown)


iosonofabio:
I started working on this but there seemed to be a strong opinion that each function deserves its own PR.

My opinion is we should fix the whole file in a single PR. If you do it it's great, or I can do it too.

vtraag:
Ah, sorry, I though you were still looking for what the error was! But let me indeed follow up on the other functions. I think it makes sense to treat all functions in this PR, and we can then do the larger overhaul in another PR.

iosonofabio:
That'd be fantastic. I had a look at your changes and it looks good except maybe some corner case when the cutoff is set to exactly 0?

vtraag:
Well, yes, I made a slight modification to the cutoff, which makes more sense I think. As the documentation says, the cutoff  indicates

> The maximal length of paths that will be considered.

That means that only paths that are strictly longer than the cutoff should not be considered. Paths that are equal to the cutoff should then be included still.

For a cutoff of zero, this implies that no-paths should be explored at all. In other words, only for strictly negative cutoff the cutoff will not be used. I now realize that the explanation for the zero-cutoff should still be corrected in the documentation.

src/centrality.c
```diff
@@ -1971,7 +1991,9 @@ int igraph_betweenness_estimate(const igraph_t *graph, igraph_vector_t *res,
                     distance[neighbor] = distance[actnode] + 1;
                     IGRAPH_CHECK(igraph_dqueue_push(&q, neighbor));
                 }
-                if (distance[neighbor] == distance[actnode] + 1) {
+                if (distance[neighbor] == distance[actnode] + 1 &&
+                    (distance[neighbor] <= cutoff + 1 || cutoff < 0)) {
```
iosonofabio, line 1995: Hi Vincent, sorry this is what I meant with the zero cutoff. Most places in the code divide into > 0 and <=0, but here it's different, just double checking this is what you meant.

vtraag: Ah, yes, good catch, thanks!

vtraag:
Note that the test fails because I have not corrected them yet. @szhorvat, if you agree that this solution works, I will update those tests, and also correct the other functions.

szhorvat:
I can confirm that the examples I looked at in #1316 do work as I would expect them to.

https://github.com/igraph/igraph/issues/1316#issuecomment-581790460

I'll do more checks later.

szhorvat:
I wrote a dead slow, but exact (rational numbers) and easy to verify implementation of betweenness with cutoff in Mathematica, and tested this for several random graphs of different densities. It's all good! Go ahead and update the tests @vtraag.

szhorvat:
Maybe it's worth noting that the precision of the results is not that great. The difference between the exact results and the ones returned by igraph is tends to be on the order of 10^-8 to 10^-9 for graphs on up to 100 vertices.

I should have noted that the verification program I wrote only works for unweighted graphs. I did not test the weighted case extensively.

szhorvat:
Thanks so much @vtraag and @iosonofabio! I was really bothered by this bug.

vtraag:
I was now just working on improving things for other related functions. It seems that for `igraph_closeness` the same problem does not hold. 

However, I did want to change the cutoff to be consistent with the implementation as I now made it for `igraph_betweenness`. I decided that only negative cutoffs are used to indicate that no cutoff will be used. I just came across issue #1050, in which @ntamas states to prefer not to use `cutoff = 0` to calculate results. To me, the documentation 

> The maximal length of paths that will be considered.

indicates that lengths that are strictly larger than the maximal length should *not* be considered. Hence, paths of length 4 will be considered if the `cutoff = 4`, but not if `cutoff = 3.9`. That is also why I used `cutoff = 0` to indicate that essentially *no* paths will be considered, because it is consistent with this interpretation. @ntamas, @szhorvat , @iosonofabio , can we all agree that a `cutoff < 0` will be used to indicate that no cutoff will be used, and that we use a strict inequality, i.e. paths for which `distance > cutoff` will *not* be considered.

vtraag:
On a different, yet slightly related note, I was checking `igraph_edge_betweenness`. The problem with the cutoff does appear for the weighted variant. Incidentally, some things work better for the unweighted variant, because some of the vectors are cleared repeatedly.

https://github.com/igraph/igraph/blob/87f4142d3fe62e25b2abed156bba2ad076ab7d8f/src/centrality.c#L2421-L2424

@ntamas, I assume this is just legacy code, and that I can change it so as to work similar to the current implementation of `igraph_betweenness_estimate`?

ntamas:
> can we all agree that a `cutoff < 0` will be used to indicate that no cutoff will be used

Yes, it seems logical to me.

ntamas:
> I assume this is just legacy code, and that I can change it so as to work similar to the current implementation of `igraph_betweenness_estimate`?

Yes, feel free to change this. Thanks!

PR #1346, opened 2020-03-02
Basic draft for eulerian paths/circuits

YasirKusay:
I started to implement eulerian cycles/paths. I made:

- .c file in src with a few functions, will refine later on
- .h in include, containing only one function for now
- a .c file in examples/tests
- a .at file inside tests

I am not sure as how I can include my files into the compiling and tests steps, can you please help me?
After that, I will refine the code that I have.

PR #1400, opened 2020-06-08
test_utilities: add support for checking that the FINALLY stack is empty

szhorvat:
In response to https://github.com/igraph/igraph/pull/1346#discussion_r436287686


ntamas:
You beat me to it -- I wanted to do the same thing this afternoon :)

PR #1402, opened 2020-06-08
Corrected edge attribute description in the documentation

vtraag:
Accidentally, some edge attribute were described as being set for all *vertices*, instead of all *edges*.


ntamas:
Thanks a lot!

Issue #1358, opened 2020-03-07
Implement directed modularity


szhorvat:
`igraph_modularity` should support directed graphs.

@vtraag I need your input on whether there are multiple variations on this concept.

vtraag:
I am only familiar with the definition provided by [Leicht and Newman (2008)](https://doi.org/10.1103/PhysRevLett.100.118703), Eq. (3), and I haven't come across any other definition. This is also the definition that makes most sense, which should be straightforward to implement. Similar adjustments need to be made for the weighted case, but these are again quite straightforward. I'll put it on my list.

PR #1403, opened 2020-06-09
Support for directed graphs when calculating modularity

vtraag:
I implemented the calculation of modularity for directed graphs, as requested in #1358. Note that many functions currently do not support directed graphs, and hence, for those functions `igraph_modularity` now returns the undirected version of modularity. One function does support directed graphs (`igraph_community_edge_betweenness`) and for that it does return the directed version of modularity (if so indicated).

I've also provided some tests to check whether the directed definition provides the same modularity as the undirected definition when simply expanding an undirected graph to a directed graph with both directions present. Additionally, I check whether the aggregation also works correctly and provides the same modularity.

Note that for this test it seems that the stack is not empty. I am still checking where the problem is (not in `igraph_modularity`). But great improvement to catch these things early on @szhorvat and @ntamas ! Until that is sorted out, I mark this as a draft PR, but feel free to already comment if you want.

### DIFF
--- a/examples/tests/igraph_modularity.c
+++ b/examples/tests/igraph_modularity.c
@@ -1,0 +1,113 @@
+/* -*- mode: C -*-  */
+/*
+   IGraph library.
+   Copyright (C) 2006-2012  Gabor Csardi <csardi.gabor@gmail.com>
+   334 Harvard street, Cambridge, MA 02139 USA
+
+   This program is free software; you can redistribute it and/or modify
+   it under the terms of the GNU General Public License as published by
+   the Free Software Foundation; either version 2 of the License, or
+   (at your option) any later version.
+
+   This program is distributed in the hope that it will be useful,
+   but WITHOUT ANY WARRANTY; without even the implied warranty of
+   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+   GNU General Public License for more details.
+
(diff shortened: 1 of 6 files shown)


vtraag:
The problem of the incorrect FINAL stack size is resolved in PR #1405. I have pushed that solution separately, because it should also be merged to `master`, while this PR should only be pushed to `develop`.

szhorvat:
#1405 is also merged in `develop` now, you can pull it into this PR

vtraag:
> #1405 is also merged in `develop` now, you can pull it into this PR

Thanks! I force-pushed commits that are rebased against the updated `develop` branch.

szhorvat:
Why was a force-push necessary? Would it be sufficient to just pull into this PR branch?

vtraag:
> Why was a force-push necessary? Would it be sufficient to just pull into this PR branch?

I just wanted to keep the commits neat, so I rebased, that's all.

ntamas:
Thanks a lot!

szhorvat:
@vtraag Do we want to update `modularity_matrix` as well? While the analogous generalization is trivial, I have not used this concept in my work. Any changes should be made with real applications in mind.

vtraag:
@szhorvat, yes I was thinking it would be most reasonable to do that as well. I left that for another occasion though. We could open a separate issue for that to keep track.

PR #1404, opened 2020-06-09
Verify FINAL stack size when debugging.

vtraag:
Following up on the excellent idea in PR #1400, I thought it would be good to test whether stack sizes are the same when entering and exiting functions. This is most readily done in the `IGRAPH_CHECK` macro, because these are already defined in many places. Obviously, missing `IGRAPH_CHECK` will continue to be a problem.

Note that this degrades performance significantly, so this should *never* be used in actual compilation, and only when compiled in debug mode (i.e. `NDEBUG` should be defined to prevent using this), but it does make it a lot easier to track down bugs in the error handling. This PR reveals a number of problems across `igraph`, which we should check out in more detail. I think it would be best to solve those problems in separate PRs to keep everything neat.

Curious to hear what you all think of this suggestion!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,14 +43,14 @@ jobs:
       os: linux
       install:
         - ./bootstrap.sh
-        - ./configure --with-external-glpk
+        - ./configure --enable-verify-finally --with-external-glpk
         - make
     - name: "Linux external deps"
       os: linux
       dist: bionic
       install:
         - ./bootstrap.sh
-        - ./configure --with-external-glpk --with-external-blas --with-external-lapack --with-external-arpack
+        - ./configure --enable-verify-finally --with-external-glpk --with-external-blas --with-external-lapack --with-external-arpack
         - make        
     - name: "Linux x87"
       os: linux
@@ -58,13 +58,13 @@ jobs:
         - CFLAGS="-mfpmath=387" CXXFLAGS="-mfpmath=387"
       install:
         - ./bootstrap.sh
-        - ./configure --with-external-glpk
+        - ./configure --enable-verify-finally --with-external-glpk
         - make
     - name: "macOS"
       os: osx
       install:
         - ./bootstrap.sh
-        - ./configure --with-external-glpk --enable-asan
+        - ./configure --enable-verify-finally --enable-asan --with-external-glpk 
         - make
     - name: "Documentation"
       language: shell
--- a/configure.ac
+++ b/configure.ac
@@ -131,6 +131,11 @@ AC_ARG_ENABLE(asan,
               AC_HELP_STRING([--enable-asan], [Enable Clang address sanitizer]),
               [use_asan=$enableval], [use_asan=no])
 
+verify_finally=no
+AC_ARG_ENABLE(verify-finally,
+              AC_HELP_STRING([--enable-verify-finally], [Enable verifying the ``finally'' stack size]),
+              [verify_finally=$enableval], [verify_finally=no])
+
 debug=no
 AC_ARG_ENABLE(debug,
               AC_HELP_STRING([--enable-debug], [Enable debug build]),
@@ -384,6 +389,14 @@ if test "$use_asan" != "yes" -a "$use_gprof" != "yes" -a "$debug" != "yes"; then
   CXXFLAGS="${CXXFLAGS} -O3"
 fi
 
+if test "$verify_finally" = "yes"; then
+  CPPFLAGS="${CPPFLAGS} -DIGRAPH_VERIFY_FINALLY_STACK=1"
+fi
+
+if test "$verify_finally" = "no"; then
+  CPPFLAGS="${CPPFLAGS} -DIGRAPH_VERIFY_FINALLY_STACK=0"
+fi
+
 AC_CONFIG_FILES([Makefile src/Makefile igraph.pc igraph_Info.plist doc/Makefile include/igraph_version.h include/igraph_threading.h])
 AC_OUTPUT
 
@@ -405,6 +418,7 @@ if test "$glpk_support" != "no"; then
 fi
 AC_MSG_RESULT([  Debug build            -- $debug])
 AC_MSG_RESULT([  Clang AddressSanitizer -- $use_asan])
+AC_MSG_RESULT([  Verify finally stack   -- $verify_finally])
 AC_MSG_RESULT([  Profiling              -- $use_gprof])
 AC_MSG_RESULT([  Link time optimization -- $use_lto])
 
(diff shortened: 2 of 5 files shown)


szhorvat:
One short comment:

I would rather not use `NDEBUG` for this as I am not convinced that it is a good idea to compile release binaries with `NDEBUG` defined (and therefore all `assert`s disabled).

I would suggest introducing a separate flag, i.e. something ike

```c
#if (IGRAPH_VERIFY_FINALLY == 1)
...
#endif
```

This can then be toggled explicitly, and can also be coupled to another flag, should we decide to have a global debug flag (NDEBUG or something else) later. `#if` is better than `#ifdef` here because one can also explicitly toggle it off on the compiler command line as `gcc -DIGRAPH_VERIFY_CHECK=0`, should it be necessary.

szhorvat:
Great idea BTW! 👍 

First let's concentrate on fixing everything detected by this, and then we can polish the details.

szhorvat:
Can you please pull the latest master branch into this one as well so we can see how many of the tests failures were fixed by #1405?

szhorvat:
This broke the doc build. I think it wants the doc comment to come before the `#define` line (it won't tolerate an `#ifdef` intruding between the doc comment and the `#define`).

szhorvat:
Here's the list of failing tests. When you start working on any of them, edit the list below and add your name to it. This way we can avoid accidentally duplicating work.

 - [ ] 140: LAD subgraph isomorphism algorithm              FAILED (topology.at:47)
 - [ ] 182: All minimal s-t cuts of a graph (igraph_all_st_mincuts):  FAILED (flow.at:58)
 - [ ] 235: Minimum size separators (igraph_minimum_size_separators):  FAILED (separators.at:46)
 - [ ] 236: Cohesive blocking (igraph_cohesive_blocks):     FAILED (separators.at:52)

szhorvat:
@vtraag Let's print the difference between the initial and final stack sizes. It will aid debugging.

szhorvat:
The LAD thing is not technically a bug. There is not expected to be a match in the stack size at the function call boundary.

It concerns the `igraph_i_lad_createGraph` function. There is no corresponding `destroyGraph` function which could be added to FINALLY. Therefore, each member of LAD's Tgraph struct is freed separately.

We can fix it up so that stack sizes are enforced to match at function call boundaries, but I can not do it today. It will require care. This is a port of the original LAD code to use igraph's memory management, so it's easy to break things.

szhorvat:
@vtraag Can I push fixes to the issues above directly to this PR branch? It'll make it simpler to verify that they're indeed fixed.

vtraag:
@szhorvat, yes, feel free to push to this PR! And also feel free to adjust the definition as you see fit.

ntamas:
Great idea guys! I'll also add this to the CMake build branch once it has made its way to `master`.

ntamas:
Is it OK if I convert this PR to draft until all the test failures are resolved?

szhorvat:
@ntamas please take a look at the last commit I made, and check that it's fine. As I remember you made the LAD port originally (sorry if I'm wrong).

ntamas:
It wasn't me, but it seems okay to me, thanks!

ntamas:
#1406 is now in `master`; rebasing this against `master` should allow us to check whether #1406 fixed anything.

ntamas:
The remaining failures can all be attributed to `st-cuts.c`; I'll try to fix this.

ntamas:
@vtraag seems like I can't push to this branch; you need to merge this to fix the failing tests:

```patch
diff --git a/src/st-cuts.c b/src/st-cuts.c
index 2f055ea1..3a4d4cc7 100644
--- a/src/st-cuts.c
+++ b/src/st-cuts.c
@@ -1170,7 +1170,7 @@ static int igraph_i_all_st_mincuts_minimal(const igraph_t *Sbar,

     igraph_vector_destroy(&indeg);
     igraph_vector_destroy(&neis);
-    IGRAPH_FINALLY_CLEAN(3);
+    IGRAPH_FINALLY_CLEAN(2);

     return 0;
 }
```

vtraag:
I was just pulling it in, only to be looking at your commit :)

ntamas:
Regarding the `NDEBUG` flag: I would also prefer to have a separate compile-time switch for this; something we can also configure from `./configure`, similarly to `--enable-lto` (although we can simply connect it to `--enable-debug`, that's also okay for me).

vtraag:
I'm fine with either `--enable-debug` or with a new flag.

szhorvat:
I integrated it into the configure script as a separate option: `--enable-verify-finally`. This will add `-DIGRAPH_VERIFY_FINALLY_STACK=0` to `CPPFLAGS`.

First I added the define to `config.h`, but I realized that `config.h` is not installed. It is only used during compilation. The alternative is to have a new header (`something.h.in`) that `./configure` can modify. That seemed too complicated.

@vtraag  I'll leave squashing and reordering commits to you. I did not want to do a force push.

Take this as a proposal. We can still modify it.

szhorvat:
I also added it to Travis, but *not* to AppVeyor. I'll leave that to @ntamas. It's not clear to me if the version uploaded to PyPI is from AppVeyor—if yes, there should be no performance-ruining debug options enabled there.

I need to leave before the CI is done, hope I didn't break anything.

szhorvat:
I had to return because of a thunderstorm. I am happy to report that after running the Mathematica test suite (with some modifications so that direct calls from the Mma interface are also checked), everything is fine. No other mismatches were detected.

ntamas:
> The alternative is to have a new header (something.h.in) that `./configure` can modify. That seemed too complicated.

This is what we are doing with `igraph_threading.h` (which is generated from `igraph_threading.h.in` at configuration time. Maybe we could rename that header (as it only contains one `#define` regarding thread safety) and use it?

As for the Appveyor CI, tbh I have no idea what we are doing there with different Python versions as it has nothing to do with the Python interface - that one is built and published from `python-igraph`. I think it's only used to check whether our Python-based build scripts work fine on Windows with varying versions of Pythons, so it is safe to update `appveyor.yml` as well. (I'll do it).

ntamas:
Also ported to the `feature/cmake-build` branch in e91296d8

vtraag:
> I have no idea what we are doing there with different Python versions

One reason to keep it to ensure that `igraph` is properly build against all relevant MSVC versions (e.g. C89 style declarations). Indeed another one is that we need to ensure that `make msvc` works properly in all Python versions. Finally,`make check` is also checked in MSYS, but this is done only once.

szhorvat:
@vtraag I'm quite happy about this addition. It was a really good idea 👍

vtraag:
Thanks @szhorvat ! Perhaps there are similar improvements possible that might make it easier to identify potential bugs.

PR #1406, opened 2020-06-09
Check that the arguments of IGRAPH_FINALLY have the correct type

szhorvat:
I tend to make the mistake that I swap the arguments of `IGRAPH_FINALLY` and I write `IGRAPH_FINALLY(ptr, fun)` instead of the correct `IGRAPH_FINALLY(fun, ptr)`.

Normally, my IDE (or the compiler) would warn when something like this is wrong. But `IGRAPH_FINALLY` is casting its arguments, which masks any errors.

This PR attempts to allow `IGRAPH_FINALLY` to check its args, like this:

```c
#define IGRAPH_FINALLY(func,ptr) \
    { \
        if (0) { func(ptr); } \
        IGRAPH_FINALLY_REAL((igraph_finally_func_t*)(func), (ptr)); \
    }
```

The `if (0)` should be optimized away by the compiler, so performance is not a concern (unless I am missing something). But this line will cause the compiler to complain if the types don't match.

As expected, this keeps revealing errors. I did not fix them yet. Even if we do not decide to keep this modification to `IGRAPH_FINALLY`, we should fix these errors. I invite everyone to push to this PR and contribute. Note that we must pay attention to compiler warnings as well as to errors. Some problems show up as warnings only.

If you plan to work on a particular issue, post a note so we don't end up working on the same thing in parallel.

### DIFF
--- a/include/igraph_error.h
+++ b/include/igraph_error.h
@@ -564,7 +564,10 @@ DECLDIR int IGRAPH_FINALLY_STACK_SIZE(void);
  */
 
 #define IGRAPH_FINALLY(func,ptr) \
-    IGRAPH_FINALLY_REAL((igraph_finally_func_t*)(func), (ptr))
+    { \
+        if (0) { func(ptr); } \
+        IGRAPH_FINALLY_REAL((igraph_finally_func_t*)(func), (ptr)); \
+    }
 
 #if !defined(GCC_VERSION_MAJOR) && defined(__GNUC__)
     #define GCC_VERSION_MAJOR  __GNUC__
--- a/include/igraph_error.h
+++ b/include/igraph_error.h
@@ -565,6 +565,9 @@ DECLDIR int IGRAPH_FINALLY_STACK_SIZE(void);
 
 #define IGRAPH_FINALLY(func,ptr) \
     { \
+        /* the following branch makes the compiler check the compatibility of \
+         * func and ptr to detect cases when we are accidentally invoking an \
+         * incorrect destructor function with the pointer */ \
         if (0) { func(ptr); } \
         IGRAPH_FINALLY_REAL((igraph_finally_func_t*)(func), (ptr)); \
     }
--- a/examples/simple/tls1.c
+++ b/examples/simple/tls1.c
@@ -26,7 +26,7 @@
 #include <unistd.h>
 
 void *thread_function(void *arg) {
-    IGRAPH_FINALLY(0, 0);
+    IGRAPH_FINALLY(igraph_free, NULL);
     return 0;
 }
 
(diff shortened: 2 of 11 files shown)


ntamas:
Nice trick!

Is it OK if I convert this PR to draft until all the test failures are resolved?

szhorvat:
Converted to draft.

szhorvat:
I'll collect a list of warnings/issues in this post. Feel free to edit.

----

```
foreign.c:1158:48: warning: incompatible pointer types passing 'igraph_vector_ptr_t *(*)[3]' to parameter of type 'igraph_vector_ptr_t **'
      (aka 'struct s_vector_ptr **') [-Wincompatible-pointer-types]
    IGRAPH_FINALLY(igraph_i_gml_destroy_attrs, &attrs);
                                               ^~~~~~
../include/igraph_error.h:568:23: note: expanded from macro 'IGRAPH_FINALLY'
        if (0) { func(ptr); } \
                      ^~~
foreign.c:980:62: note: passing argument to parameter 'ptr' here
static void igraph_i_gml_destroy_attrs(igraph_vector_ptr_t **ptr) {
```

szhorvat:
I won't be able to do more igraph stuff today, so everyone feel free to fix some of the above.

ntamas:
I'll take up a few of these now.

ntamas:
Okay, fixed all of these, now I only need to figure out how to push to this PR :)

szhorvat:
Clone `https://github.com/szhorvat/igraph.git`, check out `fixes/finally-arg-check`, then you can just push.

ntamas:
Turns out there's an easier way: `git push git@github.com:szhorvat/igraph fixes/finally-arg-check:fixes/finally-arg-check`, no need for a separate checkout.

(I was using `gh pr checkout 1406` on the CLI to get a copy of this PR in my own repo, I was sort-of-hoping that there is a `gh pr push` shortcut but it looks like there isn't).

vtraag:
I actually have your repositories as remotes as well so that I can easily push to your PRs as well. But indeed, a `gh pr push` should be really the easiest!

szhorvat:
Please keep sharing these tips so we can all learn. I'm not a programmer, just a scientist 😁

szhorvat:
Looks like ready for review then. Can we merge this now?

When merged, it should also be pulled into #1404 to aid with all that work on `FINALLY`s

ntamas:
Let's wait for the CI checks; we can merge if they look all right.

include/igraph_error.h
```diff
@@ -564,7 +564,10 @@ DECLDIR int IGRAPH_FINALLY_STACK_SIZE(void);
  */
 
 #define IGRAPH_FINALLY(func,ptr) \
-    IGRAPH_FINALLY_REAL((igraph_finally_func_t*)(func), (ptr))
+    { \
+        if (0) { func(ptr); } \
```
ntamas, line 568: Can we add a comment here explaining what this hack is for (in case we won't remember 20 years later)?

ntamas: Sure, it's done now :)

ntamas:
Travis builds passed; I think this is good to go.

PR #1408, opened 2020-06-10
Separate tips on writing igraph code onto wiki

iosonofabio:


vtraag:
Good idea to concentrate all support for development in the Wiki! I added the new page also to the sidebar.

iosonofabio:
I started this: https://github.com/igraph/igraph/wiki/Contributing-your-first-function-to-igraph

ntamas:
Thanks @iosonofabio, that looks great!

I'll merge this PR soon. A few comments to the wiki page that you have started:

* How about using a `feat/` or `feature/` prefix for branches that contain new features and `fix/` or `bugfix/` for branches that contain bugfixes?

* Regarding the copyright at the top of new files: we did not really bother updating the copyright notices in old files so in almost all places you see fairly old year numbers in the files (basically we added the copyright notice once when the file was added, and did not bother updating it after that, so the end year of the copyright notice reflects the year the file was added in). Assuming that the copyright notice refers to the contents of the _file_ and not the entire _library_, I think it's enough to put `The igraph development team` in the header for new files, no need to include Gábor separately.

* I also made a few changes to the example here and there; for instance, when an igraph function takes a vector or any other complex data structure as an output argument, it must always be initialized by the _caller_ (except if the very purpose of the function is to create something anew). I also added missing `IGRAPH_CHECK()` calls and mentioned that internal functions are to be declared `static` unless needed from another compilation unit.

* Just a note to myself: if (or, rather, _when_) we switch to CMake, we need to update these guidelines as well because adding a new test would look a bit different in CMake.

PR #1409, opened 2020-06-10
Corrected cutoff for edge_betweenness

vtraag:
Similar to #1316 and as implemented in #1392, this changes the cutoff as used for `igraph_edge_beweenness`.

There are two changes in this PR. For `igraph_edge_betweenness` all vectors were always being cleared (unlike for `igraph_betweenness`), making the cutoff not problematic. This is changed, so that only the actually used nodes are being reset, similar to the implementation in `igraph_betweenness`. The resetting of these nodes is now correctly implemented, so that the cutoff works properly.

Secondly, the cutoff that is now used is consistent with the implementation in #1392, and considers all paths for which `distance <= cutoff`, as opposed to `distance < cutoff`, which it was before. Because of this, `cutoff = 0` is also used, and will return all zero results (this reverts #1050). When `cutff < 0`, no cutoff is used at all, and paths of all lengths are considered.


szhorvat:
@vtraag The docs should be updated too.

szhorvat:
I'm sorry, they're already updated. I was looking at the wrong version of the source file.

PR #1410, opened 2020-06-10
Corrected cutoff for closenes

vtraag:
Similar to #1316 and as implemented in #1392, this changes the cutoff as used for `igraph_closeness`.

Unlike for `igraph_betweenness`, the cutoff was not problematic for closeness.

The cutoff that is now used is consistent with the implementation in #1392, and considers all paths for which `distance <= cutoff`, as opposed to `distance < cutoff`, which it was before. Because of this, `cutoff = 0` is also used, and will return all zero results (this reverts #1050). When `cutff < 0`, no cutoff is used at all, and paths of all lengths are considered.

### DIFF
--- a/examples/tests/igraph_closeness.c
+++ b/examples/tests/igraph_closeness.c
@@ -18,7 +18,7 @@ void simple_test_case_no_weights_undirected() {
 
     igraph_closeness_estimate(&g, &vector_actual_results /*store results here*/, 
                              /*calculating for all vectors in the graph*/ igraph_vss_all(), 
-                             IGRAPH_ALL  /*graph is "undirected"*/, /*cutoff*/ 0 /*calculating exact centrality*/, 
+                             IGRAPH_ALL  /*graph is "undirected"*/, /*cutoff*/ -1 /*calculating exact centrality*/,
                              NULL /*unweighted*/, /*not normalised*/ 0);
 
     printf("Non normalised results below\n");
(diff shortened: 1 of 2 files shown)


szhorvat:
How about this?

 1. We keep these changes on `develop` (I agree with @ntamas)
 2. For #1392 we add an extra change on `master` only: we check if the cutoff is 0 and if so, we change it to -1. In the vast majority of cases, people would not want to use `cutoff=0` anyway, as it's an extreme edge case. The priority is to maintain the interpretation of 0 as infinity for 0.8. Of course, #1392 does change the interpretation of cutoffs in other ways too, but that is part of a bugfix.
 3. Since we periodically merge `master` into `develop`, after the above change we need to do this merge and immediately undo this change on `develop` .
 4. Expand the changelog, and carefully explain how the use of `cutoff` has changed.

vtraag:
OK, so we will consider something an API breaking change if its behaviour changes with respect to what is stated in the documentation? A correction of an implementation (e.g. something that did not work as documented) would then be considered not an API breaking change, but a bugfix. This also implies that the documentation is leading in what is (or should be) done in the function, which is a good thing. That is a principle that I can live with.

I'm fine with reverting the change in #1392 back to using `cutoff = 0` indicating that no cutoff will be used in `master`. Indeed we need to be careful to change it back again in `develop`. Will you do this @szhorvat ?

szhorvat:
> Will you do this @szhorvat ?

OK, if others agree, I'll do it next week. https://github.com/igraph/igraph/issues/1412

ntamas:
Great, I'm happy that we are on the same page. Can I merge this into `develop` then?

vtraag:
> Can I merge this into `develop` then?

As far as I am concerned, yes.

PR #1416, opened 2020-06-12
degree_sequence_game: speedup for DEGSEQ_SIMPLE_NO_MULTIPLE_UNIFORM

szhorvat:
This speeds up the configuration model implementation significantly when sampling simple graphs.

The old implementation first shuffled the stubs vector, then checked for self-loops or multi-edges.

The new implementation checks for self-loops or multi-edges concurrently with shuffling.

The speedup depends on the degree sequence and the speed of the RNG. In the Mathematica interface, the default RNG is not very fast. In a small test I got a 3x speedup for undirected and 5x speedup for directed. This will vary considerably depending on the deg. seq. but it will always be faster than the old one.

Does not break the API so goes on master.

@ntamas Something to note is the `SWAP_INT_ELEM` macro, which is also used in `tree_game`. Here I added a second identical copy. It always undef'd after use.

### DIFF
--- a/src/games.c
+++ b/src/games.c
@@ -1214,15 +1214,24 @@ int igraph_degree_sequence_game_no_multiple_directed(igraph_t *graph,
     return IGRAPH_SUCCESS;
 }
 
+/* swap two elements of a vector_int */
+#define SWAP_INT_ELEM(vec, i, j) \
+    { \
+        igraph_integer_t temp; \
+        temp = VECTOR(vec)[i]; \
+        VECTOR(vec)[i] = VECTOR(vec)[j]; \
+        VECTOR(vec)[j] = temp; \
+    }
+
 int igraph_degree_sequence_game_no_multiple_undirected_uniform(igraph_t *graph, const igraph_vector_t *degseq) {
     igraph_vector_int_t stubs;
     igraph_vector_t edges;
     igraph_bool_t degseq_ok;
     igraph_vector_ptr_t adjlist;
-    long i, j, k;
+    long i, j;
     long vcount, ecount, stub_count;
 
-    IGRAPH_CHECK(igraph_is_graphical_degree_sequence(degseq, 0, &degseq_ok));
+    IGRAPH_CHECK(igraph_is_graphical_degree_sequence(degseq, NULL, &degseq_ok));
     if (!degseq_ok) {
         IGRAPH_ERROR("No simple undirected graph can realize the given degree sequence", IGRAPH_EINVAL);
     }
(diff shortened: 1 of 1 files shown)


src/games.c
```diff
@@ -1252,35 +1266,31 @@ int igraph_degree_sequence_game_no_multiple_undirected_uniform(igraph_t *graph,
...
+    }    
 
     RNG_BEGIN();
 
     for (;;) {
         igraph_bool_t success = 1;
-        IGRAPH_CHECK(igraph_vector_int_shuffle(&stubs));
 
-        /* optimization: we do an initial pass looking for self-loops */
+        /* Shuffle stubs vector with Fisher-Yates and check for self-loops and multi-edges as we go. */
```
vtraag, line 1276: Why not simply use [`igraph_vector_shuffle`](https://igraph.org/c/doc/igraph-Data-structures.html#igraph_vector_shuffle)?

ntamas: I think that's the whole point: instead of shuffling the vector completely and _then_ checking for self-loops at the end, we check for self-loops as the shuffle goes. So, if a self-loop is produced early in the shuffle in a part of the vector that is not going to be touched any more, we can bail out earlier and start from scratch. Is that right @szhorvat?

vtraag: Ah, sorry, was too quick looking at it! :+1:

szhorvat: Yes, that's the whole point, and that's what makes the speedup possible.

vtraag:
So, my understanding now is that you shuffle until you encounter a loop or multi-edge, and then restart the shuffling completely. Why not simply skip that particular random swap and continue with another random swap? That should speed it up still more, right? Of course, this is only possible if not creating a self-loop or multi-edge is at all feasible. Or do you then get into some problems with uniform sampling?

szhorvat:
> Why not simply skip that particular random swap and continue with another random swap?

That's a very common mistake that people make. It will not ensure uniform sampling.

szhorvat:
@vtraag The `IGRAPH_DEGSEQ_SIMPLE` does something like that. If it encounters a loop or a multi-edge, it tries another connection. If it still doesn't succeed, only then does it re-try from the start. However, 1) this breaks uniform sampling 2) actually computing the sampling probabilities is not easy for an ad-hoc modification like this.

I'm literally writing a paper on this right now :-)

vtraag:
> It will not ensure uniform sampling.

Yes, this was my worry indeed.

> I'm literally writing a paper on this right now :-)

Great! Good luck with it.

PR #1418, opened 2020-06-21
Modernizing older files

iosonofabio:
...learning from #1257

Changes:

 1. `igraph_edge` converted into `IGRAPH_OTHER`
 2. `igraph_vector_size` moved out of for loop declaration

I changed a number of files that used these two patterns. Tests pass and logic appears to be preserved.

Anything else that jumps to mind?

### DIFF
--- a/src/centrality.c
+++ b/src/centrality.c
@@ -2446,7 +2446,7 @@ int igraph_edge_betweenness_estimate(const igraph_t *graph, igraph_vector_t *res
             neino = igraph_vector_int_size(neip);
             for (i = 0; i < neino; i++) {
                 igraph_integer_t edge = (igraph_integer_t) VECTOR(*neip)[i];
-                long int neighbor = (long int) IGRAPH_OTHER(graph, edge, from);
+                long int neighbor = (long int) IGRAPH_OTHER(graph, edge, actnode);
                 if (nrgeo[neighbor] != 0) {
                     /* we've already seen this node, another shortest path? */
                     if (distance[neighbor] == distance[actnode] + 1) {
--- a/src/centrality.c
+++ b/src/centrality.c
@@ -131,7 +131,7 @@ static int igraph_i_eigenvector_centrality2(igraph_real_t *to, const igraph_real
         to[i] = 0.0;
         for (j = 0; j < nlen; j++) {
             long int edge = VECTOR(*edges)[j];
-            long int nei = IGRAPH_OTHER(graph, (igraph_integer_t) edge, i);
+            long int nei = IGRAPH_OTHER(graph, edge, i);
             igraph_real_t w = VECTOR(*weights)[edge];
             to[i] += w * from[nei];
         }
@@ -644,7 +644,7 @@ static int igraph_i_kleinberg_weighted(igraph_real_t *to,
         VECTOR(*tmp)[i] = 0.0;
         for (j = 0; j < nlen; j++) {
             long int nei_edge = (long int) VECTOR(*neis)[j];
-            long int nei = IGRAPH_OTHER(g, (igraph_integer_t) nei_edge, i);
+            long int nei = IGRAPH_OTHER(g, nei_edge, i);
             VECTOR(*tmp)[i] += from[nei] * VECTOR(*weights)[nei_edge];
         }
     }
@@ -655,7 +655,7 @@ static int igraph_i_kleinberg_weighted(igraph_real_t *to,
         to[i] = 0.0;
         for (j = 0; j < nlen; j++) {
             long int nei_edge = (long int) VECTOR(*neis)[j];
-            long int nei = IGRAPH_OTHER(g, (igraph_integer_t) nei_edge, i);
+            long int nei = IGRAPH_OTHER(g, nei_edge, i);
             to[i] += VECTOR(*tmp)[nei] * VECTOR(*weights)[nei_edge];
         }
     }
@@ -1028,7 +1028,7 @@ static int igraph_i_pagerank2(igraph_real_t *to, const igraph_real_t *from,
         to[i] = 0.0;
         for (j = 0; j < nlen; j++) {
             long int edge = (long int) VECTOR(*neis)[j];
-            long int nei = IGRAPH_OTHER(graph, (igraph_integer_t) edge, i);
+            long int nei = IGRAPH_OTHER(graph, edge, i);
             to[i] += VECTOR(*weights)[edge] * VECTOR(*tmp)[nei];
         }
         to[i] *= data->damping;
@@ -1696,7 +1696,7 @@ static int igraph_i_betweenness_estimate_weighted(
             nlen = igraph_vector_int_size(neis);
             for (j = 0; j < nlen; j++) {
                 long int edge = (long int) VECTOR(*neis)[j];
-                long int to = IGRAPH_OTHER(graph, (igraph_integer_t) edge, minnei);
+                long int to = IGRAPH_OTHER(graph, edge, minnei);
                 igraph_real_t altdist = mindist + VECTOR(*weights)[edge];
                 igraph_real_t curdist = VECTOR(dist)[to];
 
@@ -2187,7 +2187,7 @@ static int igraph_i_edge_betweenness_estimate_weighted(
             nlen = igraph_vector_int_size(neis);
             for (j = 0; j < nlen; j++) {
                 long int edge = (long int) VECTOR(*neis)[j];
-                long int to = IGRAPH_OTHER(graph, (igraph_integer_t) edge, minnei);
+                long int to = IGRAPH_OTHER(graph, edge, minnei);
                 igraph_real_t altdist = mindist + VECTOR(*weights)[edge];
                 igraph_real_t curdist = VECTOR(distance)[to];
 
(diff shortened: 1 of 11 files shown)


szhorvat:
I haven't looked through this yet. Let us know when it's ready (it's still marked as draft).

ntamas:
> `igraph_edge` is used in a few more .c files, shall I go over them and fix it?

I don't think it's necessary; when you have an edge ID and you need both endpoints, I think it's totally OK to use `igraph_edge`. I think you can gain the most (in terms of readability) in cases when you have _one_ endpoint of the edge and an edge ID, and you need the _other_ endpoint - in that case `IGRAPH_OTHER()` is much more readable than the alternatives.

szhorvat:
@ntamas When I added the IGRAPH_TO/FROM/OTHER docs, I wrote that they don't check the input (i.e. is the edge ID valid?), but `igraph_edge` does.  That turns out not to be true. I assumed it does error checking because it returns an error code, but in fact it always returns `IGRAPH_SUCCESS`. Now that `igraph_edge` is abolished from the internal functions, what do you think about actually adding that check to it? I.e., if the supplied edge ID is not valid, return `IGRAPH_EINVAL`.

If we don't do that, I'll fix up the docs instead.

ntamas:
Since `IGRAPH_TO`, `IGRAPH_FROM` and `IGRAPH_OTHER` are also public, I think it's okay to make the distinction such that `igraph_edge()` checks for errors while the others don't. (Python also does this in its C API; e.g., `PyLIST_SET_ITEM()` does not check for errors while `PyList_SetItem()` does). So, please go ahead and update `igraph_edge()` to do the checks.

iosonofabio:
Related to this: since igraph_edge will check, I assume we can remove it from or internal functions when it's safe to do so

iosonofabio:
I am surveying the codebase for `IGRAPH_OTHER` and it seems like the most common signature is:

```C
igraph_integer_t neighbor = IGRAPH_OTHER(graph, (igraph_integer_t) edge, actnode);
```

but also found failry often that `edge` is used as `long int` without recasting to `igraph_integer_t`.

Any idea what's going on? I've starting enforcing the `igraph_integer_t` cast for consistency, but am wondering if that makes sense.

src/conversion.c
```diff
@@ -299,15 +291,28 @@ int igraph_get_edgelist(const igraph_t *graph, igraph_vector_t *res, igraph_bool
 
     if (bycol) {
         while (!IGRAPH_EIT_END(edgeit)) {
-            igraph_edge(graph, IGRAPH_EIT_GET(edgeit), &from, &to);
+            long int edge = IGRAPH_EIT_GET(edgeit);
+            igraph_edge(graph, (igraph_integer_t) edge, &from, &to);
+	    /* FIXME: why are the two following lines not equivalent to igraph_edge??
```
iosonofabio, line 296: This one is really a mystery to me, @szhorvat @ntamas do you understand why the commented lines crash the tests?

ntamas:
> Any idea what's going on? I've starting enforcing the `igraph_integer_t` cast for consistency, but am wondering if that makes sense.

Quite confusingly, `igraph_integer_t` used to be a _double_ in older version of igraph. (I think it had to do something with R and the fact that certain 32-bit platforms that we used at the research institute years ago used 16 bits for ints, but I don't remember any more). This is not the case any more, but several explicit casts to `igraph_integer_t` (or to `long int`) are related to this and are not necessary any more. (It also caused utter confusion when you had to use `%f` or `%g` when printing an `igraph_integer_t`).

I think we should try to reduce the clutter in the codebase by not casting stuff unless we really have to, so I'd probably vote for removing the casts where possible.

In the case of `IGRAPH_OTHER()`, it is not necessary to cast because `IGRAPH_OTHER()` eventually boils down to `IGRAPH_FROM()` or `IGRAPH_TO()`, both of which cast their edge ID argument to a `long int` anyway because it is used as an array index. Back in the old days, when `igraph_integer_t` was a `double`, we needed an explicit cast to `long int` because you cannot use doubles as array indices. In this particular place, I think we need to keep the cast because `IGRAPH_FROM()`, `IGRAPH_TO()` and `IGRAPH_OTHER()` are part of the public API and removing the cast would disallow certain variable types that compile correctly with the current version. However, when you use `IGRAPH_OTHER()` internally in some function, I think it's not necessary to do any casting at all.

src/conversion.c
ntamas: `igraph_edge()` does something weird: if the graph is directed, it works as expected (the result of `IGRAPH_FROM()` is placed in `from` and `IGRAPH_TO()` is placed in `to`), but if the graph is undirected, it does the _opposite_ (the result of `IGRAPH_TO()` goes in `from` and vice versa. I don't know why it is implemented this way, but the pattern is quite consistent in `type_indexededgelist.c`.

One thing that I figured out is that there are two vectors in `igraph_t` (`from` and `to`) that represent the source and target vertices of the edges in the graph. When igraph is adding an edge to an undirected graph, it always makes sure that the _larger_ vertex ID goes in `from` and the _smaller_ vertex ID goes in `to`. Assuming that the user's expectation is to see the smaller vertex ID first for an undirected edge, we need to swap `from` and `to` for undirected graphs.

iosonofabio: OMG.

@gaborcsardi can you perhaps remember why/how this came to be? For your convenience here the igraph_edge function which seems to do the "unexpected" thing for undirected graphs:

```C
int igraph_edge(const igraph_t *graph, igraph_integer_t eid,
                igraph_integer_t *from, igraph_integer_t *to) {

    if (igraph_is_directed(graph)) {
        *from = IGRAPH_FROM(graph, eid);
        *to   = IGRAPH_TO(graph, eid);
    } else {
        *from = IGRAPH_TO(graph, eid);
        *to   = IGRAPH_FROM(graph, eid);
    }

    return IGRAPH_SUCCESS;
}
```

@ntamas Rightio, for now I'll just triple check that all changes in this PR are consistent with the previous ordering. However, I am not sure everyone on the team (let away the users!) know this tricky thing and that the convention is:

**For undirected graphs, igraph stores the larger vertex id in `graph->from`, but `igraph_edge` and `igraph_edges` will swap the order to give the smaller vertex id first.**

I'll write a post on the forum just as a reminder.

gaborcsardi: I am not sure what you mean specifically. It makes sense to always store the larger vertex first, because it makes operations simpler. Or always store the smaller first, it does not matter. I guess the choice was arbitrary, and then hard to change, because in R this data structure can be saved to disk.

I guess the `igraph_edge()` behavior came later, because it is indeed more natural for people to see the smaller vertex first.

iosonofabio: I see, thanks for that. I just didn't know that igraph_edge swaps things so it was surprising.

iosonofabio:
I think this is ready to merge

src/conversion.c
ntamas: > in R this data structure can be saved to disk.

I did not know that - we need to remember this in case we ever decide to change the graph representation. (I was under the impression that the internals of `igraph_t` are an implementation detail and we are free to change it if the basic interface does not change). It means that the best we can do if we change the representation is to provide _both_ the old and the new one, and let the user select it at compile time. The R interface could then stick to the old representation.

iosonofabio:
Awesome! Thank you!

src/conversion.c
gaborcsardi: You can still change it, just need to do more work when "converting" between `igraph_t` and the R object.

But why would you change it, anyway? There are _a lot_ of other possible improvements, with much bigger benefits than changing this, imo.

src/centrality.c
```diff
@@ -2445,10 +2445,8 @@ int igraph_edge_betweenness_estimate(const igraph_t *graph, igraph_vector_t *res
             neip = igraph_inclist_get(elist_out_p, actnode);
             neino = igraph_vector_int_size(neip);
             for (i = 0; i < neino; i++) {
-                igraph_integer_t edge = (igraph_integer_t) VECTOR(*neip)[i], from, to;
-                long int neighbor;
-                igraph_edge(graph, edge, &from, &to);
-                neighbor = actnode != from ? from : to;
+                igraph_integer_t edge = (igraph_integer_t) VECTOR(*neip)[i];
```
vtraag, line 2448: I think the `igraph_integer_t` cast is unnecessary here, since `neip` is a `igraph_vector_int_t*`. Or am I missing something here?

src/community.c
```diff
@@ -129,8 +129,8 @@ static int igraph_i_community_eb_get_merges2(const igraph_t *graph,
 
     for (i = igraph_vector_size(edges) - 1; i >= 0; i--) {
         long int edge = (long int) VECTOR(*edges)[i];
-        long int from = IGRAPH_FROM(graph, edge);
-        long int to = IGRAPH_TO(graph, edge);
+        long int from = IGRAPH_FROM(graph, (igraph_integer_t) edge);
```
vtraag, line 132: These casts to `igraph_integer_t` are also unnecessary I think?

src/centrality.c
vtraag: (The same thing also occurs elsewhere btw, but I first wanted to make sure this is correct).

ntamas: There are probably many other places, I only looked for `IGRAPH_OTHER()` instances and removed the explicit cast from the edge ID there (because `IGRAPH_OTHER()` is casting internally anyway).

src/community.c
ntamas: Yes, they are, `IGRAPH_FROM()` and `IGRAPH_TO()` cast their arguments internally.

vtraag: What should I do, shall I correct these in a new PR?

ntamas: Yes, please go ahead if you have the perseverance to do so :)

vtraag: Well, it's a bit waste of time to look at it and then do nothing with it :)

Issue #1070, opened 2018-03-01
Test on Alpine Linux


flying-sheep:
[OpenBLAS](https://github.com/xianyi/OpenBLAS/blob/486a485bb781b6a2c017c9924197911c31f7b1f4/.travis.yml#L62-L88) does that via Travis infrastructure as well and this would serve as a regression test for #1066

PR #1419, opened 2020-06-26
added Alpine Linux test run to Travis CI

iosonofabio:
I added a chroot into Alpine linux as suggested in the open issue #1070 

Currently one test only fails: conversion from sparse to dense matrix.

I'm trying to get the exact log out but the chrooting seems to interfere a little

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -81,6 +81,8 @@ jobs:
         - alpine make check
       env:
         - CFLAGS="-Wno-misleading-indentation -Wno-sign-conversion -Wno-incompatible-pointer-types"
+      after_failure:
+        - alpine find tests/testsuite.dir -name testsuite.log -exec cat \{\} \;
 
     - name: "macOS"
       os: osx
--- a/.travis.yml
+++ b/.travis.yml
@@ -70,7 +70,7 @@ jobs:
         - "wget 'https://raw.githubusercontent.com/alpinelinux/alpine-chroot-install/v0.6.0/alpine-chroot-install' \
               && echo 'a827a4ba3d0817e7c88bae17fe34e50204983d1e  alpine-chroot-install' | sha1sum -c || exit 1"
         - alpine() { /alpine/enter-chroot -u "$USER" "$@"; }
-        - sudo sh alpine-chroot-install -p 'build-base gfortran perl linux-headers libtool'
+        - sudo sh alpine-chroot-install -p 'build-base gfortran perl linux-headers libtool git'
       install:
         - set -e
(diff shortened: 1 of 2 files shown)


iosonofabio:
I tried all kinds of bisections on test 27, i.e. conversions between sparse and dense matrices, but can't for the best of me diagnose the problem. If anyone could have a look that'd be fantastic!

ntamas:
I'll try to reproduce this in Docker on an `alpine` image.

ntamas:
I think https://github.com/igraph/igraph/commit/f6797e82fa984980e1a3a1336a0537eb195f4cac fixes this.

iosonofabio:
Should be fine I think

ntamas:
Tests seem to pass in CI -- is this ready to merge?

iosonofabio:
The only thing is I'm not sure if the `after_failure` trigger works correctly from inside the `chroot`. The CI fails alright when at least one test fails, but it causes an exclamation mark instead of an X. Do you have any clue what this could be? Maybe an exit code in bash?

ntamas:
Hmmm, good point. I think the Travis policy is to mark the build as "errored" if any of the commands in the build steps returns a non-zero exit code, _except_ the script section, where a non-zero exit code would simply indicate a test failure.

ntamas:
Do you have a link to a recently failed Travis build that errored out? I can't find one in the build history in my account.

iosonofabio:
This would be one:

https://travis-ci.org/github/igraph/igraph/jobs/702599842

It says the test failed unexpectedly, but fails to show the log...

ntamas:
Okay, this is totally a stab in the dark, but basically Travis marks the build as "errored" if there is an error in the `before_install`, `install` or `before_script` phases (i.e. any command that is invoked in these stages returns with a non-zero error code). Any non-zero error code in the `script` phase should mark the build as "failed". The reason why this build is marked as "errored" _might_ be because you invoke the build command via `alpine()`, which is a function that you defined in the `install` phase. Try invoking `make check` without using the function defined in the `install` phase, or re-define it in the `script` phase.

iosonofabio:
I redefined the function in the script phase, but it still errors with an exclamation mark instead of a cross, so it does not trigger the `after_failure`. I'll try asking Travis folks...

iosonofabio:
Alright, it seems like `set -e` messes up the Travis error codes. They had a bug report in 2013 which they seem to have fixed, but there must be something still broken about that...

either way it's fixed now, ready for merge

ntamas:
Thanks a lot!

PR #1420, opened 2020-06-26
Appveyor: rewriting config to run with and without MSVC, irrespectively of Python version

ntamas:
This PR rewrites the Appveyor CI config file to run only once on MSVC and one with MinGW/MSYS.

@iosonofabio Is that what you had in mind?
@vtraag Is this in line with what we've discussed in the Discourse group?


vtraag:
Yes, I believe this covers it.

There is one thing that we can perhaps now solve, since we are changing things anyway. It states that we are using `conda` to install suitable libraries, but I don't think that is the case. We install `libxml2` when but I don't think the `libxml2` from `conda` is actually used when compiling from MSVC. We can simply get rid of all conda dependencies here I think.

vtraag:
Great! As far as I'm concerned we are good to go.

PR #1421, opened 2020-06-26
Removed some unnecessary casts.

vtraag:
Continuing from https://github.com/igraph/igraph/pull/1418#pullrequestreview-438154796, this PR simply removes some unnecessary casts.

### DIFF
--- a/src/centrality.c
+++ b/src/centrality.c
@@ -1618,8 +1618,8 @@ static int igraph_i_betweenness_estimate_weighted(
         igraph_bool_t nobigint) {
 
     igraph_real_t minweight;
-    igraph_integer_t no_of_nodes = (igraph_integer_t) igraph_vcount(graph);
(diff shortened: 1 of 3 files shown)


ntamas:
Should this go into `master` or `develop`? On one hand, this does not break anything so it _could_ go into `master`. On the other hand, since this does not _fix_ anything either, it should probably go to `develop` instead and we merge stuff directly into `master` only if it's a bugfix (and then we merge `master` into `develop` immediately afterwards). I'm just worried that the two branches go out of sync if we don't stick to a strict policy about which branch new PRs should primarily target.

vtraag:
Yes, sorry, PR #1418 was targeted at `develop` so it probably makes more sense to merge it there.

vtraag:
Forget about this one, I will close it and start a new PR, properly rebased first against `develop` (some other time though). Sorry for messing up.

PR #1425, opened 2020-07-01
Added missing DECLDIR specs to fix DLL exports on Windows

vtraag:
As discussed during the `CMake` development, for building a shared library on Windows, some `DECLDIR` seemed to be missing. This PR corrects this in the `master` branch. Note that there are a few "border" cases of what could be considered "public" and what not. Some functions were commented to be "internal", yet were included in the public headers, and some were even explicitly included in the documentation. Some of them I marked with `DECLDIR` if I thought it would be appropriate. But feel free to comment on this.


szhorvat:
> Some functions were commented to be "internal", yet were included in the public headers, and some were even explicitly included in the documentation. Some of them I marked with `DECLDIR` if I thought it would be appropriate. But feel free to comment on this.

IMO we should go by the public headers, which are always installed with the library, and *not* the documentation.

If it's in the public headers, it *should* be exported, regardless of whether it's documented or not.

ntamas:
@szhorvat made a good point here; if we put something in the public headers but don't export it, that bit of code is basically useless. So yes, I think that if there is something that is already in the public headers, we need to export it.

I'll go through this PR in detail soon, thanks!

ntamas:
I think that `DECLDIR` can't replace `extern` entirely, so if something was originally declared as `extern`, we now need to declare it as `DECLDIR extern` instead of simply `DECLDIR`.

vtraag:
>  I think that if there is something that is already in the public headers, we need to export it.

OK, I will go through it again to see if there were any other missing symbols from public include files (i.e. from the `include/` directory).

> I think that `DECLDIR` can't replace `extern` entirely, so if something was originally declared as `extern`, we now need to declare it as `DECLDIR extern` instead of simply `DECLDIR`.

I believe the only effect of `extern` is to make sure it is properly exported when compiled as part of a C++ library, right? That is already properly dealt with by the `__BEGIN_DECLS`, `__END_DECLS` symbols. But correct me if I am wrong here.

ntamas:
> I believe the only effect of extern is to make sure it is properly exported when compiled as part of a C++ library, right? 

I think for constants it has an additional meaning: "dear compiler, the value of this stuff is defined in another compilation unit". Apparently this is why the Travis build errors out now; certain constants that were declared `extern` in the headers are not `extern` any more and the compiler tries to come up with a "default constructor" for them, which is not always possible.

vtraag:
Ah yes, of course! Sorry, forgot about that...

include/igraph_error.h
```diff
@@ -217,7 +217,7 @@ typedef void igraph_error_handler_t (const char * reason, const char * file,
  * program.
  */
 
-extern igraph_error_handler_t igraph_error_handler_abort;
+DECLDIR igraph_error_handler_t igraph_error_handler_abort;
```
vtraag, line 220: @ntamas , please note that these `extern`s are not necessary. They seem to define a symbol, but they actually define a function, and if my understanding is correct `extern` is unnecessary there (it is already implied).

vtraag:
I add back one `extern`, the other `extern`s are actually unnecessary, it now compiles and checks correctly locally (i.e. for the `master` branch).

include/igraph_error.h
ntamas: Yes, it seems to compile this way so all is good.

PR #1429, opened 2020-07-07
First stub with igraph_set_t

iosonofabio:
We [recently](https://igraph.discourse.group/t/data-structure-type-usage-and-visibility/316/3) discussed the possibility of exposing and documenting a bunch of internal types.

This PR is bringing that discussion forth, currently with a single example of `igraph_set_t`.

Seems like for each type one has to:
- make a `igraph_<type>.h` file by copying from another type
- make a `igraph_<type>_pmt.h` file by moving and adapting what's currently in `igraph_types_internal.h`
- change the header include in the implementation file
- add documentation if needed.

What do you guys think? Shall we make a list or abandon this PR until 1.0?

### DIFF
--- a/include/igraph_set.h
+++ b/include/igraph_set.h
@@ -1,0 +1,73 @@
+/* -*- mode: C -*-  */
+/*
+   IGraph library.
+   Copyright (C) 2009-2012  Gabor Csardi <csardi.gabor@gmail.com>
+   334 Harvard street, Cambridge, MA 02139 USA
+
+   This program is free software; you can redistribute it and/or modify
+   it under the terms of the GNU General Public License as published by
+   the Free Software Foundation; either version 2 of the License, or
+   (at your option) any later version.
+
+   This program is distributed in the hope that it will be useful,
+   but WITHOUT ANY WARRANTY; without even the implied warranty of
+   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+   GNU General Public License for more details.
+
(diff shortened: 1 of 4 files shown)


ntamas:
This is something that we should do sooner or later, but personally I won't be able to work on this until I finish and merge the CMake build branch. (I can do code review of course). But if you would like to go ahead with this, feel free to do so and I'll be happy to do a review.

One other approach that we could consider instead of the "poor man's templates" (in case you ever wondered what "pmt" stands for :)) is to use a pure macro-based, header-only approach, similarly to how hash tables and linked lists are done in the [uthash](https://troydhanson.github.io/uthash/) and [utlist](https://troydhanson.github.io/uthash/utlist.html#_introduction) libraries. The general idea here is that the library provides a set of macros that make certain assumptions about the types they operate on -- e.g., for singly linked lists the struct must have a member named `next` that points to the next element of the list, but otherwise the struct can contain any additional data members. Not sure how this would work for sets (or whether it would work at all), and also I'm not sure whether we should use it in igraph (for sake of consistency), but it's a possibility.

iosonofabio:
OMG that's what PMT stands for. https://github.com/troydhanson/uthash/blob/master/src/utlist.h looks somewhere between amazing and witchcraft, holy cow.

For now I just implemented this stub to collect opinions around the table. Since @szhorvat recently expressed the desire for linked lists, we might consider including `utlist` for that instead of implementing a new PMT ourselves?

Issue #1407, opened 2020-06-09
igraph_i_lad_initDomains needs more IGRAPH_FINALLY


szhorvat:
In `igraph_i_lad_initDomains`, in `lad.c`, the variabled `val` and `dom` don't have a proper `IGRAPH_FINALLY` on them, so there may be a memory leak if the function exits in the middle due to an error.

Don't do this until #1404 is merged, as it touches the same code.

PR #1430, opened 2020-07-07
Wrap val/dom in LAD algorithm in ALLOC_ARRAY

iosonofabio:
Addressing #1407

I wrapped the two variables in `ALLOC_ARRAY` which seems to basically allocate the memory, check for failure, and put the variable on the stack.


szhorvat:
You missed an exit point of the function. It's a good idea to look for `return`s that are not at the very end whenever touchnig any FINALLY counts. Fixed this, and will merge when the tests pass. Thanks!

Issue #297, opened 2013-05-04
Eliminate all gcc warnings


gaborcsardi:
For the external code, we might need some patches....

---

Imported from Launchpad using lp2gh.
- date created: 2012-06-14T15:39:11Z
- owner: gabor.csardi
- the launchpad url was https://bugs.launchpad.net/bugs/1013231

ntamas:
igraph now compiles without warnings with recent `gcc` and `clang` versions, even with `-Wall -Wextra -pedantic -Werror`.

PR #1431, opened 2020-07-08
Gcc warnings

iosonofabio:
Addressing issue #297.

There are quite a few more warnings, but they are related to dependencies (f2c or lapack in particular). Probably not going to fix those.

Also, the fix for `triangles_template.h` is a little questionable - a sometimes unused variable is now defined under an `#ifdef`. Suggestions to improve that one are welcome.

### DIFF
--- a/src/centrality.c
+++ b/src/centrality.c
@@ -2622,7 +2622,8 @@ static int igraph_i_closeness_estimate_weighted(const igraph_t *graph,
 
     int cmp_result;
     const double eps = IGRAPH_SHORTEST_PATH_EPSILON;
-    igraph_real_t mindist;
+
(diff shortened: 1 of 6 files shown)


ntamas:
Thanks!

ntamas:
I have merged this but I realized that I've already committed something in `src/triangles_template.h` in the `feature/cmake-build` branch that was supposed to fix this; basically I have added this:

```
/* deg1 is not used in all instantiations of this template */
IGRAPH_UNUSED(deg1);
```

Can you please check whether this also fixes things with your version of gcc? (It worked for me). If it does, I believe this is cleaner than the `#ifdef` trickery.

Issue #1200, opened 2019-06-09
Example for igraph_full is missing


szhorvat:
`examples/simple/igraph_full.c` is included as an example in the docs, but the example only has an empty `main` function.

Either add a meaningful example or remove this (as it's trivial anyway).

This is just a TODO item (so we won't forget).

PR #1434, opened 2020-07-09
Write example for igraph_full

iosonofabio:
Fixing #1200 with a few examples for undirected/directed full graphs with and without self loops.


ntamas:
Not sure whether this is really meant as an example or as a test case. If we don't have a test case for `igraph_full()` yet, maybe it's worth moving (or copying) this to `examples/tests/` and check whether the graph that was created is indeed a full graph? (Printing the edge list to a file and providing an `.out` file for the test case would be enough I think).

iosonofabio:
Alright, I copied the file into `example/tests` and added checks that the edges are really there with an output file. I think we can merge now?

szhorvat:
Once again, tab-indentation and space-indentation are being mixed randomly.

iosonofabio:
Indeed, my editor fixes it for me. Please feel free to improve my humble, incorregible habits.

szhorvat:
You can use `print_graph` from `test_utilities.inc` to output the graph.

Please always add `VERYIFY_FINALLY_STACK();` at the end of tests.

iosonofabio:
Yes, added the verify statement at the end of the test.

Issue #51, opened 2013-05-04
Check functions for integer overflow


gaborcsardi:
In particular, all the functions that consider number of geodesics at some point might contain an integer overflow. Let's start with betweenness and edge_betweenness, I am sure that there are more, of course.

In case of integer overflow we just return an error with a new error code, IGRAPH_EINTOVERFLOW or something like that.

---

Imported from Launchpad using lp2gh.
- date created: 2009-08-03T18:27:24Z
- owner: gabor.csardi
- the launchpad url was https://bugs.launchpad.net/bugs/408521

iosonofabio:
1. do we need a constant that stores the largest integer for this?
2. what type are the elements of `igraph_vector_t`? (I feel stupid asking this...)

gaborcsardi:
1. There are constants in the standard C lib: INT_MAX, UINT_MAX, etc. But these are not very useful for checking integer overflow, because no int can be bigger than INT_MAX, by definition.
2. doubles

iosonofabio:
Thank you. I thought one could do something like:

```C
int var_at_risk = ...;
int increment = ...;

if (increment > (INT_MAX - var_at_risk)) {
    IGRAPH_ERROR("Integer overflow", IGRAPH_EINTOVERFLOW);
}

var_at_risk += increment;
```

The number of geodesics in betweenness is an `igraph_vector_t`, so it's a double:

```C
    IGRAPH_VECTOR_INIT_FINALLY(&nrgeo, no_of_nodes);
```

so maybe I'm not understanding what kind of integer overflow you are afraid of?

gaborcsardi:
Yeah, that would work I think. Maybe it is a double to avoid the overflow, and the issue is older?

iosonofabio:
Alright, thanks, so I'll go around the codebase a little to identify integers that are at risk of overflowing

gaborcsardi:
We can also deal with this if it comes up, I guess?

iosonofabio:
I'll have a look around since I'm on it, then we can close this issue (marked for 0.9.0) and open new ones if that problem arises again.

gaborcsardi:
Btw. another common way to check for integer overflow for an A + B sum is to check the sum is smaller than A. If it is, that's an overflow.

iosonofabio:
Ok, found that `edge_betweenness` uses `unsigned long long int` which can in theory overflow. So I put checks for overflow there, in PR #1436

szhorvat:
[...]

PR #1436, opened 2020-07-10
Check for int overflow in edge_betweenness

iosonofabio:
Addressing #51 

Problems found:
 - [x] `edge_betweenness`
 - [ ] `betweenness` is already using doubles... is that alright??
 - [x] `community_edge_betweenness`

There are no other instances of `unsigned long long int` or `unsigned long int` or `long long int` in the `src` folder, so I take it that these should be the only two obvious cases.

Is that fine that the `betweenness` is using doubles?

### DIFF
--- a/include/igraph_error.h
+++ b/include/igraph_error.h
@@ -328,6 +328,7 @@ DECLDIR igraph_error_handler_t* igraph_set_error_handler(igraph_error_handler_t*
  * \enumval IGRAPH_CPUTIME CPU time exceeded.
  * \enumval IGRAPH_EUNDERFLOW Integer or double underflow.
  * \enumval IGRAPH_ERWSTUCK Random walk got stuck.
+ * \enumval IGRAPH_EINTOVERFLOW Integer overflow
  */
 
 typedef enum {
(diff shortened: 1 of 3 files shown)


vtraag:
Might it not be better to add an additional internal function like `igraph_i_add_int_safe` which does the overflow (and underflow) checks and raises an error? Otherwise the pattern needs to be repeated every time, inviting possible bugs. A macro instead of a function would also work of course.

iosonofabio:
I'd be happy with that, a macro probably less problematic in terms of performance hit. Where do you think the macro could sit?

vtraag:
Good question! I don't have any file in mind immediately, perhaps @ntamas has a suggestion. Perhaps we need a different header for this?

szhorvat:
Testing tip:

Overflow in (edge or vertex) betweenness calculations should occur with a grid graph of size 35*35, but _not_ at size 34*35.

szhorvat:
@iosonofabio In case you were still working on this: I tested out changing the `long long` to `double` and it does not seem to have any noticeable precision penalty while preventing overflow. It even makes the use of bigints unnecessary (though I don't want to remove bigint support as it was exactly that that allowed me to verify that `double`s work well). Details are in the chat.

Thus this check won't be necessary.

iosonofabio:
Sounds good, thanks for that.

szhorvat:
Closing in favour of #1576.

PR #1437, opened 2020-07-10
Add search box to html C manual

iosonofabio:
I added an additional filter with sed to the docs `Makefile.am` such that only in the html version, a customized google search box for the C manual is inserted before the table of contents.

That should partially address (for now) the lack of searchability of the docs.

I also renamed the shebang in the custom python filter to use python2.7. Notice that is past end of life, so we should do something about that within a year or so.

I also added Vincent, Szabolcs and myself to the list of authors. @gaborcsardi and @ntamas , you might want to update your affiliations. If so please push to this PR or just write in the comments and I'll update them.

### DIFF
--- a/doc/Makefile.am
+++ b/doc/Makefile.am
@@ -223,8 +223,10 @@ graphlets.xml: graphlets.xxml $(SRCDIR)/glet.c
 
 html/stamp: igraph-docs0.xml $(DOCFIX2) gtk-doc.xsl
 	if [ "x$(top_srcdir)" != "x$(top_builddir)" ]; then cp $(DOCFIX2) . ; fi && \
-	xmlto -x $(top_srcdir)/doc/gtk-doc.xsl -o html xhtml igraph-docs0.xml \
-	&& touch html/stamp
+	xmlto -x $(top_srcdir)/doc/gtk-doc.xsl -o html xhtml igraph-docs0.xml && \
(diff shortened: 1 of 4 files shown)


ntamas:
`doxrox.py` is being revamped in the `feature/cmake-build` branch so it is actually 3.x-compatible in that branch. The reason why I did not merge it back immediately is because I slightly changed the invocation and the workflow to speed up the CMake build a bit. But anyway, we don't need to about this as soon as we switch to CMake.

I'll review the rest of this PR today, thanks!

ntamas:
Let's break this PR down into multiple parts:

* the shebang was updated in `doxrox.py` to use Python 2.7; this is OK but I suggest you simply add this to the `develop` branch as a separate commit (no need to send a PR, you have write access).

* the list of authors in the documentation was extended to reflect the recent CZI grant, this is also okay -- feel free to commit this to the `develop` branch

* the search box was added to the generated HTML pages. I would rather make this part of the webpage (i.e. the `igraph.org` repo), not this repo. Right now the idea is that the scripts in this repo generate a "plain" DocBook HTML docset without any additional styling, and the `igraph.org` repo adds the styling and any additional things that are needed to make the documentation behave nicely within the context of the webpage.

I'll close this PR and add the search box to the webpage based on the snippet found in this PR, so no need to take care of that yourself. As for the remaining two points above, just add a commit to `develop`.

ntamas:
Search bar is now added to the C documentation: https://github.com/igraph/igraph.org/commit/c69ebe0afd78b2e24e7e776f752600010dbd178c

Issue #1435, opened 2020-07-09
Crash of Leiden algorithm on 600 million vertices and 3.8 billion edge graphs.


munhyunsu:
I'm trying to run the Leiden algorithm on 600 million vertices and 3.8 billion edge graphs.
The graph is constructed with `read_graph_edge_list()` and the leiden algorithm is executed. The following message is displayed and a segmentation fault occurs.

I don't know if iGraph can't handle this graph or if there is an issue with the implementation of the leiden algorithm.
However, if the iGraph developer advises how to compile libigraph.so.0 in debug mode, as in the backtrace below, to find the location of the error (debug mode compilation method), I might contribute myself.

summary:
-Leiden algorithm fails on large graphs (600 million vertices, 3.8 billion Edge)

Backtrace:
```
#0 0x00007ffff786480c in igraph_i_community_leiden ()
   from /opt/igraph/lib/libigraph.so.0
#1 0x00007ffff78668ed in igraph_community_leiden ()
   from /opt/igraph/lib/libigraph.so.0
#2 0x0000555555554ee3 in main (argc=3, argv=0x7fffffffe0b8)
    at find_community_leiden.c:40
```

vtraag:
Thanks for the report. Which version of `igraph` are you running?

munhyunsu:
Hi @vtraag 
I used igraph-0.8.2

Thanks.

vtraag:
Can you share the graph in one way or another? It is difficult to debug the problem without more detailed information.

You can also compile the `igraph` library yourself and enable debugging, preferably from the `master` branch of this repository. See https://igraph.org/c/#cinstall for the general instructions. Essentially, make sure you have installed the necessary packages. Assuming you are on Ubuntu or similar, simply run:
```
sudo apt-get install build-essential libxml2-dev libgmp-dev automake autoconf
```
after that, you should then run
```
./configure --enable-debug
make
sudo make install
```
You can also run `make check` to test the package. If you are indeed compiling from the `master` branch, you additionally need to install `flex` and `bison` and run `./bootstrap.sh` before running `./configure`.

munhyunsu:
I was already compiling and using iGraph.
I will reinstall `igraph` with the `--enable-debug` tag to find the problem.

If you are interested in this data, please let me know.
Edge data is about 70 gigabytes, so I'll upload it to Google Drive and email the link to you.

vtraag:
[...]

PR #1438, opened 2020-07-12
Memory and error handling in Leiden algorithm

vtraag:
This PR may addresses some issues around memory allocation and error handling that might be causing some problem that were identified in #1435. 

The first commit already releases some chunk of memory before allocating an additional chunk of memory, thereby decreasing the overall memory footprint.

The second commit changes the order of an `IGRAPH_FINALLY(igraph_Free, ...)` and the initialization of a vector. In the current code, it might be that the memory to create the vector itself was not allocated, in which case the vector might have been incorrectly attempted to be initalised. This is now corrected.

However, I believe there is an issue with the way that FINALLY error handling currently works. As it stands, it is implemented as a stack, meaning that the last one to be pushed on should be the first one to be popped off. However, this is not the case in the Leiden algorithm I notice (and it might also not be the case elsewhere). In particular, the Leiden algorithm uses something as follows:

```c
do {

    /* Get incidence list for fast iteration */
    IGRAPH_CHECK(igraph_inclist_init(aggregated_graph, &edges_per_node, IGRAPH_ALL));
    IGRAPH_FINALLY(igraph_inclist_destroy, &edges_per_node);

    ...

    /* Allocate temporary graph */
    tmp_graph = igraph_Calloc(1, igraph_t);
    if (tmp_graph == 0) {
        IGRAPH_ERROR("Leiden algorithm failed, could not allocate memory for aggregate graph", IGRAPH_ENOMEM);
    }
    IGRAPH_FINALLY(igraph_free, tmp_graph);

    IGRAPH_CHECK(igraph_i_community_leiden_aggregate(
                        aggregated_graph, &edges_per_node, aggregated_edge_weights, aggregated_node_weights,
                        aggregated_membership, &refined_membership, nb_refined_clusters,
                        tmp_graph, &tmp_edge_weights, &tmp_node_weights, &tmp_membership));

    /* Graph has been created by aggregation, ensure it is properly destroyed if
        * an error occurs. */
    IGRAPH_FINALLY(igraph_destroy, tmp_graph);

    if (level >= 1) {
        /* Destroy previously allocated graph (note that aggregated_graph points to
            * the previously allocated tmp_graph). */
        igraph_destroy(aggregated_graph);
        igraph_Free(aggregated_graph);
        IGRAPH_FINALLY_CLEAN(2);
    }

    /* Set the aggregated graph correctly */
    aggregated_graph = tmp_graph;    

    ...

    /* We are done iterating, so we destroy the incidence list */
    igraph_inclist_destroy(&edges_per_node);
    IGRAPH_FINALLY_CLEAN(1);
} while (continue_clustering);
```
(Here, the `...` indicate some code that is nonessential for this discussion).

The problem now is the following. What gets pushed on the stack is (in this order):
1. `igraph_inclist_destroy(&edges_per_node)` 
2. `igraph_free(tmp_graph)`
3. `igraph_destroy(tmp_graph)`

The last line that cleans one `FINALLY`, meaning the last one is popped off the stack, meaning the stack will read
1. `igraph_inclist_destroy(&edges_per_node)` 
2. `igraph_free(tmp_graph)`

When the `do .. while` loop continues in its next iteration, it will become
1. `igraph_inclist_destroy(&edges_per_node)` 
2. `igraph_free(tmp_graph)`
3. `igraph_inclist_destroy(&edges_per_node)` 
4. `igraph_free(tmp_graph2)`
5. `igraph_destroy(tmp_graph2)`

(here `tmp_graph2` refers to a newly allocated `igraph_t` object), and the last two will be popped off (when destroying `tmp_graph`), so that the FINALLY stack becomes
1. `igraph_inclist_destroy(&edges_per_node)` 
2. `igraph_free(tmp_graph)`
3. `igraph_inclist_destroy(&edges_per_node)` 

after which the last one is correctly popped off.

This explains the error with the `inclist` being destroyed when an out-of-memory error occurs (which I noted [here](https://github.com/igraph/igraph/issues/1435#issuecomment-656919699)). While executing everything from the FINALLY` stack, the `igraph_inclist_destroy(&edges_per_node)` is on the stack one time too many (and some graph objects are actually not properly destroyed).

I have been thinking of finding a way to rewrite this part without causing a problem, but I'm afraid I don't see an easy solution. Do you have any idea whether this problem might occur more often? Does anybody have any suggestion for how to solve this problem?

### DIFF
--- a/src/community_leiden.c
+++ b/src/community_leiden.c
@@ -1046,8 +1046,8 @@ int igraph_community_leiden(const igraph_t *graph,
         if (i_edge_weights == 0) {
             IGRAPH_ERROR("Leiden algorithm failed, could not allocate memory for edge weights", IGRAPH_ENOMEM);
         }
-        IGRAPH_CHECK(igraph_vector_init(i_edge_weights, igraph_ecount(graph)));
         IGRAPH_FINALLY(igraph_free, i_edge_weights);
+        IGRAPH_CHECK(igraph_vector_init(i_edge_weights, igraph_ecount(graph)));
         IGRAPH_FINALLY(igraph_vector_destroy, i_edge_weights);
         igraph_vector_fill(i_edge_weights, 1);
     } else {
@@ -1060,8 +1060,8 @@ int igraph_community_leiden(const igraph_t *graph,
         if (i_node_weights == 0) {
             IGRAPH_ERROR("Leiden algorithm failed, could not allocate memory for node weights", IGRAPH_ENOMEM);
         }
-        IGRAPH_CHECK(igraph_vector_init(i_node_weights, n));
         IGRAPH_FINALLY(igraph_free, i_node_weights);
+        IGRAPH_CHECK(igraph_vector_init(i_node_weights, n));
         IGRAPH_FINALLY(igraph_vector_destroy, i_node_weights);
         igraph_vector_fill(i_node_weights, 1);
     } else {
--- a/src/community_leiden.c
+++ b/src/community_leiden.c
@@ -610,16 +610,19 @@ static int igraph_i_community_leiden_aggregate(
 
     }
 
-    IGRAPH_CHECK(igraph_create(aggregated_graph, &aggregated_edges, nb_refined_clusters,
-                               IGRAPH_UNDIRECTED));
-
     igraph_vector_destroy(&neighbor_clusters);
     igraph_vector_bool_destroy(&neighbor_cluster_added);
     igraph_vector_destroy(&edge_weight_to_cluster);
-    igraph_vector_destroy(&aggregated_edges);
     igraph_vector_ptr_destroy_all(&refined_clusters);
 
-    IGRAPH_FINALLY_CLEAN(5);
+    IGRAPH_FINALLY_CLEAN(4);
+
+    IGRAPH_CHECK(igraph_create(aggregated_graph, &aggregated_edges, nb_refined_clusters,
+                               IGRAPH_UNDIRECTED));
+
+    igraph_vector_destroy(&aggregated_edges);
+
+    IGRAPH_FINALLY_CLEAN(1);
 
     return IGRAPH_SUCCESS;
 }


iosonofabio:
This one leaves me a little confused.

If I understand correctly, in the first iteration `level` is zero and `aggregated_graph` is actually the initial graph, so of course you don't want to destroy it. But the edge list is created first and destroyed first since the new `tmp_graph` is needed in the next iteration.

The naive suggestion would be of course to first allocate the `tmp_graph` and put it on the stack before the inclist. That makes sense logically since it's longer lived than the inclist. The problem with that is that for iterations except the first one you want to destroy the old `aggregated_graph` before destroying its own inclist. Now I'm wondering since you destroy the graph, chances are you can destroy its inclist at the same time?

Even if you cannot do that, the inclist is created in this iteration, so you know how to destroy it. So you could easily remove three things from the stack and then put the inclist back on it:

```C
    if (level >= 1) {
        /* Destroy previously allocated graph (note that aggregated_graph points to
         * the previously allocated tmp_graph). */
        igraph_destroy(aggregated_graph);
        igraph_Free(aggregated_graph);
        IGRAPH_FINALLY_CLEAN(3);
        IGRAPH_FINALLY(igraph_inclist_destroy, &edges_per_node);
    }
```

Perhaps not the prettiest, but it would work.

More generally, of course a stack cannot cover all cases. Think about vectors of pointers, those are basically unchecked memory leak landmines waiting to explode - since we don't put each item in the vector on the stack. However I am not sure redesigning the whole finally stack is our top priority now. So I'd vote for having a look across the codebase for similar interleaved cases like this and fixing them to agree with the stack rather than vice versa. For now ;-)

ntamas:
The stack model has worked fine for 99% of the use-cases so far, and maintaining hardcoded position indices in the finally stack will become hard if we ever have to refactor any of these algorithms, so I'd vote for not changing anything in the low-level code and just try to find a workaround for it in the Leiden algorithm.

If we really want to extend the existing solution: how about allowing the user to create multiple "finally stacks" such that when a new stack is created, a function that destroys everything that was put in the newly created stack is automatically placed on top of the _old_ stack (from which the new stack was "forked")? This way one can keep on pushing items to the top of the old stack or to the _new_ stack, which would essentially "insert" or "remove" items at a specific place in the old stack (where the two stacks were forked)? I don't know how hard it would be to track multiple stacks in one's head while writing code, but it would solve the problem of "removing something from the middle of the stack" in a relatively structured way, without having to rely on hardcoded indices.

szhorvat:
@vtraag Here's an idea for a solution:

 1. `igraph_i_community_leiden_aggregate` does not create the graph. Instead, it takes an empty graph as input and adds the edges to it.
 2. Conceptually, we will operate not on graphs, but on pointers to graphs. This is what is being passed to `IGRAPH_FINALLY`, therefore we need a destructor for a "pointer to graph".

Here is an example constructor and destructor that can be used with `IGRAPH_FINALLY`:

```c
static int create_graph_ptr(igraph_t **graph) {
    *graph = igraph_Calloc(1, igraph_t);
    if (*graph == NULL) {
        IGRAPH_ERROR("Leiden algorithm failed, could not allocate memory for graph", IGRAPH_ENOMEM);
    }
    IGRAPH_FINALLY(igraph_free, *graph);
    IGRAPH_CHECK(igraph_empty(*graph, 0, IGRAPH_UNDIRECTED));
    IGRAPH_FINALLY_CLEAN(1);
    return IGRAPH_SUCCESS;
}

static void destroy_graph_ptr(igraph_t **graph) {
    if (*graph) {
        igraph_destroy(*graph);
        igraph_Free(*graph);
    }
}
```

We will work with two pointers, `aggregate_ptr` and `temp_ptr`. The loop will look like this:

 1. Loop:
    1. create `inclist_aggregate`
    2. create `temp_ptr`
    3. call `igraph_i_community_leiden_aggregate`
    4. `swap(aggregate_ptr, temp_ptr)`
    5. if `level > 0`, destroy `temp_ptr`
    6. destroy `inclist_aggregate`
 2. if `level > 0`, destroy `aggregate_ptr`

If you like the idea, I can push this tomorrow.

It's quite late here, I hope I didn't make a mistake here.

vtraag:
I am not sure if I follow entirely @szhorvat (it's getting late here as well :), but I think the crucial insight is indeed that I could simply create the graph beforehand. That way, I don't need to push/pop it off the FINALLY stack, I can simply allocate an empty graph initially (and push it to FINALLY stack), and then reuse that same graph by emptying it every time. I actually also need to do that on a couple of other variables (vectors), but for those vectors the solution was more clear.

I will try to see if I can find some time to do this later this week.

szhorvat:
That's not what I meant, but looking more closely, it seems to be a simpler solution. `aggregated_graph` and `tmp_graph` seem to be never used at the same time. Right now, `igraph_i_community_leiden_aggregate` takes `aggregated_graph` and creates `tmp_graph` from it. Do you mean that instead of doing this, `igraph_i_community_leiden_aggregate` could simply update `aggregated_graph` in-place? That does indeed seem to be the cleanest solution.

iosonofabio:
Of course, if you can allocate variables and memory beforehand and just play with what's inside simplifies things.

vtraag:
Hmm, I see that the `igraph_t` object is not able to remove all edges in an efficient manner. This makes it difficult to reuse the same variable. I could of course add that function. What do you think?

I will have to rethink your original proposal @szhorvat at some later time.

iosonofabio:
I'd be in favor of adding an `igraph_clear_graph` kind of thing that removes all edges and sets the number of boxes to zero.

szhorvat:
@vtraag Don't complicate it, just destroy and re-create the graph in-place.

Right now, we have:

```
int igraph_i_community_leiden_aggregate(const igraph_t *graph, igraph_t *aggregated_graph) {
    igraph_vector_t aggregated_edges;

    /* compute aggregated_edges */

    igraph_create(aggregated_graph, &aggregated_edges, nb_refined_clusters, IGRAPH_UNDIRECTED)
}
```

Change this to:

```
int igraph_i_community_leiden_aggregate(igraph_t *graph) {
    igraph_vector_t aggregated_edges;

    /* compute aggregated_edges from 'graph' */
    
    igraph_destroy(graph);
    igraph_create(graph, &aggregated_edges, nb_refined_clusters, IGRAPH_UNDIRECTED)
}
```

We don't need to try to gain additional efficiency, just re-shuffle the order of creation and destruction to make it compatible with how the finally stack works.

szhorvat:
@vtraag If we go with updating the graph in-place, the most straightforward implementation would need to start by copying the entire graph. Since this whole thing grew out of a user running into an out-of-memory condition, that makes me worried.

The current implementation will create a new graph after the first aggregation step anyway. Then the question is: how big is the size difference between the original graph, and the graph after the first aggregation step? (Sorry, I am not deeply familiar with the method.)

My current understanding is that after aggregation, the graph is likely to become several times smaller. Is this correct, and do you have an estimate of the size reduction? If it were to become 5 times smaller, than a 10 GB graph would currently require 10 + 10/5 = 12 GB of memory (not counting other things). If we copy the graph immediately, that means 10 + 10 = 20 GB of memory. That doesn't look good ...

Another thing to consider: is the algorithm fast enough to deal with huge graphs? In the original issue, the user was talking about *billions* of edges. Is there a chance that computation time will be the limiting factor anyway, and not memory?

vtraag:
> Don't complicate it, just destroy and re-create the graph in-place.

Yes, you are right, I can simply destroy and re-create the graph. 

> after aggregation, the graph is likely to become several times smaller.

The graph is indeed likely to become a couple of times smaller, especially in terms of edges, since those are most likely to be in between the nodes that are aggregated. I don't have exact statistics on that, but probably 4-5 times sounds reasonable.

> If we copy the graph immediately, that means 10 + 10 = 20 GB of memory. That doesn't look good ...

No, exactly, this is why I don't copy the graph to begin with. Although, to be fair, a `igraph_inclist_t` is being allocated which also takes considerable amount of memory. This touches upon the question of whether we may actually improve the current graph object (which we started to discuss on the [forum](https://igraph.discourse.group/t/igraph-edge-and-from-to-swapping/310)), instead of having this need to create fast iterators each time anyway.

> Another thing to consider: is the algorithm fast enough to deal with huge graphs? In the original issue, the user was talking about _billions_ of edges. Is there a chance that computation time will be the limiting factor anyway, and not memory?

We have run the algorithm on networks with up to about 1 billion edges (although fewer nodes), which takes up to an hour or so. I doubt that computation time will be the problem.

vtraag:
Thanks for the suggestion again @szhorvat, I believe the latest commit should solve all issues. In fact, the code only got better because of the stack constraint on the FINALLY handler, so the constraint may actually be a good one! :smile:

@munhyunsu perhaps you could try to run this PR against your network to see if all problems are then resolved? That is, you might still run out of memory of course, but there should then be some error provided (if it is not killed by an OOM killer first). To be clear, I didn't test anything yet against the network you provided.

I'm going on holidays as of tomorrow. I will be back in three weeks. I will see if things have been merged or if there are still some other remaining things that should be done. :parasol_on_ground:

szhorvat:
> I will see if things have been merged

So it's okay to merge before you come back if everything looks good and we get positive feedback from @munhyunsu?

szhorvat:
Enjoy the holidays! ☀️

vtraag:
> So it's okay to merge before you come back if everything looks good and we get positive feedback from @munhyunsu?

Yes, if think there are no more problems left, please feel free to merge at your discretion!

> Enjoy the holidays! :sunny:

Thanks!

Issue #1432, opened 2020-07-08
Graphicality for non-simple graphs


szhorvat:
`igraph_is_graphical_degree_sequence` checks if a degree sequence has a realization as a *simple* graph and `igraph_realize_degree_sequence` constructs it.

I am going to implement checking if there is a realization as a graph with/without self-loops and with/without multi-edges, and eventually also constructing these realizations (because I need it for my work).

I am looking for feedback on the API. We can:

 1. create new functions
 2. extend the existing `igraph_is_graphical_degree_sequence`

I prefer (2), as we can use a consistent extension to both `is_graphical_degree_sequence` and `realize_degree_sequence`.

If we go with (2), my first idea was to add two arguments, `igraph_bool_t multiple, igraph_bool_t loops`, similar to other existing functions. These control if multi-edges and self-loops are allowed respectively. This is not quite ideal because `multiple=false, loops=true` would naturally implies that while self-loops may be present, there may only be at most one. Implementing this case is straightforward enough, but it's kind of pointless (I do not expect many applications). The interesting cases are _loopless multigraphs_ and _loopy multigraphs_. Thus, the next possibility would be to have a single "type" variable with enum values such as `SIMPLE_GRAPH`, `LOOPLESS_MULTIGRAPH`, `LOOPY_MULTIGRAPH`. Not as pretty, but probably more aligned with actual use cases, and saves us implementing some not-so-useful cases. It would also allow specifying that we allow multiple self-loops, but we do not allow multiple edges between distinct vertices (which is probably not needed, but it is considered in papers occasionally). Yet another possibility is to go with the `multiple` and `loops` flags but simply return an `IGRAPH_UNIMPLEMENTED` for the missing cases.

So, the choices are:

 1. `multiple` and `loops` flags, complete implementation right away
 2. `multiple` and `loops` flags, partial implementation for now, return `IGRAPH_UNIMPLEMENTED` for the `multiple=false` and `loops=true` case.
 3. Single control variable with named options.

Do you have any preferences on which one to go with?

----

Note 1: We do already have `igraph_is_degree_sequence`, which is basically the loopy multigraph case. 

Note 2: Eventually, a similar extension could be implemented for `igraph_rewire` as well, but this is more work (especially the directed case).

ntamas:
[...]

PR #1441, opened 2020-07-17
Fixes igraph_is_graphical_degree_sequence for the undirected case

szhorvat:
The new undirected graphicality test based on Kiraly's paper was bugged, as I mentioned in https://github.com/igraph/igraph/issues/1432 This PR fixes the problem.

Changes:

 - fix `igraph_is_graphical_degree_sequence` for undirected case
 - explicitly check degree sequence for negative entries, and return "false" is any are found
 - add `igraph_vector_reverse_sort` (used in `igraph_is_graphical_degree_sequence`)
 - document `igraph_vector_reverse_sort` (we may remove the docs if we don't want to explicitly support a newly added function in a 0.8.x release)
 - remove claim that `igraph_vector_sort` was using `qsort` from the standard C library, as it is using `qsort.c` shipped with igraph


szhorvat:
The problems were:

 - must sort deg. seq. decreasingly
 - the code did not use the sorted version at all
 - must check that sum of degrees is even

szhorvat:
@ntamas Sorry about being a bit pushy, but the rest of PRs to come are condition on this. I'm happy to address even minor-sounding concerns. My priority is to correct the problem and move on.

ntamas:
Sorry about the delay, we went on a short holiday break with the family and this fell through the cracks somehow. I'm merging it soon.

szhorvat:
I messed up and duplicated one of the checks. Will send an extra PR soon.

szhorvat:
Fixed here 8373efcb1d91352ad8a48c099b158be8c0f860c3

Issue #1401, opened 2020-06-08
[Windows] failing tests because of EOL issues


jannick0:
I am compiling the igraph package on Windows using MSYS's MINGW64.  The package compiles fine, but there are two issues with the tests (i.e. when running `make check`):

1. Most of the tests fail because of EOL issues. This can be worked around by using the additional flag `--strip-trailing-cr` passed on to `diff` when running the tests.  This is a remedy for the situation where the tests create files with `CRLF` as EOL (unless the output stream mode is set to binary), while the original test files to be compared with have `LF` as EOL.
2. Given (1.) the remaining tests which fail are pajek-related which call `foreign.c`.  In this file `newline` is set to `\r\n` which on Windows is converted to `\r\r\n` (unless the output stream mode is set to binary which appears not to be the case in the tests at hand):  
https://github.com/igraph/igraph/blob/c8fc07f8e5f51e04dbe47a439792297efca2000d/src/foreign.c#L2094
Now step (1.) removes the latter `\r`, i.e. it converts `\r\r\n` back to `\r\n` which then creates a difference to the original test files due to the additional `\r`.

A potential way out of issue (2.) could be to define in `foreign.c`  `newline` as `\n` on Windows (when, e.g., the macro `_WIN32` is defined).

I did not go deeper into how EOLs are treated in the lexers on various platforms.  

Of course, happy to see other solutions for (1.) without tweaking `diff` flags when running `make check` on a Windows platform.

vtraag:
Thanks for the report! Could you please indicate whether you build `igraph` from the provided MSVC zip file, or whether you are building from the cloned GitHub repository? This might be relevant for issue (1).

jannick0:
I used [igraph-0.8.2.tar.gz](https://github.com/igraph/igraph/releases/download/0.8.2/igraph-0.8.2.tar.gz). Should I expect Windows-like EOLs in [igraph-0.8.2-msvc.zip](https://github.com/igraph/igraph/releases/download/0.8.2/igraph-0.8.2-msvc.zip)?  That would indeed help then, but I did not try since I was not aware of that if true.

vtraag:
OK, thanks!

It seemed somewhat strange to hear you are encountering line ending difficulties, because `igraph` is build in a [CI on Windows](https://ci.appveyor.com/project/ntamas/igraph) using `msys2`, and all tests run correctly. I thought that it might be an issue because of line endings being changed when cloning using `git`, but that is apparently not the case then.

[...]

PR #1453, opened 2020-08-01
autoconf: EOL issue on Windows (testsuite)

jannick0:
With this commit series all tests pass on my Windows 10 machine (using MSYS2 MINGW64). 
Please run testse on other platforms, too, on OSX in particular (where `\r` is special). Thanks.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -455,7 +455,28 @@ if test "$verify_finally" = "no"; then
   CPPFLAGS="${CPPFLAGS} -DIGRAPH_VERIFY_FINALLY_STACK=0"
 fi
 
-AC_CONFIG_FILES([Makefile src/Makefile igraph.pc igraph_Info.plist doc/Makefile include/igraph_version.h include/igraph_threading.h])
+AC_CONFIG_FILES([Makefile src/Makefile igraph.pc igraph_Info.plist doc/Makefile])
+
+# Generate header files (using script ./config.status), but update only upon changes.
+dnl NB: Approach requires AC_CONFIG_FILES to be called in configure.ac.
+AC_CONFIG_COMMANDS_POST([
+  for _h in include/igraph_version.h include/igraph_threading.h; do
+    AC_MSG_CHECKING([whether to generate header file $_h])
+    ./config.status --file=$_h.tmp:$_h.in > /dev/null 2>&1
+    if ! test -f $_h; then
+      mv $_h.tmp $_h
+      AC_MSG_RESULT([generated])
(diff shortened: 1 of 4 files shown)


szhorvat:
@ntamas Is it okay to use the Visual Studio 2019 image for AppVeyor? [It is claimed that it has a new (working) msys2.](https://help.appveyor.com/discussions/problems/27815-build-failing-due-to-key-error) I'm not so familiar with AppVeyor. We should somehow get it working again before merging this.

@jannick0 According to the Travis tests, it works on macOS and Linux.

ntamas:
Totally fine; I'll try to switch the Appveyor image and then rebuild the CI tests for this PR. It looks like there's quite a bit of arcane autotools magic in there - thanks for the great job @jannick0!

ntamas:
@szhorvat so it looks like we _are_ using the Visual Studio 2019 image already. I fixed things in `master` by importing the missing PGP key manually; it seems to work now.

I'll take a chance and merge this into `master` as is and see whether the CI checks pass there.

jannick0:
Great - I am glad that everything runs smoothly now. Thanks, guys! Happy to close #1401.

szhorvat:
I think we are having some problems with this on some more exotic platforms:

https://travis-ci.com/github/szhorvat/igraph/builds/178149534

jannick0:
Uuuh, let me have a look.

szhorvat:
@jannick0 The problem is that `fgetc` returns an `int` and `EOF` is generally an `int` not a `char`. We must change the type of `c` and `c_prev` into `int` in order to make it work.

szhorvat:
I pushed a fix, let's see if the tests pass.

jannick0:
OK - great. Thanks.  To be on the very safe side the while loop could be forced to be finite (e.g., using a counter).

szhorvat:
it passes now: https://travis-ci.com/github/szhorvat/igraph/builds/178154491



>  To be on the very safe side the while loop could be forced to be finite (e.g., using a counter).

I think it's fine with the `while (1)`. An `EOF` *must* come up eventually. The problem was that the value returned by `fgetc` was implicitly cast to a `char`. The program worked on platforms where `char(EOF) == EOF`, but it failed on those where this is not the case. *Why exactly* this comparison happened to be `true` on x86, I don't understand ... but with the correct type for `c` and `c_prev` everything should be fine on all platforms.

jannick0:
Yes, fully agree. Thanks!

PR #1465, opened 2020-09-14
Bliss: catch exceptions and convert to igraph error codes

### LINKED ISSUE #1178: Bliss errors should be translated to igraph error codes
Bliss errors (see `fatal_error` function) should be converted to igraph error codes.

Proposal: `fatal_error` should be modified to throw an exception, then that exception should be caught by all top-level igraph/Bliss functions. A similar use of exceptions is already planned in #1171 

`fatal_error`:

https://github.com/igraph/igraph/blob/master/src/bliss/defs.cc#L29

Uses of `fatal_error`:

https://github.com/igraph/igraph/blob/master/src/bliss/graph.cc#L45
https://github.com/igraph/igraph/blob/master/src/bliss/graph.cc#L696

See also checklist: https://github.com/igraph/igraph/issues/1177

Currently the error function prints to the console (should not happen) and terminates the process (should not happen). The R interface excludes this, so notify @gaborcsardi when this issue is resolved.

szhorvat:
This is meant to fix #1178, but I am not completely happy with it. Any comments, @ntamas?

Bliss itself does not seem to ever throw exceptions explicitly, but it may do so implicitly (e.g. out-of-memory condition).

The first commit simply removes/comments any printing to terminal (fprtintf) from the code.

### DIFF
--- a/src/bliss.cc
+++ b/src/bliss.cc
@@ -25,6 +25,7 @@
 #include "igraph_datatype.h"
 #include "igraph_interface.h"
 
+#include "igraph_handle_exceptions.h"
 
 using namespace bliss;
 using namespace std;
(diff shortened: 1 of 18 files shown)


szhorvat:
Just did a force-push. The change is that now all (hopefully) C++ code is wrapped in try ... catch blocks.

Also, one more commit to make some functions in bliss.cc static.

ntamas:
Seems okay to me, thanks!

PR #1467, opened 2020-09-15
Building shared library on Windows

vtraag:
These commits are an initial attempt at correctly building a shared library on Windows.

Note that there is one problem remaining. The compiled `igraph.dll` cannot be found when simply running `ctest`. When manually copying this file over to the proper test directory, all the tests do run correctly. This should of course not happen, but I don't immediately know how to best configure this in `CMake`. Any suggestions @ntamas ?

I wanted to check what you think of the current solution of how to enable/disable tests that rely on internal functions (i.e. which are not exported when building a shared library on Windows). Is this what you had in mind @ntamas? 

Of course, in the long term, we should correctly export symbols, and as said, we can also start doing this in `gcc` by using  [visibility markers](https://gcc.gnu.org/wiki/Visibility). Even then, we will of course not be able to test internal functions when building a shared library.

### DIFF
--- a/etc/cmake/FindGMP.cmake
+++ b/etc/cmake/FindGMP.cmake
@@ -9,7 +9,7 @@
 set(GMP_PREFIX "" CACHE PATH "Path to GMP prefix")
 
 find_path(GMP_INCLUDE_DIR gmp.h PATHS ${GMP_PREFIX}/include /usr/include /usr/local/include)
-find_library(GMP_LIBRARY libgmp.a PATHS ${GMP_PREFIX}/lib /usr/lib /usr/local/lib)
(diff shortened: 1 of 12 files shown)


ntamas (approved):

ntamas:
> The compiled igraph.dll cannot be found when simply running ctest. When manually copying this file over to the proper test directory, all the tests do run correctly.

I don't have a solution for this off the top of my head but I'll look around to see how this should be done with `ctest`. It might be the case that copying is the canonical solution in Windows unless there's a way to specify an additional directory to look for DLLs apart from the current directory and the standard system paths.

> I wanted to check what you think of the current solution of how to enable/disable tests that rely on internal functions

Seems okay to me! I'll merge this.

PR #1472, opened 2020-09-24
Further correcting tests on Windows

vtraag:
Below is the best way I could find to copy over the necessary DLLs for testing. It relies on `GET_RUNTIME_DEPENDENCIES` to get all runtime dependencies, but I exclude all dependencies that should be in the common path. This problem is also discussed on the [CMake Discourse forum](https://discourse.cmake.org/t/get-runtime-dependencies-has-difficulties-with-windows-api-sets/1768/8).

For some reason I get a missing error dependency on `vcruntime14_1d.dll`, even though it simply exists in my path in `C:\Windows\system32`. I am not sure why this is happening, but I now excluded it in order to not trigger a warning.

There is one other drawback: this is executed for each test. This is the only way I could think of that ensures that all necessary files are copied over, even if only a single test is being build. It should not physically copy the files again if the files are already present, if I am correct. However, it does trace the dependencies of `igraph.dll` each time. This makes it a bit slower, although I have not noticed it in practice (the build is taking long enough anyway).

If you have any other recommendations or tips on how to solve this, happy to hear from you!


vtraag:
Sorry, for some reason I no longer had [this](https://cmake.org/pipermail/cmake/2015-November/061978.html) solution in mind about just adapting the path. I have now done that, this is indeed a much simpler setup. This is now simplified in the latest commit.

ntamas:
Trying to pick up this thread now -- is this ready for review or are you still working on it?

vtraag:
Probably better to wait for it a little bit more. I will ping you when ready.

vtraag:
Yes, good to go. There are essentially three things here:

1. Correct RPATH setting for running tests on recent `gcc` versions (necessary because of `--enable-new-dtags` is set by default).
2. Distinguish internal tests (i.e. those which use functions that are not exported when building a shared library) from other tests. These internal tests are disable when building a shared library on Windows.
3. Export warning/error handler when building a shared library.

ntamas:
I have seen that you have consistently changed this:

```
if (BUILD_SHARED_LIBS)
  set_property(TARGET bliss PROPERTY POSITION_INDEPENDENT_CODE ON)
endif()
```

to this:

```
set_property(TARGET bliss PROPERTY POSITION_INDEPENDENT_CODE ${BUILD_SHARED_LIBS})
```

There might be a bit of semantic difference between the two; in the original version, `BUILD_SHARED_LIBS` implies `POSITION_INDEPENDENT_CODE` being set to `ON`, but it leaves it at the default setting if `BUILD_SHARED_LIBS` is `OFF`. In the new version, setting `BUILD_SHARED_LIBS` to `OFF` forces `POSITION_INDEPENDENT_CODE` to be `OFF` as well.

Did you have specific problems that were solved by this change or is this purely a cosmetic one?

For what it's worth, the CMake documentation says this:

> The `POSITION_INDEPENDENT_CODE` property determines whether position independent executables or shared libraries will be created. This property is True by default for `SHARED` and `MODULE` library targets and False otherwise.

If this is true, then there is probably no need to force it to `OFF` explicitly, as it would prevent the user from manually setting the global `CMAKE_POSITION_INDEPENDENT_CODE` to `ON` to force all targets to be built with `-fPIC` even for static libs (as we set it back to `OFF` afterwards).

etc/cmake/test_helpers.cmake
```diff
@@ -53,6 +53,8 @@ function(add_legacy_test FOLDER NAME)
     )
     set_property(TEST ${NAME} PROPERTY SKIP_RETURN_CODE 77)
   endif()
+  SET(IGRAPH_LIBDIR $<TARGET_FILE_DIR:igraph>)
```
ntamas, line 56: Please add a comment here explaining why these two lines are needed. I remember now as we've discussed it only a few weeks ago, but I will definitely forget this in a few months :)

Also, maybe it's worth putting this in an `IF(WIN32)` clause? We are messing around with the system path here, so it's probably worth limiting its impact as much as possible (and we don't need this on Linux or macOS).

ntamas:
Feel free to merge this after we've discussed the two points I've raised above.

vtraag:
> Did you have specific problems that were solved by this change or is this purely a cosmetic one?

This was purely cosmetic. I did not consider it would also change the default setting, perhaps indeed better to stick to the previous formulation then.

PR #1476, opened 2020-09-25
community_leiden: corrected self-loops

### LINKED ISSUE #1474: eigenvector_centrality: another wrong result
Consider this graph:

image

The correct eigenvector centralities are 

```
{1., 0.481194, 0.193937}
```

This is what I get in the unweighted case. However, if I assign weight 1 to each edge, I get:

```
{1., 0.249141, 0.0586332}
```

This is not correct.

The reason for this behaviour is that `igraph_inclist_remove_duplicate` fails to remove duplicate edges in this case. It assumes that the inclist is sorted and it removed only consecutive duplicates. However, this inclist is not sorted. It looks like this:

```
1 0 1 0 2 
2 3 
3 
```

`1` and `0` are the edge indices of the two self-loops.

### LINKED ISSUE #1475: Different outputs between igraph.community_leiden and leidenalg with self loops
Make sure that these boxes are checked before submitting your issue -- thank you!

- [x] This issue is for the Python interface of igraph.

Hi,
I have recently discovered that the leiden clustering algorithm was included in python-igraph, and this is very helpful for me. It mostly produces very similar results as the `leidenalg` package. However, there seems to be a systematic difference between the 2 packages when vertices have self-edges. Here's an example:

```python
import igraph # v0.8.2
import leidenalg #  v0.8.2


def get_graph():
    g = igraph.Graph()
    g.add_vertices(['V0', 'V1', 'V2', 'V3'])
    g.add_edges([('V0', 'V1'), ('V2', 'V3')])

    return g

def print_igraph_clusters(g):
    clusters = g.community_leiden('CPM', resolution_parameter=0.1)
    print([cluster for cluster in clusters])

def print_leiden_clusters(g):
    clusters = leidenalg.find_partition(
        g, leidenalg.CPMVertexPartition, resolution_parameter=0.1
    )
    print([cluster for cluster in clusters])

> g = get_graph()
# same output for both
> print_igraph_clusters(g)
[[0, 1], [2, 3]]
> print_leiden_clusters(g)
[[0, 1], [2, 3]]

> g.add_edges([(f'V{i}', f'V{i}') for i in range(4)])
# igraph produces singleton clusters while leiden produce the same output as before 
> print_igraph_clusters(g)
[[0], [1], [2], [3]]
> print_leiden_clusters(g)
[[0, 1], [2, 3]]
```

vtraag:
This fixes the problem that was identified in #1475. The problem was that self-loops were not consistently being ignored. This PR fixes #1475 by ignoring self-loops consistently throughout.


ntamas:
It looks like the tests also need an update, don’t they?

szhorvat:
Can you add something to the changelog to explain what changed?

vtraag:
Right, I didn't realise there were examples with self-loops included. I'll correct that early next week. I'll also update the CHANGELOG.

szhorvat:
I change the bug number on the "fixes" line, I think you meant to say 1475 but wrote 1474 (which will trigger auto-closing that bug once the PR is merged).

Issue #797, opened 2014-12-30
Add delimiter keyword to write_edgelist


macks22:
Currently spaces are used. It would be useful to be able to write a standard csv or tsv file, for instance.

iosonofabio:
I can't seem to find the function `write_edgelist` in the C library. Maybe R specific? If so should we move the issue to that repo?

szhorvat:
@iosonofabio I beliebe his refers to `igraph_write_graph_edgelist` and `igraph_read_graph_edgelist`.

carissableker:
I would appreciate if the same option could be added to `igraph_write_graph_ncol` as well.

sajaddp:
Hi,
Is there a plan for this issue after six years?

iosonofabio:
haha @sajaddp there is a plan! It's a simple fix, so feel free to make a PR for it. Otherwise I'll try to find some time in the next days and it'll make it for the next release

PR #1478, opened 2020-10-02
Delimiter write edgelist

iosonofabio:
This is an attempt at enabling a custom delimiter when reading/writing edgelists from/to text files (#797 )

It currently breaks the API because I am not sure as of how to handle default arguments of this kind.

Anyone can advise on this one? Shall we just break and **require** a `del` argument from 0.9, or how else would you handle this kind of trivial but API-breaking fix?

**NOTE**: I haven't changed the tests yet, so they will probably fail.

### DIFF
--- a/include/igraph_foreign.h
+++ b/include/igraph_foreign.h
@@ -39,7 +39,8 @@ __BEGIN_DECLS
 /* -------------------------------------------------- */
 
 DECLDIR int igraph_read_graph_edgelist(igraph_t *graph, FILE *instream,
-                                       igraph_integer_t n, igraph_bool_t directed);
+                                       igraph_integer_t n, igraph_bool_t directed,
+                                       char del);
 DECLDIR int igraph_read_graph_ncol(igraph_t *graph, FILE *instream,
                                    igraph_strvector_t *predefnames, igraph_bool_t names,
                                    igraph_add_weights_t weights, igraph_bool_t directed);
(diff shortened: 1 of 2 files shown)


ntamas:
We will break the API several times in 0.9 so I don't mind adding an extra argument to `igraph_write_graph_edgelist()` and `igraph_read_graph_edgelist()`.

However, note that `igraph_read_graph_edgelist()` used `isspace()` for comparisons because `isspace()` also matches newline characters -- so, it was only recommended (but not required) to use one edge per line. The user was free to put all edges on the same line. This won't work after your modifications, and you would need an extra piece of code to look for a newline after every second vertex ID.

I think that since `igraph_read_graph_edgelist()` does not mandate the usage of one edge per line, a better place for this functionality would be a function like `igraph_read_graph_edgelist_with_separators()` (but let's find a better/shorter name for that), which could be organized similarly to `igraph_read_graph_ncol()` (which is a terrible name because most people don't realize that this is essentially an edge list where the vertex names can also be strings). So, one edge per line, but the separator could be chosen arbitrarily.

ntamas:
_Footnote_: I can foresee that once we added this feature, people would start reading edge lists in CSV files with it, only to realize that quotes are not handled properly so they cannot have vertices with commas in their names. The next step would then be to add a fully fledged CSV parser :) I was wondering whether we could anticipate that and create a function that takes a delimiter character _and_ a quote character as well.

pradkrish:
It makes sense to have a function that can read edges(one per line) from a file with common delimiters(comma, semicolon, pipe) and vice-versa write to a file with delimiters. Is there still an interest in implementing this? I can take it up after my current PR is accepted.

Issue #1484, opened 2020-10-21
Mismatching FINALLY calls in igraph_i_cattribute_combine_vertices


vtraag:
[This issue](https://igraph.discourse.group/t/help-error-again-assertion-no-100-failed-in-igraph-0-8-3/487/6) on the igraph forum revealed that there was a mismatch of FINALLY calls. The problem is revealed by running the following code with `igraph` compiled with `--enable-verify-finally`.

<details>
  <summary>Code to reproduce error</summary>

```C
#include <iostream>
#define IGRAPH_STATIC 1
#include "igraph.h"
#include <vector>
#include <memory>

struct Dmatch{
	Dmatch(int query_Idx, int train_Idx) { queryIdx = query_Idx, trainIdx = train_Idx; }
	int trainIdx;
	int queryIdx;
};
bool extend1to2(igraph_t& sourceGraph, igraph_t& extendGraph, std::vector <Dmatch>& matches) {

	int n_source_vertices = GAN(&sourceGraph, "n_vertices");
	int n_exd_vertices = GAN(&extendGraph, "n_vertices");
	std::unique_ptr<igraph_vector_t, void(*)(igraph_vector_t*)>
		source_label(new igraph_vector_t(), &igraph_vector_destroy),
		exd_label(new igraph_vector_t(), &igraph_vector_destroy),
		source_edge_weight(new igraph_vector_t(), &igraph_vector_destroy),
		exd_edge_weight(new igraph_vector_t(), &igraph_vector_destroy);

	igraph_vector_init(source_label.get(), 0);
	igraph_vector_init(exd_label.get(), 0);
	igraph_vector_init(source_edge_weight.get(), 0);
	igraph_vector_init(exd_edge_weight.get(), 0);
	VANV(&sourceGraph, "label", source_label.get());
	VANV(&extendGraph, "label", exd_label.get());
	EANV(&sourceGraph, "weight", source_edge_weight.get());
	EANV(&extendGraph, "weight", exd_edge_weight.get());

	igraph_t union_graph;
	igraph_disjoint_union(&union_graph, &sourceGraph, &extendGraph);
	igraph_vector_append(source_label.get(), exd_label.get());
	igraph_vector_append(source_edge_weight.get(), exd_edge_weight.get());
	SETVANV(&union_graph, "label", source_label.get());
	SETEANV(&union_graph, "weight", source_edge_weight.get());
	igraph_vector_t	contract_vertices_list;
[...]

PR #1485, opened 2020-10-21
igraph_i_cattribute_combine_vertices: correct mismatching FINALLY calls.

vtraag:
This fixes the problem identified in #1484 of mismatching FINALLY calls in igraph_i_cattribute_combine_vertices.

### DIFF
--- a/src/cattributes.c
+++ b/src/cattributes.c
@@ -1577,7 +1577,8 @@ int igraph_i_cattribute_combine_vertices(const igraph_t *graph,
 
     igraph_free(funcs);
     igraph_free(TODO);
-    IGRAPH_FINALLY_CLEAN(2);
+    igraph_i_cattribute_permute_free(new_val);
+    IGRAPH_FINALLY_CLEAN(3);
 
     return 0;
 }


szhorvat:
Looks good, I think you should merge. We should add a test which properly exercises these functions ...  

(I should note that I never used igraph's attribute handlers. The Mma interface doesn't use them.)

ntamas (approved):

ntamas:
LGTM

The attribute handlers are exercised heavily from the Python interface, but bugs like these are hard to spot as you need to repeat the buggy operation many times to reach the size limit of the finally stack. Anyway, I should have noticed this had I queried the size of the "finally" stack at the end of the Python test suite; during normal operation, the "finally" stack must always be empty when the execution flow gets back to Python's REPL loop.

It might be a good idea to expose a function that reports the size of the "finally" stack somewhere hidden in the API of the higher-level interfaces so test suites can use them.

szhorvat:
@ntamas We already have `--enable-verify-finally`, it could be used for the Python tests as well.

ntamas:
Just realized that the bug here was in the _C_ attribute handler, which is not used by the Python interface. So yes, @vtraag is right, we need more tests for this part of the code.

vtraag:
I might have been too quick with this fix. It seems that instead that the 

```
    igraph_i_cattribute_permute_free(new_val);
```

was missing, the 

```
IGRAPH_FINALLY(igraph_i_cattribute_permute_free, new_val);
```

was uncalled for. By freeing `new_val`, I think the new values of the combined attributes would actually be freed, which should not be the intention. Hence, this should also not have been pushed to the FINALLY stack. What do you think?

See https://github.com/vtraag/igraph/commit/d94528170823b5253beb8c3b691e0934502ebf32 for a proper fix. I have now also asked the OP to test that fix, we should probably also wait until (s)he confirms.

vtraag:
The OP also confirms this is the correct fix. I think we should merge the proper fix. It seems that the same pattern (pushing `permute_free` to the FINALLY stack) is also happening in other places in `cattributes.c`. I will have a look at those other instances later today.

ntamas:
Okay - will you send another PR later today when you've had the chance to take a look at all the other instances, or shall I just fix it now on my own?

szhorvat:
A question: would adding `VERIFY_FINALLY_STACK()` to the end of the C attribute handler tests have revealed any of these issues? We should really take the time to separate tests from usage examples, move the tests to the test directory, and add  `VERIFY_FINALLY_STACK()` to the end of each test. But only after the cmake branch is merged, otherwise we just create extra work.

vtraag:
Just created a new PR (#1486). I think you are right @szhorvat : we should indeed clearly separate tests from examples.

PR #1487, opened 2020-10-22
igraph_blas_internal.h: add missing defines

szhorvat:
Fixes building with internal LAPACK, external BLAS.

When trying to build the tests, I got: 

```
Undefined symbols for architecture x86_64:
  "_igraphdsymv_", referenced from:
      _dlatrd_ in libigraph.a(dlatrd.c.o)
      _dsytd2_ in libigraph.a(dsytd2.c.o)
  "_igraphdsyr2_", referenced from:
      _dsytd2_ in libigraph.a(dsytd2.c.o)
  "_igraphdsyr2k_", referenced from:
      _dsytrd_ in libigraph.a(dsytrd.c.o)
  "_igraphdtrmv_", referenced from:
      _dlahr2_ in libigraph.a(dlahr2.c.o)
      _dlarft_ in libigraph.a(dlarft.c.o)
ld: symbol(s) not found for architecture x86_64
```

This fixes the problem, at least for the static build.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 ### Fixed
 
  - `igraph_i_cattribute_combine_vertices()`: fixed invalid cleanup code that eventually filled up the "finally" stack when combining vertices with attributes extensively.
+ - Fixed building with vendored LAPACK and external BLAS
 
 ### Other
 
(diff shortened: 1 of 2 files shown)


ntamas (approved):

vtraag:
Was this really intended for `master`? Shouldn't this have been targeted at the `feature/cmake-build` branch?

ntamas:
I think this affects the non-CMake build as well. At least that was my understanding.

szhorvat:
Yes, it was for master as well. The problem was that `--with-external-blas --with-external-lapack` worked, but `--with-external-blas` (and internal lapack) did not work. 

However, I only tested on the cmake branch, which has unrelated problems. A few more things need to be fixed. I'll push those now.

PR #1488, opened 2020-10-23
Further fixes for building with external blas, internal lapack

szhorvat:
More fixes for building with external BLAS, internal LAPACK.

Changes:

 - When building internal LAPACK, do not also include the objects belonging to internal BLAS (internal LAPACK should be able to use external BLAS)
 - `xerbla.c` and `len_trim.c` belong to LAPACK, I believe, not to BLAS, so move them there in CMakeLists.txt
 - `igraph_lapack_ddot` is renamed to `igraph_blas_ddot` and moved to `igraph_blas.c`/`igraph_blas.h`, as this is a BLAS function, not a LAPACK one. This function was exposed publicly, so this is a breaking change.
 - in `igraph_blas_internal.h`, give a prototype to `igraphddot_` (which may be changed by a `#define` to `ddot_`) so that there would be no complaints about missing prototypes. There are already other similar prototypes there.
 - `igraph_blas_internal.h` can now be removed from `lapack.c` (I added it yesterday in https://github.com/igraph/igraph/commit/dada582ba62a39b9dca49eb4566b91ced49ad9b8, but the better fix is to move the ddot stuff to blas.c instead as noted above)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@
  - `cutoff=0` is no longer interpreted as infinity (i.e. no cutoff) in `betweenness`, `edge_betweenness` and `closeness`. If no cutoff is desired, use a negative value such as `cutoff=-1`.
  - `igraph_realize_degree_sequence()` has an additional argument controlling whether multi-edges or self-loops are allowed.
  - `igraph_is_degree_sequence()` and `igraph_is_graphical_degree_sequence()` are deprecated in favour of the newly added `igraph_is_graphical()`.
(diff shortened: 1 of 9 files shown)


ntamas:
Thanks for the cleanup! I think it's all good, with one exception: `len_trim` is not BLAS but a [Fortran intrinsic](https://gcc.gnu.org/onlinedocs/gfortran/LEN_005fTRIM.html). If we want to be consistent, we should probably move it to `src/fortran_intrinsics.c` instead.

ntamas (approved):

Issue #1492, opened 2020-10-23
CMake with external SuiteSparse on Windows


vtraag:
I thought it would be more convenient to keep track of the separate issues for the CMake build, and only keep the overview of all CMake related issues in #1490. This issue is to keep track of the CMake build using external SuiteSparse libraries.

It is currently possible to build with the external SuiteSparse provided by `vcpkg`. However, there is one strange problem: for some reason, `libcxsparse.dll` is not copied to the `src\Debug` directory, as is the case for other settings. I am not sure why this is not done. After copying `libcxsparse.dll` over manually, almost all tests pass except `igraph_sparsemat8`.

![image](https://user-images.githubusercontent.com/6057804/97050508-c8451d00-157d-11eb-941c-c2f0eb671a10.png)

The problem arise at 

https://github.com/igraph/igraph/blob/0ea2c8ddebd34d9f910d7b4e18568181209f48ee/examples/simple/igraph_sparsemat8.c#L193

I don't understand what happens however when debugging the problem. Before it enters `igraph_sparsemat_add_cols`, `spmat.cs` exists, but after it enters `A->cs` does not exists, only `A->cs_di` exists. It then crashes on this line:

https://github.com/igraph/igraph/blob/0ea2c8ddebd34d9f910d7b4e18568181209f48ee/src/sparsemat.c#L2737

The memory addresses of `spmat.cs` and `A->cs_di` are identical. I have no idea what is happening here.

vtraag:
I think that the problem might be that `realloc` is reallocating memory (`A->cs->p`) that is previously allocated by `cxsparse`. In principle, both libraries should be using the same runtime, so that the memory can be reallocated across libraries. It does make it a bit fragile however if there is some mismatch in runtimes. Is there some way to let the `realloc` be invocated by `cxsparse` instead of by `igraph` (I couldn't find much documentation of `cxsparse`)? I will keep looking into what exactly creates the (presumable) mismatch in runtimes.

vtraag:
Actually, the problem was that I copied the incorrect DLL, namely the one that was not built with debug symbols (`/MD` instead of `/MDd`), hence the mismatch in runtimes. Nonetheless, I think that switching to `cs_realloc`, as proposed in #1500 will be an improvement. Generally, I think it would be good to not mix free/alloc calls across libraries.

vtraag:
[...]

PR #1500, opened 2020-10-24
Use realloc from CXSparse

vtraag:
This PR changes `igraph_sparsemat_add_cols` to use `cs_realloc` to ensure that the reallocation is happening from the same runtime from `CXSparse`. This ensures that the reallocation is done with the same runtime that originally allocated the memory. If this is not done, this might lead to memory problem such as noted in #1492, which is fixed by this PR.


szhorvat:
Looks good.

PR #1502, opened 2020-10-25
DECLDIR cleanup

szhorvat:
Justification:

 - `_WIN32` is defined both on 32-bit and 64-bit Windows. It is define by the compiler itself, and does not depend on the inclusion of any header. It appears to me that it is the correct way to test for Windows. Reference: https://docs.microsoft.com/en-us/cpp/preprocessor/predefined-macros?view=vs-2019

 - `WIN32` is not a standard macro (it may be defined by some headers though). We don't need it. Reference: see above and https://stackoverflow.com/q/9025708/695132 and https://stackoverflow.com/q/662084/695132 

 - `declspec` is supported by MINGW, so no need to exclude it from MINGW. Reference: https://stackoverflow.com/q/22285240/695132

 - Cygwin is pretending to be Unix as much as possible, and does not actually define `_WIN32`, unless using the `-mwin32` option. However, there are suggestion that in the past it did, so it is better to keep explicitly excluding it. Reference: https://stackoverflow.com/q/47150492/695132 and https://cygwin.com/faq/faq.html#faq.programming.preprocessor

 - While `__CYGWIN32__` exists (both on 32- and 64-bit), the Cygwin documentation suggests using `__CYGWIN__`.  Reference: https://cygwin.com/faq/faq.html#faq.programming.preprocessor

### DIFF
--- a/include/igraph_decls.h
+++ b/include/igraph_decls.h
@@ -8,9 +8,12 @@
     #define __END_DECLS /* empty */
 #endif
 
+/* _WIN32 is always defined on Windows (both 32- and 64-bit systems).
+ * It is defined by the compiler itself, thus it does not depend on the inclusion of headers.
+ * Reference: https://docs.microsoft.com/en-us/cpp/preprocessor/predefined-macros */
 #undef DECLDIR
-#if defined (_WIN32) || defined (WIN32) || defined (_WIN64) || defined (WIN64)
-    #if defined (__MINGW32__) || defined (__CYGWIN32__)
+#if defined (_WIN32)
+    #if defined (__CYGWIN__)
         #define DECLDIR /**/
     #else
         #ifdef IGRAPH_EXPORTS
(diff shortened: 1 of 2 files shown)


szhorvat:
Note: this is not yet properly tested!

vtraag:
In other cases this seemed insufficiently general: https://igraph.discourse.group/t/cmake-on-windows/298/42?u=vtraag. I'll have a look though.

szhorvat:
Do you have an example where `_WIN32` was not defined on Windows (not Cygwin!)? The official documentation says that it should always be defined by the compiler itself.

szhorvat:
My motivation for this change was to go by something that is officially documented. I added a comment pointing to the documentation, to avoid future confusion. 

There is a lot of confusion out there. You'll find all sort of inconsistent attempts to check for Windows in various projects. I want to avoid this, and want to avoid relying on guesswork. I'd like to use the official way, and be confident about why we are doing it this way.

If we include other checks, we should know *why* and *when* they are needed (according to the MS documentation, they are not needed).

PR #1508, opened 2020-10-27
Properly fix path problems with CTest on MinGW

### LINKED ISSUE #1501: CMake: when building a shared library, tests don't run on MSYS2
When building a shared library, tests don't run on MSYS2 due to `libigraph.dll` not being present on the `PATH`.

@vtraag I thought you mentioned fixing this recently, but maybe I misremembered.

vtraag:
This is a continuation of #1504. There were some further problems with paths on MinGW, see https://github.com/igraph/igraph/issues/1501#issuecomment-717322063 for some more details. This now finally properly fixes #1501.

### DIFF
--- a/etc/cmake/test_helpers.cmake
+++ b/etc/cmake/test_helpers.cmake
@@ -58,7 +58,9 @@ function(add_legacy_test FOLDER NAME)
     # add the dir that contains the built igraph.dll to the path environment variable
     # so that igraph.dll is found when running the tests.
     SET(IGRAPH_LIBDIR $<TARGET_FILE_DIR:igraph>)
-    SET_TESTS_PROPERTIES( ${NAME} PROPERTIES ENVIRONMENT "PATH=${IGRAPH_LIBDIR};$ENV{PATH}" )
+    string(REPLACE "/" "\\" IGRAPH_LIBDIR ${IGRAPH_LIBDIR})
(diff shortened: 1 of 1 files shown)


vtraag:
Sorry, I now realise that we should document this monstrosity properly. I unfortunately don't have time now anymore, I have to go.

ntamas (approved):

ntamas:
Let's wait for the CI tests; if they pass, I'm okay to merge this.

vtraag:
I now tried to rerun this locally in MinGW to double check something. I rebased against the latest `cmake` branch before opening this PR (without testing actually), but I am now running into several errors similar to the following:

```
C:/Users/traagva1/src/igraph/src/adjlist.c:95:5: error: 'igraph_adjlist_init' redeclared without dllimport attribute: previous dllimport ignored [-Werror=attributes]
   95 | int igraph_adjlist_init(const igraph_t *graph, igraph_adjlist_t *al,
      |     ^~~~~~~~~~~~~~~~~~~
```

I don't know immediately what commit caused this, I think this might have been a warning instead of an error earlier.

ntamas:
This is probably unrelated so let's keep track of it in a separate issue.

The doc generation fails now, but that's also another issue, I'll fix it soon.

PR #1509, opened 2020-10-27
cmake: Condition warning options on C_COMPILER_ID

szhorvat:
This PR makes all warning options conditional on the compiler. GCC,Clang,AppleClang are included, as these are what we use, and these usually take the same options.

These options are not actually needed for compilation. They are just a nice to have during development. They alert us to problems.

However, they will also prevent other compilers from working:

 - Compilation failed with the Intel compiler, so options had to be adjusted for that one a bit. Removing them entirely would have worked as well, but adjusting them made it easier for us to make use of the Intel compiler's warning. 

 - Compilation also failed with the PGI compiler. With this PR, it passes (after adjusting the source code a bit, but that's a separate thing). Of course, it shows lots of warning, but igraph does compiler, and the tests pass, and that's what matters.

### DIFF
--- a/etc/cmake/compilers.cmake
+++ b/etc/cmake/compilers.cmake
@@ -1,11 +1,5 @@
 if(MSVC)
   add_compile_options(/FS)
-else()
-  add_compile_options(
-    # Intel compiler:
-    # Do not print remarks:
-    $<$<C_COMPILER_ID:Intel>:-diag-disable=remark>
-  )
 endif()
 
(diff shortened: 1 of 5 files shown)


szhorvat:
There is one additional change here: adding `include(compilers)` to the vendored GLPK, so that any globally set compiler options take effect for it too.

ntamas (approved):

ntamas:
Ready to merge if the CI build passes.

PR #1511, opened 2020-10-27
to_directed: add RANDOM and ACYCLIC modes

### LINKED ISSUE #1270: Feature request: as.directed random mode
Arbitrary may be useful but randomly assigning an edge direction (or two edges?) may also be useful.

Arbitrary is clearly non-random:

    g <- make_ring(10)
    as.directed(g, "arbitrary") %>% plot

arbitrary_as_directed

szhorvat:
I don't use this function in the Mma interface, so it would be nice to integrate it into Python for easier testing.

Fixes #1270

### DIFF
--- a/examples/tests/igraph_to_directed.c
+++ b/examples/tests/igraph_to_directed.c
@@ -1,0 +1,43 @@
+
+#include <igraph.h>
+
+#include "test_utilities.inc"
+
+int main() {
+    igraph_t ug;
+    igraph_t dg;
+
+    igraph_small(&ug, 6, /* directed= */ 0,
+                 2,0, 1,4, 3,2, 2,3, 0,4, 5,0, 2,3, 5,3, 2,5, 0,1,
+                 -1);
+
+    igraph_copy(&dg, &ug);
+    printf("\nARBITRARY:\n");
+    igraph_to_directed(&dg, IGRAPH_TO_DIRECTED_ARBITRARY);
+    print_graph(&dg, stdout);    
+    igraph_destroy(&dg);
+
+    igraph_copy(&dg, &ug);
+    printf("\nMUTUAL:\n");
+    igraph_to_directed(&dg, IGRAPH_TO_DIRECTED_MUTUAL);
+    print_graph(&dg, stdout);
+    igraph_destroy(&dg);
+
+    igraph_copy(&dg, &ug);
+    printf("\nACYCLIC:\n");
+    igraph_to_directed(&dg, IGRAPH_TO_DIRECTED_ACYCLIC);
+    print_graph(&dg, stdout);
+    igraph_destroy(&dg);
+
+    igraph_copy(&dg, &ug);
+    printf("\nRANDOM (edge count only):\n");
+    igraph_to_directed(&dg, IGRAPH_TO_DIRECTED_RANDOM);
+    printf("%d\n", igraph_ecount(&dg));
+    igraph_destroy(&dg);
+
+    igraph_destroy(&ug);
+
+    VERIFY_FINALLY_STACK();
+
+    return 0;
+}
--- a/src/conversion.c
+++ b/src/conversion.c
@@ -373,15 +373,7 @@ int igraph_to_directed(igraph_t *graph,
         IGRAPH_VECTOR_INIT_FINALLY(&edges, size);
         IGRAPH_CHECK(igraph_get_edgelist(graph, &edges, 0));
 
-        if (mode == IGRAPH_TO_DIRECTED_ACYCLIC) {
-            for (i=0; i < no_of_edges; ++i) {
-                if (VECTOR(edges)[2*i] > VECTOR(edges)[2*i+1]) {
-                    igraph_real_t temp = VECTOR(edges)[2*i];
-                    VECTOR(edges)[2*i] = VECTOR(edges)[2*i+1];
-                    VECTOR(edges)[2*i+1] = temp;
-                }
-            }
-        } else if (mode == IGRAPH_TO_DIRECTED_RANDOM) {
+        if (mode == IGRAPH_TO_DIRECTED_RANDOM) {
             RNG_BEGIN();
 
             for (i=0; i < no_of_edges; ++i) {
@@ -393,6 +385,21 @@ int igraph_to_directed(igraph_t *graph,
             }
 
             RNG_END();
+        } else if (mode == IGRAPH_TO_DIRECTED_ACYCLIC) {
+            /* Currently, the endpoints of undirected edges are ordered in the
+               internal graph datastructure, i.e. it is always true that from < to.
+               However, it is not guaranteed that this will not be changed in
+               the future, and this ordering should not be relied on outside of
+               the implementation of the minimal API in type_indexededgelist.c.
+
+               Therefore, we order the edge endpoints anyway in the following loop: */
+            for (i=0; i < no_of_edges; ++i) {
+                if (VECTOR(edges)[2*i] > VECTOR(edges)[2*i+1]) {
+                    igraph_real_t temp = VECTOR(edges)[2*i];
+                    VECTOR(edges)[2*i] = VECTOR(edges)[2*i+1];
+                    VECTOR(edges)[2*i+1] = temp;
+                }
+            }
         }
 
         IGRAPH_CHECK(igraph_create(&newgraph, &edges,
--- a/src/conversion.c
+++ b/src/conversion.c
@@ -368,11 +368,12 @@ int igraph_to_directed(igraph_t *graph,
         igraph_t newgraph;
         igraph_vector_t edges;
         long int size = no_of_edges * 2;
+        long int i;
+
         IGRAPH_VECTOR_INIT_FINALLY(&edges, size);
         IGRAPH_CHECK(igraph_get_edgelist(graph, &edges, 0));
 
         if (mode == IGRAPH_TO_DIRECTED_ACYCLIC) {
-            long int i;
             for (i=0; i < no_of_edges; ++i) {
                 if (VECTOR(edges)[2*i] > VECTOR(edges)[2*i+1]) {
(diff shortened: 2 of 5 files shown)


src/conversion.c
```diff
@@ -341,26 +353,49 @@ int igraph_get_edgelist(const igraph_t *graph, igraph_vector_t *res, igraph_bool
...
         igraph_t newgraph;
         igraph_vector_t edges;
-        long int no_of_edges = igraph_ecount(graph);
-        long int no_of_nodes = igraph_vcount(graph);
         long int size = no_of_edges * 2;
         IGRAPH_VECTOR_INIT_FINALLY(&edges, size);
         IGRAPH_CHECK(igraph_get_edgelist(graph, &edges, 0));
 
+        if (mode == IGRAPH_TO_DIRECTED_ACYCLIC) {
+            long int i;
```
ntamas, line 375: Basically all branches seem to have an `i` counter variable. Can we lift the declaration to the top-level?

szhorvat: OK, done that. Lots of variables could be shared between the two branches of the `switch` as well (originally an if-else). I left those mostly as they were before, and only lifted the vertex and edge-count variables. The reason is that if we add more methods in the future, not everything may be shared between them.

szhorvat:
It appears that when I create an undirected graph, the 'from' vertex will always be the smaller index one, and the 'to' the larger index one. I.e., `1,0,  1,2,  3,0` is transformed to `0,1,  1,2,  0,3`.

@ntamas @gaborcsardi Do you know if this ordering is guaranteed or not?

Also pinging @vtraag, as you've been looking into the `igraph_t` structure recently.

When the graph construction goes through `igraph_add_edges`, then endpoints of undirected edges will be ordered ("canonicalized") like this:

https://github.com/igraph/igraph/blob/master/src/type_indexededgelist.c#L277

However, I do not know if _all_ construction paths go through here, and if there are an modification paths that may violate it.

The comment here suggests that the canonicalization of undirected edges is not guaranteed:

https://github.com/igraph/igraph/blob/master/include/igraph_datatype.h#L60

However, this code suggests that it *is* guaranteed after all:

https://github.com/igraph/igraph/blob/master/src/type_indexededgelist.c#L1110

----

In this PR, the `ACYCLIC` method basically just orders the endpoints of undirected edges before creating the directed one. If undirected edges are already canonicalized, then the `ARBITRARY` method effectively does the same. This does not mean that there should not be an `ACYCLIC` method, as it's semantically different, and the behaviour of `ARBITRARY` could change in the future. But knowing whether edge canonicalization is done is important for the implementation.

gaborcsardi:
> @ntamas @gaborcsardi Do you know if this ordering is guaranteed or not?

It is guaranteeed.

szhorvat:
OK, then I will:

 - Update the comment here to note that this is currently guaranteed: https://github.com/igraph/igraph/blob/master/include/igraph_datatype.h#L60

 - Not rely on it outside of `type_indexedgelist.c`, in particular not rely on it in this PR

gaborcsardi:
Right, thanks!

szhorvat:
Ready for review. Should be squashed before merging.

examples/tests/igraph_to_directed.c
```diff
@@ -0,0 +1,43 @@
...
+#include <igraph.h>
+
+#include "test_utilities.inc"
+
+int main() {
+    igraph_t ug;
+    igraph_t dg;
+
+    igraph_small(&ug, 6, /* directed= */ 0,
+                 2,0, 1,4, 3,2, 2,3, 0,4, 5,0, 2,3, 5,3, 2,5, 0,1,
```
vtraag, line 11: The duplicate edges are explicitly in here to test the case with parallel edges?

src/conversion.c
```diff
@@ -341,26 +353,55 @@ int igraph_get_edgelist(const igraph_t *graph, igraph_vector_t *res, igraph_bool
...
         igraph_vector_t edges;
-        long int no_of_edges = igraph_ecount(graph);
-        long int no_of_nodes = igraph_vcount(graph);
         long int size = no_of_edges * 2;
+        long int i;
+
         IGRAPH_VECTOR_INIT_FINALLY(&edges, size);
         IGRAPH_CHECK(igraph_get_edgelist(graph, &edges, 0));
 
+        if (mode == IGRAPH_TO_DIRECTED_RANDOM) {
```
vtraag, line 376: It is a bit strange to have this `if` inside a `switch` statement on the exact same variable. Why not put this under the `IGRAPH_TO_DIRECTED_RANDOM` `case`?

vtraag (commented):
Looks good to me, just a few small remarks/questions.

On l. 409-412 it now says
```
igraph_vector_destroy(&edges);
IGRAPH_I_ATTRIBUTE_DESTROY(&newgraph);
IGRAPH_I_ATTRIBUTE_COPY(&newgraph, graph, 1, 1, 1);
IGRAPH_FINALLY_CLEAN(2);
```

I think it is better to but the `IGRAPH_FINALLY_CLEAN` always after destroys only. If now an error occurs in `IGRAPH_I_ATTRUBTE_COPY` (not sure this can happen though), the `FINALLY` stack is not correct. Unfortunately I cannot comment inline because this is not a part that is inside the diff.

examples/tests/igraph_to_directed.c
szhorvat: Yes.

src/conversion.c
szhorvat: Almost all code is shared between the ARBITRARY / ACYCLIC / RANDOM cases. The only difference is that we do some special ordering for the ACYCLIC / RANDOM cases in the middle of the implementation.

If I used 4 different `case`s in switch (as I originally planned to—this is why I changed the original `if` to a `switch`) then 3 of those cases would be almost identical, duplicating a lot of code.

szhorvat:
Regarding the FINALLY stuff, I wasn't paying attention as it was old code. You are correct: `IGRAPH_I_ATTRIBUTE_COPY` can fail. Pls see the last commit.

src/conversion.c
```diff
@@ -405,11 +405,12 @@ int igraph_to_directed(igraph_t *graph,
         IGRAPH_CHECK(igraph_create(&newgraph, &edges,
                                    (igraph_integer_t) no_of_nodes,
                                    IGRAPH_DIRECTED));
-        IGRAPH_FINALLY(igraph_destroy, &newgraph);
-        igraph_vector_destroy(&edges);
+        IGRAPH_FINALLY(igraph_destroy, &newgraph);        
         IGRAPH_I_ATTRIBUTE_DESTROY(&newgraph);
         IGRAPH_I_ATTRIBUTE_COPY(&newgraph, graph, 1, 1, 1);
```
vtraag, line 410: Should this also be enclosed in `IGRAPH_CHECK`?

src/conversion.c
```diff
@@ -405,11 +405,12 @@ int igraph_to_directed(igraph_t *graph,
...
+        igraph_vector_destroy(&edges);
         IGRAPH_FINALLY_CLEAN(2);
+
         igraph_destroy(graph);
```
vtraag, line 414: Should this `igraph_destroy(graph)` not be located before `IGRAPH_FINALLY_CLEAN`? Well, actually, it shouldn't matter much, we can pop from the stack before destroying it, but we almost always first destroy and then pop from the stack.

src/conversion.c
```diff
@@ -442,9 +443,10 @@ int igraph_to_directed(igraph_t *graph,
         IGRAPH_CHECK(igraph_i_attribute_permute_edges(graph, &newgraph, &index));
 
         igraph_vector_destroy(&index);
-        igraph_vector_destroy(&edges);
-        igraph_destroy(graph);
+        igraph_vector_destroy(&edges);        
         IGRAPH_FINALLY_CLEAN(3);
+
+        igraph_destroy(graph);
```
vtraag, line 449: Same here, shouldn't this be before the `IGRAPH_FINALLY_CLEAN`?

src/conversion.c
szhorvat: No, it's a macro, and has no return code (however, it does include an IGRAPH_ERROR, so there is a possibility of failure).

src/conversion.c
vtraag: Yes, that is right. But couldn't we then also reduce code duplication between the last `case`? In principle it is the same, except that the edges are included twice. It is now a bit strange mix of `if` and `switch`, that when reading back might seem a bit counterintuitive. If you want I can also do this.

src/conversion.c
szhorvat: Technically it doesn't matter, as `igraph_destroy` doesn't fail.

Semantically, I think it's correct to keep `igraph_destroy(igraph)` at the end. The finally stack contains not `igraph_destroy(graph)`, but `igraph_destroy(newgraph)`. This is what I call the "constructor-finally" (when I think about it). It's `FINALLY_CLEAN` has no corresponding explicit `destroy` call. In constructors, the idea is that we put a `destroy` call on the finally stack in case the constructor fails. But if the constructor succeeds, of course its result should not be destroyed.

szhorvat: I think it's more readable to keep it at the end. If I put it the `igraph_destroy(graph)` right before the `FINALLY_CLEAN`, it makes the impression to the reader that it is an `igraph_destroy(graph)` that is being cleaned from the stack—this is not correct. 

The mismatch between the number of destroy calls precending the `FINALLY_CLEAN` and the number in the `FINALLY_CLEAN` is expected, for the same reason as in constructors.

vtraag: Ah yes, you are right, you only destroy the old allocated `graph`, which you then supply with the new allocated stuff from `newgraph`.

src/conversion.c
szhorvat: That case is different enough that I think that entangling it with the first case would hurt readability too much.

vtraag (approved):

PR #1516, opened 2020-10-31
fix prototypes of fortran intrinsics

szhorvat:
The prototypes of fortran intrinsics seems to be bad. See how they are actually used from LAPACK:

https://github.com/igraph/igraph/blob/master/src/lapack/dlamch.c#L88

As with all fortran functions, the arguments should be pointers.

I found this while compiling with LTO on Linux.

### DIFF
--- a/src/fortran_intrinsics.c
+++ b/src/fortran_intrinsics.c
@@ -23,31 +23,31 @@
 
 #include <float.h>
 
-double digitsdbl_(double x) {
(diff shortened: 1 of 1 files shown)


szhorvat:
We could also `#include "f2c.h"` and use `doublereal` and `integer` instead of `double` and `int`.

ntamas (approved):

PR #1519, opened 2020-10-31
CMake: create parsers/lexers before packaging.

vtraag:
This PR adds a separate target to build the parser sources, called `parsersources`. The built parsers (i.e. the `.c` files) are placed in a separate directory `parsers` in the `src` folder in the build tree, and are hence used from there in the build tree. Hence, you can also separately (only) build the parser sources by calling `cmake --build . --target parsersources`.

Note that the separate `parsers` folder was necessary to make it easier to include the entire directory.

This PR now also excludes the parser (`.y`) and lexer (`.l`) files from the packaged source.

We might think about moving the original parser sources also to a separate `parser` directory in the original `src` directory to avoid confusion.

### DIFF
--- a/etc/cmake/packaging.cmake
+++ b/etc/cmake/packaging.cmake
@@ -13,6 +13,8 @@ add_custom_target(dist
   USES_TERMINAL
 )
 
+add_dependencies(dist parsersources)
+
 #############################################################################
 ## Configuration of the source package
 #############################################################################
@@ -31,6 +33,7 @@ set(
 	"${CMAKE_SOURCE_DIR}/include;/include"
 	"${CMAKE_SOURCE_DIR}/msvc/include;/msvc/include"
 	"${CMAKE_SOURCE_DIR}/optional;/optional"
+	"${CMAKE_BINARY_DIR}/src/parsers/;/src/parsers"
 	"${CMAKE_SOURCE_DIR}/src;/src"
 	"${CMAKE_SOURCE_DIR}/tests;/tests"
 )
@@ -45,6 +48,8 @@ set(CPACK_INSTALL_SCRIPT "${CMAKE_SOURCE_DIR}/etc/cmake/cpack_install_script.cma
 set(
 	CPACK_SOURCE_IGNORE_FILES
 	"\\\\..*/"
+	"\.l$"
+	"\.y$"
 	"${CMAKE_SOURCE_DIR}/build"
 	"${CMAKE_SOURCE_DIR}/optional/simpleraytracer"
 	"Makefile.am"
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -12,23 +12,25 @@ add_subdirectory(prpack)
 
 # Generate lexers and parsers
 set(PARSER_SOURCES)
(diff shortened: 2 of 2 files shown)


vtraag:
I now realise that we should then also disable the `bison` and `flex` dependencies if those files are not included.

vtraag:
This is actually a bit of a catch-22 situation. The dependencies need to be setup before including the subdirectories, but it only becomes clear whether the `flex` and `bison` dependencies are need when including the subdirectories. The easiest solution is probably to include the `flex` and `bison` dependencies only in the relevant subdirectory (i.e. `src`). Or do you have any other suggestion @ntamas?

ntamas:
Aren't Flex and Bison declared as _optional_ dependencies at the moment? I think they are, which means that the build will happily commence even if you don't have Flex and/or Bison.

We can probably improve this by moving them from the `OPTIONAL_DEPENDENCIES` variable to the `REQUIRED_DEPENDENCIES` variable if we add at least one Flex or Bison target, assuming that we can sort out the visibility problems of the variables, but the current situation is not as bad as it seems. I'm inclined to merge this PR and then experiment with it further in the CMake branch. Any objections?

(I'll merge by the end of the day if there aren't any).

vtraag:
Yes, fine to merge and experiment further! I have thought also about a more general setup, more in the line of the `generated_sources` approach. However, this seems not straightforward. Right now, the generated sources are in a separate directory (i.e. `parsers`). Adding arbitrary directories and automatically excluding the meta-source files (i.e. used to generate the source files) is also more difficult because of the way that CMake includes entire directories. It might therefore be better to also keep the current naming, so that is it clear that it concerns the parser sources (although we might call it `parser_sources` to make it more clear?)

ntamas (approved):

vtraag:
Oh, by the way, one thing that I now realise I didn't get around to anymore is that it would be preferable to simply make `dist` an `ALIAS` for `package_source`. We should then add the dependency to `package_source`. Right now, the packaging only correctly generates the parser sources if we run `cmake --build . --target dist`, not if we call `cmake --build . --target package_source`, which might be a default expectation in some cases.

ntamas:
For some reason we cannot add any dependencies to the `package_source` target, not even if I add them _after_ calling `include(CPack)`. I think this is because the `package_source` target is some sort of a "virtual" target that gets created only for the makefile and the Ninja generators.

I tried adding it to `etc/cmake/cpack_install_script.cmake` as an `execute_process()` command, but now `make package_source` fails with:

```
CMake Error at /Users/tamas/dev/igraph/igraph/etc/cmake/cpack_install_script.cmake:22 (file):
  file INSTALL cannot find "/Users/tamas/dev/igraph/igraph/IGRAPH_VERSION":
  No such file or directory.
```

and I have no idea where this is coming from. There's nothing like that in `cpack_install_script.cmake` near line 22.

vtraag:
Ok, probably better then to simple leave it as is, and clearly document that source packages should be build with `dist`, not with `package_source`.

ntamas:
Okay, in the meanwhile I managed to read the message properly and now I understand :) (I needed to get some sleep). We need to generate the `IGRAPH_VERSION` file for the tarball.

ntamas:
Fixed in 8752ab51

Issue #1528, opened 2020-11-02
CMake: Windows test and shared builds


vtraag:
On Windows there are a couple of tests that we decided we should disable. Some of them don't seem to use any internal functions, yet don't compile correctly on Windows. 

Let us take one example `igraph_random_sample`. Building that tests errors out with the error: `error LNK2019: unresolved external symbol __imp_igraph_random_sample referenced in function error_test `. This suggest that the function is not properly exported/imported (i.e. incorrectly marked with `dllspec`). However, it is perfectly well defined here:
https://github.com/igraph/igraph/blob/a4f2fec3ea25348cf7fed1507ae47381b1673ea9/include/igraph_nongraph.h#L76-L77
I even double checked whether `DECLDIR` is correctly defined when including this header, and it is.

Now here is where it gets weird. If I instead include these two lines in `igraph_random.h` instead of in `igraph_nongraph.h`, it does compile correctly! You might think it is then a problem with the whole file `igraph_nongraph.h`, but that is also incorrect. For example, `igraph_nongraph.h` also contains `igraph_fisher_yates_shuffle`, but that compiles perfectly fine with the tests.

vtraag:
I think I actually also just found out the problem: `igraph_nongraph.h` is not properly included in `random.c`.

vtraag:
There are two options of how we can solve this: either we move the relevant declarations to `igraph_random.h`, or alternatively, we include also `igraph_nongraph.h` in `random.c`. At any rate, we should clean that up.

Possibly similar problem appear in other parts of the codebase. Would there be any proper way of detecting these?

vtraag:
It actually turns out `igraph_fisher_yates_shuffle` is nowehere implemented. The test file `igraph_fisher_yates_shuffle.c` does not use that implementation. We should hence just remove the declaration of `igraph_fisher_yates_shuffle`.

szhorvat:
I am fixing many of these issues right now, will push in 15 min.

szhorvat:
I fixed a bunch of these. Can you test again @vtraag ?

Here's the method I used:

 - disable `-Werror` and enable [`-Wmissing-prototypes`](https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html#index-Wmissing-prototypes)
 - record the warnings, and filter out those which refer to functions with `_i_` in the name
 - manually sort through the rest and add any missing includes

The nastiest bug was a mismatch between the prototype and definition of `igraph_hrg_sample`.

[...]

PR #1529, opened 2020-11-03
CMake: Re-enabled some internal tests.

vtraag:
This re-enables some test as discussed previously in #1528. Additionally, it includes some comments for files for which it is less clear why they are marked as "internal".

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -88,15 +88,10 @@ add_legacy_tests(
   igraph_delete_vertices
   igraph_empty
   igraph_get_eid
+  igraph_get_eids
   igraph_is_directed
   igraph_neighbors
 )
-if (BUILD_INTERNAL_TESTS)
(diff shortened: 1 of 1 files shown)


ntamas:
LGTM. Will merge when the CI checks pass.

tests/CMakeLists.txt
```diff
@@ -170,21 +167,19 @@ add_legacy_tests(
...
 if (BUILD_INTERNAL_TESTS)
   add_legacy_tests(
     FOLDER simple NAMES
-    igraph_eccentricity
-    igraph_get_all_simple_paths
-    igraph_pagerank
-    igraph_radius
-    igraph_rewire
-    single_target_shortest_path
+    igraph_rewire # Uses internal igraph_rewire_core
```
szhorvat, line 182: One small note is to eventually use the correct name in the comment, since @ntamas already renamed `igraph_rewire_core`

PR #1530, opened 2020-11-04
CMake: MSVC warnings

vtraag:
Regarding `warning C4273: inconsistent dll linkage` in #1507, the problem is that these functions are defined multiple times. Most often these occur in `f2c`, but also once in `random.c`. For `random.c` there is already a solution, we simply detect whether `rint` and `rintf` already exist. This PR corrects that.

For `f2c` the situation is slightly more complicated. The relevant functions (`erf`, `erfc` and `ungetc`) are defined in the MSVC libraries (`cmath`). Of course, it is defined there with a (correct) `declspec` declaration, while in `f2c` it is not defined as such (only with `extern`). There are multiple possible solutions: (a) simply remove the relevant definitions (of `erf`, `erfc` and `ungetc`) completely; or (b) add the correct `declspec` declarations. What option do you think would be better? Or are there perhaps other options still?

`


vtraag:
I also addressed warning `C4100: unreferenced formal parameter` in https://github.com/igraph/igraph/pull/1530/commits/b55d3add194ab06280a38846a80a85ff7a664dca. Most often I marked this argument as unused, using `IGRAPH_UNUSED`. However, there were also a number of instances for which the arguments really did seem superfluous. In that case, I did remove the actual argument. Please do check if you are OK with this.

@szhorvat , this also holds for the callback function for `cliquer`. The graph argument (`g`) never seems to be used. I left the argument in, because it might be that some callback function in the future would actually use the graph. But perhaps you think this argument actually should be removed?

vtraag:
Let me simply consider all warnings in this PR, but you can then already provide feedback on the ones that I did solve.

vtraag:
@tamas, on another note, there seem to be some functions from `plfit` that would need to be updated for version 0.9. I have now also marked this in the comments with a `TODO`, but perhaps better to open a separate issue for that, so that we don't loose track?

szhorvat:
I am not very comfortable with the mass-removal of unused arguments. Unused argument is one of those warnings which IMO should be turned off. (Heck, with one of the compilers I used it was `IGRAPH_UNUSED` itself that was triggering un-disableable warnings!)

I think argument removal should be discussed on a case-by-case basis. Making warnings go away is just not a useful goal. We should be clear about why an argument is or isn't needed now, and why it wouldn't ever be needed. It's also not something that's worth breaking API for.

As an example, take the `inclist_remove_duplicates` function. There are considerations such as: 1. consistency with the adjlist equivalent 2. the fact that an inclist is meaningless without its graph 3. did past implementations use the graph? if yes, might future ones use it? 4. this function came up when dealing with eigenvector centrality and as I remember, the conclusion was that such functions are not a good solution to deal with duplicate self-loops—instead those should be controlled when calling the inclist/adjlist generator. I.e. this function may be removed soon anyway. Therefore, it's not clear to me that it's a good idea to change it again at this point.

szhorvat:
tl;dr Before removing arguments, we should write down the justification (and silencing warnings is a not a good justification).

ntamas:
This PR seems to have broken something in the Appveyor build:

> C:\projects\igraph\src\eigen.c(1531): error C2065: 'cmplxvector': undeclared identifier [C:\projects\igraph\build\src\igraph.vcxproj]

szhorvat:
Please let us not be overeager in silencing warnings.

We want a robust and easy to maintain library. That does not equate no warnings for any given compiler.

I also see some of the non-trivial algorithms, such as push-relabel, being touched here. Do we know that there is no useful information in these arguments and their names *for human readers*? Do we know that it would be easier, and not harder, to make changes to these algorithms in the future with these arguments removed? Maybe they communicate the intention of the original implementor.

vtraag:
> This PR seems to have broken something in the Appveyor build:
> 
> > C:\projects\igraph\src\eigen.c(1531): error C2065: 'cmplxvector': undeclared identifier [C:\projects\igraph\build\src\igraph.vcxproj]

This was an old error (fixed in https://github.com/igraph/igraph/pull/1530/commits/f5fdb333e33a70631f40513bd8d26ee05d901839), but it turned out there was another error that I fixed now in https://github.com/igraph/igraph/pull/1530/commits/d595d270a7ab58964c21f6b4c4ae653fc7c5ad1b.

> Before removing arguments, we should write down the justification (and silencing warnings is a not a good justification).

OK, that is fine. I went through all the warnings, and added `IGRAPH_UNUSED` if the argument is not used and removed it in some cases. 

I think that it generally makes sense to only include arguments that actually do something, it might otherwise lead to incorrect expectations. In some cases there are good argument for including unused arguments: 
- they *might* be used, such as in callback options:
- the actual functionality is not *yet* implemented, but is expected to be in the future.

In other cases I removed unused arguments, unless the code was too complicated to understand correctly. Perhaps it is easier to discuss details in actual comments in the code: feel free to point to specific removal of unused arguments that you would disagree with.

> Do we know that there is no useful information in these arguments and their names for human readers? Do we know that it would be easier, and not harder, to make changes to these algorithms in the future with these arguments removed? Maybe they communicate the intention of the original implementor.

I am not sure I understand how the unused argument could communicate some intention, but I'd be happy to discuss it further. The arguments that I removed in `flow.c` are never actually used, and I don't immediately see what would be communicated there. If you think these arguments *should* be in place, I would have no problem in putting them back.

> Please let us not be overeager in silencing warnings.
> We want a robust and easy to maintain library. That does not equate no warnings for any given compiler.

Well, I think it would be helpful if we would have as few warnings as possible. In some circumstances warnings are unavoidable, and we turn them off them when compiling, instead of changing the source code. However, there may be circumstances in which warnings are actually helpful to see that some things may have gone wrong. If these warnings are turned off in the compiler, we will simply not be notified about it, even though it may reveal an actual bug. If possible, I would therefore prefer to address all warnings in as far as that is reasonably possible. 

Regarding unused variables, `IGRAPH_UNUSED` was already included for this specific purpose, and used across the code base, so I think it makes sense to continue doing that (if the unused argument actually makes sense).

vtraag:
It seems I was too quick to judge with `flow.c` indeed, I now corrected these using `IGRAPH_UNUSED` in
https://github.com/igraph/igraph/pull/1530/commits/d8162df095ecc0e9e8db7204fffaa326e7adb631. The problem is that the macro's hide a lot of the arguments that are being passed around, so I didn't see clearly that some argument might be used in some places.

szhorvat:
I'm afraid my comments may have come through as a bit too harsh.  Let me try to rephrase the concern from a different perspective:

A _lot_ is changed here, and I freaked out a bit, because I cannot immediately judge if there may be negative effects. It's too overwhelming.

I would like to ask to discuss each change individually (even if extremely briefly, in just one sentence), for the sake of making the changes (and their effects / benefits / risks) more transparent for all who are involved. I won't have time for this today, but I'll try to take them one by one and write comments here noting when I'm happy with then, or raising specific concerns when not.

Maybe it would have been better to keep these 'unused-fix' changes in a separate PR from the other, more straightforward and more clearly necessary fixes (e.g. the `rint` one).

vtraag:
> I'm afraid my comments may have come through as a bit too harsh.

No worries! It didn't come off as too harsh...

> Maybe it would have been better to keep these 'unused-fix' changes in a separate PR from the other, more straightforward and more clearly necessary fixes (e.g. the `rint` one).

Yes, maybe you are right, I created PR #1531 and #1532 to separate the issues. I now close this PR, we can then continue the discussion there.

PR #1531, opened 2020-11-04
CMake: MSVC warning C4273: inconsistent DLL linkage

vtraag:
Regarding `warning C4273: inconsistent dll linkage` in #1507, the problem is that these functions are defined multiple times. Most often these occur in `f2c`, but also once in `random.c`. For `random.c` there is already a solution, we simply detect whether `rint` and `rintf` already exist. This PR corrects that.

For `f2c` the situation is slightly more complicated. The relevant functions (`erf`, `erfc` and `ungetc`) are defined in the MSVC libraries (`cmath`). Of course, it is defined there with a (correct) `declspec` declaration, while in `f2c` it is not defined as such (only with `extern`). There are multiple possible solutions: (a) simply remove the relevant definitions (of `erf`, `erfc` and `ungetc`) completely; or (b) add the correct `declspec` declarations. What option do you think would be better? Or are there perhaps other options still?
`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -56,6 +56,8 @@ check_function_exists(stpcpy HAVE_STPCPY)
 check_function_exists(strcasecmp HAVE_STRCASECMP)
 check_function_exists(_stricmp HAVE__STRICMP)
 check_function_exists(expm1 HAVE_EXPM1)
+check_function_exists(rint HAVE_RINT)
+check_function_exists(rintf HAVE_RINTF)
 check_symbol_exists(stpcpy "string.h" HAVE_STPCPY_SIGNATURE)
 check_symbol_exists(strdup "string.h" HAVE_STRDUP)
 
--- a/src/config.h.in
+++ b/src/config.h.in
@@ -7,6 +7,8 @@
 #cmakedefine HAVE__STRICMP 1
 #cmakedefine HAVE_STRDUP 1
 #cmakedefine HAVE_EXPM1 1
+#cmakedefine HAVE_RINT 1
+#cmakedefine HAVE_RINTF 1
 
 #cmakedefine HAVE_GLPK 1
 #cmakedefine HAVE_GMP 1


szhorvat:
@ntamas Looking at this makes me realize that there are a lot more `#ifndef HAVE_XXX` in the source code, e.g.

```
./src/igraph_math.h:#ifndef HAVE_LOG2
./src/igraph_math.h:#ifndef HAVE_LOG1P
./src/igraph_math.h:#ifndef HAVE_FABSL
./src/igraph_math.h:#ifndef HAVE_FMIN
./src/igraph_math.h:#ifndef HAVE_ROUND
```

These are just a few of the examples. I was grepping the sources for `HAVE_`

These aren't created by cmake though, as far as I can tell.

Shouldn't we either get rid of these (if not needed) or make sure that cmake defines them as appropriate? Is there any other way that they might get defined that I am missing?

vtraag:
I don't think these are automatically defined elsewhere. I think it would be useful to leave them in, in case there is a compiler that doesn't support these functions. I think that with recent MSVC version these are supported. We should just include some additional `check_function_exists` for these functions. Additionally, there is some stuff around infinity in `math.c` that we might want to take a closer look at.

ntamas:
@vtraag is right, all that we need is a bunch of `check_function_exists()` or `check_symbol_exists()` calls in the top-level `CMakeLists.txt` and an adjustment of `src/config.h.in` so CMake knows to export these variables into `config.h`. I vote for keeping these and adding the necessary checks in `CMakeLists.txt`.

ntamas:
@vtraag Regarding `erf`, `erfc` and `ungetc` -- I would vote for patching `f2c` such that it does not erroneously try to override the existing declaration from `math.h` or `stdio.h`.

ntamas (approved):

ntamas:
Added `HAVE_` macros for a few of the functions that @szhorvat has found, FYI.

szhorvat:
Yes, that was what I meant: need more `check_function_exists()`

PR #1532, opened 2020-11-04
CMake: MSVC warning C4100: unreferenced formal parameter

vtraag:
This PR addresses warning `C4100: unreferenced formal parameter` mentioned in #1507. Most often I marked this argument as unused, using `IGRAPH_UNUSED`. However, there were also a number of instances for which the arguments really did seem superfluous. In that case, I did remove the actual argument. Please do check if you are OK with this.

There still remain a number of `C4100` warnings that occus in `bliss` and `cs`, but I don't think we should address those, so we should just silence them I think. If you agree I will add the relevant no-warning flags for `bliss` and `cs`.


include/igraph_qsort.h
```diff
@@ -25,6 +25,7 @@
 #define IGRAPH_QSORT_H
 
 #include "igraph_decls.h"
+#include "igraph_types.h"
```
ntamas, line 28: Do you remember why this include is necessary here? The only thing that is non-standard in this file seems to be `size_t`, which should be covered by `stddef.h`.

src/layout.c
```diff
@@ -2347,6 +2347,8 @@ int igraph_layout_mds(const igraph_t* graph, igraph_matrix_t *res,
     igraph_matrix_t m;
     igraph_bool_t conn;
 
+    IGRAPH_UNUSED(options);
+
```
ntamas, line 2351: The "options" argument should be removed IMHO. It used to be used when `igraph_layout_mds` used ARPACK to calculate eigenvectors, but it was a bad decision anyway because the matrices that MDS works on are not sparse. Currently the documentation says that this argument is ununsed, and since we are breaking the API anyway for 0.9, we can remove the argument as I see zero chance that it gets re-introduced in the future.

src/other.c
```diff
@@ -260,6 +260,11 @@ static const char* igraph_i_plfit_error_message = 0;
 
 static void igraph_i_plfit_error_handler_store(const char *reason, const char *file,
         int line, int plfit_errno) {
+
+    IGRAPH_UNUSED(file);
+    IGRAPH_UNUSED(line);
+    IGRAPH_UNUSED(plfit_errno);
```
ntamas, line 266: In cases when the arguments are unused but must be there because they are part of a callback signature, I think an easier solution is just to remove the _name_ of the argument and leave its type there. Can you test whether it would work with MSVC? If so, let's do it this way because it improves readability.

src/other.c
```diff
@@ -355,7 +360,7 @@ int igraph_power_law_fit(const igraph_vector_t* data, igraph_plfit_result_t* res
     plfit_stored_error_handler = plfit_set_error_handler(igraph_i_plfit_error_handler_store);
     if (discrete) {
         plfit_discrete_options_init(&disc_options);
-        /* approximation method should be switched to PLFIT_P_VALUE_EXACT in igraph 0.9 */
+        /* TODO: pproximation method should be switched to PLFIT_P_VALUE_EXACT in igraph 0.9 */
```
ntamas, line 363: There's a typo there.

ntamas (changes requested):
A few comments here and there -- see the source. I think the biggest change that I would like to test is whether simply omitting the _name_ of unused arguments from _callback functions_ (where we are forced to add the argument because of the API that consumes the callback) would fix the warnings there. If this is the case, it is a better solution than throwing `IGRAPH_UNUSED()` everywhere.

include/igraph_qsort.h
vtraag: The `IGRAPH_UNUSED` definition is in this header. I would be more than happy to move this definition, because it does not really seem to fit in this header, but that would touch a lot of more files then presumably.

src/layout.c
vtraag: I would be fine with removing. The reason that I left it in is because I couldn't foresee whether this argument would be useful at some point when we would like to change the implementation. To me it sounds reasonable that we will never use `ARPACK` for this, so it might make sense to remove it then indeed.

include/igraph_qsort.h
ntamas: Got it. We could move it to `igraph_decls.h` -- after all, it _declares_ that a variable is unused :)

src/layout.c
ntamas: Okay, let's remove it then!

src/other.c
vtraag: Unfortunately that does not work. Or well, that is, it gives a different warning: `error C2055: expected formal parameter list, not a type list`. According to the Microsoft docs, this is also expected by ANSI C, so this is then a warning that will presumably also pop up with other compilers. So, probably best to keep this as is.

src/other.c
vtraag: Fixed.

vtraag: Fixed

include/igraph_qsort.h
vtraag: I am now doing this, I just move the declaration to `igraph_decls.h` and remove the `igraph_types.h` inclusion here. It might be that there are some superfluous includes still on `igraph_types.h`, although that is unlikely.

include/igraph_adjlist.h
```diff
@@ -188,8 +187,7 @@ DECLDIR int igraph_adjedgelist_init(const igraph_t *graph,
                                     igraph_inclist_t *il,
                                     igraph_neimode_t mode);
 DECLDIR void igraph_adjedgelist_destroy(igraph_inclist_t *il);
-DECLDIR int igraph_adjedgelist_remove_duplicate(const igraph_t *graph,
-        igraph_inclist_t *il);
+DECLDIR int igraph_adjedgelist_remove_duplicate(igraph_inclist_t *il);
```
szhorvat, line 190: If we deprecate `adjedgelist_*` functions (we should), we should not change them anymore.

include/igraph_adjlist.h
```diff
@@ -85,8 +85,7 @@ DECLDIR int igraph_inclist_init(const igraph_t *graph,
 DECLDIR int igraph_inclist_init_empty(igraph_inclist_t *il, igraph_integer_t n);
 DECLDIR void igraph_inclist_destroy(igraph_inclist_t *il);
 DECLDIR void igraph_inclist_clear(igraph_inclist_t *il);
-DECLDIR int igraph_inclist_remove_duplicate(const igraph_t *graph,
-        igraph_inclist_t *il);
+DECLDIR int igraph_inclist_remove_duplicate(igraph_inclist_t *il);
```
szhorvat, line 88: I would not change this, because:

 - Based on our discussion about counting self-loops for eigenvector centrality and other cases, `adjlist_remove_duplicate` and `inclist_remove_duplicate` are not good solutions for what they do. Self-loop counting should be controlled when generating the adjlist/inclist. Therefore, we might soon deprecate this function. Thus, I would not create the impression that it is maintained.
 - The corresponding `adjlist_remove_duplicate` _must_ take a graph argument, consistency is worth maintaining
 - inclists, defined in terms of edge IDs, are not meaningful without a graph

Overall, I would prefer not to touch this function now (and get rid of it soon).

include/igraph_embedding.h
```diff
@@ -54,7 +54,6 @@ DECLDIR int igraph_laplacian_spectral_embedding(const igraph_t *graph,
         igraph_integer_t no,
         const igraph_vector_t *weights,
         igraph_eigen_which_position_t which,
-        igraph_neimode_t degmode,
```
szhorvat, line 57: Agreed, remove this!

src/scg.c
```diff
@@ -2119,6 +2119,8 @@ int igraph_scg_laplacian(const igraph_t *graph,
     igraph_bool_t tmp_lsparse = !Lsparse, tmp_rsparse = !Rsparse;
     igraph_sparsemat_t myLsparse, myRsparse, tmpsparse, Rsparse_t;
 
+    IGRAPH_UNUSED(direction);
```
szhorvat, line 2122: We should create a wishlist issue to implement the function of this parameter, otherwise it will get forgotten. IMO, this should be done for every similar case (non-functional documented arguments in public functions)

include/igraph_adjlist.h
ntamas: Yes, that's true, although it doesn't hurt to modify them one last time here. :) I leave it up to Vincent to decide whether he tackles the removal in this PR as well.

src/triangles.c
```diff
@@ -428,7 +428,6 @@ int igraph_i_trans4_al_simplify(igraph_adjlist_t *al,
 
 int igraph_transitivity_local_undirected4(const igraph_t *graph,
         igraph_vector_t *res,
-        const igraph_vs_t vids,
         igraph_transitivity_mode_t mode) {
```
szhorvat, line 431: Does anyone fully understand the design of this function? Are we *sure* that there is no reason for, or benefit from having the four different versions of this function have the same signature?

This is one of those things I would not change without a deeper understanding. This change is literally aesthetic to silence a minor compiler warning that is not even on by default.

szhorvat:
That's all, no more comments from me.

src/triangles.c
vtraag: This function is explicitly only called in the condition that all vertices are requested. To me that suggests that its intention is to never actually do something with this argument.

Instead, having the same arguments in all functions seems just esthetics to me. But if you find that they are somehow defined elsewhere to have the same call signature, I'd be happy to learn about it.

ntamas: I vaguely remember that Gábor has experimented with several different implementations of the transitivity function and that some of them later turned out to be efficient only if you need the transitivities of (almost) all vertices -- but then we did not bother removing the `vids` argument. There is no underlying deeper design decision here that would necessitate keeping the argument there, especially because these functions are meant to be internal anyway, so if later we somehow realize that they could still be made efficient for a subset of vertices only, we can still re-add the arguments without breaking the API (if we remove them from the public API now).

So, my vote is for removing the arguments and then renaming the functions to `igraph_i_...` and making them static.

szhorvat: Yes, let us remove them from the headers in this PR instead of fussing with the develop branch. I checked and they are not used anywhere, so we will not need to create an internal header for them.

src/scg.c
szhorvat: Done: https://github.com/igraph/igraph/issues/1541

include/igraph_adjlist.h
ntamas: Agreed.

src/triangles.c
vtraag: If we remove this in this PR, then why not also remove the `*remove_duplicates` functions in this PR?

ntamas: We should rename the PR in that case to make this explicit (or at least add it in the PR description) -- but otherwise I'm fine with it.

szhorvat: `igraph_transitivity_local_undirected4` was undocumented, and one might argue that it was fairly obvious that it was never meant for public use (i.e. it was exposed mistakenly). Therefore, removing it immediately makes sense.

Functions which were meant to be part of the public API are better deprecated for a version before removing them entirely.

Since all documented functions with `adjedgelist` in the name have been deprecated since 0.6, I support removing `igraph_adjedgelist_remove_duplicate` (even if it did not have an explicit deprecation notice).

----

As for `igraph_inclist_remove_duplicate`, I would not make changes to it in this PR, as there is no replacement functionality for it yet (i.e. add a flag controlling self-loop duplication to `inclist_init`). I think this one needs more discussion.

szhorvat: Anyway, deciding whether we do the removal in this PR or later should definitely not block merging this PR! If there's a doubt, let's leave it for later and go ahead with this PR.

vtraag: OK, let's otherwise clean up those things in separate PRs, that might make it more targeted. I will undo the removal of the argument for now then, and in a next PR I will remove the deprecated `adjedgelist` functions, and in yet another we do something with `inclist_remove_duplicates`.

include/igraph_adjlist.h
vtraag: As said in other comments, let's do this in a separate PR.

include/igraph_adjlist.h
vtraag: As said in other comments, let's do this in a separate PR. I have now added back in the `graph` argument.

vtraag:
As far as I am concerned we can (squash and) merge this.

PR #1536, opened 2020-11-06
CMake: Fix incorrect DLL linkage of some functions in f2c

vtraag:
As suggested here https://github.com/igraph/igraph/pull/1531#issuecomment-721919083, we might patch `f2c` to make sure that functions are correctly linked. This PR does so. This is done for the `erf`, the `erfc` and the `ungetc` function.

### DIFF
--- a/src/f2c/derf_.c
+++ b/src/f2c/derf_.c
@@ -1,17 +1,13 @@
 #include "f2c.h"
+#include <math.h>
+
 #ifdef __cplusplus
 extern "C" {
 #endif
 
 #ifdef KR_headers
-#ifndef HAVE_ERF
-double erf();
-#endif
 double derf_(x) doublereal *x;
 #else
-#ifndef HAVE_ERF
-extern double erf(double);
-#endif
 double derf_(doublereal *x)
 #endif
 {
--- a/src/f2c/derf_.c
+++ b/src/f2c/derf_.c
@@ -4,10 +4,14 @@ extern "C" {
 #endif
 
 #ifdef KR_headers
+#ifndef HAVE_ERF
 double erf();
+#endif
 double derf_(x) doublereal *x;
 #else
+#ifndef HAVE_ERF
 extern double erf(double);
+#endif
 double derf_(doublereal *x)
 #endif
 {
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -55,6 +55,8 @@ include(CheckFunctionExists)
 include(CheckSymbolExists)
 
 # First we check for some functions and symbols
+check_function_exists(erf HAVE_ERF)
+check_function_exists(erfc HAVE_ERFC)
 check_function_exists(log2 HAVE_LOG2)
 check_function_exists(log1p HAVE_LOG1P)
 check_function_exists(fabsl HAVE_FABSL)
@@ -66,6 +68,7 @@ check_function_exists(_stricmp HAVE__STRICMP)
 check_function_exists(expm1 HAVE_EXPM1)
 check_function_exists(rint HAVE_RINT)
 check_function_exists(rintf HAVE_RINTF)
+check_function_exists(ungetc HAVE_UNGETC)
 check_symbol_exists(stpcpy "string.h" HAVE_STPCPY_SIGNATURE)
 check_symbol_exists(strdup "string.h" HAVE_STRDUP)
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -66,7 +66,6 @@ check_function_exists(_stricmp HAVE__STRICMP)
 check_function_exists(expm1 HAVE_EXPM1)
 check_function_exists(rint HAVE_RINT)
 check_function_exists(rintf HAVE_RINTF)
-check_function_exists(ungetc HAVE_UNGETC)
 check_symbol_exists(stpcpy "string.h" HAVE_STPCPY_SIGNATURE)
 check_symbol_exists(strdup "string.h" HAVE_STRDUP)
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -55,8 +55,6 @@ include(CheckFunctionExists)
 include(CheckSymbolExists)
 
 # First we check for some functions and symbols
-check_function_exists(erf HAVE_ERF)
-check_function_exists(erfc HAVE_ERFC)
 check_function_exists(log2 HAVE_LOG2)
 check_function_exists(log1p HAVE_LOG1P)
 check_function_exists(fabsl HAVE_FABSL)
--- a/src/config.h.in
+++ b/src/config.h.in
@@ -1,6 +1,8 @@
 #ifndef IGRAPH_CONFIG_H
 #define IGRAPH_CONFIG_H
 
+#cmakedefine HAVE_ERF 1
+#cmakedefine HAVE_ERFC 1
 #cmakedefine HAVE_EXPM1 1
 #cmakedefine HAVE_FABSL 1
 #cmakedefine HAVE_FMIN 1
@@ -14,6 +16,7 @@
 #cmakedefine HAVE_STRCASECMP 1
 #cmakedefine HAVE__STRICMP 1
 #cmakedefine HAVE_STRDUP 1
+#cmakedefine HAVE_UNGETC 1
 
 #cmakedefine HAVE_GLPK 1
 #cmakedefine HAVE_GMP 1
--- a/src/config.h.in
+++ b/src/config.h.in
@@ -1,8 +1,6 @@
 #ifndef IGRAPH_CONFIG_H
 #define IGRAPH_CONFIG_H
 
-#cmakedefine HAVE_ERF 1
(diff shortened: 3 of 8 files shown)


src/f2c/derf_.c
```diff
@@ -4,10 +4,14 @@ extern "C" {
 #endif
 
 #ifdef KR_headers
+#ifndef HAVE_ERF
```
szhorvat, line 7: I don't understand this.

This file appears to **assume** that the `erf()` function is available, and define `derf_` (a fortran function) in terms of it. 

Why did you condition it on an `#ifndef HAVE_ERF`?

It is my impression that this version of f2c wouldn't work at all without having `erf()`. If we don't have `erf()`, configuration should just fail.

vtraag:
Yes, I don't really understand why these functions are defined there. To me it seems completely superfluous. But then again, there might be a reason for it. Perhaps to ensure it at least compiles, even if the implementation of those functions are not available?

szhorvat:
@vtraag Could you please try the following? Instead of these changes, simply add the expected headers:

 - `#include <math.h>` in `erf_.c`, `derf_.c`,  `erfc_.c` and `derfc_.c`

Add it *right after* `#include "f2c."`

Then remove all the _declarations_ for the C functions, both the K&R and the modern ones.

Can you please try this and let me know if it fixes the problem?

In the meantime I'll look at what's going on with `ungetc()`, as that should have been declared already in `stdio.h`, which is included ...

szhorvat:
> Perhaps to ensure it at least compiles, even if the implementation of those functions are not available?

It will compile, but it will not link.

szhorvat:
If this solves it, we still *could* test for these functions with cmake and let it abort if it didn't find them, but it would only make sense to do it if f2c is compiled at all (it's needed by the internal arpack, lapack, blas). It's a bit annoying to have to fuss with this because these functions in f2c are not actually called by any of arpack, lapack, blas ... in principle we could even just remove them, though that would leave us with a partial f2c, which is strange.

vtraag:
Adding the headers won't solve the issue. The problem is that they are defined with different attributes, once without `declspec` (in the `f2c` sources and once with `declspec`). The solution is hence to simply remove the declarations if the functions are already properly defined, which is exactly what is tested for at the moment. This is also the problem with `unget.c`.

We could also simply remove the declarations all together, but in that case it won't compile if the function is missing. I am not too familiar with the details going on inside the linking step, but perhaps in situation in which this function is not defined (i.e. extern) and not invoked anywhere, the linking might still work?

szhorvat:
It makes no sense to remove these declarations _conditionally_. Declarations indicate that the function exists. If it does not exist, there will be a linking error.

`<math.h>` _should_ be included (currently it is not, unless I missed a nested include), and there should not be any additional declarations. 

Does doing this not resolve the problem?

vtraag:
I have not checked if/how/where `math.h` is included. It should be somehow, otherwise it shouldn't even compile in the current PR, right? The warning arises exactly because the function is declared twice. I'm happy to include `math.c` explicitly as well, but I won't get around to it this weekend.

The `check_function_exists` checks if the function can be linked by the way, not whether it is declared, according to the CMake documentation.

vtraag:
Removing the declarations does work (obviously). As clarified, I can either add or not the `#include <math.h>`, that does not make a difference. This begs the question how exactly these functions (i.e. `erf` and friends) are included. I still haven't figured that out yet, but I want to ask a related question. 

Many other `f2c` source files use `#include "math.h"` suggesting there is some local header file that is used. I don't see such an include file immediately, except for `math.hvc`, which does use `#include <math.h>` and undefines `complex` immediately afterwards. This file does seem to be used in the old autotools build system in some way (being copied to `math.h` effectively). As stated in `math.hvc`, it says `for VC 4.2`, presumably referring to Visual C. Additionally, in these files, `abs` is undefined before including `math.h`, perhaps in some cases this function is both included in the C library and the specific math library? At the moment, we don't use something like this in CMake. @ntamas, have you ever taken a look at this particular setup for the CMake build? It is not entirely clear to me whether the source files in `f2c` are generated in some way, and if so how.

ntamas:
As far as I know the only source file that is generated for `f2c` is a file named `arithchk.h`.

> @ntamas, have you ever taken a look at this particular setup for the CMake build? 

No, but I strongly suspect it's some legacy stuff; there are similar issues floating around from year 2008 [on the Internet](https://rwkbp.makekb.com/entry/826/). VC 4.2 is something that I don't think we should support anyway, so feel free to drop the extra `math.hvc`.

ntamas:
Regarding `HAVE_ERF` et al in this PR, I think that our goal is to check whether there already exists a _declaration_ for said function, so we should probably use [`CheckSymbolExists`](https://cmake.org/cmake/help/latest/module/CheckSymbolExists.html) - this looks for a declaration in a specific include file. We could also rename `HAVE_ERF` to `HAVE_ERF_DECLARATION` to make the intent clear.

szhorvat:
This seems to simply be a case of non-matching declspec between the declarations of `erf()`/`ercf()` that f2c comes with, and what they _actually_ are in MSVC.

All that seems to be needed is not to rely on the declarations that come with `f2c` and include `math.h` properly.

When f2c was written, `erf` was not a standard C function. Now it is.  Notice this in the f2c README:

```
Both ERF and DERF assume your C library provides the
	underlying erf() function (which not all systems do).
```

With the way f2c was written in the dark ages, your system either had `erf` and things worked, or it didn't, in which case you were still free to link in your own implementation without needing to change any of f2c's own sources. The generic declaration was there. 

----

To fix the problem here, `derf_.c` should look like this:

```
#include "f2c.h"
#include "math.h"

#ifdef __cplusplus
extern "C" {
#endif

#ifdef KR_headers
double derf_(x) doublereal *x;
#else
double derf_(doublereal *x)
#endif
{
return( erf(*x) );
}
#ifdef __cplusplus
}
#endif
```

There do not need to be any checks done with cmake.

Either the user's system has `erf` (which is practically always the case given that it is required by C99, a 20+ yr old standard, and it was often present even before than), and things work, or it doesn't, in which case there will be a linking error. There will not be an error from using the function without a declaration, as that is not an error in C. Nothing will change compared to the current situation except MSVC won't show warnings.

----

If we want to be extra nice in how errors are shown, cmake _could_ check if the function is present, and simply abort if it is not. Then basically the same error that the linker would report at the end of compilation would get reported early, during the configuration step. To do this right, it must only abort when actually compiling the internal f2c, and not otherwise. In my opinion, adding these details to an already complicated build system is pointless. igraph doesn't even use `erf`, or any function that indirectly uses `erf`. It'll just add 2-3 seconds to an already painfully slow configuration step on Windows.

vtraag:
I have been trying to understand why not using `#include <math.h>` does not end up in a compiler error/warning about a missing declaration. When using a simple test file 

```
#include <stdio.h>

int main()
{
    printf("Erf: %.2f\n", erf(1.96));
}
```

I get, as expected, the following error: `warning C4013: 'erf' undefined; assuming extern returning int`. Again, as expected, when properly including `math.h`, this warning goes away. I have no idea why *not* including `math.h` in `erf_.c` and friends does *not* show this warning. 

I gave up trying to understand this further, and simply include `math.h`, just to be sure. I also removed the `check_function_exists` calls and corresponding definitions.

szhorvat:
That is actually an important point. I do recall seeing several `C4013` warnings. You showed several examples of this warning in this issue: https://github.com/igraph/igraph/issues/1507#issue-730265506 We should indeed try to understand that ...

vtraag:
OK, I just found out. We include `f2c.h`, which includes `igraph_blas_internal.h`, which includes `igraph_types.h`, which finally includes `math.h` (and `stdio.h` for `ungetc`). I found out using this command `cl /c /P /d1PP [src.c]` (with relevant `src.c` obviously) as suggested here: https://stackoverflow.com/questions/15951728/how-to-find-issues-concerning-includes-visualize-inclusion-tree.

This suggest that we can leave out `math.h` and `stdio.h` in the inclusion in the relevant files. If you agree with this, I will do so.

szhorvat:
Yes, wanted to post the same :-)

szhorvat:
I would still suggest including `math.h` and `stdio.h` explicitly where they belong. It is not good practice to rely on too many nested includes. `igraph_blas_internal.h` could change in the future and end up not including `math.h`.

vtraag:
It would still be nice to get a nice visual overview of the entire include hierarchy. I have not managed to get something like that yet.

vtraag:
> I would still suggest including `math.h` and `stdio.h` explicitly where they belong. It is not good practice to rely on too many nested includes. `igraph_blas_internal.h` could change in the future and end up not including `math.h`.

Yes, good point.  I will add the inclusion for `stdio.h` for `ungetc` then still.

vtraag:
Note that the failing build was due to the problem that was solved by PR #1545.

ntamas:
Is this ready to merge then?

szhorvat:
I'm happy with this one (noting it because I complained the most).

PR #1540, opened 2020-11-09
igraph_diameter: add special case for null graph

### LINKED ISSUE #1538: igraph_diameter() mishandles null graph
`igraph_diameter()` mishandles null graph, and returns 0 as diameter. It should return infinity.

szhorvat:
Fixes #1538


ntamas:
Thanks!

PR #1546, opened 2020-11-10
cmake: disable uninteresting MSVC warnings

szhorvat:
This disables some uninteresting warnings from MSVC. See #1507

### DIFF
--- a/etc/cmake/compilers.cmake
+++ b/etc/cmake/compilers.cmake
@@ -4,7 +4,13 @@ endif()
 
 macro(use_all_warnings TARGET_NAME)
   if(MSVC)
-    target_compile_options(${TARGET_NAME} PRIVATE /W4)
+    target_compile_options(${TARGET_NAME} PRIVATE 
+      /W4 # enable most warnings, then disable:
+      /wd4244 # 'conversion' conversion from 'type1' to 'type2', possible loss of data
+      /wd4267 # 'var' : conversion from 'size_t' to 'type', possible loss of data 
+      /wd4996 # deprecated functions, e.g. 'sprintf': This function or variable may be unsafe. Consider using sprintf_s instead.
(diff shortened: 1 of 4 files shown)


szhorvat:
The MS compiler output should be less overwhelming now, and we should be able to see relevant warnings more easily:

https://ci.appveyor.com/project/ntamas/igraph/builds/36240213

szhorvat:
The MSVC output should hopefully be less chaotic, see here when build has finished: https://ci.appveyor.com/project/ntamas/igraph/builds/36240321

vtraag:
Sorry, I didn't get around to replying to your comment https://github.com/igraph/igraph/issues/1507#issuecomment-724746757 anymore. I'm perfectly fine with the (silence) warning flags that you put in here now.

I do want to raise the point that I think we should address error `warning C4456: declaration of variable hides previous local declaration`. This can point out potential issues with new code, and I therefore think that it would be nice if we can enable this warning again.

Finally, after we have silenced/fixed all warning, we should then enable error on warnings (`/WX`). That way, any new contributions get verified well.

ntamas:
The MSVC build output looks much better now.

How about this: we address the warnings that we think need to be addressed, based on the current CI output, and push changes to this branch until we get close to zero warnings. Then we re-enable C4456 as proposed by @vtraag, do another iteration until we are happy with the output, and then merge this PR. Basically this PR would be the clean-up-the-warnings-in-MSVC branch.

vtraag:
It might be better to merge this PR first and open separate PRs for separate warnings. The previous issues were also addressed in separate PRs as well (#1531, #1532, #1536), which might be needed, as we first discussed in PR #1530.

ntamas:
Okay, let's merge and continue in separate PRs.

PR #1548, opened 2020-11-21
Implement fatal error handling

szhorvat:
This PR adds fatal error handling, as well as igraph-specific assertions.

When a fatal error occurs, we don't do any cleanup, just bail immediately. This is necessary in some libraries where it would be difficult to integrate igraph's own error handler.

This can also be used for assertions that we _do not_ want to be disabled by `NDEBUG`. Some asserts don't hurt performance, and the messages they print can be very useful to debug user problems. Also, when an assert macro is done properly, the compiler can be aware that is never returns, and [can optimize the code much better](https://stackoverflow.com/a/47183046/695132). Thus, the idea that assertions should simply be removed for better performance is flawed.

A fatal error handler is expected to never return. Reasonable options are to `abort()` or to do a `longjmp`. The R interface would use the `error()` function provided by R, which simply returns to the R top level after printing a message. See the modification I made in the Cliquer sources—originally, R's `error()` was used.

This is still a draft, awaiting feedback before I go further.


src/cliquer/cliquer.c
```diff
@@ -938,15 +934,7 @@ static boolean store_clique(set_t clique, graph_t *g, clique_options *opts) {
...
-#ifdef USING_R
-		        error("CLIQUER INTERNAL ERROR: ",
-			      "clique_list_count has negative value!");
-#else
-			fprintf(stderr,"CLIQUER INTERNAL ERROR: "
-				"clique_list_count has negative value!\n");
-			fprintf(stderr,"Please report as a bug.\n");
-			abort();
-#endif
+                    IGRAPH_FATAL("CLIQUER INTERNAL ERROR: clique_list_count has negative value! Please report as a bug.");
```
ntamas, line 937: It seems like the Cliquer source uses tabs and this line uses spaces.

ntamas:
Initially I was a bit uneasy with this PR so let me describe how I understand the purpose of this PR to put my mind at ease.

Right now we have a few places in the codebase where we provide different implementations for R and non-R environments because we are not allowed to use `abort()` in the R code. This is one of the reasons why we have the `USING_R` macro at all (the other one is to avoid `printf()` when compiling the R interface). This PR essentially centalizes the use of `abort()` to the fatal error handler, and then in the R interface we only need to replace the fatal error handler with a call to R's `error()` instead.

The reason why I was worried about this PR a bit in the beginning is that the Python C API has no function similar to R's `error()` so I would either have to use `abort()` there or re-write lots of hand-written interface functions to set up the stage for `longjmp()` with `setjmp()` before diving down into the C layer. However, it looks like I wouldn't really "lose" anything with just having `abort()` in the code of the Python interface because they are already there now. So it seems fine for me as it is now.

The new doc chunks still have to be inserted into the documentation, though. I also vaguely remember that we have a guide somewhere in the wiki about integrating third-party code with igraph; we should make a note of replacing `abort()` calls with fatal errors there.

szhorvat:
Yes, your understanding is correct.

But it is not only about R. There is a good reason why they forbid `abort()` (and also `assert`) in R packages: if something goes wrong, ideally it should not take the whole session down with it. The same applied to all other high-level interfaces.

I think I can make use of this in the Mathematica interface as well (though `setjmp`/`longjmp`), although I have not implemented yet. I will implement it before making this PR ready, just to see that it works.

In summary, some errors cannot be handled gracefully, and _at this moment_ they result in aborting the entire process. This will make it possible for high-level interfaces to at least avoid an abort/crash, and let the user save their work. The error still won't be handled very nicely: memory won't be freed, resourced won't be cleaned up, and in the worst case, it might even leave a graph in a broken internal state. _But at least we didn't crash the session!_  We might even choose to give the user a severe warning, [like MATLAB does](https://stackoverflow.com/questions/27360178/matlab-encountered-an-internal-error-and-needs-to-close), although in the case of igraph such an error is unlikely to compromise the process.

szhorvat:
One additional type of error where this is useful is an assertion failure. With `assert`s, we don't want proper error handling (finally stack) because assertions _should_ never fail. If they do, it's a bug. But of course there are bugs in igraph. When they happen, it is best to show some informative message to the user. This is why I am strongly in favour of leaving in `assert`s. However, with the Mathematica interface, when an assert is triggered, the user won't see the message because it is printed to a terminal, not the notebook. Maybe it's the same with Jupyter notebooks? I don't know. This feature would not only make it possible not to crash the session, but it will also let the message be output in whatever way is most appropriate (e.g. print to the notebook with Mathematica).

Thus, this PR would also transition all `assert`s to `IGRAPH_ASSERT`s. I need feedback on this too before I do it.

src/igraph_error.c
```diff
@@ -289,3 +299,39 @@ igraph_set_warning_handler (igraph_warning_handler_t * new_handler) {
...
+#endif
+    }
+}
+
+void igraph_fatalf(const char *reason, const char *file, int line, ...) {
+    va_list ap;
+    va_start(ap, line);
+    vsnprintf(igraph_i_fatalmsg_buffer, sizeof(igraph_i_fatalmsg_buffer) / sizeof(char), reason, ap);
+    va_end(ap);
+    igraph_fatal(igraph_i_warningmsg_buffer, file, line);
```
ntamas, line 336: This should be `igraph_i_fatalmsg_buffer`

ntamas (approved):

ntamas:
@szhorvat fixed a minor copy-paste error above. I see no problems with the `IGRAPH_ASSERT()` implementation. Feel free to go ahead.

szhorvat:
I think it would make sense to clear the "finally stack" in a fatal error handler. But this should probably be left for the handler function to do instead of doing it automatically, the same way as calling `IGRAPH_FINALLY_FREE()` is left to the error handler.

szhorvat:
> The reason why I was worried about this PR a bit in the beginning is that the Python C API has no function similar to R's `error()` so I would either have to use `abort()` there or re-write lots of hand-written interface functions to set up the stage for `longjmp()` with `setjmp()` before diving down into the C layer. However, it looks like I wouldn't really "lose" anything with just having `abort()` in the code of the Python interface because they are already there now. So it seems fine for me as it is now.

It would even be okay to not set up a fatal error handler in Python. Then the system will use the default, which `abort`s.

The fatal error handler would only be called in situation where in the past we simply `abort`ed anyway.

include/igraph_error.h
```diff
@@ -35,13 +35,15 @@ __BEGIN_DECLS
  * prefix renamed to IGRAPH_), as I couldn't find a better way to do
  * them. */
 
+/* IGRAPH_NORETRUN indicates to the compiler that a function does not return.
```
ntamas, line 38: Typo here: "IGRAPH_NORETRUN"

szhorvat:
This should be ready now (and should be squashed-and-merged). I'll leave replacing some `assert`s with `IGRAPH_ASSERT`s for a second PR.

src/igraph_error.c
```diff
@@ -289,3 +324,42 @@ igraph_set_warning_handler (igraph_warning_handler_t * new_handler) {
...
+
+static IGRAPH_THREAD_LOCAL igraph_fatal_handler_t *igraph_i_fatal_handler = NULL;
+
+igraph_fatal_handler_t *igraph_set_fatal_handler(igraph_fatal_handler_t * new_handler) {
+    igraph_fatal_handler_t * previous_handler = igraph_i_fatal_handler;
+    igraph_i_fatal_handler = new_handler;
+    return previous_handler;
+}
+
+#ifndef USING_R
```
vtraag, line 339: Perhaps I'm missing something, but couldn't we get rid of the `USING_R` thing here? After all, the `igraph_abort` will never pose any problem to R since it will not call `abort`.

szhorvat: No, there's an `fprintf` and that's also not allowed. 

Actually I removed all these, then realized that we had the `fprintf`, put them back, and added a comment above `igraph_abort` at the top of the file.

szhorvat: To be accurate, it's not the `fprintf` but the `stderr`.

vtraag: Wouldn't it then be easier to put that `fprintf` statement in `USING_R` defs? For `R` it issues an `error` anyway.

szhorvat: The easiest thing was not to change it :-) R is not using the entire function anyway. If you think that something needs to be improved, feel free to go ahead.

PR #1549, opened 2020-11-22
Use Mini-GMP when GMP is not available

### LINKED ISSUE #1351: Make GMP non-optional
GMP is extremely useful for graph theory, and I have already done some work I want to contribute that relies on it.

We should make GMP be non-optional. Currently, the only user of GMP is Bliss, which falls back to using a `long double` (i.e. an ugly hack) when GMP is not available.

For the eventuality that an external GMP is not available, we can bundle "mini-gmp". mini-gmp comes with GMP and is intended exactly for this purpose. Quoting from its README file:

> This is "mini-gmp", a small implementation of a subset of GMP's mpn,
> mpz and mpq interfaces.
> 
> It is intended for applications which need arithmetic on numbers
> larger than a machine word, but which don't need to handle very large
> numbers very efficiently. **Those applications can include a copy of
> mini-gmp to get a GMP-compatible interface with small footprint. One
> can also arrange for optional linking with the real GMP library, using
> mini-gmp as a fallback** when for some reason GMP is not available, or
> not desired as a dependency.

It consists of only two `.h` and two `.c` files.

If we go this route, eventually we can also get rid of igraph's own bignum implementation, and replace it with GMP. GMP may provide much better performance. GMP is continually being developed: it received bugfixes and performance improvements. We don't want to keep maintaining igraph's bignum if we don't have to.

szhorvat:
Mini-GMP is a tiny partial implementation of the GMP API. It is meant to be included in projects directly. See `mini-gmp/README` within the GMP distribution.

Fixes #1351.


src/igraph_gmp.h
```diff
@@ -0,0 +1,6 @@
+
```
ntamas, line 1: Let's add the usual license header here; just copy it from one of the other source files and update the copyright years and the name if needed (use "The igraph development team" for the name).

ntamas (approved):

ntamas:
This seems okay for me; there's only one minor comment about the missing license header in `src/igraph_gmp.h`.

ntamas:
Okay, I'll just merge and then add the license :)

szhorvat:
So you're fine with calling it "vendored GMP" in the CMake files? It's of course convenient.

ntamas:
Yes, it would be too much hassle to re-write the CMake files to have another category (almost-vendored-but-simplified :)) just for the sake of GMP. If it becomes too confusing for users, we can reconsider.

PR #1554, opened 2020-11-25
V0.8.4: MINGW32/64 fixes

jannick0:


include/igraph_decls.h
```diff
@@ -13,7 +13,7 @@
  * Reference: https://docs.microsoft.com/en-us/cpp/preprocessor/predefined-macros */
 #undef DECLDIR
 #if defined (_WIN32)
-    #if defined (__CYGWIN__)
+    #if defined (__GNUC__)
```
szhorvat, line 16: Can you explain why this is necessary? MinGW supports `__decldir`

src/foreign.c
```diff
@@ -2091,7 +2091,7 @@ int igraph_write_graph_pajek(const igraph_t *graph, FILE *outstream) {
                               };
     const char *estrnames2[] = { "a", "p", "l", "lc", "c" };
 
-    const char *newline = "\x0d\x0a";
```
ntamas, line 2094: The goal with the original code was to ensure that we always print Windows-style line endings, even when we run the code on Linux or macOS. So I think it is correct here -- but it works only if the input file is opened in binary mode, not when it is opened in text mode. So, we need to fix the places where `igraph_write_graph_pajek()` is called in the test cases to ensure that the files are opened in binary mode, and also document this requirement in `igraph_write_graph_pajek()`.

szhorvat:
@jannick0 You say in the commit message that `_WIN32` and `__CYGWIN__` are mutually exclusive. Do you know if this has always been so, i.e. if Cygwin has never defined `_WIN32` (of course unless targeting Win32)?

[This SO question](https://stackoverflow.com/q/47150492/695132) suggests that Cygwin used to define `_WIN32`. This is why I left in an explicit check for it. However, we might just forget about old Cygwin to keep things simpler.

include/igraph_decls.h
vtraag: Building on the question of @szhorvat: how does MinGW work when building a shared library? From [this](http://www.mingw.org/wiki/sampledll) page it seems that it might be possible to build DLLs without using `declspec` declarations, but that would require using `-no-undefined` and `--enable-runtime-pseudo-reloc` options. Can you confirm this is the case?

If MinGW does not require `declspec` declarations for a shared library, that would simplify it a lot. Building a shared library under Windows was not yet supported, and it turned out there were quite some things that needed to be corrected before it built correctly (see PR https://github.com/igraph/igraph/pull/1467 for some of those things). Essentially, on the `develop` branch, we now have a working build system on Windows (using CMake), including a fully functional shared library (both on MSVC and MinGW). However, this is only for the next major release (0.9.0). If `declspec` declarations are required on MinGW, I'm afraid it will be difficult to support building a shared library on MinGW for the 0.8.x series.

We actually do build the library also on MinGW in the CI, see [this](https://ci.appveyor.com/project/ntamas/igraph/builds/36493659/job/n0fcy5deax7169cp) build. That build does passes without this modifications. Locally on my machine, I also need to make some modifications, so I am not sure why this is actually building correctly on the CI.

src/foreign.c
ntamas: 330f55511 addresses the problem in the Pajek tests cases where we now force `stdout` to be opened in binary mode.

jannick0: Unfortunately, using `#include <unistd.h>` in 330f555 is not portable.  

For the issue itself: Just out of curiosity - why should `igraph_write_graph_pajek()` write Windows-style EOLs on any platform? This appears to be an artificial requirement.  Is this because of the pajek lexer (not having looked into the dependencies here)?

jannick0:
> @jannick0 You say in the commit message that `_WIN32` and `__CYGWIN__` are mutually exclusive. Do you know if this has always been so, i.e. if Cygwin has never defined `_WIN32` (of course unless targeting Win32)?

Unfortunately, I do not know. However, I had checked it for the latest gcc version for CYGWIN (gcc version 10.2.0), MSYS2 (gcc version 10.2.0) and MSYS ([gcc version 3.4.4](https://github.com/gcc-mirror/gcc/releases/tag/releases%2Fgcc-3.4.4) from 2005).  The latter suggests that the mutual exclusivity must be in for 15 years already I guess.  
 
> [This SO question](https://stackoverflow.com/q/47150492/695132) suggests that Cygwin used to define `_WIN32`. This is why I left in an explicit check for it. However, we might just forget about old Cygwin to keep things simpler.

Yes - I agree. As given in the commit log, all this boils down to MSVC I believe, doesn't it?

src/foreign.c
ntamas: Earlier versions of Pajek insisted on CRLF line endings as it was (is) primarily a Windows-only tool. (You can use it on Linux or macOS with Wine, though). We've contacted the developers on the Pajek mailing list and it looks like recent versions are okay with either Unix or Windows style line endings, so we might just drop support for older Pajek versions and just use the platform-native style.

include/igraph_decls.h
jannick0: Sorry for the late and only brief reply to the two messages above.  Unfortunately, I am really tied up to the neck right now.

* While MINGW does support `_decldir`, I bumped into a link error for MINGW64 which caused me to look into this.
* I'll try to look into [this](http://www.mingw.org/wiki/sampledll), but please bear in mind that if `autoconf` and `libtool` governs the build process it does become not so easy. [GCC information](https://gcc.gnu.org/onlinedocs/gcc-10.2.0/gcc/Microsoft-Windows-Function-Attributes.html#Microsoft-Windows-Function-Attributes) on `dllimport` and `dllexport`.  
* Using autoconf/libtool, how can static and shared libraries be built? See `DLL_EXPORT`  [here](https://www.gnu.org/software/libtool/manual/libtool.html#Windows-DLLs).
* Note that appveyor (with `MSVC_VERSION=NONE`) calls MSYS2's gcc which is CYGWIN-like, i.e. it does **not** call the MINGW32 or MINGW64 gcc which would be what you might be wanting, too. To verify watch out, e.g., for the target prefix  `x86_64-pc-msys` of the used `gcc` in `config.log`. 
* I hope to get some time to look at the develop branch using `cmake` for MSVC and MinGW.  While I recently spent some time with handling `autoconf` and friends, `cmake` runs a lot more quickly, is more flexible and updated on a regular basis. 

I'll try to put together an easy CI for the various gcc's MSYS2, MINGW32, MINGW64 which might be helpful ... hopefully shortly.

src/foreign.c
jannick0: That sounds great to me.

include/igraph_decls.h
jannick0: >     * While MINGW does support `_decldir`, I bumped into a link error for MINGW64 which caused me to look into this.

The link error can be observed [here](https://github.com/jannick0/igraph/runs/1470659590?check_suite_focus=true#step:4:20202), a log of a quickly composed github workflow for MINGW32 and MINGW64.

jannick0: > I'll try to put together an easy CI for the various gcc's MSYS2, MINGW32, MINGW64 which might be helpful ... hopefully shortly.

#1558 - CI runs for MINGW32 and MINGW64 which currently fail (see my previous comment).

szhorvat: Yes, this should be fixed somehow, preferably in the simplest possible way for 0.8.x.

Now all work is being done on the `develop` branch and we use CMake for the build system. [See draft build instructions.](https://igraph.discourse.group/t/draft-document-for-installation-with-cmake/504) Now we build either a static library or a DLL, but not both at the same time. This makes thing simpler. When _using_ an igraph built as a static library on Windows, it is necessary to define `IGRAPH_STATIC` before including any headers. This switches the `DECLDIR` to empty. Otherwise it's "dllimport". The entire decldir thing _should_ work with MinGW as well.

If you have time at some point, it would be really great if you could help set up a CI workflow for MinGW, either on Appveyor, which we already use, or some other CI service. We should make sure that what happened here (broken build with MinGW) won't happen again. This is not urgent since the 0.9 release is quite some time off.

szhorvat:
> As given in the commit log, all this boils down to MSVC I believe, doesn't it?

We'd like to make the decldir thing work with MinGW as well for version 0.9, but for 0.8, maybe the best solution is to just switch it off. We don't want to make big changes to 0.8 (in case we hoped that 0.8.4 would be the last 0.8.x release)

> However, I had checked it for the latest gcc version for CYGWIN (gcc version 10.2.0), MSYS2 (gcc version 10.2.0) and MSYS ([gcc version 3.4.4](https://github.com/gcc-mirror/gcc/releases/tag/releases%2Fgcc-3.4.4) from 2005). The latter suggests that the mutual exclusivity must be in for 15 years already I guess.

Thanks for this! I'll just remove the reference to Cygwin then. Cygwin is free, so there's no good reason not to use a version more recent than 2005 :-)

include/igraph_decls.h
jannick0: > Yes, this should be fixed somehow, preferably in the simplest possible way for 0.8.x.

It could well be that gcc complains about different function declaration and and definition qualifiers, i.e. probably `DECLDIR` should be consistently used in the .h and the .c files.  ... just guessing, not checked yet.     

> [...] When _using_ an igraph built as a static library on Windows, it is necessary to define `IGRAPH_STATIC` before including any headers. This switches the `DECLDIR` to empty.

Why not putting `-DIGRAPH_STATIC` into `igraph.pc` for a static library on Windows?  Alternatively, to have simultaneously static and dynamic libraries, why not shipping `igraph-config.sh` which handles this stuff, assuming that `cmake` does not have a hard time to build both.  
 
> If you have time at some point, it would be really great if you could help set up a CI workflow for MinGW, either on Appveyor, which we already use, or some other CI service. We should make sure that what happened here (broken build with MinGW) won't happen again. This is not urgent since the 0.9 release is quite some time off.

Ok, I'll try to look into that.  Let's do this first to have CI MINGW test runs before tackling the dllimport issue.

jannick0: > > If you have time at some point, it would be really great if you could help set up a CI workflow for MinGW, either on Appveyor, which we already use, or some other CI service. We should make sure that what happened here (broken build with MinGW) won't happen again. This is not urgent since the 0.9 release is quite some time off.
> 
> Ok, I'll try to look into that. Let's do this first to have CI MINGW test runs before tackling the dllimport issue.

#1559 - First version of github workflow ... tests fail due to link issue with static igraph library.

src/foreign.c
ntamas: 330f555 is now reverted; I have simply switched to Unix line endings in 4d52c2e3d. This should produce:

* LF on Windows with binary mode
* CRLF on Windows with text mode
* LF on Unix, both in binary and in text mode

ntamas:
I'm trying to wrap my head around the thread here (without much experience in the intricacies of development on/for Windows). What's the conclusion then? Shall we remove the `#if defined (__CYGWIN__)` check as suggested by @szhorvat, or shall we replace it with `#if defined (__GNUC__)` as proposed in the original PR?

If I understand correctly, the CI runs that were recently added using Github workflows is based on MINGW32/64 using a _non_-Cygwin-style compiler, so the problem is that we are hitting the part of the `#if defined(_WIN32)` branch and start adding `__declspec()` directives in the headers. The same `__declspec()` modifiers are not present in the actual source files, and the _non_-Cygwin compiler complains about that. I don't understand the reason, though; [this answer on SO](https://stackoverflow.com/a/11550679/156771) claims that the `__declspec()` directives can be omitted if the declarations are identical (and they seem to be). It seems like the compiler in MINGW32/64 insists on repeating the `__declspec()` declarations. On the other hand, @jannick0 suggests that these declarations are not needed at all for the _non_-Cygwin compiler in MINGW32/64 and the PR aims to fix that by narrowing things down to non-GNU compilers only. Am I correct?

szhorvat:
There are two cases:

 - `develop` branch, Cmake: here everything works fine at the moment. We can use decldir without problems both with MSVC and MinGW.

 - `master` branch, autotools: Making the decldir stuff work fine with autotools seems to be a bit of a challenge, so we might want to just drop it at least for MinGW. Someone needs to test if if actually works fine with MSVC or not, and based on that decide what to do exactly.  @vtraag said in chat that in order to drop decldir safely, we need to add some new compiler options for MinGW. @vtraag Can you explain the details?

We will probably end up with different `DECLDIR` definitions on the `master` and `develop` branches, and will need to be careful when merging `master` into `develop`. 

As for CygWin, I'll remove any reference to it. It seems that all non-ancient versions of CygWin will present themselves as proper Unix, so we don't need to special case them.

jannick0:
MINGW is GNUC which does not need any DECLDIR, while on Windows it is needed by MSVC (please note that any requirements [MSDN](https://docs.microsoft.com/en-us/cpp/cpp/dllexport-dllimport) suggests to be `Windows specific` does not neccessarily extend to all Windows compilers). Any other systems which need `dllimport/dllexport`?

szhorvat:
Not exporting symbols which are not meant to be public seems like generally a good idea. Though not standard on other platforms, it is possible. 

Given that everything works perfectly on the `develop` branch, and we have the additional safety provided by not exporting non-public symbols (and risking a conflict), is there any reason not to keep `DECLDIR` as it is on `develop`?

szhorvat:
OK — on the `master` branch I changed `DECLDIR` to be active only `#ifdef _MSV_VER`. On `develop` it is as before, and it works with MinGW. In both cases, I removed any references to CygWin.

Hopefully this will work.

Compilation with MSVC succeeds on the `master` branch even when building a DLL. Compilation with MinGW also seems to work. @vtraag was referring to [this MinGW doc page](http://www.mingw.org/wiki/sampledll) which mentions that certain linking options are necessary. But things appear to work without them. Maybe they are not necessary for MinGW-w64?

If there are any concerns about this solution, please let me know.

ntamas:
LGTM. I have also merged the `XML2_CFLAGS` related change. Thanks a lot for everyone!

jannick0:
> If there are any concerns about this solution, please let me know.

In case there are are any complaints `#if defined _WIN32 && ! defined __GNUC__` instead of  `#ifdef _MSC_VER` could help (see [here](https://www.gnu.org/software/libtool/manual/libtool.html#Windows-DLLs), explanation and bottom of section). Logically equivalent to my suggestion which I drafted with the goal of minimal changes.  But I cannot see any real concern.

Many thanks to you guys!

PR #1557, opened 2020-11-28
Add a function to check potential connectedness

szhorvat:
This PR adds a function to check potential connectedness of a degree sequence.

Which name do you prefer? `igraph_is_potentially_connected` or `igraph_is_potentially_connected_degree_sequence`?

This PR should be merged as normal, _not_ squash-and-merged.

### DIFF
--- a/doc/structural.xxml
+++ b/doc/structural.xxml
@@ -81,6 +81,7 @@
 <section id="degree-sequences"><title>Degree Sequences</title>
 <!-- doxrox-include igraph_is_graphical -->
 <!-- doxrox-include igraph_is_bigraphical -->
+<!-- doxrox-include igraph_is_potentially_connected_degree_sequence -->
 <!-- doxrox-include igraph_is_degree_sequence -->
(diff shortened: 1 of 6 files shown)


ntamas:
How about `igraph_is_degree_sequence_potentially_connected`? Not sure which one is better, this one seems more natural to me, but YMMV. I am okay with the other proposals as well.

szhorvat:
> `igraph_is_potentially_connected_degree_sequence()` is the way to go; it is consistent with `igraph_is_graphical_degree_sequence()`.

These are actually deprecated in favour of `is_graphical` which is short, but doesn't say anything about degree sequences.

`is_graphical` is not released yet, so we can change its name.

We could e.g. use `is_graphical_degseq` and `is_potentially_connected_degseq`. These not so horribly long.

szhorvat:
I fixed the doc build, improved the docs, and added `\experimental` for now.

szhorvat:
Closing this, as the implementation is incorrect. Will open a new PR sometime after the source reorganization is complete.

PR #1558, opened 2020-11-29
[master/autoconf] add github workflow for MSYS2's MINGW32 and MINGW64

jannick0:
This is meant as a start for CI runs on MSYS2.

Issue #1499, opened 2020-10-24
Non-robust SCG tests


szhorvat:
The following tests tend to fail when using alternative external BLAS/LAPACK implementations:

```
	251 - igraph_scg_grouping3 (Failed)
	254 - igraph_scg_semiprojectors2 (Failed)
```

Tested with OpenBLAS on macOS and MKL on Linux. It's always these two tests. They seem not to be robust, there is probably some degeneracy.

<details><summary>MKL on Linux:</summary>

```
> cat test_igraph_scg_grouping3.diff
--- /home/szhorvat/Repos/igraph/examples/simple/igraph_scg_grouping3.out	2020-10-24 15:03:18.161052000 +0200
+++ /home/szhorvat/Repos/igraph/build/tests/test_igraph_scg_grouping3.out	2020-10-24 15:59:33.453802000 +0200
@@ -1,5 +1,5 @@
 0.433013 0.57735 0.57735 0.144338 0.144338 0.144338 0.144338 0.144338 0.144338 0.144338
-0 1 1 0 2 2 2 2 2 2
+0 1 1 0 1 2 2 2 2 2
 0 1 1 2 1 1 1 1 1 1
 0 1 1 2 1 1 1 1 1 1
 0 1 1 2 3 3 3 3 3 3
```

```
> cat test_igraph_scg_semiprojectors2.diff
--- /home/szhorvat/Repos/igraph/examples/simple/igraph_scg_semiprojectors2.out	2020-10-24 15:03:18.176041000 +0200
+++ /home/szhorvat/Repos/igraph/build/tests/test_igraph_scg_semiprojectors2.out	2020-10-24 15:59:33.645804000 +0200
@@ -1,11 +1,11 @@
 ----------------------
 0.75 0 0 0.25 0 0 0 0 0 0
-0 0.5 0.5 0 0 0 0 0 0 0
-0 0 0 0 0.166667 0.166667 0.166667 0.166667 0.166667 0.166667
+0 0.444444 0.444444 0 0.111111 0 0 0 0 0
+0 0 0 0 0 0.2 0.2 0.2 0.2 0.2
 ---
 1 0 0 1 0 0 0 0 0 0
-0 1 1 0 0 0 0 0 0 0
-0 0 0 0 1 1 1 1 1 1
+0 1 1 0 1 0 0 0 0 0
+0 0 0 0 0 1 1 1 1 1
 ---
 ----------------------
 1 0 0 0 0 0 0 0 0 0
```
</details>

[...]

PR #1559, opened 2020-11-29
[develop/cmake] Github Workflow for MSYS2's MINGW32/MINGW64

jannick0:
Here as a start a github workflow for MSYS2's MINGW32/MINGW64 
* static library only
* tests fail due to unresolvable symbol with names beginning with `__imp`, so the known MINGW issue to be fixed.

To be done:
* add shared library build
* add various scenarios for -D flags (?)

### DIFF
--- a/.github/workflows/msys2-build.yml
+++ b/.github/workflows/msys2-build.yml
@@ -1,0 +1,46 @@
+name: MINGW32 / MINGW64 (MSYS2) - cmake build
+
+# TODO: Check the BRANCHES element when merging the branch 'develop' to, say, 'master'.
+on:
+  push:
+    branches:
+      - develop
+  pull_request:
+    branches:
+      - develop
+
+jobs:
+  build:
+    runs-on: windows-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        include: [
+          { msystem: MINGW32, arch: i686   },
+          { msystem: MINGW64, arch: x86_64 }
+          ]
+
+    defaults:
+      run:
+        shell: msys2 {0}
+
+    steps:
+      - name: Init ${{ matrix.msystem }}-System
+        uses: msys2/setup-msys2@v2
+        with:
+          msystem: ${{ matrix.msystem }}
+          install: git base-devel mingw-w64-${{ matrix.arch }}-cmake mingw-w64-${{ matrix.arch }}-toolchain mingw-w64-${{ matrix.arch }}-glpk mingw-w64-${{ matrix.arch }}-gmp mingw-w64-${{ matrix.arch }}-zlib mingw-w64-${{ matrix.arch }}-libxml2 mingw-w64-${{ matrix.arch }}-suitesparse
+          update: true
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Configuration
+        run: |
+          mkdir -p build-${{ matrix.arch }}
+          cd build-${{ matrix.arch }}
+          cmake .. -G 'MSYS Makefiles' -DIGRAPH_GLPK_SUPPORT=1 -DIGRAPH_GRAPHML_SUPPORT=1 -DIGRAPH_VERIFY_FINALLY_STACK=1
+      - name: Build
+        run: make
+        working-directory: build-${{ matrix.arch }}
+      - name: Test
+        run: make check
+        working-directory: build-${{ matrix.arch }}


szhorvat:
Sorry about the stupid question, but do we need to enable Github Actions to see this work? Or do I just not see the link to the results? @ntamas, can you take a look?—I assume it needs admin access.

jannick0:
I cannot remember what exactly I did to have workflows being fired up. But try to look into `Settings > Actions > ...` to enable the feature for the repo. 

Then look into `Actions` which should show the list of repo workflows, more [here](https://docs.github.com/en/free-pro-team@latest/actions/managing-workflow-runs/disabling-and-enabling-a-workflow).  Not sure if workflows committed prior to enabling the feature immediately work, though.

HTH.

ntamas:
> do we need to enable Github Actions to see this work?

They are enabled, but I think we won't see the build results in `develop` until we actually merge the PR. Until then, we can take a look at the results in @jannick0 's branch.

ntamas:
Is #1558 a duplicate of this?

jannick0:
Sorry about the confusion: No, it is meant for the `develop` branch using `cmake`. To make things more clear, I just changed the issue title.

szhorvat:
#1558 Was for the master branch with autotools, this is for the develop branch with cmake.

ntamas:
LGTM. @szhorvat , is it okay if I merge this? I know that we are currently trying to figure out which CI system to use in the long run, but this would be quite useful until we manage to fix the MINGW32/MINGW64 issues.

jannick0:
Could you please wait merging, since I believe the CI config needs branch restrictions?  I am working on this. Thx.

szhorvat:
@ntamas Sorry, I thought I responded already. Yes, we should absolutely merge this when Jannick is ready. But it is possible that someone with admin access to the igraph organizations needs to enable the GitHub Actions CI before it starts working. Can you please look into this?

With CI services like Travis restricting the amount of services that OSS projects can use, it might be best if we split the runs across several CI services. We'll also get results faster, plus if one breaks (as it sometimes happens), the other ones will still be working. Its a win-win (we take less resources from individual providers, but gain some robustness).

jannick0:
@ntamas Ready to go.

* The workflow is restricted to the branch 'develop', although it exists on this branch only.  But I have no idea how github workflows are fired up, so this restriction should be helpful for safety reasons if other branches are forked from 'develop'.
* Since the branch in my repo underlying this PR has a name other than 'develop', the workflow is not fired up in my repo - I just checked that before the force-push.

... so, fingers crossed. Still the MINGW issue does exist, such that we should not be disappointed that the builds will fail, but they should start.

ntamas:
Let's roll then! Thanks!

jannick0:
Great - works: CI is fired up. ... Now we can try to tackle the MINGW issue.

szhorvat:
> Now we can try to tackle the MINGW issue.

It's fixed now on `develop`, see https://github.com/igraph/igraph/commit/a4d7f2630443a9f2a369cad0f5747dee38acb7a9

But I don't quite understand why this ever worked at all with MSVC ... @vtraag, any tips?

Note: the scg test failures are known, and not related to MinGW, see #1499

vtraag:
> But I don't quite understand why this ever worked at all with MSVC ... @vtraag, any tips?

No, I don't know. There were some warnings in the [log](https://ci.appveyor.com/project/ntamas/igraph/builds/36579967/job/hdnpv395692nx8vw) though

```
[00:06:40] sampling.obj : warning LNK4049: locally defined symbol igraph_rng_default imported [C:\projects\igraph\build\src\igraph.vcxproj]
[00:06:40] sampling.obj : warning LNK4217: locally defined symbol igraph_rng_get_integer imported in function plfit_walker_alias_sampler_sample [C:\projects\igraph\build\src\igraph.vcxproj]
[00:06:40] sampling.obj : warning LNK4217: locally defined symbol igraph_rng_get_unif imported in function plfit_runif [C:\projects\igraph\build\src\igraph.vcxproj]
[00:06:40] sampling.obj : warning LNK4217: locally defined symbol igraph_rng_get_unif01 imported in function plfit_rpareto [C:\projects\igraph\build\src\igraph.vcxproj]
```

All igraph calls  in `plfit/random.c` are done through a MACRO expansion. Perhaps the name mangling is done at some other stage in the MSVC compiler than with MinGW? 

At any rate, `IGRAPH_STATIC` should indeed be defined for all object libraries, including `plfit`, so [your solution](https://github.com/igraph/igraph/commit/a4d7f2630443a9f2a369cad0f5747dee38acb7a9) looks good to me. There's just one thing I don't understand: why do you include this definition in `prpack/CMakeList.txt`? Shouldn't this be included in `plfit/CMakeList.txt`? It might be just lucky that this works because `plfit` is included before `prpack` in `src/CMakeList.txt`.

szhorvat:
> There's just one thing I don't understand: why do you include this definition in `prpack/CMakeList.txt`?

Thanks for noticing. I was comparing the two files and somehow ended up putting it into the wrong one without noticing it. I fixed it.

PR #1561, opened 2020-12-02
Replace assert() by IGRAPH_ASSERT()

szhorvat:
As the title says, this PR replaces `assert()` by `IGRAPH_ASSERT()` in the sources, except in C++ code and in source borrowed from elsewhere (e.g. bliss).

The one exception is a minor variables renaming in `cliques.c`.

### DIFF
--- a/src/blas.c
+++ b/src/blas.c
@@ -22,10 +22,9 @@
 
 */
 
+#include "igraph_error.h"
 #include "igraph_blas.h"
 #include "igraph_blas_internal.h"
-
-#include <assert.h>
 
 /**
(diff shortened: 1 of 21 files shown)


szhorvat:
Added another commit disabling the `unused-variable` warning for Bliss and mini-gmp. This warning comes up when `NDEBUG` is defined. Some variables were only used by `assert`. In this code, I did not change `assert` to `IGRAPH_ASSERT`: Bliss is C++, so it's not appropriate there, and mini-gmp should not be touched by us.

ntamas (approved):

PR #1562, opened 2020-12-03
Remove functions that were deprecated in 0.6

szhorvat:
This PR removes functions that were deprecated in 0.6.

There was a disabled test for `igraph_es_adj`, but it seems to have been written for a very old version of igraph, and did not work anymore. I considered fixing and re-enabling it (for the replacement of `igraph_es_adj`, i.e. `igraph_es_incident`), but it looks like it's less work to write a new test from scratch. So I didn't fix it.


ntamas (approved):

ntamas:
LGTM.

vtraag (approved):

vtraag:
It seems that we missed a reference to `igraph_subgraph` in `structure_generators.c:45`, which we should still delete.

ntamas:
Removed.

PR #1564, opened 2020-12-03
Add Azure Pipelines

szhorvat:


szhorvat:
@vtraag I'm done with this for now. There are several improvements one could make but I'm lost in the docs. You've used Azure before, so please take a look, and if you're happy, then squash-and-merge this.

vtraag:
Great work! It looks to build quite nice, and fast! Perhaps we could also add Windows builds?

vtraag:
One minor suggestion though: perhaps we can simply write out fully `internal` instead shortening it to `int` to avoid any confusion?

szhorvat:
I would keep it short. The whole point of the template syntax was to make everything more compact, and this will mostly be used by us, and never by users. I can add a clarifying comment to the template though.

vtraag:
> this will mostly be used by us, and never by users

True of course, but my own code might still confuse me later as well :smile:

PR #1565, opened 2020-12-04
CI MINW32/MINGW64(autoconf): github workflow enhancements

jannick0:
A mixed commit series in tiny steps covering mostly CI MINGW32/MINGW64 (autoconf):  

* It appears that a github action is only triggered on a branch when the workflow file exists on that branch. So no need to restrict any branch names here.
* Branches with names starting with `github/`, `travis/` or `appveyor/` trigger the mentioned CI only or - as implemented - are ignored by all CIs other than the mentioned one.
* Fix to EOL issue (Windows).  

The implemented name convention differs from the logic suggested in the discussion  of #1558. I am open for discussion, of course. Please edit as you may find appropriate.     

NB: `git` prohibits brach names composed as `<existing branch name>/<feature>` (e.g. a branch name like `master/abc` is rejected when `master` already exists).  

More details in the commit messages.

TODO: Once this is in, badges could be added to `README.md`.


ntamas:
Thanks a lot!

jannick0:
One of the two jobs fail because of `git fetch --prune --unshallow` which might compete with the same of the other job.  If this persists, then `git fetch --prune --unshallow || true` might help.

The same could be added to the `develop` branch.  I am running out of time, so I am wondering if one of you guys could amend the github workflow there.  That would be helpful.

jannick0:
actions/checkout#400 - feature request to add fetching tags into `action/checkout`.

vtraag:
@jannick0, not sure if this helps, but in another project I used the following piece for GitHub actions to make sure we have all commits necessary to identify a version from `git` tags.
```
    - uses: actions/checkout@v2
      with:
        fetch-depth: 0
```

ntamas:
Merged this PR into `develop` and tried to reconcile the changes with CMake. We'll see how it goes.

Issue #1322, opened 2020-02-06
All graph generators should check their input for validity


szhorvat:
There are multiple graph generators that currently do not check their input for validity.

These need to be fixed:

 - [x] `callaway_traits_game`
 - [x] `establishment_game`

The following functions have been verified to check their arguments properly:

```
forest_fire_game
```

Please update the lists above as you verify functions or find new ones with problems.

This is marked as "good first issue" because the fixes are usually small and relatively easy, but they are sometimes time-consuming: one needs to read up on and understand the random graph model.

szhorvat:
@YasirKusay, if you're looking for something very easy, but somewhat boring, you can check some of these.

siddharth25pandey:
In which file the multiple graph generator is present?

vtraag:
The graph generation is implemented in [`src/games.c`](https://github.com/igraph/igraph/blob/master/src/games.c).

siddharth25pandey:
What changes we have to do in callaway_traits_game & establishment_game functions?

szhorvat:
Most of the work is not writing code here, but understanding precisely what each function does, thinking about what is valid and invalid input, and checking if the function already verified the input.

szhorvat:
The list in the OP is not exhaustive. It contains functions for which it was confirmed that checks are missing. There are certainly several other functions.

To take `igraph_callaway_traits_game`, it is obvious from the code that there are no checks, and there is even a TODO note.

If you look at the [docs](https://igraph.org/c/doc/igraph-Generators.html#igraph_callaway_traits_game), a few obvious things that should be checked that jump to mind:

 - `type_dist` and `pref_matrix` size consistent with number of types (`types`)?
 - `type_dist` entires non-negative? Are zero entires okay or not (to be determined)?
 - same as above for `pref_matrix` entires
 - `pref_matrix` symmetry consistent with `directed`?
 - okay to have `types == 0` or not?
 - etc.

This would require much more careful thought, of course.

If you would like to contribute, it will be very welcome @siddharth25pandey . I'd suggest to review the function and write a list of requirements (similar to the above, but properly checked) before implementing the checks.

vtraag:
By the way, @siddharth25pandey, feel free to open a [draft PR](https://github.blog/2019-02-14-introducing-draft-pull-requests/) so that we can assist you while coding.  It is not necessary to first try to finish everything before opening a PR. If you have any difficulties, feel free to ask here for any support!

StutiV:
[...]

PR #1567, opened 2020-12-07
added parameter check in igraph_callaway_traits_game

Keivin98:
Performed a parameter check based on the issue #1322.

- type_dist and pref_matrix size consistent with the number of types (types)?
- type_dist entires non-negative? 
- types certainly must not be negative.
- nodes must be non-negative.


src/games.c
```diff
@@ -1691,7 +1691,28 @@ int igraph_callaway_traits_game (igraph_t *graph, igraph_integer_t nodes,
     igraph_real_t maxcum;
     igraph_vector_t nodetypes;
 
-    /* TODO: parameter checks */
+    //parameter check
```
szhorvat, line 1694: Please use `/* ... */` style comments.

src/games.c
```diff
@@ -1691,7 +1691,28 @@ int igraph_callaway_traits_game (igraph_t *graph, igraph_integer_t nodes,
     igraph_real_t maxcum;
     igraph_vector_t nodetypes;
 
-    /* TODO: parameter checks */
+    //parameter check 
+    if(nodes < 0){
+        IGRAPH_ERROR("Invalid number of vertices", IGRAPH_EINVAL);
```
szhorvat, line 1696: It's best to be clear about _why_ a value is invalid. I would suggest "The number of vertices must be non-negative". Of course, it is obvious that it should be non-negative, but this clearly communicates to the user that all non-negative values are acceptable.

src/games.c
```diff
@@ -1691,7 +1691,28 @@ int igraph_callaway_traits_game (igraph_t *graph, igraph_integer_t nodes,
...
+    }
+    
+    if (types < 1) {
+        IGRAPH_ERROR("types must be >= 1", IGRAPH_EINVAL);
```
szhorvat, line 1700: Again, it's best to spell out the problem: "The number of vertex types must be at least 1". This message will be seen by users of high-level interfaces, and it may not be clear at all to them what "types" means, as there may not be a parameter with this name.

src/games.c
```diff
@@ -1691,7 +1691,28 @@ int igraph_callaway_traits_game (igraph_t *graph, igraph_integer_t nodes,
...
+    }
+
+    if (igraph_vector_size(type_dist) != types) {
+        IGRAPH_ERROR("Number of types is not consistent to type_dist", IGRAPH_EINVAL);
```
szhorvat, line 1704: "The vertex type distribution vector must agree in length with the number of types"

The error concerns `types_dist`, so let's put it first. Again, we spell out what `types_dist` is, for the sake of high-level interfaces.

src/games.c
```diff
@@ -1691,7 +1691,28 @@ int igraph_callaway_traits_game (igraph_t *graph, igraph_integer_t nodes,
...
+    }
+    
+    if (igraph_matrix_size(type_dist) != types * types) {
+        IGRAPH_ERROR("Size of matrix is not consistent to type_dist", IGRAPH_EINVAL);
```
szhorvat, line 1708: Please check that the matrix is square (i.e. number of rows and columns must both be the same as `types`).

src/games.c
```diff
@@ -1691,7 +1691,28 @@ int igraph_callaway_traits_game (igraph_t *graph, igraph_integer_t nodes,
...
+    }
+
+    for (i = 0; i < types; i++) {
+        if(VECTOR(*type_dist)[i] < 0){
+            IGRAPH_ERROR("Negative component in type_dist", IGRAPH_EINVAL);
```
szhorvat, line 1713: Spell out what `type_dist` is.

szhorvat (changes requested):
Thanks for taking the time to submit a PR!

I left a few comments on the code. Can you make these changes please?

Additionally, remove all changes from the PR that affect files other than `games.c`.

Finally, if we do add the checks, let's make them complete:

 - Check the elements of the preference matrix and ensure that they are between 0 and 1 (inclusive). They must be probabilities.

 - When the graph is undirected, check that the preference matrix is symmetric.

Keivin98:
Thank you for the comments @szhorvat! I modified the code according to your instructions.

src/games.c
```diff
@@ -1691,7 +1691,47 @@ int igraph_callaway_traits_game (igraph_t *graph, igraph_integer_t nodes,
...
+    }
+    
+    for (i = 0; i < types; i++) {
+        if(VECTOR(*type_dist)[i] < 0){
+            IGRAPH_ERROR("Negative component in the vertex type distribution vector", IGRAPH_EINVAL);
+        }
+        igraph_vector_t row_i;
+        igraph_matrix_get_row(pref_matrix, &row_i, i); 
+        for(j = 0; j < types; j++){
+            if(VECTOR(row_i)[j] < 0 || VECTOR(row_i)[j] > 1){
```
ntamas, line 1721: We have `igraph_matrix_minmax()` to retrieve the smallest and largest elements of the matrix; please use that to test whether all elements of the matrix fall between 0 and 1.

src/games.c
```diff
@@ -1691,7 +1691,47 @@ int igraph_callaway_traits_game (igraph_t *graph, igraph_integer_t nodes,
...
+                IGRAPH_ERROR("Probability of index in preference matrix not in [0,1]", IGRAPH_EINVAL);
+            } 
+        }
+    }
+
+    if(!directed){
+        igraph_matrix_t pref_matrix_copy;
+        igraph_matrix_copy(pref_matrix, &pref_matrix_copy);
```
ntamas, line 1729: `igraph_matrix_is_symmetric()` can be used to test whether the matrix is symmetric; no need to make a copy.

ntamas (changes requested):
I have also proposed a few changes that would re-use existing igraph functions instead of implementing the sanity checks from scratch.

There are still four extra files in the PR that are unrelated to the changes proposed in the PR. I have never really found a good way to remove unrelated files from a PR without having to create a new PR from scratch -- @szhorvat, any ideas about how this could be done easily?

vtraag:
>  I have never really found a good way to remove unrelated files from a PR without having to create a new PR from scratch

We could push a commit to this PR that deletes the extra files, and then squash-and-merge?

szhorvat:
The wording of some error message still needs to be changed, so please do not merge yet, even after removing the accidentally committed changes (I don't have time to fix that now).

ntamas:
> We could push a commit to this PR that deletes the extra files, and then squash-and-merge?

The problem is that these are not really "extra files" -- they are alreaedy in the repo but the modifications do not belong to this PR. So one would need to commit a patch to this PR that reverts the extra files, and then squash-and-merge at the end.

szhorvat:
Maybe it's cleanest to start a new PR then.

szhorvat:
@Keivin98 Can you please start a new PR where you do not include any other changes than to the file `games.c`? An alternative solution is that we include the changes manually, and not through this PR, but then it would not be recorded you _your_ contribution by GitHub. Let me know how you want to proceed.

szhorvat:
@Keivin98 Let us know if you want to open a new PR with the unneeded files excluded. The purpose of this would be so that you 'get credit', as GitHub will records your contribution. Otherwise, I can merge the changes manually. Let me know how you want to proceed.

vtraag:
@szhorvat, wouldn't it be the easiest to simple revert the files that should not have been touched in this PR?

vtraag:
I'd be happy to do this tomorrow btw.

ntamas (approved):

PR #1576, opened 2020-12-18
betweenness: improve precision and speed

szhorvat:
Changes in this PR:

 - Betweenness functions now use `double` instead of `long long` to count paths. This avoids overflow.
 - Betweenness tests are moved to `tests/unit`; `VERIFY_FINALLY_STACK` is added to each; some tests are improves/extended.
 - Support for using bigints to count paths is removed.  I verified that the previous implementation was in fact _less_ precise because dividing two bigints to get a `double` was done as `bigint a, b; bigint c = (a * shift) / b; double res = double(c) / shift` where `shift  = 1000000000`. This shift was not large enough for good precision. After correcting that, the results between bigint and `double` agree to 14 decimal digits.
 - `nobigint` arguments are removed from all functions
 - All bigint functionality is stripped from igraph. It was only used for betweenness. If we need bigints in the future, we will use GMP instead, which is much faster.


szhorvat:
@ntamas  I suggest a plain merge, _not_ squash-and-merge.

ntamas (approved):

ntamas:
Thanks a lot!

PR #1583, opened 2020-12-21
Implement harmonic centrality

### LINKED ISSUE #1373: Implement "harmonic centrality"
The harmonic centrality of a vertex is the average inverse distance to all other vertices.

https://en.wikipedia.org/wiki/Closeness_centrality#In_disconnected_graphs

https://igraph.discourse.group/t/user-defined-distance-for-unconnected-node-in-closeness-and-harmonic-centrality/183

szhorvat:
Changes in this PR:

 - Break up `centrality_other.c` into smaller parts and remove unneeded headers
    * Update copyright notice in these files
 - Implement harmonic centrality: `igraph_harmonic_centrality` and `igraph_harmonic_centrality_range_limited`. Fixes #1373.
 - Made sure that `igraph_vector_min` is not called on empty vectors in betweenness/closeness calculations. This would result in a crash.
 - Replace `igraph_closeness_estimate` with `igraph_closeness_cutoff`:
    * `igraph_closeness_estimate` is retained, but deprecated. Now it issues a warning.
    * `igraph_closeness_cutoff` has the `cutoff` argument moved to the end.
 - Also replace `igraph_(edge)_betweenness_estimate` with `igraph_(edge)_betweenness_cutoff` and move `cutoff` to the end. The original functions are deprecated.
 - Remove the unneeded comparison-with-tolerances from the closeness calculation. A small error will not affect the result here (with betweenness it will, so using tolerances was important there).


szhorvat:
@ntamas This is not yet done, but I need input to proceed. I noticed that weights closeness uses comparison with tolerance to determine if two paths have the same length:

https://github.com/igraph/igraph/pull/1583/files#diff-3a9a78ccf7f2683f60b049001de999de2e35b3547aba20517c2f41a3fd31e97fR131

https://github.com/igraph/igraph/pull/1583/files#diff-3a9a78ccf7f2683f60b049001de999de2e35b3547aba20517c2f41a3fd31e97fR144-R146

https://github.com/igraph/igraph/pull/1583/files#diff-3a9a78ccf7f2683f60b049001de999de2e35b3547aba20517c2f41a3fd31e97fR194-R199

This seems pointless to me. It does not matter if we choose the "wrong" path as the shortest one due to numerical errors, as closeness is defined based on path length. If the error in the path length is tiny, then the error in closeness will be tiny. Am I missing something important?

The betweenness calculation also has this. There, it is indeed very necessary because the betweenness value depends on _which_ is the shortest path, not its _length_. Small errors can change the result dramatically.

In order to implement harmonic centrality, I copied and edited the closeness sources. I think it makes sense to remove this comparison-with-tolerance, and go with a normal comparison for the sake of simplicity and performance. I would also do this for plain closeness. Agreed?

ntamas:
The links that you provided are broken (they all point to this PR in general, the anchors have no effect), but I think I understand the problem that you mentioned, and I think that historically, the weighted closeness and the betweenness implementations were developed in parallel or close to each other in time, and the closeness centrality code simply copied code that worked from betweenness without thinking through whether the tolerance check is needed or not. I agree with removing the tolerance check from the closeness centrality implementation.

szhorvat:
@ntamas Please take a very careful look, but do not squash-and-merge, (just merge or leave it to me).

szhorvat:
The deprecated functions are no inserted in the docs at the moment. I can insert them, as the doc comments are retained. Let me know what you think about this.

ntamas:
Just a quick heads up that I've started reviewing this and that there is no need to keep the deprecated functions in the docs.

ntamas:
Finished with all the files except `closeness.c`; I'm starting that now.

As for the test cases, I trust that you calculated harmonic centrality for these graphs in some other tool (e.g., Mathematica?) and we are using those values in the test cases, right?

szhorvat:
Yes, I checked the values.

ntamas (approved):

ntamas:
This is great, thanks for your hard work on this! I'm not merging as requested above, I'm leaving it up to you.

PR #1584, opened 2020-12-21
Added fuzzer to integrate igraph into OSS-fuzz

AdamKorcz:
This PR adds a fuzzer with build script to run it continuously through OSS-fuzz.

Fuzzing is a way of testing programs whereby pseudo-random data is passed to a target application with the goal of finding bugs and vulnerabilities. I can see that igraph previously has been fuzzed by the community.

OSS-fuzz is a project by Google into which open-source projects can integrate and have their fuzzers run continuously. If bugs are found, maintainers get notified with an email containing a link to a detailed bug report with stack trace and reproducible test case.

I will shortly set up the integration on the OSS-fuzz side. All that is needed to complete the integration is to:

1. Merge in the files in this PR
2. Add at least one maintainers email address for potential bug reports.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,22 @@
 # igraph C library changelog
 
+## [0.8.5]
+
+### Added
+
+### Changed
+
+ - `igraph_write_graph_pajek()`: the function now always uses the platform-native line endings (CRLF on Windows, LF on Unix and macOS). Earlier versions tried to enforce Windows line endings, but this was error-prone, and since all recent versions of Pajek support both line endings, enforcing Windows line endings is not necessary any more.
+
+### Fixed
+
+ - Fixed several compilation issues with MINGW32/64 (PR #1554)
+ - `igraph_layout_davidson_harel()` was not interruptible; now it is.
+ - Added a missing memory cleanup call in `igraph_i_cattribute_combine_vertices()`.
+ - Fixed a few memory leaks in test cases.
+
+### Other
+
 ## [0.8.4]
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,16 @@
 # igraph C library changelog
 
-## [0.8.5]
+## [Unreleased]
 
 ### Added
+
+### Changed
+
+### Fixed
+
+### Other
+
+## [0.8.5] - 2020-12-07
 
 ### Changed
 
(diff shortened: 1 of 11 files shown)


AdamKorcz:
The integration files on the OSS-fuzz side have been added here: https://github.com/google/oss-fuzz/pull/4877

szhorvat:
Hello @AdamKorcz, thanks for this nice contribution! Currently, all development happens on the `develop` branch, which will become igraph 0.9. We have switched to a CMake-based build system. Build instructions are roughly: 

```
mkdir build && cd build
cmake ..
make
```

More info [here](https://igraph.discourse.group/t/draft-document-for-installation-with-cmake/504).

`libigraph.a` is created in `build/src`.

Would you be able to re-base this PR against `develop` and make sure that oss-fuzzer uses that branch?

szhorvat:
Alternatively, we can do it on `master` first then I'll merge this into `develop` with all the necessary changes. I'll get back to this later today. Feel free to add my email (as shown in GitHub) in the other PR. Perhaps @ntamas would prefer his to be included as well?

AdamKorcz:
I would be happy to merge this is an the merge it into master. I will get your email address added @szhorvat

AdamKorcz:
Closing to reopen

szhorvat:
@AdamKorcz Why did you use `if(size<5) return 0;`? I would like to understand if there is any good reason for this before adding more fuzzers.

AdamKorcz:
> @AdamKorcz Why did you use `if(size<5) return 0;`? I would like to understand if there is any good reason for this before adding more fuzzers.

Generally there may be some false positives in the case of short byte arrays. I cannot remember whether it applies to igraph though. If you add more fuzzers, you might be able to leave that out.

szhorvat:
@AdamKorcz I removed the limits and so far everything seems to be good.

Another question:

Do you know why [this performance report](https://oss-fuzz.com/performance-report/libFuzzer_igraph_bliss_fuzzer/libfuzzer_asan_igraph/latest) complains about logging? The log files linked from the report do not contain any messages written by the program. In fact I do not think this program is outputting anything at all (or at least it shouldn't).

szhorvat:
@AdamKorcz Is there a way to completely remove any generated corpus for a target and start from zero again?

I was experimenting with adding targets for some of the more complicated functions that process graphs. These need to receive the graph in some appropriate binary format that the fuzzer can mutate in a useful way. I think my first attempt was not very successful, and the graph format should be changed. This would make any existing corpus useless.

(In fact, I am not sure that fuzzing such functions is useful at all, even with a better input format and custom mutators. But it's always worth experimenting a bit. If you don't seek, you won't find anything.)

PR #1587, opened 2020-12-26
external SuiteSparse library suite

jgmbenoit:
I understand that externalizing SuiteSparse is already done in the coming version 0.9 (develop branch).
This is mainly for the record as it can be useful to others meanwhile.

Description: external SuiteSparse library suite
 Allow to use the system wide SuiteSparse library suite
 instead of using the one furnished in the upstream source.
Origin: debian
Last-Update: 2020-12-26


ntamas (approved):
Thanks a lot!

PR #1588, opened 2020-12-30
Added fuzzer and OSS-fuzz build script

AdamKorcz:
Reopens https://github.com/igraph/igraph/pull/1584 - I apologize for double-posting. Rebasing to `develop` in last PR became a little messy.

This PR adds a fuzzer with build script to run it continuously through OSS-fuzz.

Fuzzing is a way of testing programs whereby pseudo-random data is passed to a target application with the goal of finding bugs and vulnerabilities. I can see that igraph previously has been fuzzed by the community.

OSS-fuzz is a project by Google into which open-source projects can integrate and have their fuzzers run continuously. If bugs are found, maintainers get notified with an email containing a link to a detailed bug report with stack trace and reproducible test case.

The integration files on the OSS-fuzz side have been added here: google/oss-fuzz#4877

### DIFF
--- a/fuzzing/build.sh
+++ b/fuzzing/build.sh
@@ -1,0 +1,16 @@
+#!/bin/bash -eu
+
+sed 's/use_all_warnings/\#use_all_warnings/g' -i $SRC/igraph/src/CMakeLists.txt
+mkdir build && cd build
+cmake ..
+make -j$(nproc)
+
+# Create seed corpus
+zip $OUT/read_lgl_fuzzer_seed_corpus.zip \
+        $SRC/igraph/examples/simple/karate.gml
+
+cd $SRC/igraph
+
+$CC $CFLAGS -I$SRC/igraph/build/include -I$SRC/igraph/include -o read_lgl_fuzzer.o -c ./fuzzing/read_lgl_fuzzer.c
+$CC $CFLAGS $LIB_FUZZING_ENGINE read_lgl_fuzzer.o \
+        -o $OUT/read_lgl_fuzzer ./build/src/libigraph.a
--- a/fuzzing/read_lgl_fuzzer.c
+++ b/fuzzing/read_lgl_fuzzer.c
@@ -1,0 +1,57 @@
+/*
+   IGraph library.
+   Copyright (C) 2021  The igraph development team
+   
+   This program is free software; you can redistribute it and/or modify
+   it under the terms of the GNU General Public License as published by
+   the Free Software Foundation; either version 2 of the License, or
+   (at your option) any later version.
+   
+   This program is distributed in the hope that it will be useful,
+   but WITHOUT ANY WARRANTY; without even the implied warranty of
+   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+   GNU General Public License for more details.
+   
+   You should have received a copy of the GNU General Public License
+   along with this program; if not, write to the Free Software
+   Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
+   02110-1301 USA
+*/
+
+#include <stdint.h>
+#include <string.h>
+#include <stdlib.h>
+#include "igraph.h"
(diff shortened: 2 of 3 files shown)


fuzzing/read_lgl_fuzzer.c
```diff
@@ -0,0 +1,57 @@
+/*
+  Heuristic graph coloring algorithms.
+  Copyright (C) 2020 Szabolcs Horvat <szhorvat@gmail.com>
```
ntamas, line 3: Please fix the header and the copyright notice here.

fuzzing/build.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash -eu
+
+sed 's/use_all_warnings/\#use_all_warnings/g' -i $SRC/igraph/src/CMakeLists.txt
+mkdir build && cd build
+cmake ..
+make -j$(nproc)
```
ntamas, line 6: FYI, if the build machine has `ninja`, you can use `cmake -GNinja ..` followed by `ninja -$(nproc)`; it is much faster on my machine than `make`. Not sure whether it matters in the build env, though.

ntamas (approved):

ntamas:
Thanks a lot -- see my two minor comments above. Please fix the copyright notice in the newly added file and then I'll merge.

fuzzing/build.sh
AdamKorcz: Good to know. I suggest checking that out in a separate iteration on this as there might be other modifications required to build with Ninja. Let's merge this commit in first and then I will be happy to check it out.

AdamKorcz: Ideally though, the next step should be integrating the fuzzers into igraphs build system. I tried the make build just now, and it definitely builds quickly enough on the OSS-fuzz platform.

ntamas (approved):

ntamas:
Thanks a lot!

ntamas:
@AdamKorcz seeing the crash reports from today, I realized that the logic in `fuzzing/read_lgl_fuzzer.c` might be incorrect (and it should be called `read_gml_fuzzer.c` as it is targeting the `igraph_read_graph_gml` function). The problem is that in case of an invalid input, it attempts to call `igraph_destroy(&g)` anyway, even if `igraph_read_graph_gml()` returned an error code. When `igraph_read_graph_gml()` returns an error code, the graph object is left in an undefined state so it is not safe to call `igraph_destroy(&g)` on it; what we should do instead is as follows:

* if `igraph_read_graph_gml()` returns zero, the code should proceed and call `igraph_destroy(&g)` as before.

* if `igraph_read_graph_gml()` returns nonzero, the code should _not_ call `igraph_destroy(&g)`.

* in both cases, it should somehow be checked that there was no memory leaked, even if the input was invalid (in other words, `igraph_read_graph_gml()` must clean up after itself for invalid inputs).

Questions:

1. If I modify `fuzzing/read_lgl_fuzzer.c` in the repo, will that be picked up automatically by OSS-fuzz?

2. Can I mark today's reports as "stale" somehow, at least for those where I suspect this could be the cause of the problem (i.e. incorrect logic in the fuzzer, not a bug in igraph's code).

AdamKorcz:
> If I modify fuzzing/read_lgl_fuzzer.c in the repo, will that be picked up automatically by OSS-fuzz?

Yup.

> Can I mark today's reports as "stale" somehow, at least for those where I suspect this could be the cause of the problem (i.e. incorrect logic in the fuzzer, not a bug in igraph's code).

Yes, there should be some command to leave in the comment on the bug report, however I can't remember what it is. The OSS-fuzz maintainers should answer quickly if you raise an issue over there. I am sure they will be happy to help.

Your suggestions sound good to me  from a fuzzing-perspective as well.

szhorvat:
@AdamKorcz Thanks so much for this contribution, and thanks @ntamas for merging it and fixing the bugs it revealed! It took me longer to get back to igraph in the new year than I expected.

PR #1589, opened 2021-01-04
Add support for AFL++

AdamKorcz:
This fixes the broken afl++ build: https://github.com/google/oss-fuzz/pull/4877

### DIFF
--- a/fuzzing/read_lgl_fuzzer.c
+++ b/fuzzing/read_lgl_fuzzer.c
@@ -24,6 +24,7 @@
 #include "igraph.h"
 #include <stdio.h>
 
+extern "C" 
 int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size){
 	if(size<5) return 0;
 
@@ -41,7 +42,7 @@ int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size){
 	FILE *ifile;
 	ifile = fopen("/tmp/libfuzzer.gml", "r");
 	if(ifile == 0){
-		unlink(filename);
+		remove(filename);
 		return 0;
 	}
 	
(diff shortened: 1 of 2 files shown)


fuzzing/read_lgl_fuzzer.c
```diff
@@ -41,7 +43,7 @@ int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size){
 	FILE *ifile;
 	ifile = fopen("/tmp/libfuzzer.gml", "r");
 	if(ifile == 0){
-		unlink(filename);
```
ntamas, line 44: What's wrong with `unlink()`? Is it not supported with AFL++?

Also, I'm not familiar with `std::__fs` but the double-underscore seems to imply to me that this is an implementation detail. If we have to use `std::` stuff, can't we get away with `std::filesystem::remove_all()` instead?

AdamKorcz: We are compiling the file as C++ instead of as C and `unlink()` is therefore not supported.

> can't we get away with `std::filesystem::remove_all()` instead?

Unfortunately not. Tested locally just now.

ntamas: Okay, how about [`remove()`](https://en.cppreference.com/w/c/io/remove) then? That one is supposed to be portable since C89/C90 and is equivalent to `unlink()`. Sorry for nitpicking, I'm just generally uneasy about using internal stuff.

AdamKorcz: remove() works fine locally. It has been added!

ntamas:
Thanks, merged!

AdamKorcz:
thanks! @ntamas Would you like to be on the mailing list for bug reports?

ntamas:
Yes please.

AdamKorcz:
What is your email address?

ntamas:
ntamas@gmail.com

AdamKorcz:
Hereby added!

Issue #1592, opened 2021-01-08
Public functions that are never used/tested


szhorvat:
This is a list of public non-deprecated functions that are never used either directly or indirectly in the main codebase or the tests/examples.

Tests should be written for some of these functions.

If you are looking to contribute to igraph, fixing some of these is an easy way to get started. If you are unsure which one to take, just ask. 

 - [ ] igraph_adjlist_fprint
 - [ ] igraph_inclist_print
 - [ ] igraph_inclist_fprint
 - [ ] igraph_arpack_unpack_complex
 - [ ] igraph_attribute_combination_remove
 - [ ] igraph_is_bipartite
 - [ ] igraph_bipartite_game
 - [ ] igraph_hub_score
 - [ ] igraph_authority_score
 - [ ] igraph_maximal_cliques_file
 - [ ] igraph_clique_size_hist
 - [ ] igraph_compare_communities
 - [ ] igraph_split_join_distance
 - [ ] igraph_le_community_to_membership
 - [ ] igraph_modularity_matrix
 - [ ] igraph_complex_csc
 - [ ] igraph_complex_cot
 - [ ] igraph_full_citation
 - [ ] igraph_extended_chordal_ring
 - [ ] igraph_linegraph
 - [ ] igraph_kautz
 - [ ] igraph_get_adjacency_sparse
 - [ ] igraph_eigen_adjacency
 - [ ] igraph_eigen_laplacian
 - [ ] igraph_laplacian_spectral_embedding
 - [ ] igraph_dim_select
 - [ ] igraph_sir
 - [ ] igraph_graphlets
 - [ ] igraph_solve_lsap
 - [ ] igraph_maximum_matching
 - [ ] igraph_motifs_randesu_estimate
 - [ ] igraph_motifs_randesu_no
 - [ ] igraph_dyad_census
 - [ ] igraph_list_triangles
 - [ ] igraph_neighborhood_size
 - [ ] igraph_neighborhood
 - [ ] igraph_neighborhood_graphs
 - [ ] igraph_local_scan_k_ecount
 - [ ] igraph_local_scan_k_ecount_them
 - [ ] igraph_local_scan_neighborhood_ecount
 - [ ] igraph_scg_norm_eps
 - [ ] igraph_stack_ptr_destroy_all
 - [ ] igraph_random_spanning_tree
 - [ ] igraph_rewire
 - [ ] igraph_unfold_tree
 - [ ] igraph_is_mutual
 - [ ] igraph_is_chordal
[...]

PR #1594, opened 2021-01-09
Add unit test for st_mincut_value

GroteGnoom:
Part of issue #1592.

I was planning to test the actual cut instead of just the value, but usually there are multiple min cuts possible and I didn't want to have a failed test when a different min cut is returned. (A test with only one possible min cut can of course still be added)

The functionality is already tested in maxflow, but this is a simple check to see that everything is correctly connected.


szhorvat:
Thank you! I made one small modification to the copyright notice. Please use this one as the template for future contributions.

PR #1595, opened 2021-01-11
Add unit test for st_mincut

GroteGnoom:
Part of #1592


tests/unit/igraph_st_mincut.c
```diff
@@ -0,0 +1,57 @@
...
+    int source = 0;
+    int target = 4;
+
+    igraph_vector_init(&partition, 0);
+    igraph_vector_init(&partition2, 0);
+    igraph_vector_init(&cut, 0);
+
+    igraph_small(&g, 5, IGRAPH_DIRECTED, 0, 1, 1, 2, 1, 3, 2, 4, 3, 4, -1);
+
+    igraph_st_mincut(&g, &value, &cut, &partition, &partition2, source, target, /*capacity*/ 0);
```
szhorvat, line 36: Please use `NULL` instead of `0` for better readability.

tests/unit/igraph_st_mincut.c
```diff
@@ -0,0 +1,57 @@
...
+    assert(partition_vertex == 0);
+
+    /*     cleanup     */
+
+    igraph_vector_destroy(&cut);
+    igraph_vector_destroy(&partition);
+    igraph_vector_destroy(&partition2);
+    igraph_destroy(&g);
+
+}
```
szhorvat, line 57: `return 0` is missing.

tests/unit/igraph_st_mincut.c
```diff
@@ -0,0 +1,57 @@
...
+    igraph_st_mincut(&g, &value, &cut, &partition, &partition2, source, target, /*capacity*/ 0);
+
+    /*     tests     */
+
+    assert(igraph_vector_size(&cut) == 1);
+    assert(igraph_vector_size(&cut) == 1);
+    assert(igraph_vector_size(&partition2) == 4);
+
+    int cut_edge = VECTOR(cut)[0];
+    int partition_vertex = VECTOR(partition)[0];
```
szhorvat, line 45: Please move _declarations_ to the beginning of the block, to comply with the C90 standard.

We will probably start allowing mid-block declarations soon, but the decision has not been made yet.

@ntamas Any opinions on this?

szhorvat, line 45: Regarding `partition_vertex = VECTOR(partition)[0]`, there is no documented guarantee that the first vertex of the partition would be the source vertex (only that the source vertex is in the first partition). Of course, _in this specific case_, the first partition has only one vertex, so `VECTOR(partition)[0]` is necessarily the source vertex. This was at first confusing to me. Please add a short comment pointing out that the first partition contains only one vertex, namely the source.

tests/unit/igraph_st_mincut.c
```diff
@@ -0,0 +1,57 @@
...
+    igraph_vector_init(&cut, 0);
+
+    igraph_small(&g, 5, IGRAPH_DIRECTED, 0, 1, 1, 2, 1, 3, 2, 4, 3, 4, -1);
+
+    igraph_st_mincut(&g, &value, &cut, &partition, &partition2, source, target, /*capacity*/ 0);
+
+    /*     tests     */
+
+    assert(igraph_vector_size(&cut) == 1);
+    assert(igraph_vector_size(&cut) == 1);
```
szhorvat, line 41: This line is duplicated. I assume you meant to write `igraph_vector_size(&partition)`.

szhorvat (changes requested):
Thanks for this contribution!

I added some comments on the code, and here's one more:

You mentioned that in general, there is more than one min-cut, so one can't easily test that the result is correct. However, if we choose the graph wisely, then there will be precisely one min-cut. I would suggest outputting the cut, and verifying that it is what it should be (0->1 here). 

It would also be nice to add a weighted version (i.e. having capacities) which also has precisely one solution, and where the solution would be different from the unweighted one.

Helper functions to print vectors are in `test_utilities.inc`. Trivial cases can be verified with `assert`, but if the min-cut consists of more than a single edge, then it becomes more convenient to print. You can sort the edge index vector to canonicalize it (since there are no guarantees about ordering).

szhorvat:
This is one possible assignmetn of capacities:

image

The solution is then 1->2, 3->4.

GroteGnoom:
> Thanks for this contribution!

Thank you for taking the time to check and correct everything!

> You mentioned that in general, there is more than one min-cut, so one can't easily test that the result is correct. However, if we choose the graph wisely, then there will be precisely one min-cut. I would suggest outputting the cut, and verifying that it is what it should be (0->1 here).
> 
> It would also be nice to add a weighted version (i.e. having capacities) which also has precisely one solution, and where the solution would be different from the unweighted one.

Allright, I will work on that tomorrow.

> Helper functions to print vectors are in `test_utilities.inc`. Trivial cases can be verified with `assert`, but if the min-cut consists of more than a single edge, then it becomes more convenient to print. You can sort the edge index vector to canonicalize it (since there are no guarantees about ordering).

Thanks!

tests/unit/igraph_st_mincut.c
ntamas: I feel a general uneasiness about mid-block declarations even if C90 allows it. If we always declare everything at the top of a block, then it is easy to get an overview of the variables that a particular block uses as one only needs to look them up at the top of the block. It becomes harder if the variables can be declared anywhere (although syntax highlighting obviously helps).

szhorvat: C90 (same as C89) does not allow it. Older (but not very old) MSVC doesn't allow it. Only C99 and new MSVC allows it.

OK, we keep things at the top of the block then.

When moving declarations mid-block would improve readability, there is an easy workaround: add a new block, i.e.

```
{
     int i;

    {
        int a=1;
        f(a);
    }

    ...

    {
        int b=3;
        f(b);
     }
}
```

An added readability advantage is that these extra syntactic blocks can clearly delimit semantic blocks in the code.

szhorvat:
Please, in the future do use an explicit `#include <igraph.h>` even if we already have an `#include "test_utilities.inc"`. 

No need to do anything for this PR. I'll do it quickly before I merge.

PR #1596, opened 2021-01-11
Add st_edge_connectivity unit test

GroteGnoom:
part of #1592


ntamas (changes requested):

GroteGnoom:
I will :)

I did notice that if I forget a igraph_destroy(&g) and then call VERIFY_FINALLY_STACK(), there are no errors. Is this the expected behavior? Should there be a different check for that?

ntamas:
Yes, that's normal; `VERIFY_FINALLY_STACK()` simply checks whether the so-called "finally" stack is empty. This stack is an internal data structure where igraph pushes pointers that should be cleaned up _in case an error happens in an igraph function_, and these pointers are freed automatically when igraph returns from any of its own functions with `IGRAPH_ERROR()`. During normal termination of an igraph function, the "finally" stack is processed "manually" because there could be cases (e.g., in constructor functions) when we want to return a pointer to the user whose associated memory area was allocated by igraph so we cannot blindly free everything that is in the stack. This leads to the problem: if the manual cleanup code is incorrect, there could be a pointer left in the "finally" stack when the execution exits igraph's turf, leading to a memory leak. Furthermore, if the user calls _another_ igraph function _later on_, and this other function returns with `IGRAPH_ERROR()`, stuff that was accidentally left in the "finally" gets cleaned up at an incorrect time.

ntamas (approved):

ntamas:
@szhorvat is it okay to merge?

PR #1597, opened 2021-01-11
Add edge_disjoint_paths unit test

GroteGnoom:
part of #1592


tests/unit/igraph_edge_disjoint_paths.c
```diff
@@ -0,0 +1,32 @@
...
+
+    igraph_small(&g, 6, IGRAPH_DIRECTED,
+                 0, 1, 0, 2, 1, 2, 1, 3, 2, 4, 3, 4, 3, 5, 4, 5, -1);
+
+    igraph_edge_disjoint_paths(&g, &value, 0, 5);
+
+    igraph_destroy(&g);
+
+    assert(value == 2);
+}
```
szhorvat, line 32: Please add the `return 0`.

szhorvat:
I'll get to these PRs European-time evening today.

In the future, feel free to add more than a single test through a PR.

ntamas:

PR #1598, opened 2021-01-11
Add 4 unit tests, fix residual_graph bug

GroteGnoom:
Part of #1592
residual_graph didn't give the residual capacity, but the original
capacity of the residual graph.


tests/unit/igraph_cohesion.c
```diff
@@ -0,0 +1,44 @@
...
+    igraph_integer_t value;
+    igraph_bool_t checks = 1;
+
+    igraph_small(&g, 7, IGRAPH_DIRECTED,
+                 0, 1, 0, 2, 1, 2, 1, 3, 2, 4, 3, 4, 3, 5, 4, 5, 1, 6, 6, 3, 5, 0, -1);
+
+    igraph_cohesion(&g, &value, checks);
+
+    assert(value == 1);
+
```
szhorvat, line 33: The graph needs to be destroyed here because creating a new one in its place. This is the cause of the test failure.

tests/unit/igraph_adhesion.c
```diff
@@ -0,0 +1,44 @@
...
+    igraph_integer_t value;
+    igraph_bool_t checks = 1;
+
+    igraph_small(&g, 7, IGRAPH_DIRECTED,
+                 0, 1, 0, 2, 1, 2, 1, 3, 2, 4, 3, 4, 3, 5, 4, 5, 1, 6, 6, 3, -1);
+
+    igraph_adhesion(&g, &value, checks);
+
+    assert(value == 0);
+
```
szhorvat, line 33: The graph needs to be destroyed here before creating a new one in its place. This is the cause of the test failure.

ntamas:

PR #1603, opened 2021-01-13
Add write_graph_dimacs unit test

GroteGnoom:
Part of #1592
Yes, it just outputs the source and target, even if they don't exist

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -316,6 +316,7 @@ add_legacy_tests(
 add_legacy_tests(
   FOLDER tests/unit NAMES
   igraph_write_graph_leda
+  igraph_write_graph_dimacs
   lineendings
   pajek
   pajek2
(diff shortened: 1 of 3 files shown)


szhorvat:
@ntamas It seems a bit unusual that the DIMACS writer takes `source`/`target` as parameters and writes specifically a max-flow problem file. This format is not used only for max-flow problems.

Would it make sense to remove these arguments, or perhaps make them optional in some sense, and make this into a generic DIMACS graph format writer? Here's a more generic format description: http://prolland.free.fr/works/research/dsat/dimacs.html

The difference would be that:

 - The `n` line would not be included. This specified the source/target for the flow problem.
 - The `p` line would look like `p edge NODES EDGES` instead of `p max NODES EDGES`.

ntamas:
Totally fine for me; I'm not familiar with the DIMACS format and I think it was primarily added to gain access to lots of max-flow problems that we could test our max-flow implementation with. If it makes sense to make it into a more generic reader/writer, feel free to go ahead.

szhorvat:
There are many different DIMACS challenge formats, designed not so much for graph exchange, but for use during the competitions. Here is the most complete description I could find:

http://dimacs.rutgers.edu/archive/Challenges/

It would be nice to have a generic reader that can extract at least the graph portion of most versions. As you say, the main use of this is to access various benchmark sets.

To do all this will be quite a bit of work, so for now I'll merge this PR, and will open a new issue for the idea of a more general DIMACS reader.

szhorvat:
Thanks!

Issue #1599, opened 2021-01-12
igraph_is_same_graph: a function that checks if two graphs are the same


szhorvat:
**What is the feature or improvement you would like to see?**

Detect if two graphs are the same, i.e. have the same vertex and edge sets. This is not the same as isomorphism.

`1-2-3` and `1-3-2` are isomorphic but not the same (edge sets differ).

`1-2-3` has potentially multiple representations, e.g. `1-2, 2-3` or `2-3, 1-2` (different edge order). The function should detect that these are the same graph.

**Use cases for the feature**

It is especially useful for verification (e.g. writing tests)

Ref: https://igraph.discourse.group/t/is-the-order-of-print-graph-guarateed-to-be-the-same/587

iosonofabio:
I'd be happy to implement this, it seems easy enough. It is a little more tricky is you also want to check all vertex and edge attributes, but still should be doable.

szhorvat:
Thanks @iosonofabio.  Can you first please write a purely topological version that ignores any attributes? This is what I was looking for. 

Additionally, it would be useful to have a function that canonicalizes the representation (i.e. makes it possible to directly compare two instances after both were canonicalized). Let me know if you want to tackle this related problem at the same time or leave it for later.

----

Comparing attributes is useful, but it's a somewhat different operation. IMO it's best done in the high-level interfaces and not the C core. Note that the C core doesn't know how to compare two attribute values. We couldn't just use `==`, as attributes can be (and in practice are) non-trivial data structures. I see two options: 

 1. We'd either need to allow for a custom comparison function, passed in as an extra argument by the user, which allows the C core to test the equality of two attribute values
 2. Or we will have a canonicalization function (which I'd like anyway) that is used as a basis of an attribute-comparing implementation in a high-level interface. The high-level interface knows how to compare objects in its host language.

One argument for doing the attribute comparing version in the high-level interface is the following: What does it mean that two graphs are the same? Their vertex sets and edge sets must be the same.  In the C core, vertices don't have names, only indices, thus "same vertex set" and "same edge set" are a different concept than if vertex names were possible.

As you can see the version with attributes is quite a bit more complicated. Let's leave that for later.

iosonofabio:
Ok no attributes for now. Just to make sure, all we have got to do is:

- check the number of vertices and edges is the same
- check both graphs are directed/undirected
- for loop over edges and check that source/target of each is the same

If all pass return true, else false

That correct?

szhorvat:
[...]

PR #1604, opened 2021-01-13
Add function igraph_is_same_graph for tests

iosonofabio:
Addressing #1599.

This relies on `graph->ii` which is a vector of edge ids, double sorted by source and target vertices. If that representation goes, we should adapt this function. The easiest way to do that would be to sort the edges ourselves in the same way inside the function.

edit: I should put a header somewhere. Which header file would be appropriate for this to be considered internal?

### DIFF
--- a/src/graph/type_indexededgelist.c
+++ b/src/graph/type_indexededgelist.c
@@ -1691,3 +1691,62 @@ int igraph_incident(const igraph_t *graph, igraph_vector_t *eids,
 
     return 0;
 }
+
+
+/**
+ * \function igraph_is_same_graph
+ * \brief Check if two graphs are topologically the same
+ *
+ * \param graph1 The first graph object.
+ * \param graph2 The second graph object.
+ * \return 1 if they are the same, 0 if they are different.
+ *
+ * NOTE: this function is mostly for internal use in unit tests.
+ * Users should not rely on it for production code.</para><para>
+ *
+ * Added in version 0.9.</para><para>
+ *
+ * Time complexity: O(E), the number of edges in the graphs.
+ */
+
+int igraph_is_same_graph(const igraph_t *graph1, const igraph_t *graph2) {
+    long int nv1 = igraph_vcount(graph1);
+    long int nv2 = igraph_vcount(graph2);
+    long int ne1 = igraph_ecount(graph1);
+    long int ne2 = igraph_ecount(graph2);
+    long int i, eid1, eid2;
+
+    /* Check for same number of vertices/edges */
+    if ((nv1 != nv2) || (ne1 != ne2)) {
+        return 0;
+    }
+
+    /* Check for same directedness */
+    if (igraph_is_directed(graph1) != igraph_is_directed(graph2)) {
+        return 0;
+    }
+
+    /* Vertices have no names, so no they must be 0 to nv - 1 */
+
+    /* Edges are double sorted in the current representations ii/oi of
+     * igraph_t (ii: by incoming, then outgoing, oi: vice versa), so
+     * we just need to check them one by one. If that representation
+     * changes, this part will need to change too. */
+    for (i = 0; i < ne1; i++) {
+        eid1 = (long int) VECTOR(graph1->ii)[i];
+        eid2 = (long int) VECTOR(graph2->ii)[i];
+
+        /* check they have the same source */
+        if (IGRAPH_FROM(eid1) != IGRAPH_FROM(eid2)) {
+            return 0;
+        }
+
+        /* check they have the same target */
+        if (IGRAPH_TO(eid1) != IGRAPH_TO(eid2)) {
+            return 0;
+        }
+
+    }
+
+    return 1;
+}
--- a/src/graph/type_indexededgelist.c
+++ b/src/graph/type_indexededgelist.c
@@ -1695,35 +1695,53 @@ int igraph_incident(const igraph_t *graph, igraph_vector_t *eids,
 
 /**
  * \function igraph_is_same_graph
- * \brief Check if two graphs are topologically the same
+ * \brief Are two graphs identical as labelled graphs?
+ *
+ * Two graphs are considered to be the same if they have the same vertex and edge sets.
+ * Graphs which are the same may have multiple different representations in igraph,
+ * hence the need for this function.
+ *
+ * <para></para>
+ * This function verifies that the two graphs have the same directnedness, the same
+ * number of vertices, and that they contain precise the same edges (regardless of their ordering)
+ * when written in terms of vertex indices. Graph attributes are not taken into account.
+ *
+ * <para></para<
+ * This concept is different from isomorphism. For example, the graphs
+ * <code>0-1, 2-1</code> and <code>1-2, 0-1</code> are considered the same
+ * because they only differ in the ordering of their edge lists and the ordering
+ * of vertices in an undirected edge. However, they are not the same as
+ * <code>0-2, 1-2</code>, even though they are isomorphic to it.
+ * Note that this latter graph contains the edge <code>0-2</code>
+ * while the former two do not—thus their edge sets differ.
  *
  * \param graph1 The first graph object.
  * \param graph2 The second graph object.
- * \return 1 if they are the same, 0 if they are different.
- *
- * NOTE: this function is mostly for internal use in unit tests.
- * Users should not rely on it for production code.</para><para>
- *
- * Added in version 0.9.</para><para>
+ * \param res The result will be stored here.
+ * \return Error code.
  *
  * Time complexity: O(E), the number of edges in the graphs.
+ *
+ * \sa igraph_isomorphic() to test if two graphs are isomorphic.
  */
 
-igraph_bool_t igraph_is_same_graph(const igraph_t *graph1, const igraph_t *graph2) {
+int igraph_is_same_graph(const igraph_t *graph1, const igraph_t *graph2, igraph_bool_t *res) {
     long int nv1 = igraph_vcount(graph1);
     long int nv2 = igraph_vcount(graph2);
     long int ne1 = igraph_ecount(graph1);
     long int ne2 = igraph_ecount(graph2);
     long int i, eid1, eid2;
 
+    *res = 0; /* Assume failure */
+
     /* Check for same number of vertices/edges */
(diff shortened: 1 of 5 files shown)


vtraag:
> edit: I should put a header somewhere. Which header file would be appropriate for this to be considered internal?

I believe all other functions that work directly on the `igraph_t` object are stored in `igraph_interface.h`.

iosonofabio:
two quick questions:

1. do we want this function to return `igraph_bool_t` instead of `int`?
2. I haven't learned how to run tests locally yet with the new CMake system. I tried to follow the pattern in `.github/workflows/cmake...` but it says 0% tests passed. Sorry about that, having a working local test suite would reduce the number of commits in these PRs.

iosonofabio:
hmm the test failure seems to be due to *memory leaks in the erdos-renyi test*? I'm not sure what's going on here... @ntamas?

vtraag:
> do we want this function to return `igraph_bool_t` instead of `int`?

That's probably a good idea.

> I haven't learned how to run tests locally yet with the new CMake system.

You did compile the tests first with `cmake --build . --target check` (or simply `make check`)? Running them should be done using `ctest`, possibly including a configuration, e.g. `ctest -C Debug`, but I think that is Windows only.

iosonofabio:
Thanks, now it returns `igraph_bool_t`. I got it to test locally by using:

```
mkdir build
cd build
cmake ..
make
make check
```
and that already tests it. All good here.

vtraag:
> that already tests it.

Ah, so `make check` both builds and runs the tests, even better. Just FYI, with `ctest` you can run specific tests with a regex specifier, e.g. `ctest -R is_same_graph`. That way, you can easily test only one test case. SImilarly, you can build only that test by invoking `cmake --build . --target test_igraph_is_same_graph`.

src/graph/type_indexededgelist.c
```diff
@@ -1691,3 +1691,62 @@ int igraph_incident(const igraph_t *graph, igraph_vector_t *eids,
...
+
+/**
+ * \function igraph_is_same_graph
+ * \brief Check if two graphs are topologically the same
+ *
+ * \param graph1 The first graph object.
+ * \param graph2 The second graph object.
+ * \return 1 if they are the same, 0 if they are different.
+ *
+ * NOTE: this function is mostly for internal use in unit tests.
```
szhorvat, line 1704: Is there any good reason for this? If the function is robust, its use should not be discouraged (and it _should_ be robust). Testing is the most common, but by far not the only use case.

There are several other changes I want to make to the docs. Is it okay if I make edits to this text on this branch?

iosonofabio:
There is a reason, which is that it is the truth ;-)

But sure edit at will, nothing personal

ntamas:
> do we want this function to return igraph_bool_t instead of int?

The rule of thumb is: if the function can fail, ever, then it should return an `int` with an igraph error code (zero if everything was OK), and the _real_ result should be returned in an output argument. Failure can mean things like "not enough memory to allocate auxiliary data structures" as well. So, if we decide to return `igraph_bool_t` here, we commit ourselves to never use memory allocations inside the function, and to never use any igraph function that can fail on its own. I'm okay with that, just wanted to make sure that you are aware of it.

ntamas:
I'm not particularly happy with the direct usage of `graph->ii`, but there might be valid reasons for that.

Roughly speaking, anything that accesses the internals of an `igraph_t` object is considered to be part of the "basic graph interface". We are aiming for this interface to be as small as possible in order to make it easier for us to experiment with alternative graph representations. My plans for the future are to make the basic graph interface smaller, not larger; we already have several methods in the basic graph interface that could be moved outside:

* `igraph_empty()` simply calls `igraph_empty_attrs()`, and I cannot really conceive any graph data structure where this would not be true, so `igraph_empty()` could be moved outside the basic API, leaving only `igraph_empty_attrs()` in there.

* `igraph_delete_vertices()` delegates everything to `igraph_delete_vertices_idx()` so this could be removed from the basic API as well.

...and so on, and so on.

The point I'm trying to make here is that if we decide to reach into the internals of `igraph_t` in `igraph_is_same_graph()`, and then later on we decide to implement a new graph data structure to see whether we gain something in terms of efficiency, we would need to deal with the implementation of `igraph_is_same_graph()` _again_ with this new graph data structure.

ntamas:
By the way, I have seen the comment from @vtraag in the original issue where he advocates for this function being part of the core graph API. I'm okay with that as well, just wanted to make sure that we understand the implications. Also, if we do decide that this function becomes part of the core graph API, then we should make sure to document very carefully what the function needs to do (compare directednes, compare number of nodes, ensure that the edge sets are the same, not necessarily in the same order, but with same edge multiplicities -- so basically a "multiset equality"), and what the function does not need to do (e.g., compare attributes). This is to help future implementors of other graph data structures to know what to deal with and what not to deal with.

iosonofabio:
Yes @ntamas that is clear. I agree with the goal to reduce the basic interface.

Double sorted edges are useful for this function, whether (1) they are sorted using an internal data structure like current `ii` or (2) by calling a public function to get the sorted edge ids (which is how `ii` is made in the first place, o'course).

If you prefer option 2 then such a public function needs to be written. Its implementation would currently rely on `ii` for speed but would need to change if we drop `ii`. Or we could make it slower and recompute the sorted edges pretending that we don't know about `ii`.

The bottom line is that if `ii` is just a source of worry we should just dump it and consider it deprecated. But if it's useful for internal functions like this then there's nothing wrong in using it.

Btw @szhorvat this whole discussion is exactly why I added the documentation line warning users that this function is internal.

iosonofabio:
I'm thinking that if this function creates such headaches in the public API, we could move it to the file that contains other test utility functions and be done with it

vtraag:
I think we need a more precise terminology. Yes, this function `igraph_is_same_graph` should be considered part of the core graph API, as @ntamas formulated it. This is what I referred to as "internal" previously, but this is a misleading nomenclature it now seems, which is causing some confusion. It should be a *public* function, that can be used by users of the library, hence the declaration should be made in `include/igraph_interface.h`, as it is done now. If the function is *not* public, it should *not* be declared in `include/igraph_interface.h`. Hence, the comment by @szhorvat that we should not state in the documentation that this function is only for internal use, which I agree with.

It does imply, as @ntamas already observed, that this implementation is part of the core graph API, and hence, should be changed along with a different implementation.

Although it could technically be possible to move it outside the current graph representation, I think that would sacrifice too much on speed. At the moment the graph is already internally sorted, and you essentially need to perform exactly this operation in order to identify whether two graphs are identical (i.e. the exact same edges, even if the original edge ordering is different). That was the original reason why I proposed it should be part of the core graph API.

vtraag:
> if we decide to return `igraph_bool_t` here, we commit ourselves to never use memory allocations inside the function, and to never use any igraph function that can fail on its own. I'm okay with that, just wanted to make sure that you are aware of it.

Yes, this is a good point. I assumed that it could never fail because it does not call any other functions that can fail at the moment. However, that is not necessarily the case for future representations, because they are able to fail, for example because they do need to sort some edges. For that reason, I think you are correct, and the API should be declared with returning an `int`, and the result should be provided in a pointer/reference.

szhorvat:
A lot has happened here while I was out. In short, here are my preferences:

 - During the winter break I did think a bit about whether to make this part of the core API (i.e. messing with internal data structures), but didn't reach a definitive conclusion. **I am weakly in favour of making it part of the core API,** for performance reasons.
 - **I am strongly in favour of returning an error code** instead of returning a bool. Especially for the core API, stability is paramount, and it is not at all unlikely that if we change the graph data structure, it will become possible for this function to fail.

szhorvat:
> hmm the test failure seems to be due to _memory leaks in the erdos-renyi test_? I'm not sure what's going on here... @ntamas?

This is now fixed (it wasn't a leak, but there was a problem, see #1606).

iosonofabio:
Aaargh, I was pushing my changes at the same time...

szhorvat:
> Aaargh, I was pushing my changes at the same time...

I'm sorry about that, I wanted to push these before you woke up, but my internet stopped working for a while ... 

Feel free to overwrite my version (but please keep the docs).

iosonofabio:
Alright, it's too messy to have two people editing the code in parallel.

@szhorvat please ask before changing the code in my PRs next time. My understanding was that you were going to change the documentation text, not a bunch of other things too. In any case, please keep pushing here and then we can merge, I'm stopping my contribution to this PR now.

And error codes are ints, not `igraph_bool_t`

iosonofabio:
@szhorvat in the tests, you still have to assert that the error code returned by the function is zero, to avoid issues in the future.

szhorvat:
I'm sorry, my intention was to merge it. I considered this to be basically complete save for documentation and a tedious detail about how it returns it result.

iosonofabio:
that's ok, but please add those asserts in the tests if that's our practice now.

szhorvat:
> that's ok, but please add those asserts in the tests if that's our practice now.

That's done now. Can you take a look with a fresh morning mind and if you're happy and the CI tests pass, then squash-and-merge it?

iosonofabio:
Thank you @szhorvat!

It looks good to me.

@ntamas is the current practice to squash and merge or just merge? I remember a previous conversation on this topic but can't quite recall what the conclusion was, sorry about that

szhorvat:
> is the current practice to squash and merge or just merge?

It depends on the situation. When the commits are messy, and the commit messages assume a specific context that won't be apparent when reading them a year later, it should be squash-and-merge. I find it easier to work this way as I can just push many commits without having to keep all of them organized. The tradeoff is that the PR should be relatively small.

Sometimes, when the PR is too big, it's better to keep commits separate. But then the PR author should make sure that the commits are all nice and well-structured, which can be a pain. This is the case in PR #1583 (and I very much regret choosing this path because it's a pain)

iosonofabio:
Personally I never squash because you can see in the git DAG that those commits all came together by the same author(s) anyway, and a git DAG is not a showroom, I'd rather have it a little verbose than too terse.

Anyway, I don't usually merge my own PRs out of courtesy, @ntamas feel free to do either one, your call.

ntamas (approved):

ntamas:
Thanks a lot guys!

PR #1605, opened 2021-01-13
Fix/stimulus

vtraag:
This PR upgrades `stimulus.py` to work correctly with Python 3 and removes the dependency on `seqdist`, by replacing it with `OrderedDict`.

It appeared there were many functions had missing `DEPS` specifications in `functions.def`, which lead to some problems during the upgrading of the code to Python 3. It turned out most were actually in functions that were not automatically translated by stimulus in the R interfaces. Nevertheless, they are now all corrected in the `functions.def`.

The upgraded code was tested in the R interface, and builds correctly. There still seem to be some issues, which I am looking into.


tools/stimulus.py
```diff
@@ -243,13 +247,13 @@ def generate(self, inputs, output):
         out.close()
 
     def generate_function(self, f, out):
-        print "Error: invalid code generator, this method should be overridden"
+        print("Error: invalid code generator, this method should be overridden")
```
ntamas, line 250: Idiomatic Python 3.x is to 1) `raise NotImplementedError` here, 2) mark the `generate_function` method with the `@abstractmethod` decorator and 3) construct the class as `class CodeGenerator(metaclass=ABCMeta)` (where `ABCMeta` is imported from `abc`). This prevents instantiation of `CodeGenerator` until the method is overridden.

vtraag: Yes, good point. I haven't gone through all the code to make it Python 3.x idiomatic, just functional. But this might be a good occasion to revise the entire code.

iosonofabio: I'd be happy to check for Python-ness of code here and in `python-igraph`, if you guys would like me to

vtraag: This PR only involves `stimulus.py`, which is the tool that generates the code for interfaces. At the moment it is mostly used by the R interface, but in principle we could also use it to build any other interface (e.g. also for `python-igraph`). I wouldn't go through the entire `python-igraph` library, but if you want to check this `stimulus.py`, please go ahead! Feel free to push immediately to this PR (or open a PR against my branch :smile: ).

vtraag:
OK, I improved a number of things in `stimulus.py` and `functions.def`. It now does build correctly, and creates correct R code for the interface, so in principle, this could be merged. But since we are touching this now anyway, we might see if we can improve it in any other way. In particular, there seems to be a problem still on line 670, because `pname` is not properly defined for that case. That doesn't happen to be ever invoked, so it is not a problem currently, but it still is incorrect.

I won't be working on this anymore today, so if you want to do a further check with regards to this issue and other non-idiomatic Python 3 stuff @iosonofabio or @ntamas , feel free to have a go at it.

iosonofabio:
alright, @ntamas has already formatted everything ;-)

vtraag:
I think this is mostly done. I thought about updating all string formatting to use `''.format(...)`, but this seems quite a lot of work, and I cannot find a way how to do this automatically. Perhaps @ntamas or @iosonofabio you know of a way? Otherwise, as far as I am concerned, this is good to go.

ntamas:
> I thought about updating all string formatting to use `''.format(...)`

It depends on which Python version you are targeting anyway. `str.format()` is soooo 2016 now that Python 3.6 has [f-strings](https://realpython.com/python-f-strings/) :)

So I'd say if we want to support older Python versions, then let's just leave the string formatting as it is because we need to use obsolete constructs to support old versions. If we drop support for Python 3.5 and earlier, then we should use f-strings where possible -- but maybe it's enough to upgrade the code if we stumble upon instances of old percent-style formatting and we are working on that part of the code anyway.

tools/stimulus.py
```diff
@@ -1,160 +1,178 @@
...
 
+
 ################################################################################
 
+
 def main():
     # Command line arguments
     try:
-        optlist, args = getopt.getopt(sys.argv[1:], 't:f:l:i:o:hv', ['help'])
+        optlist, args = getopt.getopt(sys.argv[1:], "t:f:l:i:o:hv", ["help"])
```
ntamas, line 163: `getopt` is obsolete, most Python code uses `argparse` nowadays, but this is not a priority.  We can do it any time in the future if we have some time.

tools/stimulus.py
```diff
@@ -163,152 +181,162 @@ def main():
...
         # Parse function and type files
-        parser=PParser()
-        self.func=seqdict.seqdict()
+        parser = PParser()
+        self.func = OrderedDict()
         for f in func:
-            ff=open(f)
-            newfunc=parser.parse(ff)
-            self.func.extend(newfunc)
+            ff = open(f)
```
ntamas, line 236: `open()` is mostly used nowadays as a context manager, like:

```
with open(f) as ff:
    ...do something with ff here, close() is called automatically when exiting the context...
```

This is not urgent either, just something for the future.

ntamas (approved):

ntamas:
Okay, I went through the PR and it looks good to me, apart from a few obsolete Python constructs that do not affect functionality. I think I'll just merge it now and we can tidy it up if we plan to work more on it in the future.

@vtraag What do you think about separating it to a repo on its own and including it as a submodule in `rigraph`?

vtraag:
Thanks for all the observations still about some obsolete things. Clearly, I hand't seen f-strings yet, lagging behind apparently :)



> @vtraag What do you think about separating it to a repo on its own and including it as a submodule in `rigraph`?

Yes, that might be a good idea. At least it detaches the functionality of `stimulus` from the `igraph` repository, which makes it a bit easier to separate the maintenance of the two. However, it also raises a couple of questions. The most essential piece of the functionality is the `functions.def`, and I am not yet sure what to do with it. If we keep it in the `igraph` repository, we still have the dependency on the `igraph` repository for that part still. If we put it in a separate `stimulus` repository, it would be strange to update that repository if `igraph` is updated. The other question regards the language specific code and type definitions. Right now, the language specific code is in `stimulus.py`, while ideally you would take that out and put it in the language specific repository (i.e. `rigraph`), as is now already the case for the language specific type definitions. I am not sure what is the best approach here, I will have to think about it. If you have any thoughts on it, please do share.

It seems most elegant to keep everything separate: `functions.def` in `igraph`, `stimulus.py` (the core) in `stimulus`, language type definitions (e.g. `types-RR.def`) and language code (e.g. `RRCodeGenerator.py` or something, to be developed) in language specific interfaces, i.e. `rigraph`. However, it may also make things a bit more complex to manage and coordinate.

ntamas:
Well, yes, ideally the core `stimulus` package should be a language-independent backend that provides some functionality that allows us to generate glue code for higher-level interfaces. The part of the generator that is specific to a higher-level language should live in the repository that belongs to that particular higher-level language.

On the other hand, it doesn't hurt to do the first step now (i.e. moving `stimulus.py` out of the C igraph repo to its own) and then pluck out the R-specific parts in a later step.

vtraag:
OK, let's see if we can move it outside to its own `stimulus` package. Do you have any idea why it is called `stimulus`? Something like `CodeGenerator` sounds a lot more logical, if a bit less poetic.

There's one more idea that I had. Perhaps we could make a simply `C` CodeGenerator that generates the appropriate C declarations for all functions. That way, we can check if the `functions.def` is still correct. What do you think?

ntamas:
> Do you have any idea why it is called `stimulus`?

That's a pun that makes sense only in Hungarian. "Interface generator", abbreviated as `inger` (the original name of the script) means "stimulus" in Hungarian. `stimulus` sounded better :) I'm okay with renaming it to `igraph-glue-code-generator` or something like that if you can find a name that is short and snappy. :)

> Perhaps we could make a simply C CodeGenerator

I'm not that familiar with the syntax of `functions.def` to decide whether it's possible at all. Even if it is, how would we compare the generated C declarations with the original ones? Shall we try to replace the headers with a generated one and see if the compilation still works?

vtraag:
> That's a pun that makes sense only in Hungarian. "Interface generator", abbreviated as `inger` (the original name of the script) means "stimulus" in Hungarian. `stimulus` sounded better :) I'm okay with renaming it to `igraph-glue-code-generator` or something like that if you can find a name that is short and snappy. :)

That's a nice one. We should definitely keep this name :)

> I'm not that familiar with the syntax of `functions.def` to decide whether it's possible at all.

I believe it should not be too difficult.

>  Shall we try to replace the headers with a generated one and see if the compilation still works?

Yes, that was the idea indeed. I wasn't sure exactly how to do this, because when actually compiling the library, we of course rely not on the generic `igraph.h` include file, but on all other individual include files. This would imply we would need to generate separate include files, but this then becomes much more complex to generate and replace easily. So we will have to give this a bit more thought.

PR #1609, opened 2021-01-16
Remove nonlinear_barabasi_game declaration

GroteGnoom:
It was not implemented, because it had already been moved to
barabasi_game.

See also https://igraph.discourse.group/t/should-undefined-declaration-be-removed/592

igraph_nonlinear_barabasi_game is part of #1592


szhorvat:
Thanks again!

PR #1615, opened 2021-01-19
Update Bliss to 0.74

### LINKED ISSUE #1171: Bliss should be interruptible
Bliss can take very long to run on some difficult graphs (see #1170).

It should be interruptible.

Bliss is written in C++, so it might be possible to just throw an exception to break out of it.

szhorvat:
Upgrades Bliss to the just-released version 0.74, and therefore fixes #1171.


szhorvat:
@ntamas I'd like to get a review on this, but please do not merge. I did not yet test the interruption support. I will do that tomorrow, then I'll merge it myself.

You can ignore the first commit. Look at the `igraph-changes.md` instead.

Please review the subsequent commits, in particular interruption support (note the global variable—avoiding this would require a few more changes to Bliss, I think).

Note that Bliss now uses C++11, therefore I had to enable support for that in the top-level CMakeLists.txt. Are we fine with this? I'm a but reluctant to try to patch Bliss for C++98 support ...

ntamas:
@szhorvat Looks good to me so far! C++11 is okay, as well as the global variable for interruption support.

PR #1616, opened 2021-01-19
Add new extended chordal ring unit test

GroteGnoom:
Part of #1592

See also:
https://igraph.discourse.group/t/extended-chordal-ring-ouputs-a-multigraph/596/2


szhorvat:
Thanks!

PR #1617, opened 2021-01-20
Disallowing NaN weights for several path based functions

### LINKED ISSUE #1560: Make sure that passing in NaN weights do not cause a crash
Passing NaNs as weights will cause some functions to crash, see e.g. https://github.com/igraph/rigraph/issues/415

Functions which can't handle NaN should check the weight vector for NaNs.

Note that some functions may only access a small part of the weight vector in some cases. Ideally, they should not loop over the entire weight vector just to check for NaNs then.

----

List of functions that are likely affected (please amend this list):

 - Anything based on shortest paths, such as betweenness, closeness, efficiency, etc.

vtraag:
This partially fixes #1560 for path based functions by disallowing NaN edge weights. Although it could theoretically be argued that some functions may still return proper value for some limited NaN edge weights, this quickly becomes quite challenging. This PR ensures that proper errors are returned in case there is any NaN edge weight, and also corrects the documentation.

In addition, the vector functions `igraph_vector_min` and `igraph_vector_max` are now guaranteed to return NaN whenever an NaN value appears anywhere in the vector. Previously, this was only the case if the NaN happened to appear in the first position (i.e. index 0).

### DIFF
--- a/src/centrality/betweenness.c
+++ b/src/centrality/betweenness.c
@@ -50,8 +50,8 @@
  * \param directed Logical, if true directed paths will be considered
  *        for directed graphs. It is ignored for undirected graphs.
  * \param weights An optional vector containing edge weights for
- *        calculating weighted betweenness. Supply a null pointer here
- *        for unweighted betweenness.
+ *        calculating weighted betweenness. No edge weight may be NaN.
+ *        Supply a null pointer here for unweighted betweenness.
  * \return Error code:
  *        \c IGRAPH_ENOMEM, not enough memory for
  *        temporary data.
@@ -105,6 +105,8 @@ static int igraph_i_betweenness_cutoff_weighted(
         igraph_real_t minweight = igraph_vector_min(weights);
         if (minweight <= 0) {
             IGRAPH_ERROR("Weight vector must be positive", IGRAPH_EINVAL);
+        } else if (igraph_is_nan(minweight)) {
+            IGRAPH_ERROR("Weight vector must not contain NaN values", IGRAPH_EINVAL);
         } else if (minweight <= eps) {
             IGRAPH_WARNING("Some weights are smaller than epsilon, calculations may suffer from numerical precision.");
         }
@@ -280,8 +282,8 @@ static int igraph_i_betweenness_cutoff_weighted(
  * \param directed Logical, if true directed paths will be considered
  *        for directed graphs. It is ignored for undirected graphs.
  * \param weights An optional vector containing edge weights for
- *        calculating weighted betweenness. Supply a null pointer here
- *        for unweighted betweenness.
+ *        calculating weighted betweenness. No edge weight may be NaN.
+ *        Supply a null pointer here for unweighted betweenness.
  * \param cutoff The maximal length of paths that will be considered.
  *        If negative, the exact betweenness will be calculated, and
  *        there will be no upper limit on path lengths.
@@ -522,8 +524,8 @@ int igraph_betweenness_cutoff(const igraph_t *graph, igraph_vector_t *res,
  *        If negative, the exact betweenness will be calculated, and
  *        there will be no upper limit on path lengths.
  * \param weights An optional vector containing edge weights for
- *        calculating weighted betweenness. Supply a null pointer here
- *        for unweighted betweenness.
+ *        calculating weighted betweenness. No edge weight may be NaN.
+ *        Supply a null pointer here for unweighted betweenness.
  * \return Error code:
  *        \c IGRAPH_ENOMEM, not enough memory for
  *        temporary data.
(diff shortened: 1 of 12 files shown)


vtraag:
I just amended the list of functions in #1560. After this PR has merged, we can tick off individual items from that list.

vtraag:
Ha, for macOS the build fails with the message: `error: self-comparison always evaluates to false` for ` !(max != max)`. 
The whole point of this construction was to check for `nan` for `max` without explicitly relying on `igraph_is_nan`, because that introduces a dependency on the base type of the vector. Traditionally, the `x != x` would evaluate to true for `x == nan`, while for any other value this would evaluate to false. @ntamas, @szhorvat , you are on macOS, do you have any suggestions?

ntamas:
Nope, other than turning off the `tautological-compare` warning. It would be great if we could be selective about turning off the warning from CMake, preferably only for this single file and only for Clang, because this warning has the potential to catch bugs that would otherwise go unnoticed (one example I can think of is a comparison of `if (x >= 0)` where `x` is assumed to be signed but is in fact unsigned).

If it's too big of a hassle to be selective, then let's just turn off the warning for good.

src/centrality/betweenness.c
```diff
@@ -105,6 +105,8 @@ static int igraph_i_betweenness_cutoff_weighted(
         igraph_real_t minweight = igraph_vector_min(weights);
         if (minweight <= 0) {
             IGRAPH_ERROR("Weight vector must be positive", IGRAPH_EINVAL);
+        } else if (igraph_is_nan(minweight)) {
+            IGRAPH_ERROR("Weight vector must not contain NaN values", IGRAPH_EINVAL);
```
ntamas, line 109: Won't we get a misleading warning here if there are no edges in the graph? In that case, the weight vector is empty, so the min of the vector is NaN, while technically it does not _contain_ any NaN values.

ntamas: Sorry, my bad, I did not notice the `no_of_edges > 0` condition.

vtraag:
OK, I will think about whether there is another way to solve this issue, without requiring to turn off warning flags. The only reason for introducing this is that I wanted to be consistent with the NaN reporting: it should report the first such element, as it would for the maximal/minimal element. Perhaps there is a way to do this differently still.

ntamas:
Looks good to me, but can we add unit tests for the behaviour of `igraph_vector_{which_}_{min,max}` to ensure that it deals consistently with NaN values? (Especially considering that you have spotted inconsistent behaviour that you had to fix in this PR).

vtraag:
> Looks good to me, but can we add unit tests for the behaviour of `igraph_vector_{which_}_{min,max}` to ensure that it deals consistently with NaN values? (Especially considering that you have spotted inconsistent behaviour that you had to fix in this PR).

Yes, good point, I though of doing this as well, but didn't get around to it anymore. I will include unit test for this, probably tomorrow.

src/core/vector.pmt
```diff
@@ -1034,12 +1037,13 @@ long int FUNCTION(igraph_vector, which_max)(const TYPE(igraph_vector)* v) {
...
 
 /**
+ * \ingroup vector
  * \function igraph_vector_min
  * \brief Smallest element of a vector.
  *
  * The vector must be non-empty.
  * \param v The input vector.
- * \return The smallest element of \p v.
+ * \return The smallest element of \p v, or nan if any element is nan.
```
szhorvat, line 1046: Can you please spell it as NaN ("nan" will be confusing to some)?

vtraag:
I added a test and corrected the spelling in the documentation. Please do not merge yet, I will still think about how to deal with the NaN comparisons on macOS.

iosonofabio:
Perhaps

```C
x + 1 > x
```
would be true except for nan? Or even

```C
((x < 1) && (x + 1 > x)) || (x - 1 < x)
```

vtraag:
Sorry, the implementation is not correct anyway, I will correct it.

vtraag:
OK, now it is correct. Unfortunately, I see no way of working around a comparison that involves doing something like `x != x`. If we really don't want to turn off the warning for this file, I could make specific implementations that differ for the integer and floating point base types.

ntamas:
Almost there, I'm fixing the docs and then I'll merge.

GroteGnoom:
> Ha, for macOS the build fails with the message: `error: self-comparison always evaluates to false` for ` !(max != max)`.
> The whole point of this construction was to check for `nan` for `max` without explicitly relying on `igraph_is_nan`, because that introduces a dependency on the base type of the vector.
Are you sure it doesn't just work? 
```
#include <math.h>

int my_isnan(double a) {
	return isnan(a);
}

int main() {
	return my_isnan(5);
}
```
works. If the base type is int, it gets changed into a double, and still gives the correct result.

ntamas:
Docs are now fixed; I think the build will succeed this time. @vtraag once you think that the NaN testing is sorted out adequately, this is ready to merge.

vtraag:
OK, I now changed it to an explicit `isnan`, to avoid any potential probles with `x != x` type of formulation. Possibly, this might have been optimised away by clang on macOS. This way, I think it should build fine on all platforms.

And thanks for the documentation correction!

PR #1621, opened 2021-01-21
Add callaway traits unit tests, add error on zero distribution

GroteGnoom:
Part of #1592


GroteGnoom:
@szhorvat I think the documentation fails after https://github.com/igraph/igraph/commit/6f6b6a2de3abab5b1e0271c358c89248298ba321

Issue #1620, opened 2021-01-21
Redundant parameters in some random graph generators


szhorvat:
`callaway_traits_game`, `establishment_game`, `preference_game` all have both a `types` (number of vertex types) and `type_dist` (distribution of types) parameter. It is necessary that `types == igraph_vector_size(type_dist)`. Therefore, `types` is redundant—the number of types is already defined by the type distribution vector.

`asymmetric_preference_game` is similar, but it has a `type_dist_matrix` instead of a `type_dist` vector.

Since we're trying to clean up the API a bit for 0.9, shall we remove the redundant `types` parameter?

----

My guess for the existence of `types` is that originally it was intended that `type_dist` can be omitted and in that case it would be auto-generated as all-ones. This is what the R interface does: https://igraph.org/r/doc/sample_traits_callaway.html However, it does it in R, not in C. I don't think there's a reason to do it in C.

szhorvat:
Also, in the case of `asymmetric_preference_game`, it is unclear to me why it is enforced that the number of in-types and out-types be the same. In principle they could be different.

szhorvat:
Regarding `igraph_preference_game` with `fixed_sizes` set, isn't this exactly the same as a stochastic block model?

szhorvat:
`preference_game` does in fact allow for `type_dist` to be `NULL`, in which case it takes the distribution to be uniform. `callaway_traits_game`, `establishment_game` do not currently allow this in the C interface (only in R).

vtraag:
> Since we're trying to clean up the API a bit for 0.9, shall we remove the redundant `types` parameter?

Yes, that makes a lot of sense to me.

> Regarding `igraph_preference_game` with `fixed_sizes` set, isn't this exactly the same as a stochastic block model?

Yes, I agree.

GroteGnoom:
> Regarding `igraph_preference_game` with `fixed_sizes` set, isn't this exactly the same as a stochastic block model?

It also seems to me like the callaway_traits_game is equal to the establishment_game (except for error handling).

szhorvat:
> It also seems to me like the callaway_traits_game is equal to the establishment_game (except for error handling).

There are three types of games based on "traits":

 - `callaway_traits`: In each step, add 1 vertex, then try adding edges between existing vertices `k` times (succeeding with probability based on their types).
 - `establishment`: In each step, add 1 vertex, then try adding edges between the just-added vertex and another ones (succeeding with probability based on their types).
[...]

PR #1624, opened 2021-01-23
Update traits-based games

szhorvat:
This PR makes the changes detailed in #1620 .

Details:

 - Add `const` to input-only arguments.
 - Improve docs (e.g. `edges_per_step` is not the number of edges added, but the number of edges tried to be added with some probability)
 - callaway_traits and establishment now return vertex types in a new output argument
 - callaway_traits and establishment: `type_dist` is now allowed to be `NULL` in which case it is taken to be uniform
 - improve parameter checks
 - asymmetic_preference: allow different number of in- and out-types, reorder argument to always be out-types, then in-types as this feel natural given the interpretation of the type distribution matrix and preference matrix (rows: out, cols: in).
 - update old tests from `return`-style to `ASSERT`-style (otherwise the leak detector complains on failure)


szhorvat:
@GroteGnoom : I'm planning to touch `establishment_game`, so please wait with tests for that function. If you're already working on it,  let me know.

GroteGnoom:
> @GroteGnoom : I'm planning to touch `establishment_game`, so please wait with tests for that function. If you're already working on it, let me know.

I've skipped it for now, thanks for asking. Go ahead!

src/games/establishment.c
```diff
@@ -54,42 +58,100 @@
...
+
+    if (! directed && ! igraph_matrix_is_symmetric(pref_matrix)) {
+        IGRAPH_ERROR("The preference matrix must be symmetric when generating undirected graphs.", IGRAPH_EINVAL);
+    }
+
+    {
+        igraph_real_t lo, hi;
+        igraph_matrix_minmax(pref_matrix, &lo, &hi);
+
+        if (lo < 0 || hi > 1) {
```
ntamas, line 104: What if a NaN somehow sneaks into the preference matrix? Both the comparisons will be `false` if I understand correctly and we won't hit this branch, right?

ntamas (approved):
Looks good to me!

src/games/establishment.c
szhorvat: OK, will update all of them to detect NaNs. I originally started this PR before we added NaN detection.

szhorvat: Next we have to add some helpers which do the NaN check and error reporting automatically. Writing the same code again and again is getting tedious.

ntamas: A quick (although awkward) fix is to reverse the comparison: `!(lo >= 0 && hi <= 1)`

ntamas: Feel free to merge when you have updated the PR.

ntamas (approved):

PR #1625, opened 2021-01-23
Add lastcit game unit tests

GroteGnoom:
Part of #1592

### DIFF
--- a/src/games/citations.c
+++ b/src/games/citations.c
@@ -27,6 +27,7 @@
 #include "igraph_memory.h"
 #include "igraph_psumtree.h"
 #include "igraph_random.h"
+#include "igraph_interface.h"
 
 typedef struct {
     long int no;
@@ -103,6 +104,10 @@ int igraph_lastcit_game(igraph_t *graph,
         IGRAPH_ERROR("`preference' vector should be of length `agebins' plus one",
                      IGRAPH_EINVAL);
     }
+    if (nodes < 0 ) {
+        IGRAPH_ERROR("nodes should be non-negative",
+                     IGRAPH_EINVAL);
+    }
     if (agebins < 1 ) {
         IGRAPH_ERROR("pagebins should be at least 1",
                      IGRAPH_EINVAL);
@@ -113,6 +118,11 @@ int igraph_lastcit_game(igraph_t *graph,
     }
     if (igraph_vector_min(preference) < 0) {
         IGRAPH_ERROR("The preference vector must contain only non-negative values.", IGRAPH_EINVAL);
+    }
+
+    if (nodes == 0) {
+        IGRAPH_CHECK(igraph_empty(graph, nodes, directed));
+        return IGRAPH_SUCCESS;
     }
 
     binwidth = no_of_nodes / agebins + 1;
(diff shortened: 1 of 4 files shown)


szhorvat:
I should have realized this earlier: asserting that the return value is `IGRAPH_SUCCESS` is usually unnecessary because the default error handling function will abort anyway in that case. The exception is when you set an error handling function yourself, e.g. to be able to test non-successful values.

That said, please do not remove any existing assertions about this, as they do not hurt, and they would become useful if the error handler were changed in a test.

szhorvat:
Also, thanks for paying attention to the details and fixing or bringing up anything that looks off!

GroteGnoom:
@szhorvat I only now realize that this has the same issue as noticed in #1620, the length of the preferences vector is 1 + pagebins. So you can infer one from the other instead of checking they are equal. Should that be fixed here too?

GroteGnoom:
> I should have realized this earlier: asserting that the return value is `IGRAPH_SUCCESS` is usually unnecessary because the default error handling function will abort anyway in that case. The exception is when you set an error handling function yourself, e.g. to be able to test non-successful values.
> 
> That said, please do not remove any existing assertions about this, as they do not hurt, and they would become useful if the error handler were changed in a test.

Yes, for this reason I think I'll keep on adding them.

szhorvat:
When you write new error messages, or touch old ones, please use start with a capital letter and end in a fullstop. I know that this is not how most error messages are at the moment, but we'll update them eventually. 

Unless the meaning of parameter names are obvious, it's best to spell out what they mean in the error message. This is because the message may be seen by e.g. a python-igraph user who doesn't know what the C function looks like.

Some guidelines are here: https://github.com/igraph/igraph/wiki/Error-reporting-guidelines

Could we change the `pagebins` parameter name to `agebins`, to make it less confusing? (In the documentation too, not just the code.) I know it will conflict with the `agebins` variable, but is that variable needed at all? It seems the purpose might be to avoid compiler warnings about mixing `long` with `igraph_integer_t`. It could either be removed or renamed.

> @szhorvat I only now realize that this has the same issue as noticed in #1620, the length of the preferences vector is 1 + pagebins. So you can infer one from the other instead of checking they are equal. Should that be fixed here too?

Let's leave it for now. Later we can add a convenient default for the `preference` vector when `NULL` is supplied, such as the cubic decay default used by the R interface now: https://igraph.org/r/doc/sample_last_cit.html

@ntamas Any opinions on this?

ntamas:
The default used by the R interface seems good to me.

GroteGnoom:
> When you write new error messages, or touch old ones, please use start with a capital letter and end in a fullstop. I know that this is not how most error messages are at the moment, but we'll update them eventually.

Can we have an IGRAPH_ERRORF macro so I can provide better info (using igraph_errorf) without adding the line and file manually? That would also clear up things a lot. 

I'm also happy to clean up error messages in general for the functions I'm working with.

ntamas:
Now that we have switched to C99 a while ago I have nothing against having `IGRAPH_ERRORF`, provided that MSVC can cope with it.

ntamas:
LGTM, except for the typo in `recieved` :)

GroteGnoom:
The failing build is because of connection time-outs to github.

ntamas:
Thanks again!

Issue #1450, opened 2020-07-27
Put integer types into order


szhorvat:
Currently, igraph uses the following types for integers:

 - `igraph_real_t`, i.e. `double`: #556 and #1255 touch on this.
 - `igraph_integer_t` which is defined as `int`. In practice, this is currently only 32-bit even on 64-bit platforms.
 - `long`
 - internally, occasionally `int`s are used too

We should standardize the use of integer types. Here are the reasons:

 - The built-in types of C (`int`, `long`) have generally unpredictable sizes. For example, on 64-bit operating systems, `long` is 64-bit on macOS and x86 Linux, but it is 32-bit on Windows. _We should have some consistency across platforms._
 - The use of integer types is commonly tied to indexing into arrays. The C standard library provides `size_t` and `ptrdiff_t` for these. igraph tends to use `long` internally, which I assume was born out of the assumption that this would match the address size of the platform, but this is not the case in general (see `long` being 32-bit on Windows). Those internal `long`s are often cast to `igraph_integer_t`, which is `int`, and therefore the cast results in truncation.  _igraph should have an analogue of `ptrdiff_t` which matches the address size of the system._
 - _There should be guidelines on what integer type to use in what situation._ Right now, someone implementing new functions would be quite lost about whether to use `igraph_integer_t`, `int`, `long` or `size_t`. There should be some clarity about what should be used in a given situation and why.
 - We need to have some certainly about the sizes of integers before we can effectively tackle issues with integer overflow. See #51, #1446, and especially #1435.
 - _We should have some clarity about igraph's limits._ What is the largest supported edge count and vertex count? There should be checks to prevent exceeding these limits. There should be `IGRAPH_MAX_VCOUNT` and `IGRAPH_MAX_ECOUNT` constants provided. See #1435, a real-world use case where a large graph is successfully created, but its edge count cannot be retrieved with `igraph_ecount` as it does not fit into `igraph_integer_t`

----

**I would prefer the following solution for its simplicity:**

 - Let `igraph_integer_t` be a signed 32-bit integer on 32-bit platforms and 64-bit on 64-bit platforms (i.e. not simply use `int` or `long` but target a specific size that matches the architecture).
 - Use only `igraph_integer_t` in all situations, including for indexing into arrays. Effectively, `igraph_integer_t` will be the analogue of `ptrdiff_t` (signed value) and we don't bother with an unsigned analogue of `size_t`.
[...]

PR #1626, opened 2021-01-24
Fix integer types

vtraag:
This PR is an attempt at fixing integer types, as previously discussed in issue #1450.

This attempt is in essence quite straightforward, I have tried to make all internal integers that were internally used 64-bits, by introducing a new `igraph_long_t` type, which is typedefed as `int64_t`. The existing `igraph_integer_t` type is changed to be named `igraph_int_t` and is typedefed as `int32_t`. Unfortunately, we must maintain both integer types for compatibility with external libraries (e.g. CXSparse), which expect `int`'s to be provided. Hence, where needed, the integer type is `igraph_int_t`, but where possible the integer type is `igraph_long_t`. Similar changes were made for integer vectors (and other types), that are now changed to `igraph_vector_long_t` where possible.

In addition, I added an additional `igraph_error_t`, which is simply typedefed as an `int`. This is to ensure that the error code can always be interpreted as an `int`, but in practice it should always refer to an item of the `igraph_error_type_t` enumeration. All functions are changed accordingly, so that functions that return an error code now return `igraph_error_t` instead of some integer type. This makes the return type a bit more clear.

The library does build correctly, as do all tests. Some still fail however, which I need to look into still. Nonetheless, I already wanted to share this with you for feedback.

Regarding the configurability of the integer types, this is not yet done. But obviously, if we only use internal types consistently anyway (i.e. never an `int` or `long`, but only `igraph_int_t` and `igraph_long_t`), the configurability is (in theory) straightforward. Do note of course that when forcing an `igraph_long_t` to be 32-bits for compatibility with R also does limit internal integer sizes. Whether this is an immediate problem is difficult to see. The alternative solution would be to have some separate `igraph_public_integer_t` or something which is always only used for the public API, so that we can always use 64-bit integers internally and 32-bit integers in the public API, if R needs it. Of course, this will take quite some additional work.

Anyway, comments are welcome!


szhorvat:
I think this is a great start, and it similar in some ways to an attempt I made privately.

There are some concerns though I would like to note:

One main goal is to make use of 64-bit indices when available. This is accomplished.

Another main goal (that I had in mind) is to have a simple and predictable way to work. Having two types is not in this spirit. I would still like to see a single public integer type. I would like to have some easy to understand properties and guarantees for the single integer type, such as:

 - It can index anything. Vector indices and sizes can be this type. Vertex and edge IDs can be this type.
 - This makes it clear what the limits are on things such as graph sizes. If something can't be indexed with this type, then it is too large, and not supported.

Side note: Should we end up using more than one type anyway, I don't like the names `int` and `long` because people will assume that these are related to C's own `int` and `long` but thy are not the same kind of thing.

The reason @vtraag gives for having two types is so that we can interface with other libraries. I encountered this problem as well, with CXSparse as well as with BLAS/LAPACK/ARPACK. However, just having two types will not solve this problem anyway because the sizes of types differ from platform to platform. If, for example, we choose `our_int` to be `int32_t` and it matched `libx_int` from libX on one platform, it may not match it on a different platform. Thus we still would not be able to simply have `our_int` and `our_long` and always use `our_int` with libX on _any_ platform. It may be (and it is!) the case that `libx_int` is the same as `our_long` on Linux but it's `our_int` on Windows.

How do we deal with external libraries then?  I think the only general and manageably simple way is to convert data to that library's format (i.e. copy). Yes, this can affect performance, so we need to consider carefully when this solution is or isn't acceptable. A slightly faster but also slightly more complicated solution is to copy when there is no match, and pass arrays directly when there is a match. We would need to set up facilities for this.

Which libraries matter in practice?

 - CXSparse. Good news: this library is in fact adaptable to different integer sizes. It is not going to be a problem.
 - Fortran-origin things, namely BLAS/LAPACK/ARPACK. This is where it got unpleasant when I did my own exercise in changing igraph's integer type. However, the use of integers does not seem to be performance-critical with these libraries, meaning that it is okay to convert/copy without a major peformance impact.  Additionally, some implementations of these libraries come in multiple versions, with different integer sizes (MKL always does, but some other more common implementations don't). In the end, this stems from the fact that these libraries use Fortran's `integer` type which may have different sizes depending on the compiler ...

One more note about having multiple types:  When I did my own exercise in this, it seemed to me that CXSparse can be handled, but I have not done it (so I may have missed some things). I focused on BLAS mostly. When I did, I found it useful to introduce an _internal_  (not public!) `fortran_int` and `igraph_vector_fortran_int_t` type just so I can manage vectors more easily, can convert from an `igraph_vector_int_t` to an  `igraph_vector_fortran_int_t`.  The purpose of these types was simply to make it easier to interface with BLAS and convert igraph's native integers to Fortran-compatible integers.

----

In summary:

 - I would still like to have one single predictable and easy to understand integer type for igraph's public API, as well as the majority of internal igraph programming.
 - When interfacing with other libraries, it may be _convenient_ to introduce additional `vector` types, so we can write more concise code when dealing with those libraries. I think these types should be strictly private. It seems to me that _in theory_ we may need different such types in order to interface with different libraries. However, at present, the only one to deal with is BLAS/LAPACK/ARPACK. So there would be only one such thing for now.
 - CXSparse comes with multiple APIs, using multiple types. There are all combinations of `double` / `complex double` and `int` / `UF_long` where `UF_long` is designed to be precisely what we also want to do in igraph: 32-bit on 32-bit platforms and 64-bit on 64-bit platforms (including Windows 64). We should use the `dl` API, i.e. `double` with `UF_long`, and let CMake verify at configure time that `UF_long` is indeed the same as igraph's native integer (currently we use the `di` API). This is true on every system I am familiar with, but there are weird things out there ...  Sparse arrays are very useful when working with graphs, and I would like to ensure that _any_ graph that igraph supports can be converted to sparse adjacency matrix. The `dl` API will allow this.

Additionally:

 - It would be a nice optimization to avoid copying/converting to Fortran libraries (BLAS, etc.) when their integer type matches igraph's. I think this is doable without too much trouble, but if not (if problems arise), I think that there will be no serious performance impact from just copying on every platform.

ntamas:
I am about to jump on this bandwagon but I was wondering whether you guys know about ways to make compilers be super picky about integer upcasts / downcasts (i.e. implied conversion from 32-bit to 64-bit or vice versa). I know that there are tons of MSVC warnings that are related to this, but I would be interested to know whether it's possible to turn on something like this for Clang / `gcc`. I believe it would be useful in this stage so we can spot places where we could be running into problems by mixing integer types.

ntamas:
Also, I tend to agree with @szhorvat, it would be great if we could have one public integer type and make that configurable by the user. I would even try to keep on calling it `igraph_integer_t` to avoid confusing the user (just imagine how carefully we would need to explain that a new `igraph_int_t` is not the same as the old `igraph_integer_t`, and no matter how careful we are, many users would not read it anyway because not everyone is fond of reading changelogs).

@vtraag when you were creating this PR, were you simply running search-and-replace on the code w.r.t. the integer types or did you go through all source files manually? What was the general process? I would try to replicate what you did on the current `develop` branch just to get a feeling of it; merging `develop` into your PR creates lots of conflicts so I would not attempt that.

vtraag:
Overall, I agree with @szhorvat that it would be preferable to have a single (public) integer type. 

The suggestion to make integer types that are necessary to deal with external libraries strictly private sounds quite reasonable. Doing something similar for vector types that require different integer types is similarly reasonable. Dealing with non-matching and matching integer types separately sounds slightly more complicated, but it might be doable. I am not sure if this introduces fragilities into the build process and/or during runtime, and how large the gain would be of doing this in more complicated ways. I would be in favour of keeping it as simple as possible, if the performance hit is not too large.

I agree that `igraph_integer_t` is the preferable name. The only reason that I started using `igraph_int_t` and `igraph_long_t` is because I needed different types. If we make any other integer types strictly internal, I agree that the proper name for the public type should be `igraph_integer_t`.

I had not noticed that `CXSparse` actually has 64-bit type functionality, I believe that the X actually stands for that extension. That will already make that part of the codebase easier to work with. Nonetheless, if the integer type is configurable, we then might have to switch implementations, depending on whether the integer type is 32 or 64-bit, i.e. call `cs_di_xxx` or `cs_dl_xxx` respectively.

@ntamas, most of this PR indeed needs to be redone against the most recent `develop` branch. My approach was pretty simple, and you can mostly trace this via the long list of commits :) I basically just substituted every integer type to the new integer type, which initially was still called `igraph_integer_t`. This raises a whole list of issues, which later commits simply try to correct. There are number of issues that you will run into again when redoing this:
- Variadic arguments are never correctly interpreted as `int64_t`, they should always be interpreted as `int` (assuming `int` type will always work with variadic arguments). Perhaps we should further research this in order to be 100% sure that `int` will always be correctly interpreted. Since variadic argument will be limited anyway, this is not so problematic I believe.
- Random number generators use particular types. Since they may do something with bitwise operators, I think it is best to leave these as is. Presumably, anything that is every done with any bitwise operators will have to be carefully checked. I had not yet check this in-depth, but this should obviously this done for a proper implementation.
- There are a few instances in which unsigned types are used. As we all agree, we should preferably have only a single (public) integer type. I think it is fine to leave the unsigned types as they are, instead of creating separate private unsigned integer types, assuming they work well with the integer types that will be implemented.
- There are a few instances of `short`s, I believe I mostly left them as they were, but I am not 100% sure.

I haven't tried any tool that checks upcasting/downcasting and reports potential problems. I did do this for Java using IntelliJ, perhaps CLion would also have similar capabilities, I'm not sure. I guess that something like that should also be possible using `clang`, it seems that [`wimplicit-int-conversion`](https://clang.llvm.org/docs/DiagnosticsReference.html#wimplicit-int-conversion) might do the trick.

szhorvat:
> I had not noticed that `CXSparse` actually has 64-bit type functionality, I believe that the X actually stands for that extension. That will already make that part of the codebase easier to work with. **Nonetheless, if the integer type is configurable, we then might have to switch implementations, depending on whether the integer type is 32 or 64-bit, i.e. call `cs_di_xxx` or `cs_dl_xxx` respectively.**

The `dl` interface is not quite a 64-bit version one, but one that is 64-bit on 64-bit platforms and 32-bit on 32-bit one. We only need to use the `dl` functions, never the `di` ones. The `dl` functions use 32-bit integers on 32-bit platforms.

In other words, we are lucky and CXSparse's "UFlong" type aligns with what we want igraph's integer type to be.

> Dealing with non-matching and matching integer types separately sounds slightly more complicated, but it might be doable. I am not sure if this introduces fragilities into the build process and/or during runtime, and how large the gain would be of doing this in more complicated ways. **I would be in favour of keeping it as simple as possible, if the performance hit is not too large.**

Agreed.

This will only affect BLAS/LAPACK/ARPACK, which do not use integers extensively. Therefore, copying/covnerting integer types before passing to BLAS/LAPACK/ARPACK functions will _likely_ not affect performance noticeably.

It's good to note that these Fortran packages often come in two versions on 64-bit platforms: The [LP64](https://en.wikipedia.org/wiki/64-bit_computing#64-bit_data_models) version uses 32-bit integers and the ILP64 one uses 64-bit integers. LP64 is more common, and some BLAS implementations do not seem to support ILP64 at all (e.g. Apple's). Also, most distros / package managers include LP64 versions only.

For all these reasons it might be easier to just go with LP64 and copy/convert integers. However, given the amount of work we need to spend on the integer conversion anyway, it would be good to prepare for all eventualities, and make it so that it would not be too much work to move to ILP64 in the future.

vtraag:
> The `dl` interface is not quite a 64-bit version one, but one that is 64-bit on 64-bit platforms and 32-bit on 32-bit one. We only need to use the `dl` functions, never the `di` ones. The `dl` functions use 32-bit integers on 32-bit platforms.
> 
> In other words, we are lucky and CXSparse's "UFlong" type aligns with what we want igraph's integer type to be.

For most of the cases, yes, I would agree. However, R may be the exception here. If I understood correctly, the R interface doesn't have 64-bit integers. There are two solutions to this. The easiest one it to make the type configurable so that it *does* represent 32-bit integers, also on 64-bit platforms in order to accommodate for this for R. This way, it seamlessly integrated with R. The downside of this is that R will always need a special treatment. The alternative would be to always translate integers to whatever type R desires in the C-R interface. 

The latter solution would have my preference I think. I would be in favour of getting rid of all R specifics in the C core library, so that, ultimately, it should be possible for the R interface to simply link to an external igraph library (which then also has fewer CRAN-related requirements). Realistically, we cannot expect to do this until a 1.0 release is made. Nonetheless, if we want to make the integer type configurable, we might still need to accommodate for this in our interaction with CXSparse. The question is whether we really want to make it configurable or not.

szhorvat:
Right, of course: UFlong doesn't fully align with our type as we want ours to be configurable (potentially 32-bit on 64-bit).

ntamas:
Note to ourselves: we need to make `IGRAPH_PRId` (the printf format specifier for `igraph_integer_t`) public if we go down this road. Right now it's in a public header but the comment says that it is "for internal use only". We need to provide a portable way for users to print an `igraph_integer_t` without casting, no matter whether it's 32-bit or 64-bit (and we need to explain in the documentation that `igraph_integer_t` can be 32-bit or 64-bit).

szhorvat:
For reference, this also has to be changed, and in practice it's much more important than the typedef of igraph_integer_t:

https://github.com/igraph/igraph/blob/master/include/igraph_pmt.h#L77

ntamas:
@szhorvat  This is probably something we can change right now, even in `master`, without breaking things:

```diff
diff --git a/include/igraph_pmt.h b/include/igraph_pmt.h
index e97b27422..35f492cdb 100644
--- a/include/igraph_pmt.h
+++ b/include/igraph_pmt.h
@@ -74,9 +74,9 @@
     #define EQ(a,b) ((a && b) || (!a && !b))

 #elif defined(BASE_INT)
-    #define BASE int
+    #define BASE igraph_integer_t
     #define SHORT int
-    #define OUT_FORMAT "%d"
+    #define OUT_FORMAT "%" IGRAPH_PRId
     #define ZERO 0
     #define ONE 1
     #define MULTIPLICITY 1
```

All the tests still pass after making this change; no wonder since `igraph_integer_t` is an alias for `int` at the moment. Things start to break if I also change `igraph_integer_t` to be an `int64_t`, because then people who potentially used `printf("%d")` to print an element of an `igraph_vector_int_t` will use the wrong format string.

ntamas:
So I'm going ahead with the patch proposed above. Note that strictly speaking this change invalidates all code where we have an `igraph_vector_int_t` and we take the address of one of its elements and store it in an `int*`. It works right now because `igraph_integer_t` aligns with `int`, but it will break in the future when we change the type that `igraph_integer_t` points to.

Therefore, what I'm doing now is that I'm temporarily changing `igraph_integer_t` to an `int64_t` and identify all the places in the code where the compiler complains about incompatible pointer types, and I'm fixing it where possible. Then I'll switch `igraph_integer_t` back to `int` in order not to break the API. I was wondering whether I should commit this to `master` or `develop`; I'm leaning towards `develop` even if it is not a breaking change. Any objections?

szhorvat:
I think this is fine for `master`. I was actually quite surprised that you considered `vector_int_t` to have a different base type than `igraph_integer_t`. I always assumed that these two are the same.

ntamas:
> In other words, we are lucky and CXSparse's "UFlong" type aligns with what we want igraph's integer type to be.

Well, the match is not 100% perfect. It is true that `UFlong` is 64 bits on 64-bit platforms, _but_ on macOS both `long int` and `long long int` are 64 bits. `UFlong` is defined to be a `long`. If I naively change `igraph_integer_t` to be an `int64_t`, then I get a `long long` on macOS and the compiler throws warnings all around the places because `long long*` is incompatible with `long*`. So it is not enough to match the sizes; we need to match the exact _types_.

szhorvat:
We have to work with whatever type CXSparse supports. It supports a 32-bit one and a 64-bit one. On all platforms of relevance, all 32-bit types have the same representation, and all 64-bit ones have the same representation again.

There is the annoyance that to C, they may be different types and there will be warnings. That is what it is: one has to deal with it in some way. I have seen the same problem before, and I do not think there is any nice and clean way around it. Libraries (including the C standard library) will sometimes use different types to get an integer of a certain width. One library's 64-bit type may be `long` and another's a `long long`, and whether there is a match will be platform dependent. There will always be some platform where there's a mismatch.

ntamas:
CXSparse actually allows you to override `UF_long` to whatever you want, so one option is to pick a type in igraph and then override `UF_long`, `UF_long_max` and `UF_long_id` (which is the printf format specifier) as needed.

szhorvat:
> CXSparse actually allows you to override `UF_long` to whatever you want, so one option is to pick a type in igraph and then override `UF_long`, `UF_long_max` and `UF_long_id` (which is the printf format specifier) as needed.

That is an option for a bundled CXSparse, but not an external CXSparse. Or am I missing something?

(However: I do not think there's anything wrong with enforcing using a bundled CXSparse.)

ntamas:
You are right, indeed.

ntamas:
People interested in the progress of the 64-bit integer transition should follow the [`feature/int-size-64-bit` branch](https://github.com/igraph/igraph/tree/feature/int-size-64-bit) instead of this PR so I'm closing this.

PR #1629, opened 2021-01-26
Add new barabasi aging game unit test 

GroteGnoom:
One issue that remains is that the first number in the outseq vector is ignored. Is that supposed to happen?

A bug I fixed is that the definition of age_bin didn't fit with the documentation. It divided up the nodes into bins instead of specifying the nodes required for a time step.

I find this function really confusing. I can barely get it to do what I want, and usually require trial and error. Maybe that's just because I'm using it in an unusual way. I guess the main issue is that I didn't realize soon enough that you can't change the relative importance of the degree and age factors because they're multiplied. The degree and age coefficients are only used to scale it compared to the zero appeals. The mathematical function is of course not that complicated but maybe the naming and interpretation threw me off.

Aha! I have an index out of bounds because the code assumes there are nodes here:
IGRAPH_CHECK(igraph_psumtree_update(&sumtree, 0, zero_deg_appeal * (1 + zero_age_appeal)));  
I'll fix that too.


szhorvat:
Thanks for the PR! I'd like to take a closer look tomorrow before merging. Something looks off in the documentation (including in the original version).

szhorvat:
@gaborcsardi :

The documentation of the `aging_bin` parameter in the C interface was inconsistent with what this argument actually did.

The (detailed) C docs said that `aging_bin` controls the bin width, i.e. age increases by 1 after `aging_bin` steps. In reality, `aging_bin` controlled the number of bins. The bin width was computed as `no_of_nodes / aging_bin + 1` (which IMO is a bit weird due to the integer division). [The R docs](https://igraph.org/r/doc/sample_pa_age.html) are consistent with what the function actually did.

@gaborcsardi Which fix do you prefer? Change the C docs or change what the function does? This PR now changes what the function does.

The same question applies to `igraph_recent_degree_aging_game`. These two functions should be consistent with each other.

This is an old function with probably a niche application, so perhaps we don't want to change what it does at this point.

gaborcsardi:
> This is an old function with probably a niche application, so perhaps we don't want to change what it does at this point.

Yeah, agreed.

szhorvat:
@GroteGnoom I reverted the meaning of `aging_bins`. Please have another look at the test program before merging.

GroteGnoom:
It all seems to be tested as intended :+1:

PR #1630, opened 2021-01-26
Generalize the closeness centrality computation

szhorvat:
This PR generalizes the closeness calculation and addresses #1053.

Reminder: closeness is the inverse of the mean distance to all other vertices.

In short, closeness centrality is problematic for disconnected graphs. Currently, igraph takes the distance to non-reachable nodes to be `n` (the number of vertices). This is an ad-hoc choice that makes no sense for the weighted case.

The simplest solution would be to throw an error for disconnected graphs, but: (1) this is not useful for `closeness_cutoff`, where some vertices are always unreachable for a low-enough cutoff (2) doesn't serve users the best.

Here are possible ways to deal with this issue:

 1. igraph's current approach (not good).
 2. Use the mean distance to reachable vertices, not all vertices. In the undirected case, this effectively computes the closeness per-component. Problem: the centrality scores will be much larger in small components, and not comparable to those in large components.
 3. Refuse to work with disconnected graphs and throw an error. Problems: (1) not useful with `closeness_cutoff` (2) in the directed case, when computing the closeness only for some vertices, we can't immediately detect if the graph is connected. Note that igraph currently does show a warning if it detects that the graph is disconnected (but it won't always detect this).
 4. There are approaches which first use the mean distance only to reachable vertices, but then re-scale the scores using some function of the number of vertices that could be reached. I've seen more than one way. [networkx](https://networkx.org/documentation/stable/reference/algorithms/generated/networkx.algorithms.centrality.closeness_centrality.html) does this and points to a certain reference (Wasserman and Faust). Problem: The Wasserman and Faust approach effectively assumes that the mean distance scales linearly with the graph size (component size) and attempt to correct for this. This is simply not true. Consider a spatial network: the scaling depends on the spatial dimension.

My conclusion: There are many way to handle this and igraph should not choose to use one. Instead, it should provide maximum flexibility to use either of them.

My solution:

 - Use the mean distance to reachable vertices (not all vertices).
 - Also return the number of vertices that could be reached.

With this information, one can easily get the result of any of approaches 1-4 from above.  I would leave it to high-level interfaces to make a choice on what to implement, or provide multiple options.

### DIFF
--- a/examples/simple/centralization.c
+++ b/examples/simple/centralization.c
@@ -54,17 +54,7 @@ int main() {
         return 2;
     }
 
-    igraph_set_warning_handler(igraph_warning_handler_ignore);
-    igraph_centralization_closeness(&g, /*res=*/ 0,
-                                    IGRAPH_IN, &cent,
-                                    /*theoretical_max=*/ 0,
-                                    /*normalization=*/ 1);
-    igraph_set_warning_handler(igraph_warning_handler_print);
-
-    if (!ALMOST_EQUALS(cent, 1.0)) {
-        fprintf(stderr, "in-star, closeness: %g\n", cent);
-        return 3;
-    }
+    /* Skip closeness, as it is not well-defined for disconnected graphs such as an in-star. */
 
     igraph_destroy(&g);
 
@@ -90,22 +80,12 @@ int main() {
         return 12;
     }
 
-    igraph_set_warning_handler(igraph_warning_handler_ignore);
-    igraph_centralization_closeness(&g, /*res=*/ 0,
-                                    IGRAPH_OUT, &cent,
-                                    /*theoretical_max=*/ 0,
-                                    /*normalization=*/ 1);
-    igraph_set_warning_handler(igraph_warning_handler_print);
-
-    if (!ALMOST_EQUALS(cent, 1.0)) {
-        fprintf(stderr, "out-star, closeness: %g\n", cent);
-        return 13;
-    }
+    /* Skip closeness, as it is not well-defined for disconnected graphs such as an out-star. */
 
     igraph_destroy(&g);
 
     /****************************/
-    /* undricted star */
+    /* undirected star */
     igraph_star(&g, 10, IGRAPH_STAR_UNDIRECTED, /*center=*/ 0);
 
     igraph_centralization_degree(&g, /*res=*/ 0,
--- a/include/igraph_centrality.h
+++ b/include/igraph_centrality.h
@@ -38,9 +38,11 @@ __BEGIN_DECLS
 /* -------------------------------------------------- */
 
 DECLDIR int igraph_closeness(const igraph_t *graph, igraph_vector_t *res,
+                             igraph_vector_t *reachable_count, igraph_bool_t *all_reachable,
                              const igraph_vs_t vids, igraph_neimode_t mode,
                              const igraph_vector_t *weights, igraph_bool_t normalized);
 DECLDIR int igraph_closeness_cutoff(const igraph_t *graph, igraph_vector_t *res,
+                                    igraph_vector_t *reachable_count, igraph_bool_t *all_reachable,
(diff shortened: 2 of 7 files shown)


szhorvat:
Changes in this PR so far:

 - Fix the progress reporting. 
 - Add two optional (i.e. can be `NULL`) output arguments: (1) `reachable_count`, number of reachable vertices for each vertex. This is precisely the normalization factor used by `normalize` (2) `all_reachable` is set to true if all vertices of the graph could be reached from all vertices for which we do the computation. This is for convenience. A high-level interface could use this to issue a warning if we noticed that the graph was not connected.
 - For vertices with no paths (undirected: isolated vertices, directed: zero in- or out-degree depending on `mode`) we return `NaN`. This necessitated a change for the "centralization" test.
 - Re-write documentation.

This PR is draft, but I'd like feedback on this approach ASAP.

Still to do:
 - Add more tests
 - Fix an off-by-one issue with the cutoff parameter

szhorvat:
This is now done and ready for review, but please don't merge. I want to add more tests for the newly added parameters (it's informal tested from Mathematica).

ntamas (approved):

GroteGnoom:
I'm not sure how useful this is, but I tried to break your code, read the documentation and looked at all the changes (not the whole algorithm) and I couldn't really find anything wrong.

I can get a segfault at https://github.com/igraph/igraph/blob/34727b58e4d19d27cd4e3fcb8eb62f9962bd683d/src/centrality/closeness.c#L222 by putting a NAN in the real_edges of simple_test_case_with_weights_undirected, but that's probably not too important.

I was also wondering if infinity isn't sometimes more sensible than a NaN when replacing 1/0.

ntamas:
Looks good to me! The argument list is getting scary long now, but I don't really have a better idea about what to do here, and it's C so we can make the arguments optional in the higher level interfaces.

`functions.def` still needs to be updated for the new declaration.

szhorvat:
I chose to use NaN in order to have a consistent output regardless of whether we have normalization or not. Mathematica and networkx both return 0 for this case (isolated vertices). It's a special case that we need to make a choice for. 

Generally, closeness doesn't make a lot of sense for isolated vertices, and this is IMO better communicated by returning NaN than returning zero. If the user does some arithmetic with the closeness values, a zero will go unnoticed (but still influence the results) while a NaN will wreak havoc and draw attention (it won't go unnoticed). 

An example where one does arithmetic with *all* centrality values is the computation of "centralization"

GroteGnoom:
> I chose to use NaN in order to have a consistent output regardless of whether we have normalization or not. Mathematica and networkx both return 0 for this case (isolated vertices). It's a special case that we need to make a choice for.
> 
> Generally, closeness doesn't make a lot of sense for isolated vertices, and this is IMO better communicated by returning NaN than returning zero. 
Definitely. I was mainly looking at `infinity` as an alternative. I'm guessing that if NaN's are supported that `infinity` is too.

szhorvat:
There are two distinct issues:

 - What to return for isolated vertices (or vertices with zero in- or out-degree)? Could be 0, NaN, Inf ...
 - Should there be a difference in what we return based on normalization?

The direct calculation gives 0/0 for the normalized case and 1/0 for the unnormalized case. Naively, this would give us NaN and Infinity.

While this is subjective, I think that it is more useful to return the same value both in the normalized and unnormalized cases. This is less confusing for users. Now the question is: should we return Infinity or NaN? Given that the value isn't really meaningful, I would choose NaN. NaN indicates that "this makes no sense, don't use it" better. Furthermore, the usual definition of closeness is the normalized one, which would naively give NaN.

For these edge cases, usually there is no single correct solution, but neither is the choice we make critically important. It's good to try to make a choice that's most useful during practical numerical work though (as opposed to analytic calculations on paper).

If there's a good case to be made against NaN, of course I'm happy to hear it :-)

GroteGnoom:
Allright, thanks, now I understand 🙂.

PR #1632, opened 2021-01-27
Add cited_type_game unit tests

GroteGnoom:
Part of #1592


src/games/citations.c
```diff
@@ -275,7 +281,8 @@ int igraph_cited_type_game(igraph_t *graph, igraph_integer_t nodes,
     }
     nnval = VECTOR(*pref)[type];
     if (nnval < 0) {
-        IGRAPH_ERROR("pref contains negative entries", IGRAPH_EINVAL);
+        IGRAPH_ERRORF("Preferences should be non-negative, but found %f.", IGRAPH_EINVAL,
```
szhorvat, line 284: Let's use `%g` instead of `%f`. It's typically nicer. Examples (from a cling command line):

```
[cling]$ printf("%g\n", 1e5)
100000
(int) 7
[cling]$ printf("%f\n", 1e5)
100000.000000
(int) 14
[cling]$ printf("%g\n", 1e10)
1e+10
(int) 6
[cling]$ printf("%f\n", 1e10)
10000000000.000000
(int) 19
```

szhorvat:
@gaborcsardi I'm playing with this function in the R interface to reproduce the bug that was fixed in this PR by making `nnval` be real instead of integral, but I don't understand the behaviour I see:

 - Why does this not throw an error? `cited.type.game(n=10, edges=1, types=c(1,2), pref=c(3,4))`  The length of `types` should be the same as the `n`, and this is already checked in the C interface.

 - Why is the default value of `pref` [in the R interface](https://igraph.org/r/doc/sample_last_cit.html) `pref = rep(1,
  length(types))` instead of `pref = rep(1,
  max(types))`? It does not seem right.

gaborcsardi:
> Why does this not throw an error?

Because there is no check in the C code of the R package yet (?).

> Why is the default value of ...

Seems like a bug to me.

PR #1635, opened 2021-01-28
Add recent degree aging unit tests

GroteGnoom:
Part of #1592.
This should wait for the results of #1629


GroteGnoom:
#1629 has been cleared up, so I'll change this code according to those results.

PR #1638, opened 2021-01-29
Improve NaN documentation and fixing various functions for vectors

vtraag:
This PR fixes some remaining functions of the vector API, as was previously started in #1617, aimed at addressing issue #1560 (although this PR doesn't fix any other functions).

For any function that defines any ordering, or depends on any ordering, I included a note in the documentation that the order is undefined when NaN elements are included (which I think is the best solution). There are still a number of functions that are not documented, I have not touched these.


ntamas (approved):
Is this good to go?

szhorvat:
@vtraag Can you also fix up `vector_(which_)minmax()` please? I was just fixing some error handling, and noticed this function. It's often used with probabilities. 

We also have `matrix` min/max function, which call the `vector` functions. Can you check if these need a doc update about NaN?

szhorvat:
It's fine by me.

PR #1640, opened 2021-01-30
PageRank: the walk should be restarted consistently between the ARPACK and PRPACK methods

szhorvat:
This PR fixes the problem where _personalized_ PageRank results for _disconnected_ graphs differ between the ARPACK and PRPACK methods.

A hint about what goes wrong is from examining an graph with no edges. With damping `d` approaching 1, the PRPACK result is uniform. With `d -> 0`, the PRPACK result is the same as the `reset` vector. PRPACK has two parameters, `u` and `v`. `u` was set to NULL and `v` was set to `reset`. Thus, `u` controls the vertex choice after resetting due to being stuck in a sink, and `v` controls the vertex choice during "normal" reset.

This PR sets both to be the same and explains the behaviour in the docs.

Additionally, the tests are moved to `tests/unit`, updated with a more modern approach, and expanded. I added a new small example in `examples/simple`.


src/centrality/centrality_other.c
```diff
@@ -1228,11 +1235,15 @@ int igraph_personalized_pagerank_vs(const igraph_t *graph,
  * \brief Calculates the personalized Google PageRank for the specified vertices.
  *
  * The personalized PageRank is similar to the original PageRank measure, but the
```
ntamas, line 1237: "the" is not necessary at the end of this line after the modifications

ntamas (approved):

Issue #1569, opened 2020-12-15
will Graph.get_shortest_paths() support bellman_ford algorithm


chenzhiw:
Graph.get_shortest_paths() also supports edge weights (**only Dijkstra's algorithm yet**)
Will it support bellman_ford algorithm for negative weights in the future?
I have to convert it to SciPy CSR matrix first, and traceback myself now.

ntamas:
There are no immediate plans to implement `igraph_get_shortest_paths_bellman_ford()` yet; pull requests are welcome. I have added this to the wishlist.

ayanginet:
I am a bit confused. Isn't it already implemented?
https://github.com/igraph/igraph/blob/520ee0c550cf7f9b4d6dd379e6f20f8000c03792/src/structural_properties.c#L5396-L5401

ntamas:
That function will only give you the _lengths_ of the shortest paths, not the paths themselves. `get_shortest_paths_bellman_ford()` should give you the paths as well.

szhorvat:
Indeed, this has been implemented by @guyroznb

PR #1642, opened 2021-01-30
igraph_get_shortest_path() using Bellman-Ford algorithm

guyroznb:
Hi,
This PR is answering the PR welcome #1569.
thanks!


src/paths/bellman_ford.c
```diff
@@ -203,3 +210,285 @@ int igraph_shortest_paths_bellman_ford(const igraph_t *graph,
...
+    }
+
+    igraph_dqueue_destroy(&Q);
+    igraph_vector_destroy(&clean_vertices);
+    igraph_vector_destroy(&num_queued);
+    igraph_lazy_inclist_destroy(&inclist);
+    IGRAPH_FINALLY_CLEAN(5);
+
+    return 0;
+}
```
szhorvat, line 494: This needs a newline to make the compiler happy.

szhorvat:
Thanks for the PR! We'll take a look likely next week.

Here are some tips for dealing with the test failures:

The macOS failure is trivial: you just need to add a newline at the end.

For the Linux one with sanitizers, I suggest some rewriting of the tests before you try to solve this:

 - Please move the test to `test/unit` instead of `examples/simple` and remove the reference to the file from the docs. It is not that useful as an example—let's make it purely a test.
 - Add `#include "test_utilities.inc"` and eventually we should transition to using the vector printing functions from there. Transitioning to these printing functions can be done as the very last step.
 - Do add `VERIFY_FINALLY_STACK()` to just before the end of the test.
 - Now comes the important part to help you track down the leaks: do not use `exit()` for failing the test. This _will_ trigger the leak sanitizer which will mask the true cause of the failure. Instead, use `IGRAPH_ASSERT` to check assumptions. Once this has been done, you will see if the leaks are true leaks or are caused by an early exit due to a test failure.

szhorvat:
There are updated tips for writing unit tests here: https://github.com/igraph/igraph/wiki/How-to-write-unit-tests%3F

examples/simple/igraph_get_shortest_paths_bellman_ford.c
```diff
@@ -0,0 +1,237 @@
...
+    igraph_vector_long_init(&pred, 0);
+    igraph_vector_long_init(&inbound, 0);
+
+    for (i = 0; i < igraph_vector_ptr_size(&vecs); i++) {
+        VECTOR(vecs)[i] = calloc(1, sizeof(igraph_vector_t));
+        igraph_vector_init(VECTOR(vecs)[i], 0);VECTOR(evecs)[i] = calloc(1, sizeof(igraph_vector_t));
+        igraph_vector_init(VECTOR(evecs)[i], 0);
+    }
+    igraph_vs_vector_small(&vs, 0, 1, 3, 2, 1,  -1);
+    igraph_small(&g, 5, IGRAPH_DIRECTED,
```
szhorvat, line 189: One leak is from here: this line is re-allocating `g` without having destroyed it first.

CMakeFiles/cmake.check_cache
```diff
@@ -0,0 +1 @@
+# This file is generated by cmake for dependency checking of the CMakeCache.txt file
```
szhorvat, line 1: Please remove this file from the PR (just add a new commit that deletes it).

src/paths/bellman_ford.c
```diff
@@ -203,3 +210,285 @@ int igraph_shortest_paths_bellman_ford(const igraph_t *graph,
...
+    }
+
+    if (vertices && IGRAPH_VIT_SIZE(tovit) != igraph_vector_ptr_size(vertices)) {
+        IGRAPH_ERROR("Size of `vertices' and `to' should match", IGRAPH_EINVAL);
+    }
+    if (edges && IGRAPH_VIT_SIZE(tovit) != igraph_vector_ptr_size(edges)) {
+        IGRAPH_ERROR("Size of `edges' and `to' should match", IGRAPH_EINVAL);
+    }
+
+    parents = igraph_Calloc(no_of_nodes, long int);
```
szhorvat, line 348: It looks like this is never freed, that's another leak.

src/paths/bellman_ford.c
```diff
@@ -203,3 +210,285 @@ int igraph_shortest_paths_bellman_ford(const igraph_t *graph,
...
+    if (!all_to) {
+        igraph_vit_destroy(&tovit);
+        IGRAPH_FINALLY_CLEAN(1);
+    }
+
+    igraph_dqueue_destroy(&Q);
+    igraph_vector_destroy(&clean_vertices);
+    igraph_vector_destroy(&num_queued);
+    igraph_lazy_inclist_destroy(&inclist);
+    IGRAPH_FINALLY_CLEAN(5);
```
szhorvat, line 491: This line cleans 5 entries from the finally stack, but there are only 4 destroy calls above. Is this where `parents` should be destroyed?

examples/simple/igraph_get_shortest_paths_bellman_ford.c
```diff
@@ -0,0 +1,237 @@
...
+    igraph_vector_ptr_init(&evecs, 5);
+    igraph_vector_long_init(&pred, 0);
+    igraph_vector_long_init(&inbound, 0);
+
+    for (i = 0; i < igraph_vector_ptr_size(&vecs); i++) {
+        VECTOR(vecs)[i] = calloc(1, sizeof(igraph_vector_t));
+        igraph_vector_init(VECTOR(vecs)[i], 0);VECTOR(evecs)[i] = calloc(1, sizeof(igraph_vector_t));
+        igraph_vector_init(VECTOR(evecs)[i], 0);
+    }
+    igraph_vs_vector_small(&vs, 0, 1, 3, 2, 1,  -1);
```
szhorvat, line 188: Another leak is here: this re-allocates on top of the original `vs` without destroying it first.

src/paths/bellman_ford.c
guyroznb: changed - added free at the end of the program.

src/paths/bellman_ford.c
guyroznb: You are right... I change the number of entries to 4

examples/simple/igraph_get_shortest_paths_bellman_ford.c
guyroznb: added destroy function to vs

examples/simple/igraph_get_shortest_paths_bellman_ford.c
guyroznb: added destroy to g before re - allocation

src/paths/bellman_ford.c
```diff
@@ -203,3 +210,284 @@ int igraph_shortest_paths_bellman_ford(const igraph_t *graph,
...
+    }
+
+    igraph_vector_destroy(&dist);
+    IGRAPH_FINALLY_CLEAN(1);
+
+    if (!all_to) {
+        igraph_vit_destroy(&tovit);
+        IGRAPH_FINALLY_CLEAN(1);
+    }
+    free(parents);
```
szhorvat, line 485: Just some quick comments (I'll take a proper look next week): please use `igraph_Free` here.

src/paths/bellman_ford.c
```diff
@@ -203,3 +210,284 @@ int igraph_shortest_paths_bellman_ford(const igraph_t *graph,
...
+    igraph_dqueue_destroy(&Q);
+    igraph_vector_destroy(&clean_vertices);
+    igraph_vector_destroy(&num_queued);
+    igraph_lazy_inclist_destroy(&inclist);
+    IGRAPH_FINALLY_CLEAN(4);
```
szhorvat, line 490: Now we _do_ have 5 things to clean, including the `igraph_Free(parents)`, so this should indeed be 5 now.

src/paths/bellman_ford.c
```diff
@@ -203,3 +210,284 @@ int igraph_shortest_paths_bellman_ford(const igraph_t *graph,
...
+
+    return 0;
```
szhorvat, line 492: Please use `return IGRAPH_SUCCESS`.

szhorvat:
There still seem to be some memory leaks. I'll take a detailed look around mid-next week (unless someone else gets to it first) and will help you finalize the PR then.

guyroznb:
Hi @szhorvat I managed to solve the memory leaks.. thanks!

szhorvat:
> * Please move the test to `test/unit` instead of `examples/simple` and remove the reference to the file from the docs. It is not that useful as an example—let's make it purely a test.
> 
> * Add `#include "test_utilities.inc"` and eventually we should transition to using the vector printing functions from there. Transitioning to these printing functions can be done as the very last step.
> 
> * Do add `VERIFY_FINALLY_STACK()` to just before the end of the test.
> 
> * Now comes the important part to help you track down the leaks: do not use `exit()` for failing the test. This _will_ trigger the leak sanitizer which will mask the true cause of the failure. Instead, use `IGRAPH_ASSERT` to check assumptions. Once this has been done, you will see if the leaks are true leaks or are caused by an early exit due to a test failure.

If you'd be able to rewrite the tests using the above guidelines, that would be really helpful. We would need to make this change before merging these feature. While it's not difficult, it takes some time to do.

guyroznb:
Hi @szhorvat.
So, I managed to move the example to tests (and pull the latest igraph_path.h from develop).
I left with this weird problem ("Failed to connect to github.com port 443: Timed out") which I don't know how to resolve.
Do you have any Idea?
thanks.....

GroteGnoom:
> I left with this weird problem ("Failed to connect to github.com port 443: Timed out") which I don't know how to resolve.

I think usually when you get this error you didn't do anything wrong and can't fix it, but you just have to try again later. 
You can trigger the automated checks again by amending your last commit, changing nothing and then force-pushing.

guyroznb:
thanks @GroteGnoom...

szhorvat:
Thanks for working on fixing up the tests! It seems there was a misunderstanding about the purpose of `IGRAPH_ASSERT`: it is used to verify that a condition is true.

Thus, instead of writing

```
if (expected_to_be_one != 1) {
    printf("Value was not 1!\n");
    return 13;
}
```

we should use

```
IGRAPH_ASSERT(expected_to_be_one == 1);
```

If the condition within the `IGRAPH_ASSERT` is not true, the program will be aborted, and a message containing the condition, as well as its location within the source file will be printed.

If anything is unclear, feel free to ask.

I have a deadline coming up, so I ask for some more time before dealing with the details.

interfaces/functions.def
```diff
@@ -463,6 +463,16 @@ igraph_get_shortest_paths_dijkstra:
               weights ON graph
         IGNORE: RR, RC, RNamespace
 
+igraph_get_shortest_paths_bellman:
```
ntamas, line 466: The function that you've added is called `igraph_get_shortest_paths_bellman_ford`

interfaces/functions.def
```diff
@@ -476,6 +486,7 @@ igraph_shortest_paths_bellman_ford:
         DEPS: from ON graph, to ON graph, weights ON graph
         IGNORE: RR, RC, RNamespace
 
+
```
ntamas, line 489: This newline is not needed here

src/paths/bellman_ford.c
```diff
@@ -203,3 +210,285 @@ int igraph_shortest_paths_bellman_ford(const igraph_t *graph,
...
 }
+
+
+/**
+ * \ingroup structural
+ * \function igraph_get_shortest_paths_bellman_ford
+ * \brief Calculates the weighted shortest paths from/to one vertex 
+ * allowing negative weights.
+ *
+ * If there are no negative ]weights, you are better off with
```
ntamas, line 221: Please remove the extra `]` from this line

src/paths/bellman_ford.c
```diff
@@ -203,3 +210,285 @@ int igraph_shortest_paths_bellman_ford(const igraph_t *graph,
 
     return 0;
 }
+
+
+/**
+ * \ingroup structural
+ * \function igraph_get_shortest_paths_bellman_ford
+ * \brief Calculates the weighted shortest paths from/to one vertex
```
ntamas, line 218: Not sure if our doc parser can parse multi-line `\brief` directives; please write it on one line and add a comma before "allowing"

src/paths/bellman_ford.c
```diff
@@ -203,3 +210,285 @@ int igraph_shortest_paths_bellman_ford(const igraph_t *graph,
...
+ *        if they are both null pointers.
+ * \param from The id of the vertex from/to which the geodesics are
+ *        calculated.
+ * \param to Vertex sequence with the ids of the vertices to/from which the
+ *        shortest paths will be calculated. A vertex might be given multiple
+ *        times.
+ * \param weights The edge weights. There mustn't be any closed loop in
+ *    the graph that has a negative total weight (since this would allow
+ *    us to decrease the weight of any path containing at least a single
+ *    vertex of this loop infinitely). If this is a null pointer, then the
```
ntamas, line 251: Let's document what the function will do if there is such a closed loop in the graph. (I guess it would just spin infinitely, but let's make it explicit).

ntamas: Scratch that, I see that we have an error code to indicate negative loops.

src/paths/bellman_ford.c
```diff
@@ -203,3 +210,285 @@ int igraph_shortest_paths_bellman_ford(const igraph_t *graph,
...
+    igraph_vector_null(&clean_vertices);
+    igraph_vector_null(&num_queued);
+
+    /* Fill the queue with vertices to be checked */
+    for (j = 0; j < no_of_nodes; j++) {
+        IGRAPH_CHECK(igraph_dqueue_push(&Q, j));
+    }
+
+    while (!igraph_dqueue_empty(&Q)) {
+        igraph_vector_t *neis;
```
ntamas, line 366: `igraph_lazy_inclist_t` has recently been rewritten to use `igraph_vector_int_t` for sake of consistency with the other adjlist/inclist data structures.

tests/unit/igraph_get_shortest_paths_bellman_ford.c
```diff
@@ -0,0 +1,226 @@
+/* -*- mode: C -*-  */
+/*
+   IGraph library.
+   Copyright (C) 2006-2012  Gabor Csardi <csardi.gabor@gmail.com>
```
ntamas, line 4: This should now be `(C) 2006-2021 The igraph development team`, and the address in the next line is not necessary.

tests/unit/igraph_get_shortest_paths_bellman_ford.c
```diff
@@ -0,0 +1,226 @@
...
+   Foundation, Inc.,  51 Franklin Street, Fifth Floor, Boston, MA
+   02110-1301 USA
+
+*/
+
+#include <igraph.h>
+#include "test_utilities.inc"
+#include <stdlib.h>
+
+void print_vector_li(igraph_vector_t *v) {
```
ntamas, line 28: Why is `print_vector_round()` not suitable from `test_utilities.inc`?

ntamas (changes requested):
Looks good to me in general; a few minor changes requested here and there before we can merge this.

guyroznb:
thanks for the review @ntamas

ntamas (approved):

ntamas:
Thanks a lot!

szhorvat:
Something that was missing is adding this to the documentation. This is now done in efc6e2b50b296102c929ab38c2991a719dfe17c5  Some problems that needed to be fixed:

 - We don't support `\test` (and tests are not of concern to users, so don't need to be linked from docs).
 - There were some small syntax errors that prevented the docs from being built.
 
Since all the other similar functions has a `get_shortest_path` (not `paths`) version, it would be nice to add one here too, for consistency.

szhorvat:
@guyroznb Let me know if you plan to add `igraph_get_shortest_path_bellman_ford` (note `path` not `paths`) today. If not, I'll add it. Just want to avoid stepping on each others' toes.

szhorvat:
`igraph_get_shortest_path_bellman_ford` is now implemented.

PR #1644, opened 2021-02-01
Do not export non-public symbols from shared library

szhorvat:
This uses CMake's own features to handle dllimpor/dllexport, and uses visibility on Unix as well.

This is not even close to being done, it's just a proof of concept. Comments are welcome.


szhorvat:
This is working now, so I am marking it as ready for review.

We may want to make some cosmetic changes, such as the location of the `#include` in `igraph_decls.h`, and renaming `DECLDIR` to avoid conflict with other libraries.

I suggest renaming `DECLDIR` to `IGRAPH_EXPORT` as this is what CMake produces by default (not the difference from `igraph_EXPORTS` which used to be called `IGRAPH_EXPORTS`). Alternatively, we can us `IGRAPH_DECLDIR` (although `DIR`/direction is not meaningful for visibility on non-Windows platforms).

szhorvat:
You can check that this works as intended by listing the symbols in the shared library with something like `nm -gUC libigraph.0.dylib` (macOS) and checking that only public symbols are show. There will be a few more "typeinfo" in there, for C++ exceptions (I don't understand the technical details of why these appear).

vtraag:
Thanks, great to see this! I'll try to take a closer look shortly. 

> I suggest renaming `DECLDIR` to `IGRAPH_EXPORT` as this is what CMake produces by default

Yes, that sounds good.

One small remark: we should check that the generated include file `igraph_export.h` is properly packaged. Now that I am looking at this, we should double check the same for the other generated files (i.e. `igraph_version.h` and `igraph_threading.h` are at least in `include/` in my build folder). At least, I'm now not 100% sure if these are being packaged. @ntamas, what do you think?

vtraag:
Or actually, we of course don't need to package these files since they will be regenerated.

ntamas (approved):

src/cliques/cliquer/CMakeLists.txt
```diff
@@ -12,18 +12,17 @@ target_include_directories(
...
   set_property(TARGET cliquer PROPERTY POSITION_INDEPENDENT_CODE ON)
 endif()
 
-if (WIN32)
-  # Since these are included as object files, they should call the
-  # function as is (without declspec)
-  target_compile_definitions(cliquer PRIVATE IGRAPH_STATIC)
-endif()
+# Since these are included as object files, they should call the
+# function as is (without declspec)
```
vtraag, line 24: Now that `IGRAPH_STATIC` or `igraph_EXPORTS` are also valid on other platforms besides Windows, should we perhaps clarify this comment a bit? (The same of course holds for the other locations).

vtraag (commented):
It all looks good to me! Only one minor comment regarding, well, a comment :)

src/cliques/cliquer/CMakeLists.txt
szhorvat: Done (see last commit)

vtraag (approved):

PR #1645, opened 2021-02-02
Add compare_communities unit tests

GroteGnoom:
Part of #1592

The Adjusted Rand on 2 vertices in 2 clusters still outputs -NaN,
but that seems correct and a bit excessive to check for up front.

I'm not sure why this function is used instead of just having separate
function calls. Separate function calls would also clear up the
documentation.

### DIFF
--- a/src/community/misc.c
+++ b/src/community/misc.c
@@ -727,7 +727,7 @@ static int igraph_i_compare_communities_rand(
 
     if (igraph_vector_size(v1) <= 1) {
         IGRAPH_ERRORF("Rand indices not defined for only zero or one vertices. "
-        "Found memberschip vector of size %ld", IGRAPH_EINVAL, igraph_vector_size(v1));
+        "Found membership vector of size %ld", IGRAPH_EINVAL, igraph_vector_size(v1));
     }
 
     /* Calculate the confusion matrix */
--- a/src/community/misc.c
+++ b/src/community/misc.c
@@ -230,7 +230,8 @@ int igraph_reindex_membership(igraph_vector_t *membership,
         long int c = (long int)VECTOR(*membership)[i];
 
         if (c >= n) {
-            IGRAPH_ERROR("Cluster out of range", IGRAPH_EINVAL);
+            IGRAPH_ERRORF("Membership indices should be less than total number of vertices. "
+            "Found member of cluster %ld, but only %ld vertices.", IGRAPH_EINVAL, c, n);
         }
 
         if (VECTOR(new_cluster)[c] == 0) {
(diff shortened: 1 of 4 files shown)


GroteGnoom:
Allright, linux outputs:
`Adjusted Rand result: -nan`
and Windows and Mac do:
`Adjusted Rand result: -nan(ind)`
I'll see what's going on there.

ntamas:
We have `igraph_real_printf()` in `igraph_types.h` that is meant to work around platform-specific printing styles for NaN. We also have `print_real()` in `test_utilities.inc` for the same purpose.

GroteGnoom:
Ah, perfect, thank you!

src/community/misc.c
```diff
@@ -703,6 +725,11 @@ static int igraph_i_compare_communities_rand(
     double rand, n;
     double frac_pairs_in_1, frac_pairs_in_2;
 
+    if (igraph_vector_size(v1) <= 1) {
+        IGRAPH_ERRORF("Rand indices not defined for only zero or one vertices. "
+        "Found memberschip vector of size %ld", IGRAPH_EINVAL, igraph_vector_size(v1));
```
ntamas, line 730: Typo here :) (`memberschip`)

tests/unit/igraph_compare_communities.c
```diff
@@ -0,0 +1,102 @@
...
+
+   You should have received a copy of the GNU General Public License
+   along with this program.  If not, see <https://www.gnu.org/licenses/>.
+*/
+
+#include <igraph.h>
+#include "test_utilities.inc"
+
+char *names[] = {
+    [IGRAPH_COMMCMP_VI] = "VI",
```
ntamas, line 23: Hmmm, this is nice but I wonder whether this construct is understood by most C compilers. Is this a standard thing? @szhorvat ?

szhorvat:
All is fine with this PR, but I'm confused about how one gets a `-NaN`. I thought NaNs don't have a sign. Perhaps some implementations print the sign bit anyway?

ntamas:
Are we using `-ffast-math`? I think that NaNs get all screwed up with that.

tests/unit/igraph_compare_communities.c
GroteGnoom: Designated initializers are in C99, not in earlier standards. I hope I remember correctly we use C99.

Figuring out if MSVC actually supports it is hard, but I found a blog post comment which references another comment that doesn't exist anymore which seems to indicate that VS 2013 supports it.

https://devblogs.microsoft.com/cppblog/c11-and-c17-standard-support-arriving-in-msvc/#comment-1230

I used it to not depend on the enum order.

GroteGnoom:
Thanks for that website 🙂 It's kind of strange how complicated a NaN is. 

I do agree that -NaN doesn't make sense, but I guess in practice (as in this case) it just means -0.0/0.0.

tests/unit/igraph_compare_communities.c
szhorvat: I tried it here with VS2015, seems to be good: https://godbolt.org/z/z3qWa1

szhorvat: Note that it's only available in C mode, not in C++ mode, but that's fine.

We should keep supporting VS2015, i.e. version 19.0 (not 19.x where x > 0), for the time being. This seems to be fine with 19.0

PR #1647, opened 2021-02-04
Add le_community_to_membership unit tests

GroteGnoom:
Part of #1592

I removed the
`igraph_matrix_nrow(merges) < steps`
test from le_community_to membership, because doesn't do any good there and is already tested in community_to_membership.

I added a 'already merged' test to community_to_membership because it shouldn't happen and gives unexpected results.

### DIFF
--- a/src/community/leading_eigenvector.c
+++ b/src/community/leading_eigenvector.c
@@ -1007,13 +1007,13 @@ int igraph_le_community_to_membership(const igraph_matrix_t *merges,
     long int components, i;
 
     if (no_of_nodes > 0) {
-            components = (long int) igraph_vector_max(membership) + 1;
+        components = (long int) igraph_vector_max(membership) + 1;
     } else {
         components = 0;
     }
     if (components > no_of_nodes) {
         IGRAPH_ERRORF("Invalid membership vector, number of components is %ld, but should "
-         "not be greather than number of nodes, which is %ld.", IGRAPH_EINVAL, components, no_of_nodes);
+         "not be greater than the number of nodes, which is %ld.", IGRAPH_EINVAL, components, no_of_nodes);
     }
     if (steps >= components) {
         IGRAPH_ERRORF("Number of steps should be smaller than number of components. "
--- a/src/community/leading_eigenvector.c
+++ b/src/community/leading_eigenvector.c
@@ -984,9 +984,10 @@ int igraph_community_leading_eigenvector(const igraph_t *graph,
  * It takes \c membership
  * and performs \c steps merges, according to the supplied
  * \c merges matrix.
- * \param merges The matrix defining the merges to make.
- *     This is usually from the output of the leading eigenvector community
- *     structure detection routines.
+ * \param merges The two-column matrix containing the merge
+ *    operations. See \ref igraph_community_walktrap() for the
+ *    detailed syntax. This is usually from the output of the
+ *    leading eigenvector community structure detection routines.
  * \param steps The number of steps to make according to \c merges.
  * \param membership Initially the starting membership vector,
  *     on output the resulting membership vector, after performing \c steps merges.
@@ -1005,17 +1006,19 @@ int igraph_le_community_to_membership(const igraph_matrix_t *merges,
     igraph_vector_t fake_memb;
     long int components, i;
 
-    if (igraph_matrix_nrow(merges) < steps) {
-        IGRAPH_ERROR("`steps' to big or `merges' matrix too short", IGRAPH_EINVAL);
+    if (no_of_nodes > 0) {
+            components = (long int) igraph_vector_max(membership) + 1;
+    } else {
+        components = 0;
     }
-
-    components = (long int) igraph_vector_max(membership) + 1;
     if (components > no_of_nodes) {
-        IGRAPH_ERROR("Invalid membership vector, too many components", IGRAPH_EINVAL);
+        IGRAPH_ERRORF("Invalid membership vector, number of components is %ld, but should "
+         "not be greather than number of nodes, which is %ld.", IGRAPH_EINVAL, components, no_of_nodes);
     }
     if (steps >= components) {
-        IGRAPH_ERROR("Cannot make `steps' steps from supplied membership vector",
-                     IGRAPH_EINVAL);
+        IGRAPH_ERRORF("Number of steps should be smaller than number of components. "
+                     "Found %" IGRAPH_PRId " steps, %ld components.",
+                     IGRAPH_EINVAL, steps, components);
     }
 
     IGRAPH_VECTOR_INIT_FINALLY(&fake_memb, components);
@@ -1023,13 +1026,13 @@ int igraph_le_community_to_membership(const igraph_matrix_t *merges,
     /* Check membership vector */
     for (i = 0; i < no_of_nodes; i++) {
         if (VECTOR(*membership)[i] < 0) {
-            IGRAPH_ERROR("Invalid membership vector, negative id", IGRAPH_EINVAL);
+            IGRAPH_ERRORF("Invalid membership vector, negative id found: %g.", IGRAPH_EINVAL, VECTOR(*membership)[i]);
         }
         VECTOR(fake_memb)[ (long int) VECTOR(*membership)[i] ] += 1;
     }
     for (i = 0; i < components; i++) {
         if (VECTOR(fake_memb)[i] == 0) {
-            IGRAPH_ERROR("Invalid membership vector, empty cluster", IGRAPH_EINVAL);
+            IGRAPH_ERRORF("Invalid membership vector, empty cluster found, with index %ld.", IGRAPH_EINVAL, i);
         }
     }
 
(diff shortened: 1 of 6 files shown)


src/community/leading_eigenvector.c
```diff
@@ -1005,31 +1006,33 @@ int igraph_le_community_to_membership(const igraph_matrix_t *merges,
     igraph_vector_t fake_memb;
     long int components, i;
 
-    if (igraph_matrix_nrow(merges) < steps) {
```
ntamas, line 1008: `igraph_le_community_to_membership()` is part of the public API (although I don't quite understand why, but this ship has sailed), so in theory anyone can call this function with weird merge matrices. Therefore, this check should be kept, even though it is already checked in the "normal" route where we call `igraph_le_community_to_membership()` from within the library.

src/community/leading_eigenvector.c
```diff
@@ -1005,31 +1006,33 @@ int igraph_le_community_to_membership(const igraph_matrix_t *merges,
...
+            components = (long int) igraph_vector_max(membership) + 1;
+    } else {
+        components = 0;
     }
-
-    components = (long int) igraph_vector_max(membership) + 1;
     if (components > no_of_nodes) {
-        IGRAPH_ERROR("Invalid membership vector, too many components", IGRAPH_EINVAL);
+        IGRAPH_ERRORF("Invalid membership vector, number of components is %ld, but should "
+         "not be greather than number of nodes, which is %ld.", IGRAPH_EINVAL, components, no_of_nodes);
```
ntamas, line 1016: Typo here (`greather`). Also, a `the` is missing from before `number of nodes`

src/community/misc.c
```diff
@@ -120,12 +131,24 @@ int igraph_community_to_membership(const igraph_matrix_t *merges,
...
 
+    IGRAPH_VECTOR_INIT_FINALLY(&already_merged, steps + no_of_nodes);
     IGRAPH_VECTOR_INIT_FINALLY(&tmp, steps);
 
     for (i = steps - 1; i >= 0; i--) {
         long int c1 = (long int) MATRIX(*merges, i, 0);
         long int c2 = (long int) MATRIX(*merges, i, 1);
 
+        if (VECTOR(already_merged)[c1] == 0) {
+                VECTOR(already_merged)[c1] = 1;
```
ntamas, line 142: Indentation inconsistent with the rest of the file here

src/community/leading_eigenvector.c
```diff
@@ -1005,31 +1006,33 @@ int igraph_le_community_to_membership(const igraph_matrix_t *merges,
     igraph_vector_t fake_memb;
     long int components, i;
 
-    if (igraph_matrix_nrow(merges) < steps) {
-        IGRAPH_ERROR("`steps' to big or `merges' matrix too short", IGRAPH_EINVAL);
+    if (no_of_nodes > 0) {
+            components = (long int) igraph_vector_max(membership) + 1;
```
ntamas, line 1010: Indentation inconsistent with the rest of the file here

src/community/misc.c
```diff
@@ -120,12 +131,24 @@ int igraph_community_to_membership(const igraph_matrix_t *merges,
...
+        } else {
+            IGRAPH_ERRORF("Merges matrix contains multiple merges of cluster %ld.", IGRAPH_EINVAL, c1);
+        }
+        if (VECTOR(already_merged)[c2] == 0) {
+                VECTOR(already_merged)[c2] = 1;
```
ntamas, line 147: Indentation inconsistent with the rest of the file here

src/community/misc.c
```diff
@@ -120,12 +131,24 @@ int igraph_community_to_membership(const igraph_matrix_t *merges,
         igraph_vector_null(csize);
     }
 
+    IGRAPH_VECTOR_INIT_FINALLY(&already_merged, steps + no_of_nodes);
```
ntamas, line 134: Since you only ever going to store `0` or `1` in this vector, an `igraph_vector_bool_t` would suffice.

ntamas: This would save 3 bytes per entry, and in the future it might save even more if we rewrite `igraph_vector_bool_t` as a bit vector.

tests/unit/test_utilities.inc
```diff
@@ -296,6 +296,18 @@ void print_matrix_complex_first_row_positive(const igraph_matrix_complex_t *matr
     igraph_matrix_complex_destroy(&copy);
 }
 
+void matrix_init_int(igraph_matrix_t *mat, int ncol, int nrow, int* elem) {
```
ntamas, line 299: Please specify in the function name whether the elements are in row-major or column-major order. (Something like `matrix_init_int_rowwise`).

ntamas (changes requested):

ntamas:
Thanks a lot! I've requested some changes here and there, hopefully they are easy to fix.

src/community/leading_eigenvector.c
```diff
@@ -1005,31 +1006,33 @@ int igraph_le_community_to_membership(const igraph_matrix_t *merges,
...
         if (VECTOR(*membership)[i] < 0) {
-            IGRAPH_ERROR("Invalid membership vector, negative id", IGRAPH_EINVAL);
+            IGRAPH_ERRORF("Invalid membership vector, negative id found: %g.", IGRAPH_EINVAL, VECTOR(*membership)[i]);
         }
         VECTOR(fake_memb)[ (long int) VECTOR(*membership)[i] ] += 1;
     }
     for (i = 0; i < components; i++) {
         if (VECTOR(fake_memb)[i] == 0) {
-            IGRAPH_ERROR("Invalid membership vector, empty cluster", IGRAPH_EINVAL);
+            IGRAPH_ERRORF("Invalid membership vector, empty cluster found, with index %ld.", IGRAPH_EINVAL, i);
```
szhorvat, line 1035: One concern I have is that some of the igraph interfaces use 0-based indexing (Python, C) and others use 1-based indexing (R, Mathematica). This is useful—they match the host language.

But if we report a zero-based index in an error message, it would be confusing to R users.

This didn't really come up before as we didn't report specific values in error messages. Of course, being specific is generally good ...

@ntamas @vtraag Any suggestions about how to proceed here?

 1. Not report indices at all (not ideal—dimishes usefuless)
 2. Spell out that this is a zero-based index (pretty ugly, but at least useful ... )

tests/unit/test_utilities.inc
szhorvat: (Internally, igraph uses column-major storage, but this function interprets `elem` in row-major order, which is potentially confusing.)

src/community/misc.c
szhorvat: @ntamas `igraph_bool_t` is an `int`, so typically 4 bytes, because of R.

But yes, we might rewrite it in the future to be a `char`.

src/community/leading_eigenvector.c
ntamas: I would go for 1 for sake of simplicity; we were quite okay so far without spelling these out in error messages.

src/community/leading_eigenvector.c
GroteGnoom: Every call to le_community_to_membership will also call community_to_membership. This call is done before anything else is done with `steps` or `merges`. So community_to_membership will always do the check, and le_community_to_membership will fail properly without this test. The same thing would simply be tested twice with these lines added.

src/community/leading_eigenvector.c
vtraag: I'm a bit torn between the two. 

It might make debugging some errors easy if we immediately get an error reported that reflects the index. However, it might indeed confuse some users, even if we explicitly state that the index is zero-based. Still, it might also save some users some headache if they have an idea where a particular problem may occur. All considering, I think that I would have a (slight) preference of option (2) over option (1).

So @szhorvat, you can break the tie in the decision :)

src/community/leading_eigenvector.c
ntamas: Okay, fair enough, I misunderstood this part then!

ntamas (approved):
All my concerns have been addressed, I'm okay with merging this.

src/community/leading_eigenvector.c
GroteGnoom: That would indeed be horribly confusing. I'm not sure how much we want to limit error lengths.
We could do `"...index %ld (for 0-based indexing) / %ld (for 1-based indexing)", i, i+1` .
Which can still be confusing for users who are not familiar with this, but at least they can look it up.

vtraag: We actually also already have an ugly `USING_R` in place, so we could also add a conversion function `igraph_index(long idx)` that either simply returns `idx` or `idx + 1` is `USING_R` is defined. Something similar could potentially be done for Mathematica. The downside of this is that is renders unclear for a particular error whether it was thrown from an R / Mathematica session or from a C / Python session, so it may actually only complicate things.

Perhaps it would be simplest to only report the zero-based index, and explicitly mention that.

Issue #1301, opened 2020-01-27
Internal functions in public API


vtraag:
At the moment there are a number of internal functions exposed in the public API:

- [x] `igraph_flow.h`
https://github.com/igraph/igraph/blob/41f34a9f01f34965cc918f3ec07a728c5a32edf3/include/igraph_flow.h#L125
https://github.com/igraph/igraph/blob/41f34a9f01f34965cc918f3ec07a728c5a32edf3/include/igraph_flow.h#L132

- [x] `igraph_sparsemat.h`
https://github.com/igraph/igraph/blob/41f34a9f01f34965cc918f3ec07a728c5a32edf3/include/igraph_sparsemat.h#L263

    **[Decision](https://github.com/igraph/igraph/issues/1301#issuecomment-726251473):** Rename and expose in public API.

- [x] `igraph_spmatrix.h`
https://github.com/igraph/igraph/blob/41f34a9f01f34965cc918f3ec07a728c5a32edf3/include/igraph_spmatrix.h#L92-L95

- [x] `igraph_structural.h`
https://github.com/igraph/igraph/blob/41f34a9f01f34965cc918f3ec07a728c5a32edf3/include/igraph_structural.h#L144-L147

- [x] `igraph_visitor.h`
https://github.com/igraph/igraph/blob/41f34a9f01f34965cc918f3ec07a728c5a32edf3/include/igraph_visitor.h#L88

    **Note:** `igraph_i_bfs()` is currently used by the Python interface.

There are also a number of occurrences of `igraph_i_*` declarations in `igraph_attributes.h`, but some of these do seem to be intended for public use. These should probably be cleaned up.

szhorvat:
This we leave for 0.9 or 0.8.x, right? I think it's not critical and we're out of time.

vtraag:
Yes, definitely!

iosonofabio:
Since we just started a discussion on internal types and exposure, this is related.

I'd be for making everything private unless you guys tell me not to. Does anyone use any of the above functions?

vtraag:
I would agree to make all `igraph_i` functions private, in principle.

ntamas:
I can comment on a few of these:

* The feedback arc set related stuff is used by the Sugiyama layout, which is in a different compilation unit. If we move them from `igraph_structural.h`, we need to create a private header that can be included by the Sugiyama layout (which is totally okay for me).

[...]

PR #1648, opened 2021-02-04
private functions used in tests are now exported from the shared library

szhorvat:
Private functions used in tests are now exported from the shared library (i.e. you can see them in the `.dylib`/`.so` using the `nm` command and programs linking to the library can use them), but they _do not_ appear in public headers and users have no way of discovering them (other than looking at internals and adding the function prototypes to their programs explicitly).

All of these have `IGRAPH_PRIVATE_EXPORT` instead of `IGRAPH_EXPORT` so it is easy to disable exporting them if we want to do that.

Exceptionally, `igraph_i_bfs` gets `IGRAPH_PRIVATE_EXPORT` in a public header. This is the only use of `IGRAPH_PRIVATE_EXPORT` in a public header. This function is in a public header because the Python interface uses it (see #1301). I assume Debian wants its version of the Python interface to link to the C/igraph shared library. Thus for `igraph_i_bfs` maybe a plain `IGRAPH_EXPORT` makes more sense.

### DIFF
--- a/include/igraph_decls.h
+++ b/include/igraph_decls.h
@@ -13,3 +13,7 @@
 
 /* Include the definition of macros controlling symbol visibility */
 #include "igraph_export.h"
+
+/* Used instead of IGRAPH_EXPORT with functions that need to be tested,
+ * but are not part of the public API. */
+#define IGRAPH_PRIVATE_EXPORT IGRAPH_EXPORT
--- a/include/igraph_flow.h
+++ b/include/igraph_flow.h
@@ -122,18 +122,7 @@ IGRAPH_EXPORT int igraph_residual_graph(const igraph_t *graph,
                                         igraph_t *residual,
                                         igraph_vector_t *residual_capacity,
                                         const igraph_vector_t *flow);
-int igraph_i_residual_graph(const igraph_t *graph,
-                            const igraph_vector_t *capacity,
-                            igraph_t *residual,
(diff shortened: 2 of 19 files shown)


szhorvat:
As part of addressing #1301, I just removed some functions from public headers.

I was forced to comment out the `_i_spmatrix` functions because if I just make them `static`, the compiler will throw a warning (which is taken as an error with current settings) about the function not being used.

ntamas (approved):

ntamas:
`igraph_i_bfs()` is special and I'm not sure what to do about it. The Python interface exposes `igraph_i_bfs()` as _the_ BFS function, and does not use `igraph_bfs()` at all. At the same time, the Python interface also reimplements BFS in a `BFSIter` class because Python users naturally expect the BFS traversal to behave like an iterator, and there's no way we can simulate that with the current C implementation.

So, considering that `igraph_i_bfs()` is actually the function that is exposed in Python as _the_ BFS function, it might be worth renaming it to something like `igraph_bfs_simple()`, because this is essentially what it is -- it's a simpler interface to BFS for use-cases when one does not need the "full" BFS solution with `order`, `rank`, `pred`, `succ`, `father`, `dist` and `callback` as they are in `igraph_bfs()`.

ntamas:
I'm going to merge this and we can continue the discussion about `igraph_i_bfs()` on Slack.

PR #1649, opened 2021-02-05
Add modularity_matrix unit tests

GroteGnoom:
Part of #1592

I wasn't too sure if I should leave the NaN's and infs in (https://igraph.discourse.group/t/the-unit-tests-thread/597/17), but it could be just fine.


szhorvat:
We recently made the choice to return NaN for the modularity when the number of edges is zero, and document this.

I would suggest catching this case (`sw == 0`) and returning a matrix filled with all NaNs. This is a bit more consistent than letting NaNs and Infs happen arbitrarily.

Also, it should be documented that the modularity matrix is not well-defined when there are zero edges. This is what's important.

What to return for the meaningless case is not that critical (for as long as it's clear that it's meaningless, as it is now).

@vtraag any comments?

szhorvat:
@vtraag Since `igraph_modularity` was updated for the directed case, do you want to update this function as well? It's a trivial, very low risk change and would make the system a bit more consistent. Also, maybe @GroteGnoom would enjoy adding some features instead of just tests.

If we _don't_ do it, then it is important to document that "edge directions are ignored"..

vtraag:
> I would suggest catching this case (`sw == 0`) and returning a matrix filled with all NaNs. This is a bit more consistent than letting NaNs and Infs happen arbitrarily.

Yes, I agree, this makes sense.

I think it would make sense to update `igraph_modularity_matrix` to also work for the directed case. The function itself is never used anywhere in `igraph` itself it seems, to there are no dependencies to think about. I don't have time to do so now however. Perhaps @GroteGnoom, you would be interested in doing this?

szhorvat:
IMO it is also fine to skip testing the zero-edge case and state in the documentation that the result for this case is undefined.

GroteGnoom:
> IMO it is also fine to skip testing the zero-edge case and state in the documentation that the result for this case is undefined.

Allright, I will go with this option :slightly_smiling_face:

> I think it would make sense to update igraph_modularity_matrix to also work for the directed case. The function itself is never used anywhere in igraph itself it seems, to there are no dependencies to think about. I don't have time to do so now however. Perhaps @GroteGnoom, you would be interested in doing this?

Sure!
If I understand correctly, the adjacency matrix does currently take the directions in to account, so what needs to be changed is to use in-degrees and out-degrees for the next part (the edge-probability matrix).

vtraag:
> If I understand correctly, the adjacency matrix does currently take the directions in to account, so what needs to be changed is to use in-degrees and out-degrees for the next part (the edge-probability matrix).

Yes, the `igraph_get_adjacency` call already returns a directed matrix (if the graph is directed). You then indeed need to get in- and out-degree (or strenghts) separately and multiply them as `out[i] * in[j] / (2 *m)`, similar to how it is done in `igraph_modularity`. You can make it a bit more efficient by scaling the degree/strength vectors by `2 * m` before the double loop, requiring only `2 n` operations instead of `n^2`, similar to how it is done in `igraph_modularity`.

For a unit test, it would be nice to check if `igraph_modularity` and `igraph_matrix_modularity` matrix give the same outcome for a given partition.

szhorvat:
@vtraag I would call the current version buggy actually ... It uses a _directed_ adjacency matrix with undirected (i.e. total) degrees.

vtraag:
> @vtraag I would call the current version buggy actually ... It uses a _directed_ adjacency matrix with undirected (i.e. total) degrees.

Yes, quite right. I think it was never intended to be used for directed graphs, but this is neither checked nor documented. Improving this is then all the more important.

szhorvat:
OK, a bit of care is needed to make sure we get this right.

`modularity` has a `directed` argument that can be used to ignore edge directions. This is useful as most community detection methods ignore edge directions. `modularity_matrix` should also have such an argument.

Ignoring edge directions in directed graph will lead to multigraphs when we have reciprocal edges. This case must not be forgotten. Thus, if `A` is the adj. mat. then:

 - We retrieve `A` using `get_adjacency`.
 - If the graph is undirected set `directed=false`.
 - If the graph itself is undirected (not the same as `directed == false`!), multiply `A_ii` by 2.
 - If:
     * `directed == false`, compute `(A_ij + A_ji) / 2 - gamma * k_i * k_j / (2*m)`. This works for undirected adjacency matrices (thanks to multiplying the diagonal) and effectively ignores edge directions for the directed adjacency matrices.
     * `directed == true`, compute `A_ij - gamma * k_out_i * k_in_j / (m)`

vtraag:
> `directed == true`, compute `A_ij - gamma * k_out_i * k_in_j / (m)`. Notice that there is no factor `2` in front of `m` (@vtraag you got this wrong above)

Yes, correct, there should be no `2` for the directed case.

GroteGnoom:
I notice I haven't been careful enough in checking the output for the weighted graph. I think the adjacency matrix should take the weights into account, but it doesn't.

Is there a function that generates the weighted adjacency matrix?

szhorvat:
Wow, this function is in bad repair ... 

> Is there a function that generates the weighted adjacency matrix?

I can't find one either, except for sparse matrices (`igraph_weighted_sparsemat`, but that takes the weight from an attribute).

I checked the R source code and it is implemented directly in R.

I suggest writing a helper function that does whatever `modularity_matrix` needs, but not more. Since we are so close to the release of version 0.9, I would not try to polish it and make it public. That is something we can do later.

`igraph_get_adjacency` is a good model for how to write such a function: iterate through the edges and do `MATRIX(mat, IGRAPH_FROM(graph, eid), IGRAPH_TO(graph, eid)) += weights[eid]` where `eid` is the edge index.

vtraag:
> In the undirected case, also add it to the transpose matrix element.

Is this also consistent with the output from the unweighted `igraph_get_adjacency`?

> I suggest writing a helper function that does whatever modularity_matrix needs, but not more

Most functions with weights would just have a `weights` argument which could be `NULL` to note take weights into account. Once 0.9 is released we fix the API, so perhaps is makes sense to simply add this `weights` argument to the existing `igraph_get_adjacency` function instead of creating a new one?

> Wow, this function is in bad repair ...

Clearly. I doubt that anyone ever actually uses this function though.

szhorvat:
> Is this also consistent with the output from the unweighted `igraph_get_adjacency`?

No, it's not. `igraph_get_adjacency` specifically avoids doubling the diagonal elements.

But for now we just need to fix this function, and later (post-0.9) we can think about how to expand the API.

GroteGnoom:
I have a new version.

I also added the option to ignore direction in the adjacency matrix.

I think igraph_modularity is not implemented correctly, because it scales k_in and k_out both, which would result in a division by m^2 instead of m.

Tests are now failing because of tiny floating point offsets.

src/community/modularity.c
```diff
@@ -216,6 +216,46 @@ int igraph_modularity(const igraph_t *graph,
...
+           const igraph_t *graph, igraph_matrix_t *res,
+           const igraph_vector_t *weights, igraph_bool_t directed) {
+    /* Specifically used to handle weights and/or ignore direction */
+    igraph_eit_t edgeit;
+    long int no_of_nodes = igraph_vcount(graph);
+    igraph_integer_t from, to;
+
+    IGRAPH_CHECK(igraph_matrix_resize(res, no_of_nodes, no_of_nodes));
+    igraph_matrix_null(res);
+    IGRAPH_CHECK(igraph_eit_create(graph, igraph_ess_all(0), &edgeit));
```
szhorvat, line 229: Let's spell out the argument of `ess_all` here, `IGRAPH_EDGEORDER_ID`.

src/community/modularity.c
```diff
@@ -216,6 +216,46 @@ int igraph_modularity(const igraph_t *graph,
...
+    IGRAPH_FINALLY(igraph_eit_destroy, &edgeit);
+
+    if (weights) {
+        while (!IGRAPH_EIT_END(edgeit)) {
+            igraph_integer_t edge = IGRAPH_EIT_GET(edgeit);
+            igraph_edge(graph, edge, &from, &to);
```
szhorvat, line 235: Let's use the newer `IGRAPH_FROM` and `IGRAPH_TO`. These are macros, which means that they will always be inlined. `igraph_edge` will cause a function call (unless link-time optimization is used and the compiler can inline it) which is potentially slow.

While performance is likely not critical here, I'd prefer `IGRAPH_FROM` and `IGRAPH_TO` in new code.

src/community/modularity.c
```diff
@@ -216,6 +216,46 @@ int igraph_modularity(const igraph_t *graph,
...
+            MATRIX(*res, from, to) += VECTOR(*weights)[edge];
+            if (!directed) {
+                MATRIX(*res, to, from) += VECTOR(*weights)[edge];
+            }
+        IGRAPH_EIT_NEXT(edgeit);
```
szhorvat, line 240: Indentation is inconsistent here.

With edge/vertex iterators, one can also use `for`:

```
for (; !IGRAPH_EITEND(edgeit); IGRAPH_EIT_NEXT(edgeit))
```

This is a stylistic matter and it is up to you. I mentioned it in case you used this indentation on purpose, to highlight the last line. In that case, `for` would be better.

src/community/modularity.c
```diff
@@ -262,28 +307,61 @@ int igraph_modularity_matrix(const igraph_t *graph,
...
-                                      /*eids=*/ 0));
+    IGRAPH_CHECK(igraph_i_modularity_matrix_get_adjacency(graph, modmat, weights, directed));
 
-    for (i = 0; i < no_of_nodes; i++) {
-        MATRIX(*modmat, i, i) *= 2;
-    }
-    for (i = 0; i < no_of_nodes; i++) {
-        for (j = 0; j < no_of_nodes; j++) {
-            MATRIX(*modmat, i, j) -= resolution * VECTOR(deg)[i] * VECTOR(deg)[j] / 2.0 / sw;
+   if (directed) {
```
szhorvat, line 315: Indentation nitpick.

src/community/modularity.c
```diff
@@ -228,8 +268,9 @@ int igraph_modularity(const igraph_t *graph,
  * </para><para>
```
szhorvat, line 268: Can you please add the formula of the directed definition in the docs?

tests/unit/igraph_modularity_matrix.c
```diff
@@ -0,0 +1,93 @@
...
+    printf("Triangle and point with self-loop, directed:\n");
+    igraph_small(&g, 4, /*directed*/1, 0,1, 0,2, 1,2, 3,3, -1);
+    igraph_matrix_init(&modmat, 0, 0);
+    test_print_destroy(&g, NULL, 1.0, &modmat, 1);
+
+    printf("Triangle with weights 0, 1, 2:\n");
+    igraph_small(&g, 3, /*directed*/0, 0,1, 0,2, 1,2, -1);
+    igraph_vector_init_int(&weights, 3, 0, 1, 2);
+    igraph_matrix_init(&modmat, 0, 0);
+    test_print_destroy(&g, &weights, 1.0, &modmat, 0);
```
szhorvat, line 71: This is the location of the test "failure" (which is of course not a false positive).

When I encounter such a situation in Mathematica, I usually use its `Chop` function. That function simply replaced numbers smaller than some `delta` (default: `1e-10`) by an exact `0`. Perhaps we could do the same here: add a helper function in the test program that scans the matrix and replaces everything smaller than some threshold with 0.0. I think it would be okay to do this directly in `test_print_destroy` and thus do it for all cases.

Just be sure to do `if (fabs(el) < delta) el = 0;` and _not_ `if (fabs(el) > delta) leave_el_alone(); else el = 0` because the latter will mess up NaNs.

szhorvat: Easy way to iterate through a matrix `mat` is to use `VECTOR(mat.data)[i]`.

src/community/modularity.c
```diff
@@ -262,28 +307,61 @@ int igraph_modularity_matrix(const igraph_t *graph,
...
+        if (!weights) {
+            IGRAPH_CHECK(igraph_degree(graph, &deg, igraph_vss_all(), IGRAPH_ALL,
+                                       IGRAPH_LOOPS));
+        } else {
+            IGRAPH_CHECK(igraph_strength(graph, &deg, igraph_vss_all(), IGRAPH_ALL,
+                                         IGRAPH_LOOPS, weights));
+        }
+
+        /* Scaling one degree factor so every element gets scaled. */
+        igraph_vector_copy(&deg_unscaled, &deg);
```
szhorvat, line 352: If we scale with `sqrt(2*sw)` then we can avoid making a copy of the degree vector.

GroteGnoom: I did try a square root before, but I am allowing negative weights, and then things get messy :)

szhorvat:
> I think igraph_modularity is not implemented correctly, because it scales k_in and k_out both, which would result in a division by m^2 instead of m.

I believe that the modularity calculation is correct because there is an extra 1/m (or 1/2m) factor in front of the definition. Thus effectively it is `sum_ij (A_ij / m - k_in/m * k_out/m) * delta_c(i)_c(j)`

szhorvat:
It looks correct to me. It would be nice to add a test to compare with `igraph_modularity` by explicitly calculation the modularity from the modularity matrix and an arbitrary partitioning.

Let me when you're done with this and I'll finish the remaining nitpicks and merge.

Thanks again for all your work on the tests and for carefully checking all cases!

GroteGnoom:
I think I fixed everything.

PR #1652, opened 2021-02-06
adjlist/inclist loop handling update

szhorvat:
Creating a PR to make review/commenting easier.


src/graph/adjlist.c
```diff
@@ -362,6 +391,12 @@ int igraph_adjlist_remove_duplicate(const igraph_t *graph,
                                     igraph_adjlist_t *al) {
     long int i, j, l, n, p;
     igraph_vector_int_t *v;
+    
+    IGRAPH_WARNING(
+        "igraph_adjlist_remove_duplicate() is deprecated; use the constructor "
+        "arguments of igraph_adjlist_init() to specify whether you want loop "
+        "edges to appear once or twice in the adjacency list"
```
szhorvat, line 398: Add fullstop.

src/graph/adjlist.c
```diff
@@ -456,32 +491,96 @@ int igraph_adjlist_replace_edge(igraph_adjlist_t* al, igraph_integer_t from, igr
...
+    return IGRAPH_SUCCESS;
+}
+
+int igraph_inclist_remove_duplicate(const igraph_t *graph, igraph_inclist_t *il) {
+    long int i, n;
+
+    IGRAPH_WARNING(
+        "igraph_inclist_remove_duplicate() is deprecated; use the constructor "
+        "arguments of igraph_adjlist_init() to specify whether you want loop "
+        "edges to appear once or twice in the adjacency list"
```
szhorvat, line 570: Fullstop.

include/igraph_adjlist.h
```diff
@@ -81,8 +80,10 @@ typedef struct igraph_inclist_t {
 
 IGRAPH_EXPORT int igraph_inclist_init(const igraph_t *graph,
                                       igraph_inclist_t *il,
-                                      igraph_neimode_t mode);
+                                      igraph_neimode_t mode,
+                                      igraph_loops_t loops);
 IGRAPH_EXPORT int igraph_inclist_init_empty(igraph_inclist_t *il, igraph_integer_t n);
+IGRAPH_EXPORT igraph_integer_t igraph_inclist_size(const igraph_inclist_t *al);
```
szhorvat, line 86: Add to CHANGELOG

include/igraph_adjlist.h
```diff
@@ -108,19 +109,22 @@ IGRAPH_EXPORT int igraph_inclist_fprint(const igraph_inclist_t *il, FILE *outfil
...
                                            igraph_lazy_adjlist_t *al,
                                            igraph_neimode_t mode,
-                                           igraph_lazy_adlist_simplify_t simplify);
+                                           igraph_loops_t loops,
+                                           igraph_multiple_t multiple);
 IGRAPH_EXPORT void igraph_lazy_adjlist_destroy(igraph_lazy_adjlist_t *al);
 IGRAPH_EXPORT void igraph_lazy_adjlist_clear(igraph_lazy_adjlist_t *al);
-/* igraph_vector_t *igraph_lazy_adjlist_get(igraph_lazy_adjlist_t *al, */
-/*                     igraph_integer_t no); */
+IGRAPH_EXPORT igraph_integer_t igraph_lazy_adjlist_size(const igraph_lazy_adjlist_t *al);
```
szhorvat, line 126: Add to CHANGELOG

include/igraph_adjlist.h
```diff
@@ -139,22 +143,25 @@ IGRAPH_EXPORT void igraph_lazy_adjlist_clear(igraph_lazy_adjlist_t *al);
...
 } igraph_lazy_inclist_t;
 
 IGRAPH_EXPORT int igraph_lazy_inclist_init(const igraph_t *graph,
                                            igraph_lazy_inclist_t *il,
-                                           igraph_neimode_t mode);
+                                           igraph_neimode_t mode,
+                                           igraph_loops_t loops);
 IGRAPH_EXPORT void igraph_lazy_inclist_destroy(igraph_lazy_inclist_t *il);
 IGRAPH_EXPORT void igraph_lazy_inclist_clear(igraph_lazy_inclist_t *il);
+IGRAPH_EXPORT igraph_integer_t igraph_lazy_inclist_size(const igraph_lazy_inclist_t *il);
```
szhorvat, line 164: Add to CHANGELOG

szhorvat:
@ntamas One concern I have is performance. It seems to me that it would be very easy to generate loops only once instead of twice directly in the source at `igraph_neighbors` or `igraph_incident`. Am I mistaken? What we do now instead is to first generate double loops, them remove them.  For example, `igraph_i_remove_loops_from_incidence_vector_in_place` adds already seen loops to a plain vector and repeatedly checks if newly found loops are already in this vector. Vector search is expensive (linear). The functions we are dealing with are fundamental, used all over the library, so it makes sense to avoid unnecessary overhead.

Why do we not then tackle the problem at the source, i.e. `igraph_neighbors` / `igraph_incident`?

szhorvat:
There are two sections of these functions: `mode & IGRAPH_OUT` and `mode & IGRAPH_IN`. For loops=none, we skip adding loops in both sections. For loops=once, we skip adding loops in one section. For loops=twice, we don't skip them at all. Isn't this all that's needed, and isn't this more performant?

I keep looking at the `NO`/`ONCE`/`TWICE` choices made in some functions. For some it would actually make sense to switch to "no loops" or "once". But I feel uncomfortable about the performance implications. It would be so much simpler if we never had to think about that.

Of course, with multi-edge removal there will be some slowdown no matter what, but that's a separate case. Even there, the addition of these arguments is immensely useful. There are functions where now we added a check to see if the graph is simple and abort if not. Sometimes the better solution is to just ignore loops and multi-edges. This is very easy to do now.

src/isomorphism/lad.c
```diff
@@ -110,7 +110,7 @@ static int igraph_i_lad_createGraph(const igraph_t *igraph, Tgraph* graph) {
 
     graph->nbVertices = no_of_nodes;
 
-    IGRAPH_CHECK(igraph_adjlist_init(igraph, &graph->succ, IGRAPH_OUT));
+    IGRAPH_CHECK(igraph_adjlist_init(igraph, &graph->succ, IGRAPH_OUT, IGRAPH_LOOPS_ONCE, IGRAPH_MULTIPLE));
```
szhorvat, line 113: @ntamas There's a bit of a problem here.

Previously, when this was effectively LOOPS_TWICE, it caused the non-simple graph error to trigger a few lines below. Now that error does not trigger on graphs that have self-loops but no multi-edges.

However, now, a few lines above, the computation of `nbSucc` using `igraph_degree` is incorrect because that line effectively double-counts loops.  I fixed that and tried many examples to see if LAD can handle self-loops. I did not so far encounter any incorrect result. That's great, it looks like we just fixed a problem and gain support for loopy graphs with LAD! But: with this are taking a risk. I only verified that it works empirically ... 

So, what do we do about this?

 1. We detect loopy graphs and report an error
 2. We fix the `igraph_degree` line and let LAD work with loopy graphs

I am cautiously leaning towards solution 2.

szhorvat:
@ntamas Other than the LAD thing, the _behaviour_ seems correct to me.

src/isomorphism/lad.c
szhorvat: I just committed proposal 2 to show what I mean.

src/graph/adjlist.c
ntamas: Fixed  in 1b08abb35

src/graph/adjlist.c
ntamas: Fixed in 1b08abb35

src/isomorphism/lad.c
ntamas: Hmmm, okay, I should have probably written `IGRAPH_LOOPS_TWICE`; I was confused by `IGRAPH_OUT`, which works like `IGRAPH_LOOPS_ONCE` for directed graphs but is ignored (and treated as `IGRAPH_ALL`) for undirected graphs. I don't know off the top of my head whether LAD supports directed graphs; if it doesn't, then I should have written `IGRAPH_LOOPS_TWICE`, shouldn't I?

szhorvat: LAD supports directed graphs, but it is not optimized for them. This is why the website say "for non directed graphs": http://perso.citi-lab.fr/csolnon/LAD.html

With the commit I pushed, it seems to handle self-loops fine. I just did a bit of random fuzzing, and it does not seem to fail to detect isomorphism, nor does it return a wrong mapping. Why I did not (could not) test with the fuzzer was that it wouldn't give false negatives (i.e. say "not isomorphic" when in fact it is). But I also did quite a bit manual testing, and it seems fine.

ntamas:
> Why do we not then tackle the problem at the source, i.e. `igraph_neighbors` / `igraph_incident`?

Because that would be a significant API breakage less than 10 days before the scheduled release of 0.9.0, and I cannot judge what the implications would be. `igraph_neighbors()` and `igraph_incident()` are used pretty much everywhere from the C core to the higher level interfaces, so if I add options for handling loop/multiple edges, it would have far-reaching implications for the higher level interfaces as well.

We can do this later, for 0.10.0, although to be honest I'm not so sure about this; it would make these functions _much_ harder to use from the C interface because then you would have to specify two extra arguments with every call, _even if your particular graph does not have loop or multiple edges and will never have_. In C++ it wouldn't be a problem because we could just pick sensible defaults for these arguments, but it is not the case here.

An alternative would be to expand the `igraph_neimode_t` enum with additional bits to indicate what to deal with loop / multiple edges, so one could use something like `IGRAPH_OUT | IGRAPH_NO_MULTIPLE | IGRAPH_LOOPS_TWICE` if he wants to be super specific, but could also simply use `IGRAPH_OUT` to use reasonable defaults (we just need to specify the bit fields in a way that zero bits represent the defaults).

ntamas:
> For example, `igraph_i_remove_loops_from_incidence_vector_in_place` adds already seen loops to a plain vector and repeatedly checks if newly found loops are already in this vector. Vector search is expensive (linear).

Yes, it's linear, but in 99.99% of the cases the "already seen loops" vector will be empty, and in 99% of the remaining 0.01% it will have length 1, so I guess its performance impact is negligible. I don't expect users to frequently use graphs where it is typical for a node to have hundreds of loop edges.

Anyway, in the vein of "measure, don't guess", this comment indicates that it would indeed be useful to have a comprehensive and _representative_ set of benchmarks for igraph that touch most parts of the library so we could test whether a change like this has an impact or not. Right now we can only guess, and my opinion is no better than yours :)

include/igraph_adjlist.h
ntamas: > If using `IGRAPH_LOOPS_TWICE` looks too ugly in the directed case, then let us write `IGRAPH_LOOPS` there in acknowledgement that it doesn't matter.

I'll do this and also swap the value of `IGRAPH_LOOPS_ONCE` and `IGRAPH_LOOPS_TWICE` as suggested.

szhorvat:
We do not actually need to change `igraph_neighbors` and `igraph_incident`.

We can have an `igraph_i_neighbors(graph, neis, node, mode, loops, multiple)` just for the private use of `igraph_adjlist_init`.

My point was not to change `igraph_neighbors` or its signature, but to simply not generate double-loops instead of first generating them, then removing them.

However, if you are still concerned about breaking stuff with this change, we can leave it for after 0.9. The behaviour would not change, only (potentially) the performance. It would also be a much simpler way to implement the same behaviour.

szhorvat:
> Anyway, in the vein of "measure, don't guess",

Performance is not the only argument for solving the problem at the source. Another one is simplicity.

vtraag:
Just my two cents: I think it would be useful to have an unambiguous definition of `igraph_neighbours` and `igraph_incident` and for that reason I suggested to simply include each edge always exactly once.

I am now actually wondering whether we shouldn't simply always stick to this. We now introduce arguments to control what is being returned (multiple edges, number of loops, etc...), but it might now be more useful to simply always use all edges, also given the problems that emerge with other functions (e.g. degree/strength, weights for multigraphs).

If a user supplies a multigraph to a function that doesn't accept it, (s)he should simply first simplify the graph. There is a reason for counting loops once or twice, but we can debate whether we should have an option to ignore loops. Again, if a function does not work correctly with loops it is easier to state this clearly and require a user to first simplify the graph (or at least remove the self-loops).

szhorvat:
> to simply include each edge always exactly once

I am not very keen on this. Before making such major changes to fundamental functions, there should be a very good justification.

In my experience, the most common use for finding neighbours is to implement algorithms. Usually is it necessary to have a degree which is consistent with the way `neighbors` was computed. With the standard definition of degree (which takes loops twice), that requires loop duplication in the adjlist/inclist as well. This is just one example that came to mind immediately.

vtraag:
> I am not very keen on this. Before making such major changes to fundamental functions, there should be a very good justification.

I agree. But I think it may be worth considering letting the interface be as simple as possible, and not see a proliferation of multiple/loop arguments everywhere. Whether there are always one or two self-loops is I think another question, as long as everything is simply always consistent with each other.

Anyway, this is probably something that ties in with other discussions about fundamental (graph) structure. So this probably goes beyond this particular PR, and we can discuss it in greater detail at some later time.

iosonofabio:
I'm late to the party, but I can see it's a good case for a synchronous discussion via teams, perhaps after 0.9.

ntamas:
Merged this as is and then we can discuss the remaining issues about multiple / loop edge handling on Teams, after we've discussed the int type transition of course :)

Issue #1515, opened 2020-10-30
get_all_shortest_paths should return edge path


szhorvat:
`get_shortest_path(s)` return the shortest paths both in terms of vertices and edges.

`get_all_shortest_paths` returns it only in terms of vertices. It should also return it in terms of edges. I explain the reason [here](https://github.com/igraph/python-igraph/issues/337#issuecomment-719403866).

szhorvat:
I haven't looked into it, but I expect this to be easy. If so, it should be done for 0.9 as it's API-breaking.

PR #1653, opened 2021-02-07
Add edges to shortest path

guyroznb:
This PR answers Issue number #1515, returning the shortest paths both in terms of edges and in vertices.

This PR affects the functions `igraph_get_all_shortest_paths` and   `igraph_get_all_shortest_paths_dijkstra`.

thanks.


src/paths/dijkstra.c
```diff
@@ -948,27 +970,56 @@ int igraph_get_all_shortest_paths_dijkstra(const igraph_t *graph,
...
         if (path == 0)
             IGRAPH_ERROR("cannot run igraph_get_all_shortest_paths_dijkstra",
-                         IGRAPH_ENOMEM);
+                        IGRAPH_ENOMEM);
         IGRAPH_FINALLY(igraph_free, path);
         IGRAPH_CHECK(igraph_vector_init(path, 1));
-        IGRAPH_CHECK(igraph_vector_ptr_push_back(res, path));
+
+        /* clear the paths vector */
+        /* If res isn't excist then create one, in order for the algorithm to work. */
```
ntamas, line 983: Suggestion: `If res doesn't exist, then create one`

src/paths/dijkstra.c
```diff
@@ -948,27 +970,56 @@ int igraph_get_all_shortest_paths_dijkstra(const igraph_t *graph,
...
+            path = igraph_Calloc(1, igraph_vector_t);
+            if (path == 0)
+                IGRAPH_ERROR("cannot run igraph_get_all_shortest_paths_dijkstra",
+                            IGRAPH_ENOMEM);
+            /* the shortest path from the source to itself consist of the edge -1 */
+            IGRAPH_FINALLY(igraph_free, path);
+            IGRAPH_CHECK(igraph_vector_init(path, 1));
+            IGRAPH_CHECK(igraph_vector_ptr_push_back(res_e, path));
+            IGRAPH_FINALLY_CLEAN(1);  /* ownership of path passed to res */
+            VECTOR(*path)[0] = -1;
```
ntamas, line 1016: Not sure if I like this -1 in the edge ID vector; couldn't we simply keep the vector empty?

src/paths/all_shortest_paths.c
```diff
@@ -175,6 +188,10 @@ int igraph_get_all_shortest_paths(const igraph_t *graph,
     IGRAPH_CHECK(igraph_vector_ptr_push_back(&paths, vptr));
     IGRAPH_CHECK(igraph_vector_init(vptr, 1));
     VECTOR(*vptr)[0] = from;
+    vptr_e = igraph_Calloc(1, igraph_vector_t);
+    IGRAPH_CHECK(igraph_vector_ptr_push_back(&path_edge, vptr_e));
+    IGRAPH_CHECK(igraph_vector_init(vptr_e, 1));
+    VECTOR(*vptr_e)[0] = -1;
```
ntamas, line 194: Not sure if I like this -1 in the edge ID vector; couldn't we simply keep the vector empty?

src/paths/dijkstra.c
```diff
@@ -1038,25 +1112,50 @@ int igraph_get_all_shortest_paths_dijkstra(const igraph_t *graph,
...
-    IGRAPH_FINALLY_CLEAN(4);
-
+    igraph_vector_ptr_destroy_all(&parents_edge);
+    if (free_res){
+        igraph_vector_ptr_destroy_all(res);
+        IGRAPH_FINALLY_CLEAN(6);
+        igraph_free(res);
+    }
+    else{
+        IGRAPH_FINALLY_CLEAN(5);
```
ntamas, line 1158: It's just a matter of taste, but I'd rather have an unconditional `IGRAPH_FINALLY_CLEAN(5)` outside the `if` condition, and an extra `IGRAPH_FINALLY_CLEAN(1)` in the `if`.

examples/simple/igraph_get_all_shortest_paths_dijkstra.c
```diff
@@ -149,6 +157,20 @@ int main() {
         VECTOR(res)[i] = 0;
     }
 
+
```
ntamas, line 160: Add a comment here explaining that we are testing the combination of `res = 0` and `res_e != 0` here.

ntamas (changes requested):
Looks good in general, but there are a few minor things that should be fixed before it is merged.

General comments:

* Please stick to the coding style used throughout the functions; in particular: use a space before the opening `{` of an `if` or `else` block, and put the `else` on the same line as the preceding `}`.

* The old version of the code was sloppy with the `igraph_Calloc()` allocations; we should check whether the result of an `igraph_Calloc()` call is NULL and raise an `IGRAPH_ENOMEM` error in that case. You don't need to fix this, it's a reminder for myself to deal with this on my own before the PR is merged.

src/paths/dijkstra.c
```diff
@@ -697,44 +705,51 @@ int igraph_get_all_shortest_paths_dijkstra(const igraph_t *graph,
...
     }
-    if (no_of_edges > 0) {
-        igraph_real_t min = igraph_vector_min(weights);
-        if (min < 0) {
-            IGRAPH_ERROR("Weight vector must be non-negative", IGRAPH_EINVAL);
-        }
-        else if (igraph_is_nan(min)) {
-            IGRAPH_ERROR("Weight vector must not contain NaN values", IGRAPH_EINVAL);
-        }
+    if (igraph_vector_min(weights) < 0) {
```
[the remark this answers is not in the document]
szhorvat: Another important point is that `igraph_vector_min` must never be called for an empty vector. That would cause a crash.

src/paths/all_shortest_paths.c
guyroznb: well I prefer not, because the vector `vptr_e` is initialized to 1.
Is it O.K?

src/paths/dijkstra.c
guyroznb: I prefer to keep it that way because path is initialized to 1.
Is this O.k?

include/igraph_paths.h
```diff
@@ -66,6 +66,7 @@ IGRAPH_EXPORT int igraph_get_shortest_path(const igraph_t *graph,
 
 IGRAPH_EXPORT int igraph_get_all_shortest_paths(const igraph_t *graph,
                                                 igraph_vector_ptr_t *res,
+                                                igraph_vector_ptr_t *res_e,
```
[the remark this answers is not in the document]
guyroznb: changed throughout the PR

interfaces/functions.def
```diff
@@ -443,7 +443,7 @@ igraph_get_shortest_paths:
         IGNORE: RR, RC, RNamespace
 
 igraph_get_all_shortest_paths:
-        PARAMS: GRAPH graph, OUT VERTEXSETLIST res, OUT VECTOR nrgeo, \
+        PARAMS: GRAPH graph, OUT VERTEXSETLIST res, OUT VERTEXSETLIST res_e, OUT VECTOR nrgeo, \
```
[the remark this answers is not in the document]
guyroznb: changed

guyroznb:
I added the `igraph_Calloc() ` NULL check where I found it missing.
thanks @ntamas for the review

src/paths/dijkstra.c
```diff
@@ -1019,44 +1074,87 @@ int igraph_get_all_shortest_paths_dijkstra(const igraph_t *graph,
...
+        if (!free_vertices){
+            IGRAPH_CHECK(igraph_vector_ptr_resize(vertices, j));
+            /* sort the paths by the target vertices */
+            igraph_vector_ptr_sort(vertices, igraph_i_vector_tail_cmp);
+        }
+        
+        if (edges){
+            IGRAPH_CHECK(igraph_vector_ptr_resize(edges, j));
+            igraph_vector_ptr_sort(edges, igraph_i_vector_tail_cmp);
+        }
```
szhorvat, line 1144: Note to self: We have a problem here: in the preceding lines, the vertex paths are sorted based on their target vertex. The edge path list should be ordered identically to the vertex path list.

guyroznb:
Hi, help is needed here. I don't know why all my tests fail (including programs which I didn't change).

thanks

ntamas:
This seems to be a problem in the logs:

```
../src/paths/dijkstra.c:913:27: error: unused variable 'k' [-Werror,-Wunused-variable]
        long int j, node, k;
```

All warnings are turned into errors by default so the compiler will complain if some leaves unused variables around.

interfaces/functions.def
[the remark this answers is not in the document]
ntamas: They are still named `res` and `res_e`; please rename them to `vertices` or `edges` to be consistent with the headers (`igraph_paths.h`).

src/paths/all_shortest_paths.c
```diff
@@ -127,6 +131,9 @@ int igraph_get_all_shortest_paths(const igraph_t *graph,
     /* paths will store the shortest paths during the search */
     IGRAPH_CHECK(igraph_vector_ptr_init(&paths, 0));
     IGRAPH_FINALLY(igraph_i_gasp_paths_destroy, &paths);
+    /* paths will store the shortest paths during the search */
```
ntamas, line 134: Please fix the comment: `"path_edge" will store the...`. Right now the comment is copied straight from three lines above and it does not reflect what `path_edge` will be used for.

src/paths/all_shortest_paths.c
```diff
@@ -174,10 +181,20 @@ int igraph_get_all_shortest_paths(const igraph_t *graph,
     }
 
     /* from -> from */
-    vptr = igraph_Calloc(1, igraph_vector_t); /* TODO: dirty */
+    vptr = igraph_Calloc(1, igraph_vector_t);
+    if (vptr == 0) {
+    IGRAPH_ERROR("cannot run igraph_get_all_shortest_paths", IGRAPH_ENOMEM);
```
ntamas, line 186: Please fix the indentation here; on my screen it is indented at the same level where the `if` is.

src/paths/all_shortest_paths.c
```diff
@@ -174,10 +181,20 @@ int igraph_get_all_shortest_paths(const igraph_t *graph,
...
+    }
     IGRAPH_CHECK(igraph_vector_ptr_push_back(&paths, vptr));
     IGRAPH_CHECK(igraph_vector_init(vptr, 1));
     VECTOR(*vptr)[0] = from;
+    vptr_e = igraph_Calloc(1, igraph_vector_t);
+    if (vptr_e == 0) {
+    IGRAPH_ERROR("cannot run igraph_get_all_shortest_paths", IGRAPH_ENOMEM);
```
ntamas, line 193: Please fix the indentation here; on my screen it is indented at the same level where the `if` is.

src/paths/all_shortest_paths.c
ntamas: But why can't we simply initialize `vptr_e` with zero elements? (i.e. `igraph_vector_init(vptr_e, 0)`.

src/paths/all_shortest_paths.c
```diff
@@ -247,11 +264,27 @@ int igraph_get_all_shortest_paths(const igraph_t *graph,
...
+                    IGRAPH_ERROR("cannot run igraph_get_all_shortest_paths", IGRAPH_ENOMEM);
+                }
                 IGRAPH_CHECK(igraph_vector_ptr_push_back(&paths, vptr));
                 IGRAPH_CHECK(igraph_vector_copy(vptr, VECTOR(paths)[fatherptr - 1]));
                 IGRAPH_CHECK(igraph_vector_reserve(vptr, actdist + 2));
                 IGRAPH_CHECK(igraph_vector_push_back(vptr, neighbor));
-
+                vptr_e = igraph_Calloc(1, igraph_vector_t);
+                if (vptr_e == 0) {
+                IGRAPH_ERROR("cannot run igraph_get_all_shortest_paths", IGRAPH_ENOMEM);
```
ntamas, line 276: Please fix the indentation of this line.

src/paths/all_shortest_paths.c
```diff
@@ -293,14 +331,30 @@ int igraph_get_all_shortest_paths(const igraph_t *graph,
...
             while (fatherptr != 0) {
-                VECTOR(*res)[j++] = VECTOR(paths)[fatherptr - 1];
+                if (vertices) {
+                    VECTOR(*vertices)[j++] = VECTOR(paths)[fatherptr - 1];
+                } 
+                if (edges) {
+                    VECTOR(*edges)[t++] = VECTOR(path_edge)[fatherptr - 1];
+                }
+                /* If dont return the vertex or the edge path then release*/
+                if (!vertices) {
```
ntamas, line 341: Please merge this as an `else` branch with the `if (vertices) {` condition above -- they logically belong together.

src/paths/all_shortest_paths.c
```diff
@@ -293,14 +331,30 @@ int igraph_get_all_shortest_paths(const igraph_t *graph,
...
+                    igraph_vector_destroy(VECTOR(paths)[fatherptr - 1]);
+                    igraph_Free(VECTOR(paths)[fatherptr - 1]);
+                }
+                if (!edges) {
```
ntamas, line 345: Please merge this as an `else` branch with the `if (edges) {` condition above -- they logically belong together.

src/paths/dijkstra.c
```diff
@@ -622,18 +620,25 @@ static int igraph_i_vector_tail_cmp(const void* path1, const void* path2) {
...
  * \brief All weighted shortest paths (geodesics) from a vertex.
  *
  * \param graph The graph object.
- * \param res Pointer to an initialized pointer vector, the result
- *   will be stored here in igraph_vector_t objects. Each vector
- *   object contains the vertices along a shortest path from \p from
- *   to another vertex. The vectors are ordered according to their
- *   target vertex: first the shortest paths to vertex 0, then to
- *   vertex 1, etc. No data is included for unreachable vertices.
+ * \param vertices Pointer to an initialized pointer vector or NULL
```
ntamas, line 630: A period (`.`) is missing from the end of this line.

src/paths/dijkstra.c
```diff
@@ -622,18 +620,25 @@ static int igraph_i_vector_tail_cmp(const void* path1, const void* path2) {
...
+ *   If not NULL Each vector object contains the vertices along a
```
ntamas, line 631: `If not NULL, then each...`

src/paths/dijkstra.c
```diff
@@ -622,18 +620,25 @@ static int igraph_i_vector_tail_cmp(const void* path1, const void* path2) {
...
- *   target vertex: first the shortest paths to vertex 0, then to
- *   vertex 1, etc. No data is included for unreachable vertices.
+ * \param vertices Pointer to an initialized pointer vector or NULL
+ *   If not NULL Each vector object contains the vertices along a 
+ *   shortest path from \p from to another vertex. The vectors are
+ *   ordered according to their target vertex: first the shortest 
+ *   paths to vertex 0, then to vertex 1, etc. No data is included 
+ *   for unreachable vertices.
+ * \param edges Pointer to an initialized pointer vector or NULL. If 
+ *   not null the Each vector object contains the edges along a
```
ntamas, line 637: `not NULL, then each vector...`

src/paths/dijkstra.c
```diff
@@ -881,10 +907,10 @@ int igraph_get_all_shortest_paths_dijkstra(const igraph_t *graph,
         }
     }
 
-    if (res) {
-        igraph_vector_t *path, *paths_index, *parent_vec;
+    if (vertices || edges) {
+        igraph_vector_t *path, *paths_index, *parent_vec, *parent_edge_vec;
         igraph_stack_t stack;
-        long int j, node;
+        long int j, node, k;
```
ntamas, line 913: This is the unused variable that the CI environment is complaining about.

src/paths/dijkstra.c
ntamas: See my comment above -- can't we simply call `igraph_vector_init(path, 0)` to create an empty vector?

src/paths/dijkstra.c
```diff
@@ -1021,44 +1089,93 @@ int igraph_get_all_shortest_paths_dijkstra(const igraph_t *graph,
...
+        if (!free_vertices){
+            IGRAPH_CHECK(igraph_vector_ptr_resize(vertices, j));
+            /* sort the paths by the target vertices */
+            igraph_vector_ptr_sort(vertices, igraph_i_vector_tail_cmp);
+        }
+        
+        if (edges){
+            IGRAPH_CHECK(igraph_vector_ptr_resize(edges, j));
+            igraph_vector_ptr_sort(edges, igraph_i_vector_tail_cmp);
+            for (i = 0; i < igraph_vector_ptr_size(edges); i++) {
```
ntamas, line 1161: Add a comment here that explains that this is the place where we remove the target vertex ID that we have added to the end of each path vector for sake of consistent ordering with `vertices`.

tests/unit/all_shortest_paths.c
```diff
@@ -46,42 +48,53 @@ int main() {
...
 
     for (i=0; i < igraph_vector_ptr_size(&paths); ++i) {
         print_vector(VECTOR(paths)[i]);
+        igraph_vector_destroy(VECTOR(paths)[i]);
+        igraph_free(VECTOR(paths)[i]);
+        VECTOR(paths)[i] = 0;
     }
     IGRAPH_ASSERT(igraph_vector_ptr_size(&paths) == VECTOR(nrgeo)[to]);
 
-    igraph_vector_ptr_free_all(&paths);
```
szhorvat, line 56: I am wondering why this needed to be removed. The test _should_ work with `vector_ptr_free_all`.

Importantly, the `get_all_shortest_paths` function should not change the item destructor of `path` (if it does, that must be corrected).

guyroznb: I did it because when I use `vector_ptr_free_all` I get memory leak. So I used the same destroy mechanism as in `igraph_get_all_shortest_paths_dijkstra.c`  tests.

szhorvat: > I did it because when I use `vector_ptr_free_all` I get memory leak.

This should not happen if everything is working properly. Let's revert this to `vector_ptr_free_all` and figure out the true cause of the memory leak.

src/paths/dijkstra.c
```diff
@@ -941,36 +972,77 @@ int igraph_get_all_shortest_paths_dijkstra(const igraph_t *graph,
...
         igraph_vector_null(paths_index);
 
-        /* clear the paths vector */
-        igraph_vector_ptr_clear(res);
-        res_item_destructor = igraph_vector_ptr_get_item_destructor(res);
-        igraph_vector_ptr_set_item_destructor(res,
-                                              (igraph_finally_func_t*)igraph_vector_destroy);
+        if (edges){
+            igraph_vector_ptr_clear(edges);
+            res_item_destructor = igraph_vector_ptr_get_item_destructor(edges);
```
szhorvat, line 985: Note that `res_item_destructor` was used to store, and later revert, the item destructor of `res`, which you now renamed `vertices`. Thus, now we need to separately save and restore the item destructor of both `vertices` and `edges`.

tests/unit/all_shortest_paths.c
guyroznb: o.k I found out what was my mistake... thanks!

src/paths/dijkstra.c
ntamas: I will shuffle the ordering around, both here and where the `vertices` vector is allocated in case the user did not provide one.

examples/simple/igraph_get_all_shortest_paths_dijkstra.c
```diff
@@ -165,21 +188,30 @@ int main() {
...
+    igraph_vector_ptr_sort(&vertices, vector_tail_cmp);
+
+    for (i = 0; i < igraph_vector_ptr_size(&vertices); i++) {
+        igraph_vector_print(VECTOR(vertices)[i]);
+        igraph_vector_destroy(VECTOR(vertices)[i]);
+        igraph_free(VECTOR(vertices)[i]);
+        VECTOR(vertices)[i] = 0;
+    }
+
+    igraph_vector_ptr_sort(&edges, vector_tail_cmp);
```
szhorvat, line 208: This is not correct.

The idea was to return the paths sorted by target vertex, i.e. effectively grouped by target vertex. This code sorts the edge result _differently_ than the vertex result, which must not be the case. The `i`th element of `vertices` must correspond to the same path as the `i`th element of `edges`. Sorting by the final edge ID doesn't really make sense.

We must be careful not to blindly copy code, but understand why it was so in the first place.

See also:

https://igraph.discourse.group/t/why-are-shortest-paths-sorted/626

szhorvat (changes requested):

examples/simple/igraph_get_all_shortest_paths_dijkstra.c
ntamas: I was wondering about this too -- I was simply looking at the diff of the expected output of the test case and I did not understand why the order of the edgeID-paths did not match the vertexID-paths.

On the other hand, aren't we sorting the paths based on the ID of the last vertex anyway (in `igraph_get_all_shortest_paths_dijkstra()`)? If we do, we don't need to sort again in the example.

ntamas: Okay, now I understand. The implementation in `igraph_get_all_shortest_paths_dijkstra()` does not do a full lexicographic ordering based on the vertex IDs from the end, it simply compares the trailing vertex IDs, but if there is a tie, it is free to order the paths in any arbitrary order. The extra sorting in the test case ensures that ties are also broken in a deterministic manner.

I will probably move this example to the `tests/unit` folder if we have merged this PR as it is overly detailed and complex to serve as a good example. We'll need to write a shorter example for the manual.

ntamas: > The ith element of vertices must correspond to the same path as the ith element of edges. Sorting by the final edge ID doesn't really make sense.

I think that for the sake of the test case and ensuring determinism in reporting the output it's okay.

On the other hand, we should actually test whether the ordering of the vectors in `vertices` and `edges` is consistent, i.e. if we ask for both, igraph must return them in an order that resolving edge IDs in the i-th vector of `edges` yields the same vertices in the same order as they are placed in the i-th vector of `vertices`. This must be tested explicitly instead of simply printing the output of `edges` to stdout because the extra post-processing (sorting) in the test case would hide problems where the ordering of the vectors in `vertices` and `edges` is not consistent.

szhorvat:
I'm sorry, I didn't realize I was commenting on the test program, and not on the function implementation. I saw this problem a few days ago and I wanted to mention the issue before the PR is merged (I don't have a lot of time today and I was in a bit of a hurry).

I now see that the implementation does indeed sort correctly.

However, the way this is implemented is really a hack: We add the vertex IDs to the end of the edge ID vectors, then sort the list of edge ID vectors, finally remove the vertex IDs. The sorting is done entirely separately on list of edge-paths and vertex-paths, and this makes me uncomfortable. Note that there are multiple different path orderings which are all considered sorted here, as only the tail vertex is compared. There's a risk that something will go wrong in the future: maybe we end up using a non-deterministic sorting algorithm (unlikely), or due to some future modification of the code the vertex-path-sort ends up using a different sorting algorithm than the edge-path-sort.

In short: the sorting should ideally be done only once, and applied to both the edge-path list and the vertex-path list. 

I would write a function that computes the permutation that sorts the paths, then apply that permutation to both the vertex-paths and the edge-paths. Create an `igraph_vector_long_t order` using `igraph_vector_long_init_seq` and sort it using a comparison function that compares `i` and `j` through `igraph_vector_tail(VECTOR(vertices)[i])` and `igraph_vector_tail(VECTOR(vertices)[j])`.  Then apply the permutation to the result: create a new vector_ptr for vertices, and set its `k`th element to `vertices[order[k]]`.  Finally copy it back into vertices. Since this only stores/copies pointers, it will not have a great memory overhead.

What do you think @ntamas ?

examples/simple/igraph_get_all_shortest_paths_dijkstra.c
guyroznb: Hi @szhorvat,
we actually don't need to sort the edge result (and the vertices result) because the function `igraph_get_all_shortest_paths_dijkstra` return a sorted edge result vector (based on the target node and sorted the same as the vertex retult vector).
@ntamas should I move the example to `test/unit` and write a short example?
thanks

szhorvat:
To illustrate that my concern is not so "out there", C++17 already has parallel sort in its standard library. Parallel sort is not deterministic. If the underlying sort implementation will ever become parallell in the future (maybe we start calling the system `qsort` again, which we have no control over), then there is a risk that the edge-paths and vertex-paths would end up being sorted differently.

ntamas:
> What do you think @ntamas ?

I agree. (Sorry, this mention has slipped through the cracks and I haven't noticed it until now when I started reviewing the state of PRs in our queue).

szhorvat:
@guyroznb Can you please make the change I requested above? The vertex and edge vectors should be sorted together. Let us know if this isn't completely clear or if you need help merging the latest develop branch into this PR.

guyroznb:
Hi @szhorvat, you wrote that you would write a function that compute the permutation for sorting the path. I can write this function or maybe sort the vertices first and, find the permutation vector and sort the edges path accordingly.

@szhorvat what do you think?

szhorvat:
It would be quite helpful if you could make this change.

You can use `igraph_qsort_r` with a comparison function to sort a vector of indices based on vertices, and effectively obtain the needed permutation. The extra argument to the comparison function can be the vector of vertex paths. Then apply this permutation to both the vertex paths and edge paths.

guyroznb:
hi @szhorvat  I don't understand the sorting function `igraph_qsort_t` and I don't see any reference to the function of the documentation. Can you please explain this function, what is the `base` and the `thunk` inputs

szhorvat:
It is the same function as described [here](https://www.freebsd.org/cgi/man.cgi?query=qsort_r&apropos=0&sektion=3&manpath=FreeBSD+11-current&format=html)

The prototype is

```
void igraph_qsort_r(void *base, size_t nel, size_t width, void *thunk,
                                  int (*compar)(void *, const void *, const void *));
```

`base` is the raw array to be sorted. Since we are creating a permutation vector, this will be an array of `long`s.

`nel` is the number of elements in the array.

`width` is the size of a single element in the array, i.e. `sizeof(long)`.

`thunk` will be passed as the 1st argument of the comparison function. We will pass a pointer to the vector_ptr of vertex paths here.

The comparison function will receive then, `thunk` (which is just a pointer to the list of vector paths) and two pointers to indices inside of the the permutation vector. It should use these indices to index into the list of vertex paths, and compare what it finds there.

src/paths/dijkstra.c
```diff
@@ -617,23 +616,35 @@ int igraph_get_shortest_path_dijkstra(const igraph_t *graph,
...
-    return (int) (igraph_vector_tail(*(const igraph_vector_t**)path1) -
-                  igraph_vector_tail(*(const igraph_vector_t**)path2));
+static int igraph_i_vector_tail_cmp(void *extra, const void* a, const void* b) {
+    igraph_vector_ptr_t *paths = (igraph_vector_ptr_t *) extra;
+    long int *aa = (long int*) a;
+    long int *bb = (long int*) b;
+    igraph_vector_t *path_a = (igraph_vector_t *) VECTOR(*paths)[*aa];
+    igraph_vector_t *path_b = (igraph_vector_t *) VECTOR(*paths)[*bb];
+
+    return (int) (igraph_vector_tail(path_a) - igraph_vector_tail(path_b));
```
ntamas, line 626: There might be a subtle bug here (and it has been there for a while now). If `igraph_vector_tail(path_a) - igraph_vector_tail(path_b)` (which is of type `igraph_real_t`) happens to be equal to a multiple of `INT_MAX+1`, that _might_ be cast into a zero. I don't know whether it actually happens for doubles; it definitely happens when casting from `long int` to `int` and the `long int` is an exact multiple of `INT_MAX+1`, and I'm not sure how compliant C compilers are, so it's probably better to spell out the comparisons explicitly. I'll fix that in a new commit.

ntamas:
Nearly done with this one; I still need to check the part at the end of `igraph_get_all_shortest_paths_dijkstra()` where the paths are permuted. I have a feeling that we could probably make use of a (not-yet-implemented) `igraph_vector_ptr_qsort_ind()` function that would be similar to `igraph_vector_qsort_ind()` but for pointer vectors. The rest seems okay now.

ntamas:
Finally! :tada: Sorry it took so long and thanks for your hard work on this!

Issue #1634, opened 2021-01-28
Several tests fail on Windows because of MSVC printf bug


szhorvat:
Several test fail on Windows, in short because of this printf bug: https://stackoverflow.com/q/65900619/695132

In short, `printf("%#g\n", 0.0)` prints one more digit with MSVC than other compilers. Note that this is almost certainly dependent not on the compiler, but the standard library.

Problem comes from here, we will likely need to change the testing code:

https://github.com/igraph/igraph/blob/develop/tests/unit/test_utilities.inc#L38

https://github.com/igraph/igraph/blob/develop/tests/unit/test_utilities.inc#L73

I tested this with VS 2019 version 16.8.4 (latest) on Windows 7 and I see the problem both in Debug and Release mode. Big question: why do we not see it on Azure Pipelines? @vtraag if you can, please do a local test on your end.

szhorvat:
I'll post a list of failing tests when I get access to a Windows machine later.

ntamas:
Damn, I accidentally assigned this to myself and now I can't de-assign myself for some reason.

PR #1656, opened 2021-02-09
Add lapack_dgetrf unit tests

GroteGnoom:
Part of #1592


GroteGnoom:
@szhorvat Does https://github.com/igraph/igraph/commit/c218f4b6ef7d49297cbcbabfb42b1e7741edffa1 cause 0.00000 to be printed out as 0.000000 on windows?

szhorvat:
Yes, MSVC bug. https://stackoverflow.com/q/65900619/695132 

I noticed it a while ago (#1634), but for some reason (that I still don't understand) it didn't affect the CI until I changed to use the Ninja generator today. It might be a Release vs Debug mode thing, not sure.

I'll fix this in some way tonight.

szhorvat:
You can now merge `develop` into this branch (or rebase it) to get rid of the test failures

szhorvat:
@GroteGnoom When you fix a bug (as here—the function didn't allocate the proper memory for the `ipiv` vector), please do add a note to the PR, to make it easier to write the changelog later.

PR #1657, opened 2021-02-10
Add lapack_dgetrs unit tests

GroteGnoom:
Part of #1592

Bugfix: No more invalid memory reads on invalid pivot vector.

Slightly problematic end results: A singular matrix seems to result in random output. Since you would be warned for this in dgetrf, and it's seems like a lot of work to check if the matrix is singular, there are no warnings or errors given for this case.


GroteGnoom:
Bugfix: No more invalid memory reads on invalid pivot vector.

Slightly problematic end results: A singular matrix seems to result in random output. Since you would be warned for this in dgetrf, and it's seems like a lot of work to check if the matrix is singular, there are no warnings or errors given for this case.

src/linalg/lapack.c
```diff
@@ -136,46 +140,52 @@ int igraph_lapack_dgetrs(igraph_bool_t transpose, const igraph_matrix_t *a,
...
     if (n != igraph_matrix_ncol(a)) {
-        IGRAPH_ERROR("Cannot LU solve matrix", IGRAPH_NONSQUARE);
+        IGRAPH_ERROR("Cannot LU solve matrix.", IGRAPH_NONSQUARE);
     }
     if (n != igraph_matrix_nrow(b)) {
-        IGRAPH_ERROR("Cannot LU solve matrix, RHS of wrong size", IGRAPH_EINVAL);
+        IGRAPH_ERROR("Cannot LU solve matrix, RHS of wrong size.", IGRAPH_EINVAL);
+    }
+    if (igraph_vector_int_size(ipiv) > 0 &&
+            (igraph_vector_int_max(ipiv) > n || igraph_vector_int_min(ipiv) < 1)) {
```
szhorvat, line 149: Let's use `igraph_vector_int_minmax` which finds the min and max in one pass.

src/linalg/lapack.c
```diff
@@ -117,10 +117,14 @@ int igraph_lapack_dgetrf(igraph_matrix_t *a, igraph_vector_int_t *ipiv,
...
  *      matrix.
  * \param a A matrix containing the L and U factors from the
- *      factorization A = P*L*U.
+ *      factorization A = P*L*U. L is expected to be unitriangular,
+ *      diagonal entries are those of U. If A is singular, no warning or
+ *      error wil be given and random output will be returned.
  * \param ipiv An integer vector, the pivot indices from \ref
- *      igraph_lapack_dgetrf must be given here.
- * \param b The right hand side matrix must be given here.
+ *      igraph_lapack_dgetrf must be given here. Row \c i of A was
```
szhorvat, line 124: Please add a `()`, i.e. `igraph_lapack_dgetrf()` (just some cosmetic nitpick)

szhorvat:
I'll make the changed from my comments (just a heads up so we don't step on each other's toes)

GroteGnoom:
Allright! I almost got started :slightly_smiling_face:
Also, the merge broke test_utilities.inc.

Issue #1577, opened 2020-12-19
Write tests for igraph_heap_ functions


szhorvat:
There are no tests for the `igraph_heap_` functions.

https://github.com/igraph/igraph/blob/develop/tests/unit/heap.c#L28

These functions are part of the public API, and generally useful for graph algorithms. They should have tests.

RamySaied1:
I would like to work on this

szhorvat:
Welcome! You will find guidance for getting started in the wiki: https://github.com/igraph/igraph/wiki Be sure to work on the `develop` branch. The first step is to learn [to build igraph](https://igraph.discourse.group/t/draft-document-for-installation-with-cmake/504). For general questions, you can use the forum at https://igraph.discourse.group/  As you go, please also verify that the documentation of these function is correct and complete.

These functions are part of the public API, so it's good to have tests for them. But it's good to note that they are not currently used at all internally, therefore fixing this task is of low priority. We are currently preparing for a release, so any PRs will likely be reviewed with 1-2 weeks of delay from now.

PR #1658, opened 2021-02-11
Add igraph_heap_ functions unit tests

RamySaied1:
issue #1577


tests/unit/heap.c
```diff
@@ -23,9 +23,66 @@
...
+    igraph_heap_reserve(&h, 5);
+    igraph_heap_destroy(&h);
+
+    /* igraph_heap_init_array*/
+    printf("Test igraph_heap_init_array\n");
+    igraph_heap_init_array(&h,arr,5);
+
+    /* igraph_heap_empty igraph_heap_size */
+    printf("Test igraph_heap_size, igraph_heap_empty\n");
+    IGRAPH_ASSERT((igraph_heap_size(&h)==5) && (!igraph_heap_empty(&h)));
```
szhorvat, line 54: Note that this won't test `heap_empty` at all if `(igraph_heap_size(&h)==5)` evaluates to true (short-circuiting `&&`). The `heap_size` and `heap_empty` calls should be separated, i.e. not joined with `&&`.

tests/unit/heap.c
```diff
@@ -23,9 +23,66 @@
...
+    igraph_heap_init(&h, 0);
+    IGRAPH_ASSERT((igraph_heap_size(&h)==0) && (igraph_heap_empty(&h)));
+
+    /* igraph_heap_push igraph_heap_top  */
+    printf("Test igraph_heap_push, igraph_heap_top\n");
+    igraph_real_t random_list[13]={-9.999,0,6,235,-2,-1000,-1,4,2000,0.1,1,-9,10};
+    igraph_vector_t top_sequence;
+    igraph_vector_init(&top_sequence,13);
+    for (i=0;i<13;i++)
+    {
```
szhorvat, line 69: Please put the `{` on the previous line to maintain a consistent style.

tests/unit/heap.c
```diff
@@ -23,9 +23,66 @@
...
+        igraph_heap_push(&h,random_list[i]);
+        VECTOR(top_sequence)[i]=igraph_heap_top(&h);
+    }
+    print_vector_format(&top_sequence,stdout,"%g");
+
+    /* igraph_heap_delete_top  */
+    printf("Test igraph_heap_delete_top\n");
+    for (i=0;i<13;i++)  VECTOR(top_sequence)[i]=(&top_sequence,igraph_heap_delete_top(&h));
```
szhorvat, line 77: Please use the usual igraph formatting style, i.e. add braces, put the `for`'s body on the next line.

I don't understand what you are trying to do with `(&top_sequence,igraph_heap_delete_top(&h))`. In C, `a,b` evaluates to `b`.

tests/unit/heap.c
```diff
@@ -23,9 +23,66 @@
...
+    printf("Test igraph_heap_push, igraph_heap_top\n");
+    igraph_real_t random_list[13]={-9.999,0,6,235,-2,-1000,-1,4,2000,0.1,1,-9,10};
+    igraph_vector_t top_sequence;
+    igraph_vector_init(&top_sequence,13);
+    for (i=0;i<13;i++)
+    {
+        igraph_heap_push(&h,random_list[i]);
+        VECTOR(top_sequence)[i]=igraph_heap_top(&h);
+    }
+    print_vector_format(&top_sequence,stdout,"%g");
```
szhorvat, line 73: This is the same as `print_vector(&top_sequence)`. However, let us keep things simple (and the risk of future mistakes low) and print the heap elements directly from the `for` loop instead of writing them into a vector first.

tests/unit/heap.c
```diff
@@ -23,9 +23,66 @@
...
+    IGRAPH_ASSERT((igraph_heap_size(&h)==5) && (!igraph_heap_empty(&h)));
+    igraph_heap_destroy(&h);
+    igraph_heap_init(&h, 0);
+    IGRAPH_ASSERT((igraph_heap_size(&h)==0) && (igraph_heap_empty(&h)));
+
+    /* igraph_heap_push igraph_heap_top  */
+    printf("Test igraph_heap_push, igraph_heap_top\n");
+    igraph_real_t random_list[13]={-9.999,0,6,235,-2,-1000,-1,4,2000,0.1,1,-9,10};
+    igraph_vector_t top_sequence;
+    igraph_vector_init(&top_sequence,13);
```
szhorvat, line 67: I would prefer to keep declarations at the top of the block, or if you want, add a new nested block here which delineates this specific part of the test.

tests/unit/heap.c
```diff
@@ -23,9 +23,66 @@
...
+    /* igraph_heap_reserve */
+    printf("Test igraph_heap_reserve\n");
+    igraph_heap_init(&h, 0);
+    igraph_heap_reserve(&h, 10);
+    igraph_heap_reserve(&h, 5);
+    igraph_heap_destroy(&h);
+
+    /* igraph_heap_init_array*/
+    printf("Test igraph_heap_init_array\n");
+    igraph_heap_init_array(&h,arr,5);
```
szhorvat, line 50: It would be nice to test that this actually copies the elements from `arr`. Therefore, I would use an array with known elements, such as `random_list` below, then check that the correct elements were added to the heap.

szhorvat (changes requested):
Thanks for the contributions! I left some comments.

In order to be complete, let us also test `igraph_heap_min_t` and its associated functions and verify that it works as a min-heap, as expected.

tests/unit/heap.c
```diff
@@ -23,9 +23,120 @@
...
+
 int main() {
 
-    /* This is not used by any functions any more,
-       no need to test it right now */
+    igraph_heap_t h_max;
+    igraph_heap_min_t h_min;
+    igraph_integer_t i;
+    igraph_real_t random_list[13]={-9.999,0,6,235,-2,-1000,-1,4,2000,0.5,1,-9,10};
+    const igraph_integer_t li_size=13;
```
szhorvat, line 34: I suggest `const igraph_integer_t li_size = sizeof(random_list) / sizeof(igraph_real_t)`.

szhorvat:
Thanks for the fixes! Be sure to comment here when you feel it's ready so we don't step on each others toes trying to edit the same files.

RamySaied1:
> Thanks for the fixes! Be sure to comment here when you feel it's ready so we don't step on each others toes trying to edit the same files.

it's done now

szhorvat:
Thanks for your contribution, it looks great!

PR #1659, opened 2021-02-12
Add layout_star unit tests

GroteGnoom:
Part of #1592

Caution: For this function the error checking time for out of range
order indices can be close to the total running time, so I'm not
100% sure about it.


szhorvat:
I think it's alright. This function is nor performance critical. It's running time would be impercetible for any graph that is small enough to be visualized. The error checking is more valuable than speed here.

PR #1660, opened 2021-02-12
Add arpack unpack complex unit tests

GroteGnoom:
There were a few bugs:

The first output vector was conjugated, even though it should be the unconjugated version (https://igraph.discourse.group/t/the-unit-tests-thread/597/23)
Memory was accessed out of bounds on normal input, and values were incorrect.

I rewrote the code. The new version has the bugs fixed, but does use more memory. I hope this is ok, but I can easily revert it. But then I would have to find a way to fix the old code.

If people were using this and it was working then I probably misunderstood something.

This requires matrix_init_real_row_major from #1657.


src/linalg/arpack.c
```diff
@@ -1374,64 +1379,52 @@ int igraph_arpack_unpack_complex(igraph_matrix_t *vectors, igraph_matrix_t *valu
...
-    /* We copy the j-th eigenvector to the (k-1)-th and k-th column */
-    k = nev * 2 - 1;
-
-    for (i = nev - 1; i >= 0; i--) {
+    new_vector_pos = 0;
+    vector_pos = 0;
+    for (i = 0; i < nev && vector_pos < igraph_matrix_ncol(vectors); i++) {
         if (MATRIX(*values, i, 1) == 0) {
+            /* Real eigenvalue */
+            memcpy(&MATRIX(new_vectors, 0, new_vector_pos), &MATRIX(*vectors, 0, vector_pos), colsize);
```
ntamas, line 1394: This assumes that igraph matrices are stored in column-major order. This is true, but if we ever decide to change it later, this part of the code will break. I'd rather use a for loop for copying the elements even if it is less efficient. A good compiler should be able to optimize it away anyway, and even if it doesn't, the performance hit is probably negligible.

We can also decide to add an `igraph_matrix_copy_column()` function in `matrix.pmt` if we really want to; this way the assumption about column-major matrices will be encapsulated in `matrix.pmt`.

src/linalg/arpack.c
```diff
@@ -1374,64 +1379,52 @@ int igraph_arpack_unpack_complex(igraph_matrix_t *vectors, igraph_matrix_t *valu
...
+            for (j = 0; j < nodes; j++) {
+                MATRIX(new_vectors, j, new_vector_pos + 3) = -MATRIX(new_vectors, j, new_vector_pos + 3);
             }
-            k -= 2;
+            new_vector_pos += 4;
+            vector_pos += 2;
         }
     }
+    igraph_matrix_destroy(vectors);
+    igraph_matrix_copy(vectors, &new_vectors);
```
ntamas, line 1425: `igraph_matrix_copy` can fail so it needs to be wrapped in an `IGRAPH_CHECK()`.

ntamas (changes requested):

ntamas:
Thanks, looks good to me, a few minor comments are above.

src/linalg/arpack.c
GroteGnoom: Good point :) Since column-copying isn't used that much ( I think), and the for loops are fine, I think for now just using a for loop every time is no problem.

ntamas (approved):

ntamas:
Thanks!

PR #1655, opened 2021-02-09
Add selector unit tests, add docs

GroteGnoom:
Part of #1592

There is still an unfreed calloc that needs to be fixed (https://igraph.discourse.group/t/the-unit-tests-thread/597/20)

PR #1663, opened 2021-02-13
add fruchterman reingold unit tests

### LINKED ISSUE #1208: Add tests for igraph_closeness

### LINKED ISSUE #1378: COPYING file is missing from MSVC zip
The `COPYING` and `ACKNOWLEDGEMENTS.md` files are missing from the MSVC source zip. This should also be included to ensure proper licences are always included. Similarly, we could perhaps also include the `CHANGELOG.md`.

### LINKED ISSUE #1444: \experimental and \deprecated for the documentation
This is a request to add features to the documentation system that makes it easy to mark a function as *experimental* (i.e. API stability not guaranteed yet) or as *deprecated*. We discussed this here: https://igraph.discourse.group/t/a-proposal-for-api-stability-marking-functions-as-experimental/248

One possible way to do this is to create the commands `\experimental` and `\deprecated` which will insert a pre-determined explanatory text (possibly with styling) in the docs.

With `\deprecated` we would usually want to add some extra information, e.g. the standard text could be "This symbol is deprecated.", followed by the custom text "Use `foo` or `bar` instead."

I'm not sure what's the best implementation is, but I hope the *intent* is clear from this issue :-)

### LINKED ISSUE #1489: cmake: GLPK_SUPPORT=OFF and USE_INTERNAL_GLPK=ON causes failure
Setting GLPK_SUPPORT=OFF and USE_INTERNAL_GLPK=ON causes a failure, but without any informative error message. This should be improved somehow (e.g. by making it clear to the user why it fails)

### LINKED ISSUE #1493: CMake reports wrong version
CMake reports version number as 0.8.3, while it should be 0.9, 0.9-pre, or similar (as with autotools on the `develop` branch).

### LINKED ISSUE #1503: CMake fails when git is not installed
At the moment, CMake fails when `git` is not installed, without a clear warning that `git` could not be found. Perhaps we can call [`FindGit`](https://cmake.org/cmake/help/v3.19/module/FindGit.html?highlight=findgit) and report a proper error when `git` is not installed? This if of course only necessary when installing from the `git` repository. From a packaged version (which should contain a fixed version), `git` should not be necessary.

### LINKED ISSUE #1505: CMake: error from "variables set to NOTFOUND"
I regularly encounter the following:

 - I run `cmake ..` in `build`
 - Configuration fails because of some missing dependencies, e.g. GLPK
 - I either run `ccmake` or edit CMakeCache.txt and set it to use internal GLPK

At this point, running `cmake ..` again results in the following error:

```
-- igraph configured successfully.
--  
-- Configuring done
CMake Error: The following variables are used in this project, but they are set to NOTFOUND.
Please set them or make sure they are set and tested correctly in the CMake files:
/Users/szhorvat/Repos/igraph-main/igraph/src/GLPK_INCLUDE_DIR
   used as include directory in directory /Users/szhorvat/Repos/igraph-main/igraph/src

-- Generating done
CMake Generate step failed.  Build files cannot be regenerated correctly.
```

Now I can manually remove `GLPK_INCLUDE_DIR`, and then not only the configure step, but also the generation step will succeed. @ntamas: I wonder if there is a way to improve this and make it more user friendly.

For some reason, when I use `ccmake` or `cmake-gui`, it still show the warning above, but the generation step doesn't actually fail. Perhaps there's an option to ignore the NOTFOUND variable if it's not used anyway? Still, this is a user experience issue. We should expect that many/most users of igraph have little to no experience with cmake.

### LINKED ISSUE #1535: CMake: ctest should show the output upon failure
This is how a failure is currently reported:

https://travis-ci.org/github/igraph/igraph/jobs/741864231#L1369

There should be more information, such as the diff between the expected outputs, and the return code.

Also, this line seems to be obsolete: https://github.com/igraph/igraph/blob/feature/cmake-build/.travis.yml#L52

### LINKED ISSUE #1539: igraph_is_connected() and null graphs
I opened this issue to discuss changing `igraph_is_connected()` to return `false` for the null graph. The null graph is the graph with no vertices.

There is no one clearly correct answer to the question of whether the null graph is connected. It is, of course, a question of convention. However, it is a common opinion that it should be considered disconnected:

 - https://math.stackexchange.com/a/50552/12384
 - [Is the null graph a pointless concept?](https://link.springer.com/chapter/10.1007%2FBFb0066433) by Harary and Read

The usual argument for considering the null graph not to be connected is that some of the properties that connected graphs with at least one vertex have break down for the null graph. One might explicitly make the exception for the null graph when discussing or defining those properties, or one might simply not consider the null graph connected. 

I think that in igraph we should make `igraph_is_connected()` return `false` for the null graph for the same reason: consistency with other functions.

To give an example, a classic argument (also in the paper I mentioned above) is this: A forest is a graph with no cycles. A tree is a connected graph with no cycles. Generally, a tree has one more vertex than edges, V = E+1. This is not true for the null graph. It makes sense to consider the null graph a forest, but not a tree, by convention, so that the V = E+1 property can hold for all trees. This is exactly what `igraph_is_tree()` does in igraph.

In order to avoid having to deal with these issues, some graph theory authors simply require at least one vertex in their definition of a graph.

----

Now we are not writing a graph theory book. We are writing graph theory software. So the considerations are different: **what is useful and safe (i.e. likely to prevent, rather than cause bugs) to users?**  

Not supporting zero-vertex graphs is not useful: it goes against expectations and would create many inconvenient error conditions, such as deleting the vertex of a singleton graph throwing an error. Thus, we *should* support null graphs (and we do).

Imagine this situation: A user is implementing a new function, using igraph's functionality. They assume that connected graphs have a number of properties that are required for the successful completion of the algorithm. Thus they write their function with `if is_connected(g) do X else ERROR`. But many of the properties that one might take for granted for connected graphs break down for the null graph, so it is likely that their function will fail in _unexpected_ ways when passing the null graph to it. To prevent this, the user: 1. has to remember this special case 2. has to implement this special case.

I assert that the safer, more user-friendly solution is that `is_connected(null_graph)` return `false`. A user who does think of special cases can still implement a nicer behaviour in their function for null graphs. A user who doesn't won't end up with a function that fails in strange ways.

One example: it is a natural assumption that a connected graph should have a diameter path, but the null graph does not. It has no paths, not even paths of length zero, in it. Compare with the singleton graph which has a path of length 0, consisting of a single vertex.

----

To be fair, I should note that no other software I have tried makes this choice. However, it is likely that they did not even consider the question seriously. I maintain that the better way is to consider the null graph disconnected.

----

Arguments against making this change:

 - Many other functions in igraph rely on `igraph_is_connected`, we must not break them. **Answer:** True, and I will check all such functions. Maybe I will even discover some edge-case behaviour that needs to be corrected. That will be a win.

 - Such a change may break existing user code. **Answer:** Since this edge case is commonly forgotten, I believe that this change is just as likely to fix existing user code as to break it.

### LINKED ISSUE #1590: Default RNG's state is not thread local
The default RNG's state is not thread-local despite the misleading IGRAPH_THREAD_LOCAL qualifier here:

https://github.com/igraph/igraph/blob/develop/src/random/random.c#L500

Notice that it references `igraph_i_rng_default_state`, which is not thread local:

https://github.com/igraph/igraph/blob/develop/src/random/random.c#L482

It cannot be made thread-local in a straightforward way because its address is used when setting `igraph_i_rng_default`.

There is no mention of this problem/limitation in the documentation about multi-threaded programming:

https://igraph.org/c/doc/igraph-Advanced.html

### LINKED ISSUE #1606: erdos_renyi_game: random test failure
**Describe the bug**
In one of the test runs, Address Sanitizer found a leak in `igraph_erdos_renyi_game()`

https://dev.azure.com/igraph-team/igraph/_build/results?buildId=201&view=logs&j=4c3a8381-b9de-5463-904f-918810769be8&t=a906eab8-d2a3-5e54-d199-606dbbfb0805&l=132

It is unclear if this is a false positive. The test does not currently set a random seed, so it *may* be real.

**To reproduce**
Unclear at the moment.

**Test output:**

```
2021-01-14T01:09:24.2979295Z  58/289 Test  #59: igraph_erdos_renyi_game ...................***Failed    0.39 sec
2021-01-14T01:09:24.2980070Z 
2021-01-14T01:09:24.2980531Z =================================================================
2021-01-14T01:09:24.2981089Z ==11974==ERROR: LeakSanitizer: detected memory leaks
2021-01-14T01:09:24.2981482Z 
2021-01-14T01:09:24.2981992Z Direct leak of 352 byte(s) in 1 object(s) allocated from:
2021-01-14T01:09:24.2983541Z     #0 0x7f46153dff30 in realloc (/usr/lib/x86_64-linux-gnu/libasan.so.4+0xdef30)
2021-01-14T01:09:24.2984216Z     #1 0x5636b6a9209e in igraph_vector_reserve ../src/core/vector.pmt:443
2021-01-14T01:09:24.2984794Z     #2 0x5636b6a2c96c in igraph_add_edges ../src/graph/type_indexededgelist.c:273
2021-01-14T01:09:24.2985379Z     #3 0x5636b6ad8897 in igraph_create ../src/constructors/basic.c:85
2021-01-14T01:09:24.2985935Z     #4 0x5636b6a3b897 in igraph_erdos_renyi_game_gnp ../src/games/games_other.c:673
2021-01-14T01:09:24.2986730Z     #5 0x5636b6a3c8ec in igraph_erdos_renyi_game ../src/games/games_other.c:821
2021-01-14T01:09:24.2987295Z     #6 0x5636b6a2826d in main ../examples/simple/igraph_erdos_renyi_game.c:107
2021-01-14T01:09:24.2988170Z     #7 0x7f46145a7bf6 in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x21bf6)
2021-01-14T01:09:24.2988697Z 
2021-01-14T01:09:24.2989149Z Direct leak of 352 byte(s) in 1 object(s) allocated from:
2021-01-14T01:09:24.2990001Z     #0 0x7f46153dff30 in realloc (/usr/lib/x86_64-linux-gnu/libasan.so.4+0xdef30)
2021-01-14T01:09:24.2990664Z     #1 0x5636b6a9209e in igraph_vector_reserve ../src/core/vector.pmt:443
2021-01-14T01:09:24.2991241Z     #2 0x5636b6a2c9b0 in igraph_add_edges ../src/graph/type_indexededgelist.c:274
2021-01-14T01:09:24.2991808Z     #3 0x5636b6ad8897 in igraph_create ../src/constructors/basic.c:85
2021-01-14T01:09:24.2992360Z     #4 0x5636b6a3b897 in igraph_erdos_renyi_game_gnp ../src/games/games_other.c:673
2021-01-14T01:09:24.2992932Z     #5 0x5636b6a3c8ec in igraph_erdos_renyi_game ../src/games/games_other.c:821
2021-01-14T01:09:24.2993532Z     #6 0x5636b6a2826d in main ../examples/simple/igraph_erdos_renyi_game.c:107
2021-01-14T01:09:24.2994375Z     #7 0x7f46145a7bf6 in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x21bf6)
2021-01-14T01:09:24.2994870Z 
2021-01-14T01:09:24.2995337Z Direct leak of 352 byte(s) in 1 object(s) allocated from:
2021-01-14T01:09:24.2996143Z     #0 0x7f46153dff30 in realloc (/usr/lib/x86_64-linux-gnu/libasan.so.4+0xdef30)
2021-01-14T01:09:24.2996807Z     #1 0x5636b6a9209e in igraph_vector_reserve ../src/core/vector.pmt:443
2021-01-14T01:09:24.2997362Z     #2 0x5636b6a92892 in igraph_vector_resize ../src/core/vector.pmt:927
2021-01-14T01:09:24.2997899Z     #3 0x5636b6ab8ca3 in igraph_vector_order ../src/core/vector.c:148
2021-01-14T01:09:24.2998474Z     #4 0x5636b6a2cdab in igraph_add_edges ../src/graph/type_indexededgelist.c:298
2021-01-14T01:09:24.3000198Z     #5 0x5636b6ad8897 in igraph_create ../src/constructors/basic.c:85
2021-01-14T01:09:24.3000663Z     #6 0x5636b6a3b897 in igraph_erdos_renyi_game_gnp ../src/games/games_other.c:673
2021-01-14T01:09:24.3001037Z     #7 0x5636b6a3c8ec in igraph_erdos_renyi_game ../src/games/games_other.c:821
2021-01-14T01:09:24.3001407Z     #8 0x5636b6a2826d in main ../examples/simple/igraph_erdos_renyi_game.c:107
2021-01-14T01:09:24.3002113Z     #9 0x7f46145a7bf6 in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x21bf6)
2021-01-14T01:09:24.3002343Z 
2021-01-14T01:09:24.3002574Z Direct leak of 352 byte(s) in 1 object(s) allocated from:
2021-01-14T01:09:24.3003140Z     #0 0x7f46153dff30 in realloc (/usr/lib/x86_64-linux-gnu/libasan.so.4+0xdef30)
2021-01-14T01:09:24.3003500Z     #1 0x5636b6a9209e in igraph_vector_reserve ../src/core/vector.pmt:443
2021-01-14T01:09:24.3003843Z     #2 0x5636b6a92892 in igraph_vector_resize ../src/core/vector.pmt:927
2021-01-14T01:09:24.3004193Z     #3 0x5636b6ab8ca3 in igraph_vector_order ../src/core/vector.c:148
2021-01-14T01:09:24.3004550Z     #4 0x5636b6a2cdf8 in igraph_add_edges ../src/graph/type_indexededgelist.c:299
2021-01-14T01:09:24.3004911Z     #5 0x5636b6ad8897 in igraph_create ../src/constructors/basic.c:85
2021-01-14T01:09:24.3005278Z     #6 0x5636b6a3b897 in igraph_erdos_renyi_game_gnp ../src/games/games_other.c:673
2021-01-14T01:09:24.3005650Z     #7 0x5636b6a3c8ec in igraph_erdos_renyi_game ../src/games/games_other.c:821
2021-01-14T01:09:24.3006021Z     #8 0x5636b6a2826d in main ../examples/simple/igraph_erdos_renyi_game.c:107
2021-01-14T01:09:24.3006802Z     #9 0x7f46145a7bf6 in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x21bf6)
2021-01-14T01:09:24.3007029Z 
2021-01-14T01:09:24.3007257Z Direct leak of 88 byte(s) in 1 object(s) allocated from:
2021-01-14T01:09:24.3007797Z     #0 0x7f46153dff30 in realloc (/usr/lib/x86_64-linux-gnu/libasan.so.4+0xdef30)
2021-01-14T01:09:24.3008171Z     #1 0x5636b6a9209e in igraph_vector_reserve ../src/core/vector.pmt:443
2021-01-14T01:09:24.3008515Z     #2 0x5636b6a92892 in igraph_vector_resize ../src/core/vector.pmt:927
2021-01-14T01:09:24.3009002Z     #3 0x5636b6a29744 in igraph_add_vertices ../src/graph/type_indexededgelist.c:372
2021-01-14T01:09:24.3009392Z     #4 0x5636b6a29cc4 in igraph_empty_attrs ../src/graph/type_indexededgelist.c:141
2021-01-14T01:09:24.3009768Z     #5 0x5636b6a29f9f in igraph_empty ../src/graph/type_indexededgelist.c:81
2021-01-14T01:09:24.3010131Z     #6 0x5636b6ad87f2 in igraph_create ../src/constructors/basic.c:78
2021-01-14T01:09:24.3010491Z     #7 0x5636b6a3b897 in igraph_erdos_renyi_game_gnp ../src/games/games_other.c:673
2021-01-14T01:09:24.3010861Z     #8 0x5636b6a3c8ec in igraph_erdos_renyi_game ../src/games/games_other.c:821
2021-01-14T01:09:24.3011294Z     #9 0x5636b6a2826d in main ../examples/simple/igraph_erdos_renyi_game.c:107
2021-01-14T01:09:24.3011900Z     #10 0x7f46145a7bf6 in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x21bf6)
2021-01-14T01:09:24.3012127Z 
2021-01-14T01:09:24.3012368Z SUMMARY: AddressSanitizer: 1496 byte(s) leaked in 5 allocation(s).
```

### LINKED ISSUE #1614: Code coverage fails on io/*-lexer.l not found
**Describe the bug**
When generating code coverage, I get a 

> genhtml: ERROR: cannot read /home/daniel/Documents/repos/igraph/io/lgl-lexer.l

I used a clean build directory, and the type of lexer switches. Sometimes it's ncol, sometimes dl, sometimes pajek.
The file
> ./coverage/io/lgl-lexer.l.gcov.html

does exist.

**To reproduce**
Follow the steps in
https://github.com/igraph/igraph/wiki/Generating-test-coverage-reports

**Version information**
The current develop version at 665d22206cfe46e03adb82c4a2096ab37f358bde
cmake 3.16.3
GNU Make 4.2.1
flex 2.6.4
bison (GNU Bison) 3.5.1
gcov (Ubuntu 9.3.0-17ubuntu1~20.04) 9.3.0
lcov: LCOV version 1.14

**other info**
I do get these warnings, which seem harmless, but are for the related parsers:

> [ 13%] [FLEX][pajek_lexer] Building scanner with flex 2.6.4
> [ 13%] [BISON][dl_parser] Building parser with bison 3.5.1
> /home/daniel/Documents/repos/igraph/src/io/dl-parser.y:78.1-12: warning: deprecated directive: ‘%pure-parser’, use ‘%define api.pure’ [-Wdeprecated]
>    78 | %pure-parser
>       | ^~~~~~~~~~~~
>       | %define api.pure
> /home/daniel/Documents/repos/igraph/src/io/dl-parser.y:81.1-27: warning: deprecated directive: ‘%name-prefix="igraph_dl_yy"’, use ‘%define api.prefix {igraph_dl_yy}’ [-Wdeprecated]
>    81 | %name-prefix="igraph_dl_yy"
>       | ^~~~~~~~~~~~~~~~~~~~~~~~~~~
>       | %define api.prefix {igraph_dl_yy}
> /home/daniel/Documents/repos/igraph/src/io/dl-parser.y:84.1-14: warning: deprecated directive: ‘%error-verbose’, use ‘%define parse.error verbose’ [-Wdeprecated]
>    84 | %error-verbose
>       | ^~~~~~~~~~~~~~
>       | %define parse.error verbose
> /home/daniel/Documents/repos/igraph/src/io/dl-parser.y: warning: 20 shift/reduce conflicts [-Wconflicts-sr]
> /home/daniel/Documents/repos/igraph/src/io/dl-parser.y: warning: 3 reduce/reduce conflicts [-Wconflicts-rr]
> /home/daniel/Documents/repos/igraph/src/io/dl-parser.y: warning: fix-its can be applied.  Rerun with option '--update'. [-Wother]
> [ 18%] [FLEX][dl_lexer] Building scanner with flex 2.6.4
> [ 18%] [BISON][gml_parser] Building parser with bison 3.5.1
> /home/daniel/Documents/repos/igraph/src/io/gml-parser.y:85.1-12: warning: deprecated directive: ‘%pure-parser’, use ‘%define api.pure’ [-Wdeprecated]
>    85 | %pure-parser
>       | ^~~~~~~~~~~~
>       | %define api.pure
> /home/daniel/Documents/repos/igraph/src/io/gml-parser.y:88.1-28: warning: deprecated directive: ‘%name-prefix="igraph_gml_yy"’, use ‘%define api.prefix {igraph_gml_yy}’ [-Wdeprecated]
>    88 | %name-prefix="igraph_gml_yy"
>       | ^~~~~~~~~~~~~~~~~~~~~~~~~~~~
>       | %define api.prefix {igraph_gml_yy}
> /home/daniel/Documents/repos/igraph/src/io/gml-parser.y:91.1-14: warning: deprecated directive: ‘%error-verbose’, use ‘%define parse.error verbose’ [-Wdeprecated]
>    91 | %error-verbose
>       | ^~~~~~~~~~~~~~
>       | %define parse.error verbose
> /home/daniel/Documents/repos/igraph/src/io/gml-parser.y: warning: fix-its can be applied.  Rerun with option '--update'. [-Wother]
> [ 18%] [FLEX][gml_lexer] Building scanner with flex 2.6.4
> [ 18%] [BISON][lgl_parser] Building parser with bison 3.5.1
> /home/daniel/Documents/repos/igraph/src/io/lgl-parser.y:73.1-12: warning: deprecated directive: ‘%pure-parser’, use ‘%define api.pure’ [-Wdeprecated]
>    73 | %pure-parser
>       | ^~~~~~~~~~~~
>       | %define api.pure
> /home/daniel/Documents/repos/igraph/src/io/lgl-parser.y:76.1-28: warning: deprecated directive: ‘%name-prefix="igraph_lgl_yy"’, use ‘%define api.prefix {igraph_lgl_yy}’ [-Wdeprecated]
>    76 | %name-prefix="igraph_lgl_yy"
>       | ^~~~~~~~~~~~~~~~~~~~~~~~~~~~
>       | %define api.prefix {igraph_lgl_yy}
> /home/daniel/Documents/repos/igraph/src/io/lgl-parser.y:79.1-14: warning: deprecated directive: ‘%error-verbose’, use ‘%define parse.error verbose’ [-Wdeprecated]
>    79 | %error-verbose
>       | ^~~~~~~~~~~~~~
>       | %define parse.error verbose
> /home/daniel/Documents/repos/igraph/src/io/lgl-parser.y: warning: fix-its can be applied.  Rerun with option '--update'. [-Wother]
> [ 18%] [FLEX][lgl_lexer] Building scanner with flex 2.6.4
> [ 18%] [BISON][ncol_parser] Building parser with bison 3.5.1
> /home/daniel/Documents/repos/igraph/src/io/ncol-parser.y:74.1-12: warning: deprecated directive: ‘%pure-parser’, use ‘%define api.pure’ [-Wdeprecated]
>    74 | %pure-parser
>       | ^~~~~~~~~~~~
>       | %define api.pure
> /home/daniel/Documents/repos/igraph/src/io/ncol-parser.y:77.1-29: warning: deprecated directive: ‘%name-prefix="igraph_ncol_yy"’, use ‘%define api.prefix {igraph_ncol_yy}’ [-Wdeprecated]
>    77 | %name-prefix="igraph_ncol_yy"
>       | ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
>       | %define api.prefix {igraph_ncol_yy}
> /home/daniel/Documents/repos/igraph/src/io/ncol-parser.y:80.1-14: warning: deprecated directive: ‘%error-verbose’, use ‘%define parse.error verbose’ [-Wdeprecated]
>    80 | %error-verbose
>       | ^~~~~~~~~~~~~~
>       | %define parse.error verbose
> /home/daniel/Documents/repos/igraph/src/io/ncol-parser.y: warning: fix-its can be applied.  Rerun with option '--update'. [-Wother]
> [ 18%] [FLEX][ncol_lexer] Building scanner with flex 2.6.4
> [ 18%] [BISON][pajek_parser] Building parser with bison 3.5.1
> /home/daniel/Documents/repos/igraph/src/io/pajek-parser.y:110.1-12: warning: deprecated directive: ‘%pure-parser’, use ‘%define api.pure’ [-Wdeprecated]
>   110 | %pure-parser
>       | ^~~~~~~~~~~~
>       | %define api.pure
> /home/daniel/Documents/repos/igraph/src/io/pajek-parser.y:113.1-30: warning: deprecated directive: ‘%name-prefix="igraph_pajek_yy"’, use ‘%define api.prefix {igraph_pajek_yy}’ [-Wdeprecated]
>   113 | %name-prefix="igraph_pajek_yy"
>       | ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
>       | %define api.prefix {igraph_pajek_yy}
> /home/daniel/Documents/repos/igraph/src/io/pajek-parser.y:116.1-14: warning: deprecated directive: ‘%error-verbose’, use ‘%define parse.error verbose’ [-Wdeprecated]
>   116 | %error-verbose
>       | ^~~~~~~~~~~~~~
>       | %define parse.error verbose
> /home/daniel/Documents/repos/igraph/src/io/pajek-parser.y: warning: 6 shift/reduce conflicts [-Wconflicts-sr]
> /home/daniel/Documents/repos/igraph/src/io/pajek-parser.y: warning: fix-its can be applied.  Rerun with option '--update'. [-Wother]
> Scanning dependencies of target igraph

I tried the suggested `--update`s, but that messed things up even more.

### LINKED ISSUE #1618: Barabasi aging game crashes on negative index from psumtree_search
**Describe the bug**
A specific barabasi_aging_game call will crash.

**To reproduce**
```
assert(igraph_barabasi_aging_game(
        &g, /*nodes*/ 5, /*m: edges_per_step*/ 2, 
        /*outseq: edges per step as vector*/ 0, /*outpref*/ 0,
        /*pa_exp*/ 1.0, /*aging_exp*/ 0.0, /*aging_bin*/ 1,
        /*zero_deg_appeal*/ 1.0, /*zero_age_appeal*/ 1.0, /*deg_coef*/ -1.0,
        /*age_coef */ 0.0, /*directed*/ 1) == IGRAPH_SUCCESS);
```
This crashes with a double free/corruption on 
`igraph_vector_destroy(&degree);`
or if I remove that, it crashes on the igraph_create, because of negative indices:
If I print the edges I get this:
`1 0 1 0 2 1 2 1 3 2 3 2 4 -1 4 -1`
which is because
`igraph_psumtree_search(&sumtree, &to, RNG_UNIF(0, sum));`
sets values of `-1` to `to`.

This does not happen with less than 5 nodes, only 1 edge per step, or a deg_coef of 1.0  (if I keep the rest the same). The negative deg_coef is maybe a little weird, but I don't think it should be impossible.

**Version information**
a0d5abe7c84155c4d1a3885589037ed2470386a5, the current develop

https://github.com/igraph/igraph/blob/a0d5abe7c84155c4d1a3885589037ed2470386a5/src/games/barabasi.c#L602

### LINKED ISSUE #1628: Edge betweenness test causes internal error in MSVC
**Describe the bug**
Compiling the test with MSVC in Release mode crashes the compiler with fatal error C1001:

https://docs.microsoft.com/en-us/cpp/error-messages/compiler-errors-1/fatal-error-c1001?view=msvc-160

**To reproduce**

```
cmake ..
cmake --build . --config Release
cmake --build . --target test_igraph_edge_betweenness --config Release --verbose
```

results in

```
    igraph_edge_betweenness.c
1>C:\...\igraph\tests\unit\igraph_edge_betweenness.c(105): fatal error C1001: Internal co
  mpiler error. [C:\Users\daetwyler\source\repos\igraph\build\tests\test_igraph_edge_betweenness.vcxproj]
```

Line 105 of this file is actually just `int main() {` so I'm really stumped on this.

**Version information**

`cl /?` says version 19.28.29334. VS version is 16.8.2. UPDATE: Also happens on 16.8.4, which is the latest version as of writing.

Current `develop` branch.

### LINKED ISSUE #1634: Several tests fail on Windows because of MSVC printf bug

### LINKED ISSUE #1650: The packaged version cannot be built because of missing documentation source files
**Describe the bug**
The packaged version ("`make dist`") cannot be configured / built because of missing documentation source files.

**To reproduce**

For completeness, I'm doing this on macOS with the Ninja generator.

In `build`,

```
cmake ..
cmake --build . --target dist
tar xf igraph-0.9.0.tar.gz 
cd igraph-0.9.0/
mkdir build && cd build
cmake ..
```

There are a bunch of errors about missing XML files such as

```
CMake Error: File /Users/szhorvat/Repos/igraph-main/igraph/build/igraph-0.9.0/doc/fdl.xml does not exist.
CMake Error at doc/CMakeLists.txt:130 (configure_file):
  configure_file Problem configuring file
```

These seem not to be included in the package.

GroteGnoom:
- efficiency: implement global/local efficiency and weighted average shortest paths
- average_path_length: change handling of disconnected graph
- diameter & diameter_dijkstra: move from structural_properties.c to shortest_paths.c
- shortest_paths.c: make internal functions static
- shortest_paths.c: change free() to igraph_free()
- change some instances of free() to igraph_free() or igraph_Free()
- local_efficiency: fix enighbourhood calculation in directed graph
- average_path_length: update test for API change
- local_efficiency: add mode argument
- local_efficiency: change vector_int to vector_char for memory efficiency
- local_efficiency: support computing only for a subset of vertices
- functions.def: added argument for average_path_length.
- version: added NEXT_VERSION to indicate develop branch.
- Appveyor: remove zip from installed packages as it seems to conflict with xz
- okay, this didn't work, let's try to upgrade xz instead
- silencing some warnings in the Travis web UI about the current configuration
- fix igraph_adjacency_spectral_embedding() test case
- ignore numeric inaccuracy in examples/simple/scg2.c test case
- fixed igraph_i_gml_tostring() that accidentally returned the address of a local variable
- fixed igraph_layout_merge2() test case; MDS layout was unstable so we use a circle now
- SCG: canonicalize the calculated groupings to make the test cases deterministic
- more SCG test case fixes: canonicalize eigenvectors as well
- watch out for negative zeros
- canonicalize the output of the Infomap algorithm as well
- Infomap test: don't check the exact partition for the Wiktionary example, it is platform-dependent
- relaxed test cases for HRG model; results are not consistent on i386
- provide a simpler case for LU solver singularity detection test case
- retabbing some test utilities and examples
- scg: relax testing a bit to get rid of numeric accuracy issues on i386
- removed and ignoring m4/libtool.m4
- implemented custom printing for matrices in test cases to avoid -0
- MSVC compilation improvements
- MSVC fixes: don't provide expm1() if we already have one
- tools/getversion.sh: infer version number from changelog if not in a Git repo
- include COPYING and all Markdown files from root in MSVC-specific ZIP, closes #1378
- initialization in C++ header is not strictly C++98 compliant
- support maxcompno in igraph_i_decompose_strong()
- don't mess with the "register" keyword; some compilers throw a warning about it
- remove profanity from one of the source files
- remove a few unused variables
- PRPACK: semicolon after namespace is not supported officially in C++98
- make prpack_igraph_graph.h independent of igraph_interface.h
- Appveyor: install zstd so we can install .zst-compressed packages
- remove trailing comma in enum (not compliant with targeted C version)
- fix a few signedness-related warnings
- ignore all build folders if we have more than one
- add delcaration of igraph_i_binsearch_slice() in community_leiden.c to avoid a compiler warning
- comment out some unused functions
- first iteration of CMake build; work in progress
- removed conflicting igraph_vector_print() from two tests
- added lots of unit tests
- fixed most of the failing unit tests
- all tests from the igraph test suite are now supported
- use libxml2 when installed
- added GLPK and GMP support; get version number from git
- tests are not built until 'make check' is invoked
- topological_sorting: raise error if the graph is not acyclic
- fix a compoiler error in MSVC when compiling for Python 2.7
- added Python 3.8 to Appveyor CI config
- "make install" now installs the library and the headers
- follow up moving topological sorting test case to a new folder
- added generated pkgconfig file to CMake build
- community_multilevel: added resolution parameter
- Update changelog
- community_leiden: remove obsolete declaration
- remove Python 3.8 from Appveyor config, it is not in the image that we are using
- document how inclist and lazy_inclist handles loop edges, refs #1396
- fix emphasis tags in documentation
- igraph_modularity: update argument checks and messages
- IGRAPH_FINALLY: check that argument types match
- IGRAPH_STRVECTOR_INIT_FINALLY: remove unneeded cast
- igraph_trie_init: fix IGRAPH_FINALLY
- modularity: implemented support for directed graphs.
- tls1.c: fix types in threaded error handling test
- changed CMake linker specification of the igraph target; most of our dependencies are now PUBLIC
- added new modularity test to CMake test suite
- fixed a few more incorrect invocations of IGRAPH_FINALLY()
- add comment explaining the type-checking trick in IGRAPH_FINALLY
- Update CHANGELOG
- modularity: fix broken doc build and polish docs
- IGRAPH_VECTOR_CHAR_INIT_FINALLY
- igraph_closeness: use negative cutoff to indicate no cutoff, for consistency.
- edge_betweenness: corrected cutoff.
- Revert "betweenness_estimate: interpret 0 as infinity in igraph 0.8"
- CHANGELOG: mention change in cutoff interpretation in betweeness & closeness
- Revert "betweenness: update cutoff=0 tests for 0.8"
- added support for FINALLY stack size checking
- Update CHANGELOG.md
- BLAS, LAPACK and ARPACK can now use the vendored copy
- CXSparse can now use the vendored copy
- GLPK can also be vendored
- trying to set up Travis CI
- use official newer CMake from Kitware as we need >=3.16
- Travis: official CMake is installed in /usr/local/bin so use that
- debugging why CMake is not found on Travis
- Travis: looks like the official CMake gets installed into /usr/bin/cmake
- CMake build fixes for Travis
- remove unused variable for cs_qr.c
- typo fix in etc/cmake/dependencies.cmake that prevented the usage of external libraries
- fix a gcc warning in cliquer/set.h
- fix deprecated directive in yacc files
- fix more compilation warnings in the Travis build
- undo accidentally removed variable removal in src/lad.c
- okay, finally understood what the problem is :)
- remove another warning
- do not remove the equals sign from behind name-prefix in yacc files even if it causes warnings on Linux
- const static must be static const instead
- do not build vendored libs if we are using an external one
- one more warning fix
- add -Wno-sign-compare to excluded warnings list because flex/bison does not generate fully compliant source files
- prevent implicit fallthrough in Pajek escape function
- fortran_intrinsics.c is actually needed for LAPACK, not igraph
- try a newer macOS image in Travis to see if it has CMake >= 3.16
- more Travis-related fixes
- fixed vendored CXSparse compilation problems
- ouch, it is GMP_LIBRARY, not GMP_LIBRARIES
- xerbla is part of BLAS, not LAPACK (it is needed by dgemv, which is in BLAS)
- len_trim is also needed for BLAS
- Removed treating errors as warning, removed m dependency for f2c, defined NCOMPLEX for cs.
- Include NCOMPLEX also for igraph. Define MSDOS and proper include dir for f2c.
- vendored cxsparse: simplify defining NCOMPLEX
- vendored GLPK: fix include path
- fix typo
- vendored cxsparse: fix include path
- Corrected GLPK include directory.
- Some corrections for compilation on Windows.
- fixed CXSparse include dirs reference
- replace strcasecmp() with _stricmp() if the former is missing
- install Git in Travis CI because CMake needs it to derive the version number
- trying to debug why the version number breaks from Travis
- Travis: more Git debugging
- modernizing doxrox.py, it is now Python 3 compatible
- DocBook XML files now built with CMake as well
- removed unused sitemap generator script
- also copy static DocBook XMLs from doc/
- Fixed strcasecmp problem.
- Start modernizing spanning_trees.c
- no more igraph_edge
- Move IGRAPH_TO to IGRAPH_OTHER
- vector_size outside of loops
- Few fixes after CI fail
- Forgot one bugfix
- A few more bugfixes from CI
- Start eroding igraph_edge from other files
- Substitute igraph_edge and more consistency with igraph_integer_t
- Some typos and unused vars
- Typo
- Typo
- Revert two lines in conversion.c: not equivalent to igraph_edge?
- Update 2 more files
- Move vector size out of loop from structural properties
- Extract vector_size from for loops for 2 more files
- igraph_edge uses IGRAPH_FROM and IGRAPH_TO
- Undo some changes since IGRAPH_FROM/TO and edge are not the same
- removed explicit igraph_integer_t casts from IGRAPH_EDGE()
- Corrected expm1 check, not testing with pthreads, defining math constants.
- Appveyor: rewriting config to run with and without MSVC, irrespectively of Python version (#1420)
- from_prufer: update test and ensure use of standard C
- from_prufer: update test and ensure use of standard C
- Changed location of test, using IGRAPH_STATIC when testing for MSVC.
- Support for dynamic library on Windows.
- clarifications in CMakeFiles.txt comments regarding static / shared compilation on Windows
- use universal line endings in stimulus.py
- added Alpine Linux test run to Travis CI (#1419)
- Gcc warnings (#1431)
- LAD: convert some raw array allocations to use ALLOC_ARRAY
- Doc improvs (#1433)
- first version where the HTML documentation build works (with no cross-references yet)
- generated HTML documentation is now cross-referenced
- added PDF documentation generation; examples not stripped from the output yet
- show which formats of the documentation will be built in the summary
- re-added disabled targets to Travis build
- Travis: fix CMake path
- make it clear that doxrox.py now needs Python 3.x
- Specify python2.7 in shebang for doxrox.py docs parser
- Add new team members to manual header in docs
- Write example for igraph_full (#1434)
- Document and add to docs: Even-Tarjan reduction (#1440)
- Add functions for Eulerian cycles/paths
- examples now stripped from the generated PDF docs
- fixed testing for the presence of certain tools needed for building the documentation
- eulerian: fix XML in doc comments
- Update gitignore
- first version of updated Appveyor file; will probably fail
- Appveyor: install winflexbison, fix directory creation
- Appveyor: use == instead of EQ
- Appveyor: okay, let's try an MSVC build only first
- added \deprecated, \deprecated-by and \experimental tags, closes #1444
- remove accidentally included file
- tests: add missing VERIFY_FINALLY_STACK()
- tests: ensure C89 standards compliance
- Add is_graphical, is_bigraphical and non-simple graph support for realize_degree_sequence
- realize_degree_sequence: update copyright notice
- trying to fix CTest invocation
- Appveyor: update keyring
- fixed test file paths
- added missing graphicality.c to source files
- add trailing newline to make clang happy
- added cycles.xxml to doc sources
- Appveyor: run tests via cmake, not via ctest (ctest does not build them)
- Appveyor: okay, also run ALL_BUILD
- Appveyor: another try with a different target
- Appveyor: ctest invocation needs build type with MSVC
- Appveyor: yet another try
- Appveyor: temporarily disabling a problematic test
- Appveyor: temporarily disabling another problematic test
- Appveyor: temporarily disabling two more problematic tests
- Appveyor: one more problematic test
- Appveyor: weeding out ARPACK-related tests to see whether at least the others pass on Appveyor
- Appveyor: weeding out ARPACK-related tests to see whether at least the others pass on Appveyor
- trying to update our LAPACK / ARPACK patches to fit the latest version
- LAPACK C source regenerated from LAPACK 3.5, fixed IGRAPH_F77_SAVE decl of some timing variables
- fix uninitialized variables in translated f2c code
- Appveyor: re-added disabled tests after LAPACK update
- realize_degree_sequence: add references to docs
- eulerian: match argument names between declaration & definition
- igraph_free: change return type to void. This is also consistent with the defintion of igraph_finally_func_t.
- DensityGrid: catch errors by reference.
- foreign.c: changed destroy call signatures to return void.
- removed debug message from GetGitRevisionDescription.cmake
- Building shared library on Windows (#1467)
- expose the BUILD_SHARED_LIBS option to the ccmake UI
- test runner now creates the diff between the expected and the observed output
- Documentation: add ID to Efficiency Measures section
- added new Kamada-Kawai-related test to test suite
- removed deprecated igraph_pagerank_old()
- there is no igraph_arpack_options_destroy()
- 'endmark' should always be a double in igraph_vector_init_real_end()
- addew Reingold-Tilford bug test
- igraph_diameter: return IGRAPH_INFINITY for non-connected graphs; change return type to igraph_real_t
- remove unused 'framework' Makefile target and igraph_Info.plist.in
- started CPack configuration for source tarball [WIP] [ci skip]
- make dist now generates a proper CMake-based tarball [ci skip]
- remove  deprecated igraph_pagerank_power_options_t from documentation [ci skip]
- igraph_local_efficiency: correct documentation
- igraph_average_local_efficiency: return 0 for graph with fewer than 3 vertices
- efficiency: add tests
- igraph_diameter: uptade signature in functions.def
- fix failing igraph_pagerank() test case on macOS
- shortest_paths.c: update copyright notice [skip ci]
- efficiency tests: improve robustness
- IGRAPH_FINALLY: more robust formulation, eliminate some IDE/compiler warnings
- structural_properties.c: make functions static where appropriate
- Change remaining uses of free() to igraph_free() where necessary
- structural_properties.c: change //-comments to /*-comments
- structural_properties: fix typo in comment
- added two missing tests and a new source file
- efficiency tests: add missing VERIFY_FINALLY_STACK
- switched to using object libraries where possible to make sure that the user only needs libigraph.a
- add include paths of vendored object libraries when needed
- remove unneeded MSVC project files [ci skip]
- add msvc/include to include path on Windows
- add msvc/include to include path on Windows even for tests
- make sure that cs.h-related test cases compile on Windows
- pkgconfig files should go in lib/pkgconfig because they are not arch-independent
- re-add INTERNAL_BLAS, INTERNAL_LAPACK, INTERNAL_ARPACK to config.h
- Added missing DECLDIR.
- Added back extern.
- CMake: enable graphml test. igraph_error.h: export warning_handler symbols.
- CMake: Changed declaration of PIC.
- CMake: copying files for testing when building shared lib on MSVC.
- Made dependency exclusion more robust. Only copy when building shared libraries on Windows.
- CMake: correcting error with RPATH for testing when dtags are enabled.
- CMake: Simplified finding DLL by simply adding to path.
- Removed unneccessary include files for MSVC.
- Make igraph_sparsemat2 an internal test.
- Incorrectly removed skip test.
- FC does not accept paths with forward slash.
- Revert "CMake: Changed declaration of PIC."
- Only add library dir to tests when building shared lib using MSVC.
- Add back /A for FC.
- Further fixes for building with external blas, internal lapack
- fix invalid reference to REMOVE_ITEM, closes #1489
- Update CHANGELOG
- version number now uses NEXT_VERSION and the Git commit hash, closes #1493
- added support for link-time optimization
- cmake: always define MSDOS for f2c on Windows
- Use realloc from CXSparse (#1500)
- fix: don't blow up during the build if Git is not installed, closes #1503
- fix: fix fatal error message in LTO check
- Support for Intel compiler
- cmake: properly quote message
- CMake add path for test on MinGW (#1504)
- fix: more descriptive error message when IGRAPH_VERSION is missing and Git is not installed, refs #1503
- feat: added support for thread-local storage build-time switch
- fix: fix IGRAPH_THREAD_SAFE define in generated include/igraph_threading.h
- fix: don't blow up on NOTFOUND entries in include paths and library lists, closes #1505
- fix: IGRAPH_THREAD_SAFE is a macro that must be 0 or 1
- fix: TLS tests on Linux need pthreads
- fix: eliminate some flex warnings about catch-all rules, refs #1506
- cmake: update some compiler-specific options
- cmake: Intel compiler should not print remarks
- Properly fix path problems with CTest on MinGW (#1508)
- fix: IGRAPH_THREAD_SAFE is now parsed properly when making the documentation
- cmake: Condition warning options on C_COMPILER_ID (#1509)
- to_directed: add RANDOM and ACYCLIC modes (#1511)
- Update CHANGELOG [ci skip]
- Correctly compiling vendored libraries as OBJECT libraries on MinGW/MSVC (#1512)
- CMake: fix compiling and testing static library on MinGW. (#1514)
- to_directed: add to ctest
- CPack: incorrect directory for config.h.in (#1518)
- first attempt at creating a proper pkg-config file even with external dependencies
- wrap ARPACK_LIBRARIES, BLAS_LIBRARIES and LAPACK_LIBRARIES in a generator expression for sake of consistency
- CMake: Re-enabled some internal tests. (#1529)
- tests: update comment [skip ci]
- ci: use TRAVIS_COMMIT to derive the version number in Travis if the latest tag is too far away in the past
- CMake: create parsers/lexers before packaging. (#1519)
- use __IGRAPH_FILE_BASENAME__ instead of __FILE__ when compiling to avoid leaking full paths
- CMake: MSVC warning C4273: inconsistent DLL linkage (#1531)
- added a few more configuration-time checks for functions [skip ci]
- be more forgiving with the dev branch checked out without Git when deriving the version number, needed for tox in python-igraph [skip ci]
- fix a typo in version.cmake [skip ci]
- renamed __IGRAPH_FILE_BASENAME__ to IGRAPH_FILE_BASENAME (double-underscore stuff is reserved)
- added AUTO for dependency-related options
- fix: IGRAPH_GLPK_SUPPORT need not be a tri-state variable [skip ci]
- fix: IGRAPH_GRAPHML_SUPPORT cannot be set from AUTO to a new value with a generator expression
- cmake: reorder features in summary output
- Travis: enable ccache
- Travis: parallelize ctest
- 'make dist' now creates parsers/lexers if needed before creating the tarball
- Update Travis configuration
- Travis: re-enable make flags
- Remove unneeded parts of the vendored SuiteSparse
- cmake: include(compilers) only at top level to avoid duplication
- CMake: fix building without GMP
- Travis: simplify choosing the generator
- Add igraph_eulerian.c and associated tests
- f2c_dummy.c is no longer needed with cmake build
- igraph_girth: clarify documentation [skip ci]
- CMake: explanatory comments for tests [skip ci]
- added new CMake script that was accidentally excluded from the previous commit
- AUTO switches are now translated to ON or OFF after evaluating dependencies, refs #1537
- fix: add -lstdc++ unconditionally to generated igraph.pc file, refs #1537 [ci skip]
- fix: handle standard C++ library differently on macOS, refs #1537 [ci skip]
- igraph_diameter: add special case for null graph (#1540)
- CMake: fix MSVC warning C4100: unreferenced formal parameter (#1532)
- tests: correct argument list of igraph_layout_merge()
- Correct layout_mds test (#1545)
- Correct Release compilation of test on MSVC (#1544)
- use proper relative paths in IGRAPH_FILE_BASENAME, refs #1524
- simplified absolute path calculation, thanks @vtraag, refs #1524
- CMake: Fix incorrect DLL linkage of some functions in f2c (#1536)
- cmake: disable uninteresting MSVC warnings (#1546)
- fix: don't add -lm to generated igraph.pc on Windows
- Update structural_properties.c
- fix: resolve a few MSVC warnings in the Bliss code
- fix: resolve a few more MSVC warnings
- fix: fix more MSVC compiler warnings
- remove obsolete autoconfig-related stuff
- chore: renamed igraph.pc.cmake.in to igraph.pc.in now that we don't have autoconf any more
- chore: removed some more unused files [ci skip]
- removed Java interface, migrated to igraph/igraph-java
- re-added .pyc to .gitignore [ci skip]
- stimulus.py: don't stop in shell generator when an unknown type is encountered
- fix more MSVC warnings
- travis: update configuration
- travis: enable basic optimization for address sanitizer build
- travis: show test output diff upon failure
- travis: correct multiline string
- fix: don't add -lstdc++ to pkgconfig file on Win32 [skip ci]
- ci: fixing WIN32 vs MSVC differences that I am not familiar with :\)
- cmake files: convert most indentation to spaces
- travis: add ccache to Alpine Linux build
- appveyor: attempt enabling gmp support
- appveyor: update vcpkg
- appveyor: add missing gmp dependencies
- fix: fix another MSVC compiler warning
- CTest output now shows the diff when the test fails, closes #1535
- disable some more MSVC warnings to reduce clutter
- appveyor: update configuration
- tests: fix igraph_scg_grouping3
- Revert "tests: fix igraph_scg_grouping3"
- fix: attempt to provide a replacement for 'cmake -E cat'
- cmake: quote generator expressions which may contain lists
- build: prevent in-tree builds
- appveyor: correct caching and reduce msbuild verbosity
- cmake: use if()s instead of generator expressions for link libraries
- Disable "potentially uninitialized local variable" MSVC warning
- Use Mini-GMP when GMP is not available (#1549)
- added license header and missing config.h include to igraph_gmp.h
- added code coverage switch
- appveyor: build both static and dynamic libraries
- plfit: use igraph's include directory when compiling
- fix: get rid of more warnings with MSVC
- chore: remove autotools test specifications
- fix: replace /Wd with /wd in MSVC argument list
- fix: busting some more warnings with MSVC
- travis: test building shared lib
- ensure that position independent code ig generated when building a shared library
- remove use of FILE from bliss, fixes compilation on RPi
- fix: use check_symbol() instead of check_function()
- check_symbol_exists requires -lm on Linux
- travis: cache snaps
- remove unused fabsl() replacement
- Revert "travis: cache snaps"
- remove HAVE_FABSL
- build: Appveyor config file incorrectly used USE_INTERNAL_GMP instead of IGRAPH_USE_INTERNAL_GMP
- fix: yet another MSVC warning fix
- fix: do not run tls2 test with external ARPACK as it is not guaranteed to be thread-safe
- add test for igraph_bliss_automorphisms
- Implement fatal error handling (#1548)
- chore: CMake code reorganization, moved feature toggles to a separate module
- [develop/cmake] Github Workflow for MSYS2's MINGW32/MINGW64 (#1559)
- define IGRAPH_STATIC when compiling plfit on Windows since plfit calls igraph functions now
- github actions: -j2 parallel build
- move plfit compile definitions into the file where it belongs
- fix: trying to fix printing of test diff on Windows
- finished merge of Github workflow, sorry
- No need to test for __CYGWIN__ when defining DECLDIR since recent CygWin never defined _WIN32
- appveyor: do not upgrade vcpkg since we no longer build with GMP
- YY_FATAL_ERROR in lexers now uses IGRAPH_FATAL
- Replace assert() by IGRAPH_ASSERT() (#1561)
- Re-add missing tests
- documentation: typos
- vector_ptr: clean up formatting, remove obsolete TODO comment
- renamed build-autoconf.yml to build-cmake.yml
- Remove functions that were deprecated in 0.6 (#1562)
- fix: fix memory leak in examples/simple/igraph_hrg3.c
- fix: fix memory leak in examples/tests/igraph_eulerian_path.c
- fix: fix memory leak in examples/tests/maximal_cliques_callback.c
- fix: fix memory leak in examples/tests/simplify_and_colorize.c
- fix: fix failure artifact paths in build-cmake.yml
- doc: remove reference to igraph_subgraph() from comments
- Remove doc reference to igraph_es_adj.c test
- fix: wrong ordering of cleanup in examples/tests/maximal_cliques_callback.c
- Add Azure Pipelines (#1564)
- layout_davidson_harel: make interruptible
- layout_gem: make interruptible
- igraph_is_connected() now returns false for the null graph, closes #1539
- fix: removed Makefile.in that accidentally crept back from the master branch
- refactor: no need for special-casing the null graph in igraph_girth() [ci skip]
- fix: trying to fix the igraph_scg_grouping3 test case with -fpmath=387
- fix: make FindGMP.cmake a better CMake citizen so we don't link to GMP unnecessarily
- fix: fix the fix of the igraph_scg_grouping3 test
- docs: improve scg docs
- azure: compile with OpenBLAS
- fix: igraph_scg_semiprojectors2 test with x87 math
- fix: make igraph_scg_grouping3 and igraph_scg_semiprojectors2 tests less sensitive to numerical instability (#1568)
- ci: increase timeout for Infomap test
- refactor: first round; moved core data types to src/core/ and math-related stuff to src/math/
- refactor: moved layout algorithms to src/layout
- refactor: moved graph reading/writing to src/io
- refactor: moved community detection to src/community
- refactor: moved interruption to src/core
- refactor: slowly started dismantling igraph_types_internal.h
- refactor: igraph_types_internal.h is no more
- refactor: fixed some tests, moved spinglass and infomap clustering to src/community
- ci: let's trigger Azure for all branches
- fix: fixed doc build after creating subdirectories in src/
- created src/constructors/, src/games/, src/cliques/ and src/topology/
- created src/centrality/
- HRG stuff separated to src/hrg/
- graph data structure and closely related stuff moved to src/graph/
- some more files moved to core/ and topology/
- fixed CMakeLists.txt for the files that I've moved around in the last commit
- more reorganization, getting closer to the finish line now
- src/ folder is almost clean now
- benchmarks moved from examples/ to tests/
- examples/tests/ moved to tests/unit/
- final cleanup
- fixed documentation generation
- rename "topology" directory to "isomorphism"
- rename gmp.h to gmp_internal.h
- move glet.c to cliques
- rename "decomposition" directory to connectivity and move separators there
- rename cliquer.h and prpack.h to cliquer_internal.h and prpack_internal.h
- Remove unused igraph_f2c.h
- MinGW: build with internal BLAS/LAPACK
- MinGW tests: switch to NetLib BLAS/LAPACK and use ARPACK
- betweenness: Use ‘double’ instead of 'long long' to count paths. This avoids overflow.
- betweenness: move tests to tests/unit
- betweenness: amend tests
- betweenness: remove bigint support, remove 'nobigint' argument
- remove bignum / igraph_biguint functions; in the future we will use GMP
- mingw: use 3 minute default timeout in ctest
- mingw ci: remove arpack
- refactor: moved tests that were not referenced from the docs from examples/simple to tests/unit, refs #1574
- refactor: moved test fixtures as well
- refactor: moved more test fixtures, tests should pass now
- refactor: move test output files as well
- refactor: rename test 'topology' to 'isoclasses'
- tests: add VERIFY_FINALLY_STACK and general cleanup
- tests: simplify functions in test_utilities and makes them more suitable for printing large values
- tests: add final missing VERIFY_FINALLY_STACKs and further test_utilities improvements
- tests: make sure that igraph_free() is used where necessary
- tests: ensure that examples/simple/igraph_community_leiden is run
- IGRAPH_ASSERT: amend documentation
- Update CHANGELOG
- refactor: split centrality_other.c, step 1
- refactor: split centrality_other.c, step 2
- examples: complete igraph_star example
- tests: move igraph_ring test to test/unit
- examples: add igraph_ring example
- example: fix copyright notice for igraph_ring and igraph_star [skip ci]
- examples: amend igraph_full and include it in tests
- example: amend igraph_community_label_propagation and include it in tests
- examples: amend igraph_topological_sorting an include in tests
- examples: clean up some example programs
- union and disjoint_union: clarify documentation
- docs: fixes to front page, add 2021 to copyright
- tests: move some "examples" which used internal functions to tests/unit
- tests: add VERIFY_FINALLY_STACK to just-moved tests
- Correct some instances of igraph_vector_min being potentially called on empty vectors.
- betweenness and closeness: ensure that vector_min is not called on empty vectors
- Implemented harmonic centrality
- closeness: replace closeness_estimate with closeness_cutoff
- splitting layout algorithms into separate files, step 1
- splitting layout algorithms into separate files, step 2
- splitting layout algorithms into separate files, step 3
- splitting layout algorithms into separate files, step 4
- re-added a missing include for Windows
- more fixed for MS Windows where apparently we need igraph_math.h for M_PI
- splitting graph IO routines into separate files, step 1
- splitting graph IO routines into separate files, step 2
- strcasecmp needs internal/hacks.h on Windows
- splitting community detection routines into separate files, step 1
- splitting community detection routines into separate files, step 2
- splitting isomorphism files
- splitting operators, part 1
- splitting operators, part 2
- Added fuzzer and OSS-fuzz build script (#1588)
- refactor: moved igraph_connect_neighborhood to operators from constructors
- splitting constructors into separate files, part 1
- fix: added missing end tag to new section in doc/operators.xxml
- splitting constructors into separate files, part 2
- splitting constructors into separate files, part 3
- Add support for AFL++ (#1589)
- moving shortest path related functions from properties/ to paths/, step 1
- moving shortest path related functions from properties/ to paths/, step 2
- moving rewiring from properties/ to operators/, part 1
- moving rewiring from properties/ to operators/, part 2
- moved some more operators from properties_other.c to operatprs/
- splitting degree-related structural properties to a separate file, step 1
- splitting degree-related structural properties to a separate file, step 2
- moved igraph_path_length_histogram to paths/
- fix: fix fuzzer so it does not attempt to destroy the graph when igraph returned an error code
- GML parser should clean up after itself if an error happens during parsing
- fix: fixed a use-after-free issue in igraph_read_graph_gml()
- refactor: moved subgraph creation to operators/
- betweenness: replace betweenness_estimate with betweenness_cutoff
- betweenness: replace edge_betweenness_estimate with edge_betweenness_cutoff
- functions.def: update betweenness/closeness:
- community/edge_betweenness.c: change comment references to _estimate functions to _cutoff ones
- centrality/betweenness.c: remove unneeded headers
- docs: typo in pagerank docs
- docs: cleanup for centrality docs.
- weighted closeness: remove unneeded comparison-with-tolerances
- Add unit test for st_mincut_value (#1594)
- Add unit test for st_mincut (#1595)
- Add st_edge_connectivity unit test (#1596)
- Add edge_disjoint_paths unit test (#1597)
- Add 4 max-flow unit tests, fix residual_graph bug (#1598)
- fix: fix a double-free problem in igraph_read_graph_gml
- test: more GML test cases
- st_mincut_value test: add missing VERIFY_FINALLY_STACK
- doc: examples are now not collapsed when the user clicks in the body
- fix: CSS/JS/image assets are now copied to build/doc/html when the HTML doc is built
- chore: fix tabs vs spaces in doc/CMakeLists.txt [ci skip]
- fix: fix assertion in igraph_st_mincut_value.c
- added VERIFY_FINALLY_STACK() to some more test cases
- fixed a memory leak in GML parser
- fix: fix failing test case on macOS
- refactor: splitting DAG-related functions, step 1
- refactor: splitting DAG-related functions, step 2
- refactor: splitting multiplicity-related functions, step 1
- refactor: splitting multiplicity-related functions, step 2
- refactor: splitting multiplicity-related functions, step 3
- refactor: splitting tree-related functions, step 1
- refactor: moved igraph_add_edge() to operators/
- refactor: moved igraph_girth() to a separate file
- refactor: separating neighborhood-related functions, step 1
- refactor: separating neighborhood-related functions, step 2
- Add igraph_write_graph_lgl example (#1601)
- Update CHANGELOG
- erdos_renyi_game: re-write tests, fixes #1606
- tests: move erdos_renyi_game test to tests/unit and make it deterministic
- erdos_renyi_game: add new example
- refactor: properties now broken down into submodules
- fix: about_structural section of docs was accidentally defined twice
- fix: fixed include of igraph_convergence_degree header
- Add function igraph_is_same_graph for tests (#1604)
- fix the author of the Vigna-Boldi paper
- Add growing_random_game test, fix docs (#1607)
- Update CHANGELOG [skip ci]
- erdos_renyi_game: consistent parameter naming between definition and declaration [skip ci]
- Add write_graph_dimacs unit test (#1603)
- spinglass: update test and documentation
- spinglass: add VERIFY_FINALLY_STACK to test
- walktrap: set random seed in test
- Add test for random_spanning_tree
- Fix memory leak in spinglass community detection
- Fix more memory leaks in spinglass community detection
- spinglass: eliminate many clang-tidy warnings
- Fix yet another memory leak in spinglass
- spinglass: ensure no memory leak when functions is interrupted
- Update CHANGELOG [skip ci]
- spinglass: make sure that large data is heap-allocated to avoid stackoverflow
- Remove nonlinear_barabasi_game declaration
- Update CHANGELOG.
- rewiring: expand tests and add internal header for igraph_i_rewire
- Add full_citation unit test
- update copyright full_citation unit test
- igraph_rewire: fix test
- linegraph: add unit tests, clarify docs (#1611)
- Add kautz unit test, add to docs (#1612)
- Add test for igraph_tree_game
- Finish igraph_complex tests
- fix: fix license header in games_other.c before starting to split it
- refactor: separating games_other.c into multiple files, step 1
- refactor: separated games into files, step 2
- fix: get rid of a few duplicated function prototypes, refs #1521
- refactor: removed a few duplicated functions that I broke while separating games_other.c
- fix: fixed duplicate documentation chunk
- Add a canonical graph printer to test_utilities (#1613)
- degree_sequence: correct missing include.
- doc: added a warning in the documentation that declares that the default RNG is not thread-safe, closes #1590
- Add new extended chordal ring unit test, update docs (#1616)
- Re-add deprecated igraph_edge_betweenness_estimate to public headers
- Update Bliss to 0.74
- Require C++11
- bliss: MSVC compatibility
- bliss: expand tabs in source code
- bliss: add interruption support
- bliss: remove printing functions unneeded by igraph
- bliss: comment out remaining uses of printing functions to comply wit CRAN rules
- vector: correctly deal with NaN.
- paths: disallowing NaN weights.
- betweenness: disallowing NaN weights.
- closeness: disallowing NaN weights.
- vector: added NaN tests, improved tests.
- vector: corrected documentation.
- vector: corrected implementation of dealing with NaN.
- fix: make Clang stop complaining about tautological comparisons (used for NaN checks)
- doc: fixed documentation build
- vector: using isnan explicitly.
- Update Bliss mention in ACKNOWLEDGEMENTS.md
- Update CHANGELOG.md
- tests: change assert() to IGRAPH_ASSERT() because assert() is disabled when compiling with NDEBUG defined in Release mode
- tests: get rid of assert() in examples as well
- use_all_warnings() should come before disabling specific warnings
- bliss: change default splitting heuristic to FL
- bliss: update capitazliation of naming and references
- Add callaway traits unit tests
- bliss documentation: add missing closing tag
- tests: fatal error handler
- tests: replace more uses of assert() by IGRAPH_ASSERT()
- build: set default build type to Release
- games: normalize declaration formatting
- Add new rewire directed edges unit test (#1622)
- Default error/warning handler style now consistent and compatible with sentence-style messages. Also reformat code for consistency.
- bliss: update to 0.75
- refactor: moved regression tests for bugs to tests/regression/
- fix: igraph_psumtree_update() should disallow negative values and NaN, refs #1618
- fix: igraph_psumtree_update() calls are now wrapped in IGRAPH_CHECK everywhere, fixes #1618
- fix: igraph_barabasi_aging_game() now requires deg_coef and age_coef to be non-negative
- build: added more excludes to the coverage configuration, fixes #1614
- bliss: final changes for the 0.75 update
- Add WARNINGF, ERRORF and FATALF macros
- formatting: added spaces (nitpicking)
- Add documentation of printf-style error macros
- bliss: match declarations with definitions
- make edge_betweenness test compatible with MSVC, fixes #1628
- Add lastcit_game unit test
- Fix segfault in lastcit_game on 0 nodes
- vertices -> edges in lastcit unit test comment
- Fix error messages. Also pagebins -> agebins.
- Also pagebins->agebins in the tests
- recieved -> received
- generalize the closeness centrality computation
- harmonic_centrality: fix progress reporting
- closeness and harmoic centrality: fix off-by-one error in cutoff
- closeness: standardize error messages
- add closeness tests, fixes #1208
- Update functions.def with new signature for closeness
- fix: added a few missing IGRAPH_CHECK() calls in scan.c
- fix: fix typo in a section ID in the documentation
- Add printf format specifier macro for igraph_integer_t: IGRAPH_PRId
- Update CHANGELOG.md
- Update CHANGELOG [skip ci]
- random.c: use igraph_finite and remove R-specific stuff
- bliss: initialize variables to shut up UBSan
- layout_reingold_tilford_circular: prevent division by zero on null graph
- test: use separate namespace for examples and unit tests
- centrality_other: renamed internal functions and made them static
- personalized_pagerank: better error checking; avoid modifying reset vector
- Update CHANGELOG [skip ci]
- added 'loops' and 'multiple' arguments to adjlist_init and inclist_init constructors
- test: simplified igraph_adjlist() example, moved the actual test to tests/unit
- pagerank: add missing IGRAPH_CHECKs
- pagerank: protect against potential exceptions thrown from PRPACK
- vector: corrected NaN for (which)_minmax.
- vector: simplified implementation of which_{min,max}
- vector: documented/improved NaN for various functions.
- vector: fix missing include for igraph_vector_is_nan.
- vector_is_nan: fix doc typo
- test: started adding unit test for igraph_adjlist_init()
- pagerank: remove unneeded conversion of ARPACK options
- pagerank: let the ARPACK method take into account self-loops, for consistency with PRPACK
- pagerank: with ARPACK method, result for edgeless graph should be the same as the 'reset' vector. Also check for NaN weights.
- Update CHANGELOG.md
- pagerank: make PRPACK method consistent with ARPACK on disconnected graphs
- fix: dots at end of URLs are now handled properly in doxrox.py
- pagerank: clarify documentation, remove dead link
- pagerank: fix edgless graph case in ARPACK method
- pagerank: update tests and examples
- pagerank: check value of damping factor
- pagerank: update copyright note of just-touched files
- Update CHANGELOG [skip ci]
- Update CHANGELOG [skip ci]
- test: tests finished for igraph_adjlist()
- refactor: lazy adj/inclists now return an igraph_vector_int_t, just like their non-lazy counterparts
- fix: fix out-of-bounds indexing in adjlist simplification
- Fix running tests with multi-config generators
- doc: added documentation for partial prefix sum trees
- Add compare_communities unit tests (#1645)
- Do not export non-public symbols from shared library
- Azure: build with TLS by default
- tests: exclude tls2 test when building a shared lib
- Fix IGRAPH_STATIC being incorrectly defined for the igraph target
- remove redudant transfer of prpack interface definitions to igraph target (they were already marked PUBLIC)
- CXSparse: define NCOMPLEX unconditionally (not just on Windows) and explain the reasons
- Clarify comments regarding visibility specifications in CMake files
- rename DECLDIR to IGRAPH_EXPORT
- export igraph_solve_lsap from shared library (add missing IGRAPH_EXPORT)
- bliss: remove unneeded 'static' from function (already have anonymous namespace)
- bliss: cosmetic cleanup
- Add split_join_distance unit tests
- Add tests for bipartite_game and is_bipartite
- Remove unimplemented igraph_layout_springs
- fix a mistake that I've made while merging the development branch
- Clean up includes in headers
- fix: igraph_inclist_init() loops argument now works
- private functions used in tests are now exported from the shared library
- igraph_sparsemat2 test is unconditionally safe to run when neither BLAS not ARPACK are vendored
- tests: remove BUILD_SHARED_LIBS
- remove _i_residual_graph and _i_reverse_residual_graph from public headers and make them static
- remove _i_spmatrix functions from public headers and comment out their definitions
- fix: igraph_i_sparsemat_view() is now public as igraph_sparsemat_view()
- fix: igraph_i_set_attribute_table() is now public as igraph_set_attribute_table()
- modularity: proofread docs
- move some SCG tests from examples to tests/unit
- refactor: cleaning up the code of igraph_neighbors() a bit
- fix: igraph_incident() now returns edges in the same order as igraph_neighbors()
- remove unimplemented igraph_vss_adj declaration
- fix: fix the inclist test case now that igraph_adjacent() and igraph_incident() produces consistent ordering
- mark deprecated functions in headers
- igraph_memory.h: remove commented obsolete part
- doc: added some more comments to igraph_incident() and igraph_neighbors()
- test: add tests for lazy adjlist and inclist
- remove private attribute handling functions from public headers
- lsap: replace igraph_warningf by IGRAPH_WARNINGF
- Extensive cleanup of header usages
- Update CHANGELOG
- modularity: minor cleanup
- lad: fix self-loops
- fix: added full stops to warning messages [ci skip]
- fix: don't use conditionals in adjlist/inclist initializations when the loops do not matter
- doc: updated changelog [ci skip]
- Add modularity_matrix unit tests
- Fix modularity matrix
- Fix typos, add test
- fix directed definition in docs, cosmetic fixes, iterate through matrix column-wise (storage order)
- doc: documented some additional changes performed in the adjlist-inclist branch
- is_chordal and maximum_cardinoality_search now support non-simple graphs
- subisomorphic_lad: make input argument const
- Clean up some tests
- feat: renamed igraph_i_bfs() to igraph_bfs_simple() and made it more user-friendly
- doc: make the igraph_bfs example look like an example, use a separate example for the callback-based version
- fix: removed unused function from examples/simple/cattributes4.c
- fix: CPack generation of source tarball now works again, fixes #1650
- Update CHANGELOG.md
- Update CHANGELOG.md
- chordality: remove restriction on directed graphs
- Update CHANGELOG.md
- some documentation proofreading
- Add le_community_to_membership unit tests (#1647)
- azure: use Ninja on Windows, add BUILD_TYPE support to templates
- work around MSVC printf bug, fixes #1634
- fix setting default build type to Release on Windows
- tests: add some missing VERIFY_FINALLY_STACK, see #1572
- convert some tabs to spaces
- Add selector unit tests, add docs (#1655)
- doc: include the HTML docs in the source tarball
- Add lapack_dgetrf unit tests (#1656)
- Add new barabasi aging game unit test  (#1629)
- build: added USE_CCACHE option to disable ccache if needed
- Clean up some error message, check for NaN weights in some functions
- add igraph_i_trans4_al_simplify to private header
- ci: update package registry before building documentation
- mark some functions as static
- doc: fix formatting of standalone HTML docs
- docs: standardize some brief descriptions
- Add layout_star unit tests (#1659)
- doc: removed unneeded shell scripts that were previously used by the igraph.org repo
- Add lapack_dgetrs unit tests (#1657)
- lapack_dgetrs: add const to ipiv argument
- correlated_pair_game: change int argument to igraph_integer_t
- grg: complete documentation.
- static_fitness_game: add const to arguments, clean up messages
- Add igraph_heap_ functions unit tests (#1658)
- Add layout_fruchterman_reingold unit tests


GroteGnoom:
Sorry, I tried using the github command line thing.

PR #1664, opened 2021-02-13
Add layout_fruchterman_reingold unit tests

GroteGnoom:
Part of #1592

This one is hard to test. I mainly put in some values to see of they
looked plausible. It also just seems very well guarded against memory
problems, which is great :)
It is easy to get into obviously suboptimal results by for example
turning the temperature up to ridiculously high values, but that seems
reasonable.

If there are suggestions for different kinds of tests I will of course
gladly add them.

----
The tests are failing because seeding didn't solve all the randomness problems.


szhorvat:
This function is likely to be sensitive to small numerical errors, so I'd say it's not really possible to test the output values. The purpose of the x87 configuration in the CI is to catch such over-sensitive tests. While x87 itself is not important, if the result is very sensitive to small perturbations, then the test is likely to give different results (thus "fail") on some non-Intel architectures as well. Linux distros like Debian do test their packages on a multitude of architectures, so this would be a problem.

This makes it difficult to test layout algorithms extensively.

I would say that the goal with layout algorithms should be (in order of decreasing priority):

 - Exercise most code paths (in this case: both `GRID` and `NOGRID`). Then AddressSanitizer has a chance to catch errors.
 - Test edge cases (null graph, singleton graphs, etc.)
 - If the output is expected to satisfy some conditions, perhaps test those.  In this case: are the vertices within the bounds? IMO this is less important.

A tip for layout algorithms: pay attention to the case of disconnected graphs. Some algorithms may not handle disconnected graphs very well, and this may not be documented in all cases. I remember that DrL does not like disconnected graphs.

----

The Windows test failure is fixed now, just merge the latest `develop` branch.

GroteGnoom:
Thanks for the clear and thorough explanation of the whole situation :)

tests/unit/igraph_layout_fruchterman_reingold.c
```diff
@@ -0,0 +1,126 @@
...
+void destroy_bounds(igraph_vector_t *bounds) {
+    for (int i = 0; i < 4; i++) {
+        igraph_vector_destroy(&bounds[i]);
+    }
+}
+
+void check_and_destroy(igraph_matrix_t *result) {
+    igraph_real_t min, max;
+    igraph_matrix_minmax(result, &min, &max);
+    IGRAPH_ASSERT(min >= -1.0);
```
szhorvat, line 41: Now the checked bounds are hard-coded (box from -1..1). It would be useful to add a parameter to this function that controls that bounds that it checks. Then we can also check the case where no bounds are set, and verify that vertices didn't fly away too far, especially in the case of disconnected graphs. For this specific function, we could check that they're within e.g. a box of -50..50.

GroteGnoom: If you have enough iterations or a high enough temperature, they are supposed to just fly away without bounds right? That is what they're doing in practice at least. I did add some -50..50 checks for the modest cases.

PR #1667, opened 2021-02-16
Add layout_random_3d unit tests

GroteGnoom:
Part of #1592


src/layout/random.c
```diff
@@ -61,7 +61,7 @@ int igraph_layout_random(const igraph_t *graph, igraph_matrix_t *res) {
 
 /**
  * \function igraph_layout_random_3d
- * \brief Random layout in 3D
```
ntamas, line 64: I'd say if the brief description of `igraph_layout_random` is "Places the vertices uniform randomly on a plane.", then let's use a similar wording here:

"Places the vertices uniform randomly in a cube."

We can then add details about the coordinates of the corners of the cube in the main docstring but it is an unnecessary detail in a brief description (which appears in many places: section titles, TOC entries of the documentation etc).

ntamas (approved):
Looks good in general. One minor comment above.

szhorvat:
@ntamas Are we sure we want to output the actual coordinates, which effectively just test the consistency of the default random number generator? It will be some work to update this if we change the default RNG.

GroteGnoom:
> It will be some work to update this if we change the default RNG.

The same goes for a huge amount of other tests. If this is a possibility in the future I should change the way I test these things in general.

ntamas:
@szhorvat Testing consistency is good anyway; users probably expect the default RNG to be consistent and reproducible with the same seed across different platforms. If we change the RNG, that's a breaking change and we should take time to update the tests.

ntamas:
Just in time to make it to 0.9.0 :)

PR #1668, opened 2021-02-16
Add layout graphopt unit tests

GroteGnoom:
I changed the interface of graphopt, because I though it would be clearer and would not lead to any actual interface changes. As far as I can tell the same thing should happen for the same arguments (except maybe less overflows if igraph_integer_t is bigger than int).

### DIFF
--- a/include/igraph_layout.h
+++ b/include/igraph_layout.h
@@ -123,7 +123,7 @@ IGRAPH_EXPORT int igraph_layout_kamada_kawai_3d(const igraph_t *graph, igraph_ma
 IGRAPH_EXPORT int igraph_layout_graphopt(const igraph_t *graph,
                                          igraph_matrix_t *res, igraph_integer_t niter,
                                          igraph_real_t node_charge, igraph_real_t node_mass,
-                                         igraph_real_t spring_length,
+                                         igraph_integer_t spring_length,
                                          igraph_real_t spring_constant,
                                          igraph_real_t max_sa_movement,
                                          igraph_bool_t use_seed);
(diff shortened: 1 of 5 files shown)


szhorvat:
It is very strange that the spring length is treated as an integer internally. As any physical length, it should be a real number. @gaborcsardi do you recall why it's so?

Here is the original graphopt: http://www.schmuhl.org/graphopt/  From a quick (and not very careful) look, my best guess is that this was made an integer so that it would be easy to adjust through a spinbox GUI.

I would propose the opposite: keep the interface as is and change all internal `spring_length` variables to real numbers.

gaborcsardi:
I don't remember, sounds good to me.

szhorvat (approved):

ntamas:
> I would propose the opposite: keep the interface as is and change all internal spring_length variables to real numbers.

What's the status of this? In the PR it is still an `igraph_real_t` but now we have a stamp of approval :) I don't mind either way, there will be API breakages in 0.10 anyway.

szhorvat:
@ntamas Previsouly it was a `real` in the API, but it was an `int` internally. Now it is a `real` internally as well. I looked through the code and I could see no reason to avoid a real. I think the original code used an `int` because the parameter was tied to a GUI control that can only set integers.

I approved but didn't merge because I thought we might want to merge most of the new tests into `master` now (unless the API had to be changed or the behaviour was changed significantly). It's very easy to rebase right here on GitHub.

ntamas:
Okay, let's merge this into `master` then! Can you do the rebase and merge?

PR #1669, opened 2021-02-16
Add layout_bipartite unit_tests

GroteGnoom:
Part of #1592

Fix: The implementation assumed bools were either 1 or 0.

New error: negative horizontal gaps resulted in x-coordinates that were all zero, so that is excluded.

### DIFF
--- a/src/layout/bipartite.c
+++ b/src/layout/bipartite.c
@@ -58,12 +58,16 @@ int igraph_layout_bipartite(const igraph_t *graph,
     igraph_vector_t layers;
 
     if (igraph_vector_bool_size(types) != no_of_nodes) {
-        IGRAPH_ERROR("Invalid vertex type vector size", IGRAPH_EINVAL);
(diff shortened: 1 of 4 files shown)


src/layout/bipartite.c
```diff
@@ -58,12 +58,16 @@ int igraph_layout_bipartite(const igraph_t *graph,
...
+                      IGRAPH_EINVAL, igraph_vector_bool_size(types), no_of_nodes);
+    }
+    if (hgap < 0) {
+        IGRAPH_ERRORF("The horizontal gap cannot be negative, got %f.", IGRAPH_EINVAL, hgap);
     }
 
     IGRAPH_VECTOR_INIT_FINALLY(&layers, no_of_nodes);
     for (i = 0; i < no_of_nodes; i++) {
-        VECTOR(layers)[i] = 1 - VECTOR(*types)[i];
+        VECTOR(layers)[i] = VECTOR(*types)[i] ? 1 : 0;
```
szhorvat, line 70: To stay consistent with the previous behaviour, this should be `? 0 : 1`.

szhorvat (approved):

Issue #881, opened 2015-10-03
3D DrL layout crashed on edgeless graph


szhorvat:
If I call `igraph_layout_drl_3d` on a graph with some vertices, but no edges, it crashes right after issuing this error:

> DensityGrid_3d.cpp:247, Exceeded density grid in DrL

For some reason the R interface does not crash, though it does show the error.
The Python interface does crash.

```python
g=Graph()
g.add_vertices(10)
g.layout_drl() # no crash
g.layout_drl(dim=3) # crash
```

szhorvat:
Actually it seems any graph with isolated vertices triggers the problem.  A graph with three vertices `{1,2,3}` and one edges `1 <-> 2` is sufficient to trigger the problem.

ntamas:
I suspect that the fact that the DrL layout does not crash for dim=2 is just a coincidence; the code for the 2D and 3D layouts is pretty similar and the cause of the crash is that the layout "diverges" for disconnected graphs.

The easiest solution is probably to say that the algorithm works for connected graphs only, and place a connectedness check before the layout algorithm is called.

ntamas:
here's a temporary workaround: 708928c

PR #1670, opened 2021-02-17
Add drl_layout unit tests

GroteGnoom:
Part of #1592

Some issues with this commit:
I don't know which x and y values should be acceptable. There seems to
be a continuous range of possible end results with different parameters,
instead of things being suddenly obviously out of range.

Negative damping coefficients seem to have a propensity to cause
segfaults. I know where the segfault takes place, but I haven't done
the research to figure out when exactly problems occur. If negative
damping should be allowed I can investigate further.

These tests are way slower than average tests.


ntamas:
The DrL layout is designed for _very_ large graphs and I suspect that it simply fails to converge for small(ish) graphs -- it just keeps on iterating forever. Also, the entire algorithm is black magic and I have no idea how to test the output or what the acceptable range of X-Y values is.

I would suggest that we test the branches in the code that verify the input arguments and bail out if the input argument is invalid. Besides that, let's just have one invocation of the DrL layout that tests the "normal" code path for a simple graph (say, the Zachary karate club graph from `igraph_famous()`), with a small number of iterations to make the test finish faster.

szhorvat:
Here's some more info on DrL: https://github.com/gephi/gephi/wiki/OpenOrd

It says the same thing as @ntamas ,

> It is one of the few force-directed layout algorithms that can scale to over 1 million nodes, making it ideal for large graphs. However, small graphs (hundreds or less) do not always end up looking so good.

DrL was at some point renamed to OpenOrd. The original sources are no longer available from Sandia, though it's possible to find some (perhaps modified?) copies on GitHub.

szhorvat:
Also, I recall DrL had problems with disconnected graphs, see #881 .

This is not unusual: many layouts are designed specifically for connected graphs, with the expectations that connected components would be laid out separate, then the results arranged nicely. igraph has some limited facilities to do this, see the `layout_merge` functions.

szhorvat:
Hm, actually I found the original sources, dates 2009. They look similar enough that at some point an update of igraph's version might even be feasible. I don't know if it's worth it though ... 

https://www.sandia.gov/smartin/software.html

szhorvat:
With layout functions that use the random number generator, we should set a seed at the beginning of the set.

tests/unit/igraph_layout_drl.c
```diff
@@ -0,0 +1,192 @@
...
+    igraph_matrix_destroy(result);
+}
+
+int main() {
+    igraph_t g;
+    igraph_matrix_t result;
+    igraph_vector_t weights;
+    igraph_layout_drl_options_t options;
+    igraph_real_t seed[20] = {0.1, 0.2, 0.3, 0.4, 0.5, 0.6, 0.7, 0.8, 0.9, 1.0, -0.1, -0.2, -0.3, -0.4, -0.5, -0.6, -0.7, -0.8, -0.9, -1.0};
+    igraph_rng_seed(igraph_rng_default(), 42);
```
szhorvat, line 100: Let's always add an empty line after the initial block of variable declaration, for readability. I just couldn't see the `rng_seed` for a while as my eye kept skipping over the initial black.

GroteGnoom:
Thanks for all the information :) I made things simpler and faster and it seems to give normal results. It's still pretty slow for very few iterations, for small graphs it seems to do most of the work up front.

ntamas (approved):

ntamas:
FWIW, the test takes about 4 seconds in the CI env, which is slower than most of the unit tests but still not one of the slowest ones. Plus, since the tests are parallelized, it probably doesn't make much of a difference. I'm okay with leaving it as is for the time being.

szhorvat:
@ntamas The DrL test didn't seem slow on my machine, so I digged in a bit. It turns out that when I compile it with GCC 10.2, it takes 1.0 seconds. When I compile with Apple Clang 11, it takes 0.06 seconds. Both tested on macOS 10.14. That's a 20x difference, which is at least worrying. This sort of thing can't be explained by one compiler doing better.

I used RelWithDebInfo configuration in both cases.

Note that on the CI you can't judge the performance on macOS as after macOS 10.15, Apple does extra checks on executables when they're first started, so all tests take at least 0.35 s or so. But you can see that the DrL one does not stand out.

szhorvat:
It's not GCC vs Clang. It's a libstdc++ vs libc++ thing. I see the bad performance with libstdc++ and Clang 11 as well.

szhorvat:
I believe it's the creation of a million `deque`s that's responsible for this. In particular this line:

https://github.com/igraph/igraph/blob/master/src/layout/drl/DensityGrid.cpp#L71

Almost all time is spent in the constructor and destructor of `drl::graph` because of this (and comparatively no time is spent in the actual layout computation).

ntamas:
I came to the same conclusion with my own investigations that I did today. It may be the case that `libc++` defers the actual creation of the queues until something is placed in the queue, while `libstdc++` does not. We could potentially work around it by constructing the queues lazily, although I'm not sure whether it's worth the fuss.

szhorvat:
It seems that this is a constant (i.e. graph size independent) overhead of 1 second for any call of DrL when using GCC. That's not nice at all, but considering that DrL is meant for huge graphs that will take much longer than a second to lay out, and doesn't produce very nice results for small graphs, it's not a big deal in the end.

The newer version of DrL, OpenOrd, has the exact same code, so upgrading won't help.

What about slower computers? On the RPi 4 it takes 2 seconds and 600 MB of memory. My RPi 1 does not have enough memory to try.

Issue #1542, opened 2020-11-10
Diameter finding functions should return edge path


szhorvat:
As in  #1515, `igraph_diameter` and `igraph_diameter_dijkstra` should return the path in terms of edge indices (not yet implemented), in addition to vertex indices (already implemented).

PR #1673, opened 2021-02-17
diameter path of edges

guyroznb:
answers issue #1542 - returning the path of the diameter in terms of edges and / or vertices.

thanks


szhorvat:
I changed the base to `develop` again. Since this PR changes the API, it's going to be merged into `develop`, and will become part of the next version, 0.10.  This change should not affect you in any way.

tests/unit/igraph_diameter.c
```diff
@@ -0,0 +1,82 @@
...
+int main() {
+
+    igraph_t g;
+    igraph_real_t result;
+    igraph_integer_t from, to;
+    igraph_vector_t path_edge, path_vertex, weights_vec;
+    igraph_real_t weights[] = { 1, 2, 3, 4, 5, 1, 1, 1, 1};
+
+
+    igraph_barabasi_game(&g, 30, /*power=*/ 1, 30, 0, 0, /*A=*/ 1,
```
szhorvat, line 37: Here we need to set a random seed, so make the test output deterministic. Something like:

```
igraph_rng_seed(igraph_rng_default(), 1234);
```

tests/unit/igraph_diameter.c
```diff
@@ -0,0 +1,82 @@
...
+   but WITHOUT ANY WARRANTY; without even the implied warranty of
+   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+   GNU General Public License for more details.
+
+   You should have received a copy of the GNU General Public License
+   along with this program; if not, write to the Free Software
+   Foundation, Inc.,  51 Franklin Street, Fifth Floor, Boston, MA
+   02110-1301 USA
+
+*/
```
szhorvat, line 22: Please use this updated copyright notice:

https://github.com/igraph/igraph/blob/master/tests/unit/igraph_st_mincut.c#L2

tests/unit/igraph_diameter.c
szhorvat: Additionally, let's print another header here, something like `printf("diameter of Barabasi graph:\n");`

tests/unit/igraph_diameter.c
```diff
@@ -0,0 +1,82 @@
...
+    igraph_destroy(&g);
+
+    printf("diameter of ring and the path in terms of edges and vertices \n");
+
+    igraph_ring(&g, 10, IGRAPH_DIRECTED, 0, 0);
+    igraph_vector_init(&path_vertex, 0);
+    igraph_vector_init(&path_edge, 0);
+    igraph_vector_view(&weights_vec, weights, sizeof(weights) / sizeof(igraph_real_t));
+    igraph_diameter(&g, &result, &from, &to, &path_vertex, &path_edge, IGRAPH_DIRECTED, 1);
+    printf("diameter: %li, from %li to %li\n", (long int) result,
```
szhorvat, line 53: Let us print the `result` as a `%g`, without casting to a `long int` (both here and below). It will not change the output.

szhorvat (changes requested):
Thanks for the contribution! I added a few comments.

Additionally, can you add a few more tests? We should test:

 - The graph with zero vertices.
 - The graph with a single vertex.
 - A disconnected graph both with `unconn=true` and `unconn=false`.
 
This will produce Inf and NaN results. These should be printed using `print_real` from test_utilities as `print_real(stdout, x, "%g")`. (Not `printf`, as `printf`'s output is platform-dependent).

src/paths/shortest_paths.c
```diff
@@ -1018,20 +1024,43 @@ int igraph_diameter(const igraph_t *graph, igraph_real_t *pres,
...
+    if ((path_vertex) || (path_edge)) {
         if (! igraph_finite(res)) {
-            igraph_vector_clear(path);
+            if (path_vertex) {
+                igraph_vector_clear(path_vertex);
+            }
+            if (path_edge){
+                igraph_vector_clear(path_edge);
+            }
         } else {
```
szhorvat, line 1035: Here too, let us update the `else` branch to use `igraph_get_shortest_path` instead of `igraph_get_shortest_paths`.

ntamas:
Looks like there are some genuine test failures so I'm putting this into draft state until these are resolved.

@guyroznb Feel free to mark it as "ready for review" when the test failures are fixed.

src/paths/shortest_paths.c
```diff
@@ -1018,20 +1024,31 @@ int igraph_diameter(const igraph_t *graph, igraph_real_t *pres,
...
+                IGRAPH_CHECK(igraph_get_shortest_path(graph, vertex_path, 0,
+                                                    (igraph_integer_t) from,
+                                                    (igraph_integer_t)to,
+                                                    dirmode));
+            } else if (edge_path != 0) {
+                IGRAPH_CHECK(igraph_get_shortest_path(graph, 0, edge_path,
+                                                    (igraph_integer_t) from,
+                                                    (igraph_integer_t)to,
+                                                    dirmode));
+            }
```
szhorvat, line 1051: It seems to me that we don't need if `if ... else` construct here. Just call `igraph_get_shortest_path` directly with `vertex_path` and `edge_path`.

src/paths/shortest_paths.c
```diff
@@ -1231,26 +1255,35 @@ int igraph_diameter_dijkstra(const igraph_t *graph,
...
+                            (igraph_integer_t) from,
+                            (igraph_integer_t) to,
+                            weights, dirmode));
+            } else if (edge_path) {
+                IGRAPH_CHECK(igraph_get_shortest_path_dijkstra(graph,
+                            /*vertices=*/ 0, /*edges=*/ edge_path,
+                            (igraph_integer_t) from,
+                            (igraph_integer_t) to,
+                            weights, dirmode));
+            }
```
szhorvat, line 1285: It seems to me that we don't need if `if ... else` construct here. Just call `igraph_get_shortest_path` directly with `vertex_path` and `edge_path`.

szhorvat (changes requested):
Thanks for the update! I added a few more comments, the main one being that we do not need an `if` with three branches when getting the actual path (am I missing some reason why you added the branches?).

guyroznb:
thanks @szhorvat and @ntamas for the review!

szhorvat (approved):

ntamas (approved):

ntamas:
Thanks a lot for the PR, much appreciated!

PR #1674, opened 2021-02-18
Add progress_handler_stderr unit tests

GroteGnoom:
Part of #1592


GroteGnoom:
Is there a proper way to check stderr? The seem to be some platform-dependent ways to redirect it to stdout, but that's probably not the way to go. Should we also have .err files instead of only .out?

ntamas:
> Is there a proper way to check stderr?

No, not in the current testing framework. But since testing stderr is rare, I would say the easiest way to proceed is like this:

```diff
--- a/etc/cmake/run_legacy_test.cmake
+++ b/etc/cmake/run_legacy_test.cmake
@@ -33,6 +33,7 @@ execute_process(
   WORKING_DIRECTORY ${WORK_DIR}
   RESULT_VARIABLE ERROR_CODE
   OUTPUT_VARIABLE OBSERVED_OUTPUT
+  ERROR_VARIABLE OBSERVED_OUTPUT
 )

 if(ERROR_CODE EQUAL 77)
```

This essentially merges stdout and stderr into the same variable in CMake, which we then compare with the expected output file in `some_test_name.out`.

If we stick to the convention that a test case writes expected output either to stdout or to stderr but never both, then the only issue that we won't catch with this is when the expected output appears on the wrong stream.

Can you test whether this solution works for you?

szhorvat:
Please add `IGRAPH_EXPORT` to this declaration in the header:

```
extern igraph_progress_handler_t igraph_progress_handler_stderr;
```

ntamas:
Note that CMake's documentation says this about using the same variable for catching stdout and stderr:

> OUTPUT_VARIABLE, ERROR_VARIABLE
> The variable named will be set with the contents of the standard output and standard error pipes, respectively. If the same variable is named for both pipes their output will be merged in the order produced.

szhorvat:
> This essentially merges stdout and stderr into the same variable in CMake, which we then compare with the expected output file in `some_test_name.out`.

I'm not sure I'm happy to start merging stdout and stderr output and comparing to the .out files. There is value in keeping the two separate. All errors e.g. from ASan go to stderr. Sometimes a wrong output and some ASan failure come together and this will make it harder to disentangle them.

I would simply skip testing what this function outputs to stderr and instead just let the test exercise the function. That is still a useful test: it just exposed a bug we had. The function was not exported from the shared library and therefore didn't work at all with shared builds. Thanks to this test we detected the problem.

ntamas:
> There is value in keeping the two separate.

Sure it is, but the code in `run_legacy_tests.cmake` would be much more convoluted if we could have an `.out` file, an `.err` file or both. If ASAN routinely writes its output to `stderr`, then this obviously changes things -- in that case, let's just validate that the test compiles and runs but let's not check what it writes to `stderr`.

szhorvat:
> let's just validate that the test compiles and runs but let's not check what it writes to `stderr`.

Yes, that is what I meant.

PR #1675, opened 2021-02-19
Add rng unit tests

GroteGnoom:
Part of #1592

It all seems to work fine. I added some doc improvements, and changed 0 error codes to IGRAPH_SUCCESS in general.


szhorvat:
Can you please try to add an `IGRAPH_EXPORT` to these lines in the header?

```
extern const igraph_rng_type_t igraph_rngtype_glibc2;
extern const igraph_rng_type_t igraph_rngtype_rand;
extern const igraph_rng_type_t igraph_rngtype_mt19937;
```

tests/unit/rng_init_destroy_max_min_name_set_default.c
```diff
@@ -0,0 +1,63 @@
...
+
+int main() {
+    int i;
+    igraph_rng_type_t rng_types[3] = {igraph_rngtype_glibc2, igraph_rngtype_mt19937, igraph_rngtype_rand};
+    igraph_rng_t rng_def;
+
+    IGRAPH_ASSERT(igraph_rng_init(&rng_def, &igraph_rngtype_glibc2) == IGRAPH_SUCCESS);
+
+    for (i = 0; i < 3; i++) {
+        test_and_destroy(rng_types + i, &rng_def);
```
szhorvat, line 56: Let's use `&rng_types[i]` for readability.

szhorvat:
Can you please fix the description of `igraph_rngtype_rand`? It says `rand/stand`, should be `rand/srand`. 

For `igraph_rngtype_glibc2`, please clarify in the docs that this is a (re)implementation of this generator. It does not actually call the one in glibc on the system.

I'd do these, but then we might edit the same file and the same time and create conflicts.

PR #1678, opened 2021-02-19
Add sparsemat_fkeep unit tests

GroteGnoom:
Part of #1592

Fix: Error checking assumed a non-zero return value meant success, but
-1 is the returned on errors.
All error checking is now done in the igraph_sparsemat_fkeep function,
so we have more information about what went wrong.


src/core/sparsemat.c
```diff
@@ -700,11 +700,18 @@ int igraph_sparsemat_fkeep(igraph_sparsemat_t *A,
                            int (*fkeep)(int, int, igraph_real_t, void*),
                            void *other) {
 
-    if (!cs_fkeep(A->cs, fkeep, other)) {
-        IGRAPH_ERROR("Cannot filter sparse matrix", IGRAPH_FAILURE);
+    if (!A) {
+        IGRAPH_ERROR("The sparse matrix pointer is NULL.", IGRAPH_EINVAL);
     }
+    if (A->cs->nz != -1) {
+        IGRAPH_ERROR("The sparse matrix is not in compressed format.", IGRAPH_EINVAL);
```
szhorvat, line 707: I would use `nz >= 0` instead of `n != -1`, as this is what the CS library itself does internally. I don't know how much we can trust the claim from the comment that this is always precisely -1 (and not say -2).

https://github.com/igraph/igraph/blob/master/vendor/cs/cs.h#L658

szhorvat: Actually, I was wrong. They use `== -1` just one line above.

https://github.com/igraph/igraph/blob/master/vendor/cs/cs.h#L657

src/core/sparsemat.c
```diff
@@ -700,11 +700,18 @@ int igraph_sparsemat_fkeep(igraph_sparsemat_t *A,
                            int (*fkeep)(int, int, igraph_real_t, void*),
                            void *other) {
 
-    if (!cs_fkeep(A->cs, fkeep, other)) {
-        IGRAPH_ERROR("Cannot filter sparse matrix", IGRAPH_FAILURE);
+    if (!A) {
```
szhorvat, line 703: I think we can skip this check, as a null pointer wouldn't purposefully be passed in here.

Most of the checks / error reports are for conditions that can be triggered from a high-level interface. Passing in a null pointer cannot.

This situation would be comparable to checking every `igraph_t *graph` not to be null, which would IMO be too much.

GroteGnoom: The specific reason I put this here is because cs_fkeep checks for it. If you don't check it here, cs_fkeep would return -1, and it's strange not to handle that error. And if you're going to handle it anyway, this is clearer.

src/core/sparsemat.c
```diff
@@ -700,11 +700,18 @@ int igraph_sparsemat_fkeep(igraph_sparsemat_t *A,
...
+        IGRAPH_ERROR("The sparse matrix pointer is NULL.", IGRAPH_EINVAL);
     }
+    if (A->cs->nz != -1) {
+        IGRAPH_ERROR("The sparse matrix is not in compressed format.", IGRAPH_EINVAL);
+    }
+    if (!fkeep) {
+        IGRAPH_ERROR("The filter function is a NULL pointer.", IGRAPH_EINVAL);
+    }
+    cs_fkeep(A->cs, fkeep, other);
```
szhorvat, line 712: It would still be good to check the return value of `cs_fkeep` and see if it is `-1`. Zero should be okay, then all elements were removed. We can link to an external version of this library, which may change in the future.

src/core/sparsemat.c
szhorvat: Issuing a proper error for this still bothers me ... I am not against checking, I am against `IGRAPH_ERROR`.  I know that this is subjective, but `IGRAPH_ERROR` generally indicates "expected errors" while `IGRAPH_ASSERT` indicates "unexpected errors".

Error messages should generally be understandable to a user of the R/Python/Mathematica interfaces, or a hypothetical GUI program using igraph. They could be triggered by something the user did. In that context "null pointer" makes no sense—what  can a user do to avoid it? If it happened, it's the fault of the programmer, not the fault of the end user. It's a bug that should be fixed in the software that depend on igraph, not a mistake made by the user of that software.

Therefore, I would go for an `IGRAPH_ASSERT` instead.

szhorvat:
Could you please add another short test that removes all elements from the matrix (i.e. the `fkeep` function returns true for everything)? This would have failed previsouly, as the 0 return was wrongly interpreted as failure.

PR #1680, opened 2021-02-19
Add sparsemat_nonzero_storage unit tests

GroteGnoom:
Part of #1592

Also added some documentation.

Biggest change: Documentation with one space missing in doxrox were
                silently ignored. Fixed the spaces, not the silent
                ignoring.

### DIFF
--- a/doc/sparsemat.xxml
+++ b/doc/sparsemat.xxml
@@ -27,13 +27,14 @@
 <!-- doxrox-include igraph_sparsemat_type -->
 <!-- doxrox-include igraph_sparsemat_is_triplet -->
 <!-- doxrox-include igraph_sparsemat_is_cc -->
-<!-- doxrox-include igraph_sparsemat_min-->
-<!-- doxrox-include igraph_sparsemat_max-->
-<!-- doxrox-include igraph_sparsemat_minmax-->
-<!-- doxrox-include igraph_sparsemat_count_nonzero-->
-<!-- doxrox-include igraph_sparsemat_count_nonzerotol-->
-<!-- doxrox-include igraph_sparsemat_rowsums-->
-<!-- doxrox-include igraph_sparsemat_colsums-->
+<!-- doxrox-include igraph_sparsemat_min -->
+<!-- doxrox-include igraph_sparsemat_max -->
+<!-- doxrox-include igraph_sparsemat_minmax -->
+<!-- doxrox-include igraph_sparsemat_count_nonzero -->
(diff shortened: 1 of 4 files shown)


szhorvat:
It is unfortunate that `igraph_sparsemat_nonzero_storage`, `igraph_sparsemat_nzmax`, `igraph_sparsemat_iterator_row`, `igraph_sparsemat_iterator_col`, `igraph_sparsemat_iterator_next`, `igraph_sparsemat_iterator_idx` return an `int`. When I see `int`, I assume error code. 

I would like to cautiously suggest making these return an `igraph_integer_t` for clarity. Currently an `igraph_integer_t` is an `int`, so this would not break API and we can have it in 0.9.1. It would be a cosmetic change for now.

If in the future we change `igraph_integer_t`, CXSparse has the flexibility to adapt (it comes with several versions, some of which use longer values).

@ntamas Any objections to making these functions return `igraph_integer_t`?

GroteGnoom:
Maybe it's not a quick practical solution, but shouldn't all the other functions actually return igraph_error_type_t?

ntamas:
There was a proposal earlier (from Vincent or me, can't remember) that we introduce `igraph_error_t` as a typedef to `int`, and then we can make it explicit which functions return an error code. This would be my preferred direction instead of replacing genuine `int`s with `igraph_integer_t`.

ntamas:
FYI, `doxrox` include bug is fixed in `4cd9dadee`

GroteGnoom:
> There was a proposal earlier (from Vincent or me, can't remember) that we introduce `igraph_error_t` as a typedef to `int`, and then we can make it explicit which functions return an error code. This would be my preferred direction instead of replacing genuine `int`s with `igraph_integer_t`.

Why doesn't `igraph_error_type_t` suffice?

szhorvat:
Vincent's proposal was to use `igraph_error_type_t`, but that was trouble to use last time I tried. I don't recall all the reasons but one reason was that C++ doesn't allow freely mixing integer values with enums, so it was very cumbersome to change it everywhere.

That's when Tamás suggested the typedef.

szhorvat:
Regarding those `int`s: when we transition to a wide `igraph_integer_t`, I would at the same time transition to a version of CXSparse that can be matched with its integer type. CXSparse supports different integer type widths. Thus at least at that point I would still argue for using `igraph_integer_t` (at least so that we can store more than 2^31-1 elements in a sparse array, i.e. more than that many edges in a graph).

PR #1681, opened 2021-02-20
Add sparsemat_getelements_sorted unit tests

GroteGnoom:
Part of #1592


src/core/sparsemat.c
```diff
@@ -2996,6 +2996,28 @@ int igraph_sparsemat_sort(const igraph_sparsemat_t *A,
...
+ * \param A A sparse matrix in either triplet or compressed form.
+ * \param i An initialized int vector. This will store the rows of the
+ *          returned elements.
+ * \param j An initialized int vector. For a triplet matrix this will
+ *          store the columns of the returned elements. For a compressed
+ *          matrix this will first store a 0, and then for each column
+ *          the cumulative number of elements that have been returned.
+ *          So if this there was 1 element returned in the first column,
+ *          and 5 in the second, this will return {0, 1, 6}.
+ * \return Error code.
```
ntamas, line 3016: Please add documentation for the `x` argument for sake of completeness.

src/core/sparsemat.c
```diff
@@ -2996,6 +2996,28 @@ int igraph_sparsemat_sort(const igraph_sparsemat_t *A,
...
+ *
+ * \param A A sparse matrix in either triplet or compressed form.
+ * \param i An initialized int vector. This will store the rows of the
+ *          returned elements.
+ * \param j An initialized int vector. For a triplet matrix this will
+ *          store the columns of the returned elements. For a compressed
+ *          matrix this will first store a 0, and then for each column
+ *          the cumulative number of elements that have been returned.
+ *          So if this there was 1 element returned in the first column,
+ *          and 5 in the second, this will return {0, 1, 6}.
```
ntamas, line 3015: I'd rather change the wording and say that `j[col_index]` is equal to the index where the `col_index`-th column starts in the returned value vector `x` (and the row vector `i` as well), and that the last element in this vector is the total number of elements. I'd also note that the number of elements in the `col_index`-th column is therefore `j[col_index + 1] - j[col_index]`.

ntamas (changes requested):
Minor fixes only, otherwise it looks good to me! Thanks a lot again!

ntamas (approved):

PR #1685, opened 2021-02-21
Add sparsemat_view unit tests and doc

GroteGnoom:
Part of #1592

This init function does not seem to have a corresponding destroy
function. sparsemat_view calls cs_calloc directly for the struct, but
doesn't allocate the index and element vectors. igraph users don't have
direct access to cs_malloc to allocate their own through the cs
interface. They also don't have access to cs_free through igraph. So
currently there's no proper destruction without assuming cs_free and
cs_malloc use free and malloc as they do now.

### DIFF
--- a/src/core/sparsemat.c
+++ b/src/core/sparsemat.c
@@ -2961,6 +2961,40 @@ int igraph_sparsemat_dense_multiply(const igraph_matrix_t *A,
     return 0;
 }
 
+/**
+ * \function igraph_sparsemat_view
+ * \brief Initialize a sparse matrix and set all parameters.
+ *
+ * This function allocates memory for a sparse matrix and sets all its
+ * parameters to the given values. For a simpler way to initialize an
+ * empty sparse matrix, see \ref igraph_sparsemat_init(). Using
+ * \ref igraph_sparsemat_destroy() on the resulting matrix will try
+ * to free \p p, \p i and \p x. To only free memory allocated in this
+ * function, use free() on the \c cs field of A.
+ *
+ * \param A The non-initialized sparse matrix.
+ * \param nzmax The maximum number of entries.
+ * \param m The number of rows.
+ * \param n The number of columns.
+ * \param p The column vector. For a triplet matrix this should contain
+  *         the columns of the elements in \x. For a compressed matrix,
+            if the column index is \c k, then <code>j[k]</code>
+ *          is the index in \p x of the start of the \c k-th column, and
+ *          the last element of \c j is the total number of elements.
+ *          The total number of elements in the \c k-th column is
+ *          therefore <code>j[k+1] - j[k]</code>. For example, if there
+ *          is one element in the first column, and five in the second,
+ *          \c j should be set to <code>{0, 1, 6}</code>.
+ * \param i The row vector. This should contain the row indices of the
+ *          elements in x.
+ * \param x The elements of the sparse matrix.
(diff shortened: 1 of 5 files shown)


szhorvat:
The purpose of this function is similar to that of `igraph_vector_view`: take existing raw data, and create a "view" to it. The compressed sparse column (or row) format is quite common, other systems use it extensively. This function provides a chance to directly access such data from another library without copying it.

With `igraph_vector_view` and `igraph_matrix_view` there is no need to have a special destructor, as the only dynamically allocated part of an `igraph_vector_t` or `igraph_matrix_t` is the raw data array. As you say, `igraph_sparsemat_t` is different, as it does not embed a `cs_di_sparse`, instead it points to one. I am not sure why. Do you @ntamas ? Life would certainly be easier if it embedded the `cs_di_sparse`, but we can't make that change for 0.9 anymore, as all 0.9.x releases should be binary-compatible with each other.

src/core/sparsemat.c
```diff
@@ -2961,6 +2961,40 @@ int igraph_sparsemat_dense_multiply(const igraph_matrix_t *A,
...
+ * \ref igraph_sparsemat_destroy() on the resulting matrix will try
+ * to free \p p, \p i and \p x. To only free memory allocated in this
+ * function, use free() on the \c cs field of A.
+ *
+ * \param A The non-initialized sparse matrix.
+ * \param nzmax The maximum number of entries.
+ * \param m The number of rows.
+ * \param n The number of columns.
+ * \param p The column vector. For a triplet matrix this should contain
+  *         the columns of the elements in \x. For a compressed matrix,
```
szhorvat, line 2980: `\x` should be `\p x`?

szhorvat:
Can you please point me to where the internal representation of the triplet format is described?

GroteGnoom:
> Can you please point me to where the internal representation of the triplet format is described?

I haven't found anything like that. CXSparse is quite concise in its documentation. Since our documentation was also not available I just tried to figure things out by running functions and looking at code. This is why some of my descriptions are a little clumsy because I don't fully understand the intent.

The best I have is the struct. https://github.com/igraph/igraph/blob/97c96f615ed4d8ab1490d4bc2fc35f080fc4c082/vendor/cs/cs.h#L61

szhorvat:
I made some changes to the docs. Could you please update the tests and plug in some data, convert to dense with `igraph_sparsemat_as_matrix` and print it?

For example, here's a dense matrix:

```
0	5
1	0
0	2
```

Here's the sparse representation as compressed format:

```
int p[] = {0, 1, 3};
int i[]  = {1, 0, 2};
double x[] = {1, 5, 2};
```

Hope I didn't make a mistake.

I change the `cs_malloc` to `igraph_Calloc`. Please change the `free` to `igraph_free` as well.

szhorvat:
Oops, it seems we also need the `igraph_memory.h` header at the top for the `igraph_Calloc`, which I just added.

GroteGnoom:
> Oops, it seems we also need the `igraph_memory.h` header at the top for the `igraph_Calloc`, which I just added.

I think the errors were because igraph_Calloc takes in a type, not a size.

szhorvat:
Thanks for finalizing it! I was using the web interface on GitHub to edit the docs, so it was a but cumbersome to make all changes.

GroteGnoom:
No problem :slightly_smiling_face: Programming is much nicer with a compiler by your side.

PR #1690, opened 2021-02-23
Add sparsemat_droptol unit tests

GroteGnoom:
Part of #1592

### DIFF
--- a/src/core/sparsemat.c
+++ b/src/core/sparsemat.c
@@ -704,7 +704,7 @@ int igraph_sparsemat_fkeep(igraph_sparsemat_t *A,
 
     IGRAPH_ASSERT(A);
     IGRAPH_ASSERT(fkeep);
-    if (A->cs->nz != -1) {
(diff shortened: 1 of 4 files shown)


src/core/sparsemat.c
```diff
@@ -751,12 +751,15 @@ int igraph_sparsemat_dropzeros(igraph_sparsemat_t *A) {
 
 int igraph_sparsemat_droptol(igraph_sparsemat_t *A, igraph_real_t tol) {
 
-    if (!cs_droptol(A->cs, tol)) {
-        IGRAPH_ERROR("Cannot drop (almost) zeros from sparse matrix",
-                     IGRAPH_FAILURE);
+    IGRAPH_ASSERT(A);
+    if (A->cs->nz != -1) {
```
szhorvat, line 755: Let's use `! igraph_sparsemat_is_cc(A)` for clarity and readability. Since that function is in the same translation unit, the compiler will be able to optimize the call away.

Issue #1686, opened 2021-02-21
Test fail for shared library on Windows


vtraag:
**Describe the bug**
After successfully building `igraph` using CMake, all tests fail, presumably because the library cannot be found. This problem only appears on Anaconda. See https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=280259&view=logs&j=a70f640f-cc53-5cd3-6cdc-236a1aa90802&t=f5d15007-a01c-5ad8-c9ce-4d519d3b275f for a log of the failure.

**To reproduce**
I am still investigating what the underlying problem is. The problem is easily reproducible in the Anaconda environment (see https://github.com/conda-forge/igraph-feedstock/pull/25 for the PR), but I have not been able to identify and replicate the problem locally. Presumably this is due to some paths that are being set in Anaconda, I am looking into this.

**Version information**
The released version 0.9.0 source tarball.

szhorvat:
Can this be closed now?

PR #1692, opened 2021-02-23
Made paths for testing in Windows more robust

vtraag:
When building a shared library on Windows, we automatically add the necessary paths to ensure the shared libraries are found when running `ctest`. This does not work well when there are paths included that end with a backslash, because the trailing backslash is then incorrectly interpreted as an escape character. This problem was raised in #1686, and this PR should fix this.

This PR changes the way that the path is constructed to ensure that each path is correctly included in a semi-colon separated string. Presumably, when 3.20 comes out, this can be replaced by a [`TO_NATIVE_PATH_LIST`](https://cmake.org/cmake/help/latest/command/cmake_path.html#to-native-path-list) construct. Of course, we may want to wait until this version of cmake is sufficiently well available. Anyway, as a reminder, I also mentioned this possibility in the comments.

Hopefully, this also still works well with msys. Let's see how the CI does.

### DIFF
--- a/etc/cmake/test_helpers.cmake
+++ b/etc/cmake/test_helpers.cmake
@@ -70,8 +70,20 @@ function(add_legacy_test FOLDER NAME NAMESPACE)
 
     # Semicolons are used as list separators in CMake so we need to escape them in the PATH,
     # otherwise the PATH envvar gets split by CMake before it passes the PATH on to CTest.
-    string(JOIN "\;" CORRECT_PATH $ENV{PATH})
-    SET_TESTS_PROPERTIES(${TEST_NAME} PROPERTIES ENVIRONMENT "PATH=${IGRAPH_LIBDIR}\;${CORRECT_PATH}" )
+    # We process each path separately to ensure it is a proper path. In particular, we need
+    # to ensure that a trailing backslash is not incorrectly interpreted as an escape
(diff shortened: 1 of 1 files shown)


szhorvat:
Why don't you use `TO_NATIVE_PATH` instead of the `string(REPLACE "/" "\\" IGRAPH_LIBDIR ${IGRAPH_LIBDIR})` too?

vtraag:
That does not work because the native path is different from the path in which the test command is being run. That was at least the outcome of the previous debugging system on MSYS. I'd be happy to give that another try though.

vtraag:
However, for some reason the msys GH actions don't seem to be triggered. Any idea why not?

szhorvat:
Not too familiar with GH Actions, but I suspect this:

https://github.com/igraph/igraph/blob/master/.github/workflows/build-cmake.yml#L10

szhorvat:
I removed that restriction, can you pull from `master` now?

vtraag:
It now seems to work well. Perhaps there are still other situations in which ctest can be invoked that may lead to a potential problem, but at least not in this current setup. As far as I'm concerned we can merge it. I left most of the comment in, to remind us of potential problems should it ever turn out to be a problem still.

ntamas (approved):

PR #1693, opened 2021-02-23
gh actions: enable cache

szhorvat:


szhorvat:
@iosonofabio Can you help set up caching and ccache for GitHub Actions?

The caching concept they use is very different from all other CI systems we used in the path. It seems to be focused on ["dependencies that don't change often"](https://docs.github.com/en/actions/guides/caching-dependencies-to-speed-up-workflows) and restoring is done based on keys. (In this PR, caches are not uploaded, "Cache hit occurred on the primary key Windows-i686-ccache, not saving cache.") I'm not sure how to employ this for ccache where file changes practically after every run.

szhorvat:
For lack of a better idea, I included the git hash in the cache key, triggering a cache save on every commit. It seems to work now, though the speedup is disappointingly small (~2 min?).

szhorvat:
CCache does appear to be working, but, strangely, it does not seem to lead to any speedup. Maybe file access is just slow with MSYS2? Perhaps it's not worth bothering.

PR #1696, opened 2021-02-24
Randomize the Louvain algorithm

### LINKED ISSUE #1633: Randomize the Louvain algorithm
At the moment, the Louvain algorithm (`igraph_community_multilevel`) is deterministic. Currently, all nodes are considered in the algorithm simply in the order of the node indices. However, it makes sense that this is randomized. At the moment, one always gets the same result for any graph, perhaps leading people to erroneously believe that this is the only "correct" partition of the Louvain algorithm. The randomization of the node order was also explicitly suggested by the original authors [1].

**Use cases for the feature**
Randomizing the node order allows to determine the stability of any particular resulting partition with regards to different runs, which can also be helpful to produce consensus clusterings [2]. 

**References**
[1] Blondel, V. D., Guillaume, J.-L., Lambiotte, R., & Lefebvre, E. (2008). Fast unfolding of communities in large networks. Journal of Statistical Mechanics: Theory and Experiment, 10008(10), 6. https://doi.org/10.1088/1742-5468/2008/10/P10008
[2] Lancichinetti, A., & Fortunato, S. (2012). Consensus clustering in complex networks. [Physics and Society; Information Retrieval]. Scientific Reports, 2, 336. https://doi.org/10.1038/srep00336

GroteGnoom:
Fixes #1633


vtraag:
Thanks for taking this up! It all looks good to me. Perhaps we can add to the documentation a small explanation that the node order is randomized?

GroteGnoom:
You're welcome :slightly_smiling_face:. I was a little bit unsure about the RNG_BEGIN() things, so I asked about it here: https://igraph.discourse.group/t/the-unit-tests-thread/597/40

I will continue to work on it tomorrow, and fix the docs.

vtraag:
I believe that because `igraph_vector_shuffle` calls `RNG_BEGIN()` and `RNG_END()` separately, they do not need to be included again, but let me double check that.

szhorvat:
@vtraag Do we want this on `master` for 0.9.1 or should this be delayed for the next version?

vtraag:
As far as I am concerned, we can leave it for later versions 0.9.2 if we want to release 0.9.1 somewhere soonish. If it happens that this already gets merged before releasing 0.9.1, that would also be perfectly fine!

vtraag (approved):

szhorvat:
I notice that this chooses a random order _once_ and then uses it for all passes. Wasn't the intention to use a different random order for each pass, @vtraag ?

vtraag:
You could, but it would increase the running time, and it is not clear directly whether it would improve the results. That is, it's a choice,  either one could be good, and I don't directly know of any analysis suggesting either the one or the other is better...

PR #1697, opened 2021-02-25
Added libxml dependency on Windows on Azure pipelines

vtraag:
This PR adds the`libxml2` dependency on Windows, hence building with GraphML support. I use `conda` for installing the `libxml2` dependency, as this should enable a faster install than `vcpkg`. Perhaps the `conda` environment can also be cached, that would be a next step to consider. Let's first see how this builds.

### DIFF
--- a/.azure/build-win.yml
+++ b/.azure/build-win.yml
@@ -39,7 +39,9 @@ parameters:
 steps:
   - script: |
       choco install winflexbison3 ninja
-      %CONDA%\Scripts\activate
+
+      call "%CONDA%\Scripts\activate" base
+
       conda install libxml2
 
     displayName: Install dependencies
--- a/.azure/build-win.yml
+++ b/.azure/build-win.yml
@@ -45,7 +45,7 @@ steps:
       key: >-
         vcpkg
         | $(Agent.Os)
-        | $(parameters.build_shared)
+        | ${{ parameters.build_shared }}
       path: $(VCPKG_INSTALLATION_ROOT)
       cacheHitVar: VcpkgRestoredFromCache
     displayName: Vcpkg Cache
--- a/.azure/build-win.yml
+++ b/.azure/build-win.yml
@@ -45,7 +45,7 @@ steps:
       key: >-
         vcpkg
         | $(Agent.Os)
-        | ${{ parameters.build_shared }}
+        | ${{ parameters.vcpkg_target_triplet }}
       path: $(VCPKG_INSTALLATION_ROOT)
       cacheHitVar: VcpkgRestoredFromCache
     displayName: Vcpkg Cache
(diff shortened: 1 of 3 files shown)


vtraag:
This is now done. The build now uses `vcpkg` instead of `conda`. This is necessary because we want to be able to provide a static build. Unfortunately that does increase the build time, and even with caching enabled, this still takes about 2 minutes.

It seems that CMake by [default uses dynamic linking to the runtime (`/MD`)](https://cmake.org/cmake/help/latest/prop_tgt/MSVC_RUNTIME_LIBRARY.html), also for static libraries. By default, for static libraries, `vcpkg` statically links to the runtime (`/MT`), and [the alternative triplet `x64-windows-static-md` was recently suggested](https://github.com/microsoft/vcpkg/issues/15122) to allow static libraries with dynamic runtime linking. Some [potential problems are mentioned](https://github.com/microsoft/vcpkg/issues/15122#issuecomment-745497720), although it is not exactly clear what the exact implications and limitations are. I think this should be fine and the build now uses `x64-windows-static-md` to build the `libxml2` library.

aggieNick02:
Just FYI, the discussion on possible static library dynamic runtime issues has been moved to a new issue, in case you want to link to it, since the previous issue ended up, in the end, being resolved by adding CI support for the triplet. Here's the new issue: https://github.com/microsoft/vcpkg/issues/16387

.azure/build-win.yml
```diff
@@ -22,7 +22,7 @@ parameters:
     default: true
   - name: build_shared
     type: boolean
-    default: false
+    default: true
```
szhorvat, line 25: I would keep this on default `false` to be consistent with the non-Windows template. Otherwise people will get confused in the future.

vtraag: Yeah, sure, no problem. I had the default `vcpkg_target_triplet` set to `x64-windows`, and I wanted to make those two consistent. I've now changed the default `vcpkg_target_triplet` to `x64-windows-static-md` to make it consistent in that way.

szhorvat:
[Appveyor seems to suggest](https://www.appveyor.com/docs/build-cache/) caching only the `vcpkg\installed` directory.  Do you think there is an opportunity for speedup here?

vtraag:
Yes, we might try that, it should at least shave off something probably.

szhorvat:
The cache is now only 46 MB, this should massively improve the cache restore times.

However, when I changed the cache directory, at first it still tried to restore all the 500+ MB of old cached data into the new directory. I don't know how to clear the cache, so I just changes the cache key.

szhorvat:
Started a second run, cache restore time is now ~5 seconds instead of 2 minutes.

Shall we merge this?

szhorvat:
There's just this annoying thing that Chocolatey often hangs for up to 3 minutes on startup. Sometimes it happens, sometimes it doesn't. It's been like this since the beginning, and unrelated to this PR.

vtraag:
We could also cache the parts of chocolatey that we need (it install things everywhere across the system, but the things we need are located in a single location I believe).

szhorvat:
It seemed that it was not the download that was slow, but simply the startup. I'm not sure. 

I'll leave it to you to squash-and-merge this. I'm very happy with it, finally everything is tested on all platforms :-)

PR #1698, opened 2021-02-25
Add get_subisomorphisms_vf2 unit tests

GroteGnoom:
Part of #1592

### DIFF
--- a/src/isomorphism/vf2.c
+++ b/src/isomorphism/vf2.c
@@ -1675,11 +1675,11 @@ static igraph_bool_t igraph_i_get_subisomorphisms_vf2(
 
 /**
  * \function igraph_get_subisomorphisms_vf2
- * Return all subgraph isomorphic mappings
+ * \brief Return all subgraph isomorphic mappings.
  *
  * This function collects all isomorphic mappings of \p graph2 to a
  * subgraph of \p graph1. It uses the \ref
- * igraph_subisomorphic_function_vf2() function.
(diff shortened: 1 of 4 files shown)


src/isomorphism/vf2.c
```diff
@@ -1675,11 +1675,11 @@ static igraph_bool_t igraph_i_get_subisomorphisms_vf2(
 
 /**
  * \function igraph_get_subisomorphisms_vf2
- * Return all subgraph isomorphic mappings
+ * \brief Return all subgraph isomorphic mappings.
  *
  * This function collects all isomorphic mappings of \p graph2 to a
  * subgraph of \p graph1. It uses the \ref
- * igraph_subisomorphic_function_vf2() function.
+ * igraph_subisomorphic_function_vf2() function. The graphs shoud be simple.
```
ntamas, line 1682: `should` :)

ntamas:
Looks good to me, but I'm not sure I was able to follow all the test cases and guarantee the correctness of the expected output. @szhorvat how do you do this usually? Do you validate it with Mathematica?

szhorvat:
It is relatively easy to check that each result is correct. Loop through the edges of the small graph, and check through the mapping that each is present in the large graph.

To check that the list is exhaustive, the only option is to compare with another subgraph finder. Luckily, we have one: LAD.

szhorvat:
Mathematica does not have built-in subgraph finding functionality. It is one of the major features that IGraph/M brings to it.

ntamas (approved):

PR #1700, opened 2021-02-25
Add igraph_citing_cited_type_game unit tests

GroteGnoom:
Part of #1592

Fix: the sumtree array itself was not freed.
Fix: sums was not destroyed

### DIFF
--- a/src/games/citations.c
+++ b/src/games/citations.c
@@ -387,59 +387,56 @@ int igraph_citing_cited_type_game(igraph_t *graph, igraph_integer_t nodes,
     igraph_i_citing_cited_type_game_struct_t str = { 0, NULL };
     igraph_psumtree_t *sumtrees;
     igraph_vector_t sums;
-    long int nocats;
+    long int no_of_types;
     long int i, j;
 
-    if (nodes == 0) {
-        return igraph_empty(graph, 0, directed);
-    }
     if (igraph_vector_size(types) != nodes) {
         IGRAPH_ERRORF("Length of types vector (%ld) not equal to number"
                       " of nodes (%" IGRAPH_PRId ").",
                       IGRAPH_EINVAL, igraph_vector_size(types), nodes);
     }
-    if (igraph_matrix_ncol(pref) != igraph_vector_max(types) + 1) {
+
+    /* avoid calling vector_max on empty vector */
+    no_of_types = nodes == 0 ? 0 : igraph_vector_max(types) + 1;
+
+    if (igraph_matrix_ncol(pref) != no_of_types) {
         IGRAPH_ERRORF("Number of preference matrix colums (%ld) not "
                       "equal to number of types (%g).",
                       IGRAPH_EINVAL,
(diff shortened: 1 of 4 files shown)


src/games/citations.c
```diff
@@ -383,14 +384,20 @@ int igraph_citing_cited_type_game(igraph_t *graph, igraph_integer_t nodes,
...
     igraph_vector_t sums;
     long int nocats;
     long int i, j;
 
+    if (nodes == 0) {
+        return igraph_empty(graph, 0, directed);
+    }
     if (igraph_vector_size(types) != nodes) {
-        IGRAPH_ERROR("Invalid size of types", IGRAPH_EINVAL);
+        IGRAPH_ERRORF("Length of types vector (%ld), not equal to number of nodes (%" IGRAPH_PRId ").", IGRAPH_EINVAL, igraph_vector_size(types), nodes);
```
szhorvat, line 397: There's an unnecessary comma in this message after the first `(%ld)`.

Let's also break the line, it's very long.

src/games/citations.c
```diff
@@ -422,7 +429,7 @@ int igraph_citing_cited_type_game(igraph_t *graph, igraph_integer_t nodes,
     for (i = 0; i < nocats; i++) {
         long int type = (long int) VECTOR(*types)[0];
         if ( MATRIX(*pref, i, type) < 0) {
-            IGRAPH_ERROR("pref contains negative entries", IGRAPH_EINVAL);
+            IGRAPH_ERRORF("Preference matrix contains negative entry: %g.", IGRAPH_EINVAL, MATRIX(*pref, i, type));
```
szhorvat, line 432: Not mandatory, but the approach I use when the very same error may be triggered from multiple places is to put the error handling after the `return` at the end of the function, put a label on it, and jump to it with `goto`. `goto` is not always evil :-)  See `igraph_cited_type_game`.

szhorvat: Of course that wouldn't work here because we print the matrix element, so ignore my comment. Sorry.

src/games/citations.c
```diff
@@ -383,14 +384,20 @@ int igraph_citing_cited_type_game(igraph_t *graph, igraph_integer_t nodes,
...
     long int i, j;
 
+    if (nodes == 0) {
+        return igraph_empty(graph, 0, directed);
+    }
     if (igraph_vector_size(types) != nodes) {
-        IGRAPH_ERROR("Invalid size of types", IGRAPH_EINVAL);
+        IGRAPH_ERRORF("Length of types vector (%ld), not equal to number of nodes (%" IGRAPH_PRId ").", IGRAPH_EINVAL, igraph_vector_size(types), nodes);
+    }
+    if (igraph_matrix_ncol(pref) != igraph_vector_max(types) + 1) {
```
szhorvat, line 399: The docs say that the matrix should be square so unless I'm missing something, we should check the number of row as well.

szhorvat (changes requested):

PR #1704, opened 2021-02-26
Add static_power_law_game unit tests

GroteGnoom:
Part of #1592

This mainly checks for memory corruption, but not much more.

### DIFF
--- a/src/games/static_fitness.c
+++ b/src/games/static_fitness.c
@@ -108,7 +108,7 @@ int igraph_static_fitness_game(igraph_t *graph, igraph_integer_t no_of_edges,
     }
 
     if (no_of_edges < 0) {
-        IGRAPH_ERROR("Invalid number of edges.", IGRAPH_EINVAL);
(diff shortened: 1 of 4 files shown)


src/games/static_fitness.c
```diff
@@ -396,7 +396,7 @@ int igraph_static_power_law_game(igraph_t *graph,
 
     if (exponent_in >= 0) {
         if (exponent_in < 2) {
-            IGRAPH_ERROR("In-degree exponent must be >= 2.", IGRAPH_EINVAL);
+            IGRAPH_ERRORF("In-degree exponent must be either negative or >= 2, got %g.", IGRAPH_EINVAL, exponent_in);
```
szhorvat, line 399: I would not mention negative values here.

Different high-level interfaces have different ways to indicate that the graph should be undirected. For some of them, talking about a negative exponent will be confusing to users. A negative exponent is just an idiosyncratic way in the C interface to indicate that the graph must be undirected.

PR #1705, opened 2021-02-26
Add missing include

antonio-rojas:
Add a missing include for igraph_structural.h, where the function igraph_subcomponent is declared. Otherwise, the symbol igraph_subcomponent is not exported when building a shared library


szhorvat:
Thank a lot! The fix will be in the soon-to-be-released version 0.9.1.

PR #1709, opened 2021-02-26
Add transitivity_avglocal_undirected unit tests

GroteGnoom:
Part of #1592

### DIFF
--- a/src/properties/triangles.c
+++ b/src/properties/triangles.c
@@ -93,6 +93,9 @@ int igraph_transitivity_avglocal_undirected(const igraph_t *graph,
     igraph_vector_t degree;
     igraph_vector_t triangles;
 
+    if (igraph_vcount(graph) == 0) {
+        IGRAPH_ERROR("Graph should have at least 1 vertex, but has 0.\n", IGRAPH_EINVAL);
+    }
     IGRAPH_VECTOR_INIT_FINALLY(&order, no_of_nodes);
     IGRAPH_VECTOR_INIT_FINALLY(&degree, no_of_nodes);
 
@@ -112,7 +115,7 @@ int igraph_transitivity_avglocal_undirected(const igraph_t *graph,
 
     neis = igraph_Calloc(no_of_nodes, long int);
     if (neis == 0) {
-        IGRAPH_ERROR("undirected average local transitivity failed",
+        IGRAPH_ERROR("undirected average local transitivity failed.",
                      IGRAPH_ENOMEM);
     }
     IGRAPH_FINALLY(igraph_free, neis);
@@ -166,7 +169,7 @@ int igraph_transitivity_avglocal_undirected(const igraph_t *graph,
     igraph_vector_destroy(&rank);
     igraph_vector_destroy(&order);
     IGRAPH_FINALLY_CLEAN(5);
-    return 0;
+    return IGRAPH_SUCCESS;
 }
 
 int igraph_transitivity_local_undirected1(const igraph_t *graph,
--- a/src/properties/triangles.c
+++ b/src/properties/triangles.c
@@ -94,8 +94,14 @@ int igraph_transitivity_avglocal_undirected(const igraph_t *graph,
     igraph_vector_t triangles;
 
     if (igraph_vcount(graph) == 0) {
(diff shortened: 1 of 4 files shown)


src/properties/triangles.c
```diff
@@ -93,6 +93,9 @@ int igraph_transitivity_avglocal_undirected(const igraph_t *graph,
     igraph_vector_t degree;
     igraph_vector_t triangles;
 
+    if (igraph_vcount(graph) == 0) {
```
ntamas, line 96: I was wondering why this is necessary. Can't we return a sort-of-sensible answer for null graphs? @szhorvat likes these corner cases, I'd be interested in what he has to say here :) For me, returning zero or NaN would both make sense, depending on the value of the `mode` parameter, no need to throw an error.

ntamas (approved):
Looks good to me, let's see whether @szhorvat has something to say about the corner case of the null graph.

szhorvat:
I would go with zero for the null graph. Good catch that we had an assertion failure in that case!

image

I also chose 0 for `igraph_average_local_efficiency`, which is a similar measure in the sense that we deal with "pairs of neighbors" during computation.

Here's the reasoning:

The local clustering coefficient of a vertex is the fraction of its neighbour-pairs that are connected, i.e. `(connected neighbour-pairs) / (all neighbour-pairs)`. That is technically meaningless for any vertex that does not have at least degree 2 (two neighbours). One might argue to return 0/0 = NaN for such vertices.

But when it comes to the computation of the _average_ local clustering, a single NaN (a single degree-1 vertex) throws a wrench into the calculation. Thus, conventionally, people have been using 0 for the local clustering of degree-0 and degree-1 vertices. That means that the average local clustering would be 0 for 2-vertex and 1-vertex graphs, which don't make a lot more sense for this calculation than the zero-vertex graph. Thus for simplicity we can return 0 for the null graph as well.

----

In the end it doesn't matter much what we return, however, I would prefer to return a numeric result instead of throwing an error. Zero is just the simplest choice that avoids having to explain why we return something different from the also meaningless cases of 1- and 2-vertex graphs.

Plus, I'd like to stay consistent with `igraph_average_local_efficiency`.

szhorvat:
Handling unimportant looking edge cases gracefully (instead of erroring out) if generally useful. It often allows simply programs, written by the end user, to generalize nicely without having to handle edge cases specially.

With a graph measure `m` that is an average of some vertex measure, one would expect the following to hold:

```
m( DisjointUnion(g1, g1) ) = 
        ( m(g1) * VertexCount(g1) + m(g2) * VertexCount(g2) ) / (VertexCount(g1) + VertexCount(g2))
```

If `VertexCount(g2) == 0` and `VertexCount(g1) > 0`, this will hold for whatever _finite_ value we return for `m(g2)`, but it would break things if we return NaN.  This is an argument for returning 0 instead of NaN.

GroteGnoom:
>Zero is just the simplest choice that avoids having to explain why we return something different from the also meaningless cases of 1- and 2-vertex graphs.

But those return a value based on the `mode` argument. For example, in the 1-vertex case, if you specify IGRAPH_TRANSITIVITY_NAN, you get a NaN back. Should the 0-vertex case also do this? Or should the 1-vertex case (and lots of other cases) be changed? I think giving a zero in that mode will still have the 'disjoint union problem', and a NaN will at least indicate that things went wrong. So my suggestion is to return the value based on the mode.

(The disjoint union example is great by the way. This shows that the problem with NaNs in this case is that internally they mean 'this vertex should not be considered when calculating the average', but it doesn't mean that in other functions.)

szhorvat:
I'm sorry, I was looking at the function in a high-level interface and forgot that it has an option to exclude degree-0 and degree-1 vertices.

When we are excluding them, I would return NaN for the null graph. This is already what is returned for graph which do not have any vertices with degree > 1.

szhorvat:
This NaN could then serve to distinguish a graph in which every vertex had to be excluded from a graph where some vertices truly have 0 local clustering (such as a cycle graph).

szhorvat:
Just a heads up that the global clustering calculation also triggers an assertion failure when using the null graph. I will fix this now.

PR #1710, opened 2021-02-28
Add transitivity barrat unit tests

GroteGnoom:
Part of #1592

No longer warns for no edges and no weights.


szhorvat:
Thanks again!

Issue #1237, opened 2019-08-17
Wishlist: Betweenness using shortest paths between a subset of vertices


szhorvat:
The betweenness of a vertex (edge) is defined, loosely, as the number of shortest paths passing through that vertex (edge) when considering shortest path _between all pairs of vertices_.

It is possible to generalize this definition in various useful ways by not considering all-pairs shortest paths, but only a subset of them.

igraph can already compute the _range limited betweenness_ (which it calls a "betweenness estimate") by only considering shortest paths not longer than a cutoff.

Another useful generalization is to take shortest paths between vertex pairs taken _from a subset `S` of vertices_.  The betweenness would still be calculated for all vertices (edges), but it would only count shortest paths between vertices contained in `S`. A variation on this theme is to take paths starting in one subset `S_1` of vertices and ending in a different subset `S_2`. This is a feature request to implement this.  Ideally, the optional cutoff feature would be present too.

Potential applications:

 - In a large homogeneous graph, we can take a uniform random subset of vertices in order to be able to get a quick estimate of betweenness values

 - In a spatially embedded graph (think a geometric graph), we may be interested in paths going from one spatial region into another spatial region, and we may want to see where these paths pass through.

PR #1711, opened 2021-02-28
Subset betweenness

guyroznb:
This PR answers the issue #1237, adding two functions which allow the calculation of betweenness centrality (of edges and vertices) using shortest paths between s subset of vertices.


tests/unit/igraph_betweenness.c
```diff
@@ -3,22 +3,18 @@
    IGraph library.
    Copyright (C) 2008-2012  Gabor Csardi <csardi.gabor@gmail.com>
    334 Harvard street, Cambridge, MA 02139 USA
-
```
szhorvat, line 6: Please do not remove the empty lines from the license notice.

src/centrality/betweenness.c
```diff
@@ -3,17 +3,14 @@
 /*
    IGraph library.
    Copyright (C) 2007-2020  The igraph development team <igraph@igraph.org>
-
```
szhorvat, line 6: Please do not remove the empty lines from the license notice.

src/centrality/betweenness.c
```diff
@@ -1018,3 +1015,841 @@ int igraph_edge_betweenness_estimate(const igraph_t *graph, igraph_vector_t *res
...
+}
+
+/**
+ * \ingroup structural
+ * \function igraph_betweenness_subset
+ * \brief betweenness centrality for subset of vertices.
+ *
+ * </para><para>
+ * This function computes the subset version of betweenness centrality
+ * by considering only those shortest paths between vertices in a givem
```
szhorvat, line 1212: givem -> given

src/centrality/betweenness.c
```diff
@@ -1018,3 +1015,841 @@ int igraph_edge_betweenness_estimate(const igraph_t *graph, igraph_vector_t *res
...
+ * subset.
+ *
+ * \param graph The graph object.
+ * \param res The result of the computation, a vector containing the
+ *         betweenness score for thr subset of vertices.
```
szhorvat, line 1217: thr -> the

src/centrality/betweenness.c
```diff
@@ -1018,3 +1015,841 @@ int igraph_edge_betweenness_estimate(const igraph_t *graph, igraph_vector_t *res
...
+    }
+
+    is_target = igraph_Calloc(no_of_nodes, unsigned char);
+    if (is_target == 0) {
+        IGRAPH_ERROR("Can't calculate shortest paths", IGRAPH_ENOMEM);
+    }
+    IGRAPH_FINALLY(igraph_free, is_target); 
+
+    for (IGRAPH_VIT_RESET(vit); !IGRAPH_VIT_END(vit); IGRAPH_VIT_NEXT(vit)){
+        if (!is_target[(long int) IGRAPH_VIT_GET(vit)]){
```
szhorvat, line 1288: Why do you check here if the vector element is false before setting it to true?

src/centrality/betweenness.c
```diff
@@ -1018,3 +1015,841 @@ int igraph_edge_betweenness_estimate(const igraph_t *graph, igraph_vector_t *res
...
+            is_target[(long int) IGRAPH_VIT_GET(vit)] = 1;
+        }
+        else{
+            to_reach --;
+        }
+    }    
+
+    for (IGRAPH_VIT_RESET(vit); !IGRAPH_VIT_END(vit); IGRAPH_VIT_NEXT(vit)) {
+        source = IGRAPH_VIT_GET(vit);
+        IGRAPH_PROGRESS("Betweenness centrality: ", 100.0 * source / no_of_nodes, 0);
```
szhorvat, line 1098: There is no guarantee that the vertex iterator goes through vertices in order of increasing IDs, therefore this `IGRAPH_PROGRESS` would not increase monotonically.

We should either drop the progress indicator, or use a separate integer variable that is incremented by one in each cycle of the `for` loop.

src/centrality/betweenness.c
```diff
@@ -1018,3 +1015,841 @@ int igraph_edge_betweenness_estimate(const igraph_t *graph, igraph_vector_t *res
...
+        nneis = igraph_vector_size(res);
+        for (j = 0; j < nneis; j++) {
+            VECTOR(*res)[j] /= 2.0;
+        }
+    }
+
+    igraph_adjlist_destroy(&adjlist_out);
+    igraph_adjlist_destroy(&adjlist_in);
+    igraph_vit_destroy(&vit);
+    igraph_free(is_target);
```
szhorvat, line 1418: Please always use the macro `igraph_Free` instead of the function `igraph_free` when doing a direct call.

src/centrality/betweenness.c
```diff
@@ -1018,3 +1015,841 @@ int igraph_edge_betweenness_estimate(const igraph_t *graph, igraph_vector_t *res
...
+    IGRAPH_FINALLY_CLEAN(4);
+
+    return 0;
```
szhorvat, line 1421: Please use explicit names for error codes in all new code (i.e `IGRAPH_SUCCESS`)

src/centrality/betweenness.c
```diff
@@ -1018,3 +1015,841 @@ int igraph_edge_betweenness_estimate(const igraph_t *graph, igraph_vector_t *res
...
+        igraph_vector_fill(res, 0);
+    } else {
+        /* only  */
+        return igraph_betweenness_cutoff(graph, res, vids, directed,
+                                                    weights, -1);
+    }
+
+    is_target = igraph_Calloc(no_of_nodes, unsigned char);
+    if (is_target == 0) {
+        IGRAPH_ERROR("Can't calculate shortest paths", IGRAPH_ENOMEM);
```
szhorvat, line 1283: In all messages, give context about where the message is coming from: `Insufficient memory for subset betweenness.` In all new code, please follow the [error reporting guidelines](https://github.com/igraph/igraph/wiki/Error-reporting-guidelines), so we don't have to spend time to fix it up later.

src/centrality/betweenness.c
```diff
@@ -1018,3 +1015,841 @@ int igraph_edge_betweenness_estimate(const igraph_t *graph, igraph_vector_t *res
...
+        adjlist_out_p = &adjlist_out;
+        adjlist_in_p = &adjlist_in;
+    }
+    for (j = 0; j < no_of_nodes; j++) {
+        igraph_vector_int_clear(igraph_adjlist_get(adjlist_in_p, j));
+    }
+
+    distance = igraph_Calloc(no_of_nodes, long int);
+    if (distance == 0) {
+        IGRAPH_ERROR("betweenness failed", IGRAPH_ENOMEM);
```
szhorvat, line 1318: It seems that these messages were copied from an old version of the standard betweenness code. Please use the new message, and refer to the [error reporting guidelines](https://github.com/igraph/igraph/wiki/Error-reporting-guidelines).

src/centrality/betweenness.c
szhorvat, line 1288: There should be a space before the `{` here, and in several other places in the code.

szhorvat:
Before continuing:

Targets and sources:

 - It would be great if we could have a separate subset for source and for target vertices. This should ideally be done now, so we don't have to break API again at a later stage.

Tests:

 - Please separate the tests for the subset betweenness into a new test program The original betweenness tests are already overloaded.
 - Include both an undirected and a directed non-trivial example graph for which it is possible to verify the results manually. This should be more complex than simply a path graph.
 - Print headers before each test output, so one can easily see what result belongs came from which part of the code.

Duplicate vertices:

 - It seems that you are trying to handle the case when some vertices appear more than once in the subset. _If_ you want to handle this case, it should be tested. However, I think it would be easier to disallow duplicates.

Function signature:

 - The argument list of `igraph_betweenness_subset` is now literally identical to that of `igraph_betweenness`, but the meaning of `vids` is entirely different. This is very confusing. I suggest using this argument list instead:
   ```
   int igraph_betweenness_subset(
                                            const igraph_t *graph, igraph_vector_t *res,
                                            igraph_vs_t vids, igraph_bool_t directed,
                                            igraph_vs_t sources, igraph_vs_t targets,
                                            const igraph_vector_t *weights);
   ```
   `vids` will control which vertices the betweenness is calculated for. It is fine to calculate it for all of them, then retain only those values which were requested, just like `igraph_betweenness` does.  `sources` and `targets` specify the subsets.

Readability:

 - Please separate the weighted and unweighted parts into two internal functions, then call them from a top-level function, similar to how this is done for the harmonic centrality: https://github.com/igraph/igraph/blob/master/src/centrality/closeness.c#L798

----

 - It would be nice to eventually think about whether some code duplication can be avoided. But this is low priority. Let us first make sure that the current implementation is correct.

guyroznb:
Hi,
I changed the `betweenness.c` library, by adding two sub function which calculate the single SSPF for the weighted and the unweighted  cases. By doing so I can merge the betweenness function for the weighted and the unweighted cases and avoid duplication in the SSPF and betweenness calculation.
I also added a new test function for the betweenness subset.

szhorvat:
Thanks for all your work on this @guyroznb! I'll try to go through it next week (as it'll take a bit more time to review).

guyroznb:
I merge this branch with updated develop branch and re-formatted the src and tests.

szhorvat:
@guyroznb Just a heads up that we have not forgotten about this PR (or the shortest paths one). This week is exceptionally busy for all of us, so it will take a bit more time.

ntamas:
This PR seems to have a negative impact on the performance of the plain betweenness function according to the results of `make benchmark`.

Results in the `develop` branch without this PR:

```
 1 Betweenness, Kautz(4,5), directed                                             0.752s  0.749s  0.001s
 2 Betweenness, Kautz(4,5), undirected                                            1.91s    1.9s  0.006s
 3 Betweenness, DeBruijn(6,5), directed                                            1.9s   1.89s  0.004s
 4 Betweenness, Grid(8,8,8,8), undirected                                         1.21s   1.21s  0.002s
 5 Betweenness, Grid(10,10,10,10), cutoff 5                                      0.336s  0.335s  0.001s
 6 Betweenness, Grid(10,10,10,10), cutoff 8                                       1.38s   1.38s  0.003s
 7 Betweenness, Barabasi n=8000 m=1, undirected                                    3.3s   3.29s  0.006s
 8 Betweenness, Barabasi n=8000 m=1, undirected, cutoff 6                        0.895s   0.89s  0.003s
 9 Betweenness, Barabasi n=30000 m=5, directed                                    1.54s   1.53s  0.008s
10 Betweenness, Barabasi n=30000 m=5, directed, cutoff 5                         0.881s  0.877s  0.002s
```

Results in the branch of this PR:

```
 1 Betweenness, Kautz(4,5), directed                                              1.11s   1.09s  0.007s
 2 Betweenness, Kautz(4,5), undirected                                            4.03s    3.8s  0.043s
 3 Betweenness, DeBruijn(6,5), directed                                           2.73s   2.71s  0.012s
 4 Betweenness, Grid(8,8,8,8), undirected                                         1.92s   1.89s  0.014s
 5 Betweenness, Grid(10,10,10,10), cutoff 5                                      0.528s  0.517s  0.004s
 6 Betweenness, Grid(10,10,10,10), cutoff 8                                       2.05s   2.04s  0.005s
 7 Betweenness, Barabasi n=8000 m=1, undirected                                   4.53s    4.5s  0.016s
 8 Betweenness, Barabasi n=8000 m=1, undirected, cutoff 6                         1.16s   1.16s  0.003s
 9 Betweenness, Barabasi n=30000 m=5, directed                                    2.19s   2.17s  0.008s
10 Betweenness, Barabasi n=30000 m=5, directed, cutoff 5                          1.25s   1.24s  0.004s
```

We need to investigate why this happens and whether we can do something to avoid the performance hit.

ntamas:
It looks like the overhead of using incidence lists even for the node betweenness case has quite a bit of overhead; according to Instruments.app, 23.5% of the time spent in `igraph_i_sspf()` is spent in this line:

```
long int neighbor = IGRAPH_OTHER(graph, edge, actnode);
```

Attached is a side-by-side comparison of the two loops; left side is this PR, right side is the original version.

Screenshot 2021-04-13 at 19 04 00

szhorvat:
Putting an unpredictable conditional in a tight loop can do that ... 

I think if there is a performance cost of this magnitude, we should avoid the refactoring that is responsible for it. Performance is supposed to be a selling point for igraph.

guyroznb:
@ntamas I don't know which graph did you used in the benchmark (weighted or unweighted). In my opinion there are four different types of betweenness when talking about adjacent vs. incident list in the SSPF phase of the betweenness algorithm:
1) edge betweenness (weighted or unweighted) - using incident list.
2) vertex betweenness weighted - using incident list. This is also done in the `develop` branch.
3) vertex betweenness unweighted - I believe that in order to remain consistent with the other types of betweenness we should use incident list. But if there is a large degradation in the running time maybe I should write a new internal SSPF function using the adjacent list instead?

ntamas:
I used the standard igraph benchmarks that you can also run yourself with `make benchmark` (or `ninja benchmark`). @szhorvat has recently added a benchmark for weighted betweenness (I was using the unweighted variants only); I have merged these into `develop` now so if you merge `develop` into this PR, you can easily run the benchmarks yourself (both with the unweighted and the weighted variant).

I agree that one probably cannot avoid using `IGRAPH_OTHER()` and incidence lists if the graph is weighted or if you need any other edge properties - there is no debate for that. I would happily go for incidence lists for sake of consistency, as long as it does not cause a performance hit. If it does, then it's better to stick to adjacency lists.

Also, I think that it's not the incidence lists themselves that cause the slowdown but the presence of the `IGRAPH_OTHER()` macro; the result of the condition this macro is unpredictable so branch prediction optimizations will perform poorly.

Anyhow, I'm going to re-run the benchmarks and post the results there. @szhorvat 's benchmarks use a good mixture of all sorts of different graphs ranging from regular grids to Barabasi-Albert scale-free networks of all sizes that I believe they cover most use-cases that we wish to optimize for.

guyroznb:
Hi, I create 4 different SSPF for every scenario in order to improve the running time. In both vertex betweenness (weighted and unweighted) I used adjacent list to represent the father of each vertex to the source. In edge betweenness I used incident list to represent edges leading to the source. unweighted vertex betweenness is the only scenario in which the input is adjacent list.

ntamas:
This is great, now the benchmark results actually improved over the version that we currently have in the `develop` branch:

```
 1 Betweenness, Kautz(4,5), directed                                             0.579s  0.577s  0.001s
 2 Betweenness, Kautz(4,5), undirected                                            1.62s   1.62s  0.001s
 3 Betweenness, DeBruijn(6,5), directed                                           1.51s   1.51s  0.001s
 4 Betweenness, Grid(8,8,8,8), undirected                                         1.07s   1.06s  0.002s
 5 Betweenness, Grid(10,10,10,10), cutoff 5                                      0.306s  0.305s  0.001s
 6 Betweenness, Grid(10,10,10,10), cutoff 8                                       1.21s   1.21s  0.001s
 7 Betweenness, Barabasi n=8000 m=1, undirected                                   2.68s   2.67s  0.005s
 8 Betweenness, Barabasi n=8000 m=1, undirected, cutoff 6                         0.72s   0.72s      0s
 9 Betweenness, Barabasi n=30000 m=5, directed                                    1.19s   1.19s  0.002s
10 Betweenness, Barabasi n=30000 m=5, directed, cutoff 5                         0.734s  0.733s  0.001s
```

I'd like to check the implementations once again for correctness so please give me some more time, but it is looking much better now.

ntamas:
Note to ourselves: the new version does not include any progress reporting. This should be restored before merging this branch.

szhorvat:
@ntamas It appears that node betweenness (`igraph_betweenness`, not just `igraph_betweenness_subset`) now ignores multi-edges.

Example:

image

What is the betweenness of 2 and 4?  The only shortest paths going through them are between 1 and 3. There are not two, but three such paths. Two pass through 2 and one passes through 4. Therefore, the betweenness of 2 should be 2/3 and the betweenness of 4 is 1/3.

This is what the old version gives. The new version (from this PR) appears to ignore multi-edges and gives 1/2 for both of them.

szhorvat:
I tested the subset _node_ betweenness on the karate club graph, as well as random directed and undirected _simple_ graphs using random subsets. The implementation appears to be correct.

The direct implementation I used (through exhaustive enumeration of shortest path) does not handle multigraphs though.

szhorvat:
For future reference, I used this implementation in Mathematica to test against:

```
findAllShortestPaths[g_, s_, t_] :=
 With[{d = GraphDistance[g, s, t]},
  FindPath[g, s, t, {d}, All]
  ]

betweenness[g_, startSet_, endSet_] :=
 Module[{paths, counts, res},
  counts = Catenate@Catenate@Table[
      paths = findAllShortestPaths[g, s, t];
      {#, 1/Length[paths]} & /@ Join @@ paths[[All, 2 ;; -2]],
      {s, startSet}, {t, endSet}
      ];
  res = Lookup[
    Total /@ GroupBy[counts, First -> Last],
    VertexList[g],
    0
    ];
  If[UndirectedGraphQ[g], res/2, res]
  ]
```

szhorvat:
I have now also tested _edge betweenness_ on simple graphs (both directed and undirected, but not weighted). That is also fine.

szhorvat:
Let us now look at the edge betweenness of the same multigraph.

image

Consider edge 1-2.

There are 3 paths between 1,3, two of which pass through it, giving 2/3. There are 3 paths between 2,4, one of which passes through it, giving 1/3. There is 1 path between 1,2, passing through it, giving 1. In total, 2/3+1/3+1 = 2.

Same for 3-4.

For 1-4 we get 1/3 + 1/3 + 1 = 5/3.

Same for both edges between 2-3.

This is exactly the result I get from `igraph_betweenness_(subset)`. It is correct. Only the node betweenness is wrong for multigraphs, not the edge betweenness.

ntamas:
I'll add these as unit tests and attempt to fix the node betweenness case again so it works properly for multigraphs. Thanks a lot for the tests!

ntamas:
FWIW, edge betweenness on your example graph from `python-igraph` 0.9.3 gives me 7/6 for the edge betweenness scores of the mutal edge pair.

ntamas:
Looks good to me now! I'll merge when the CI tests pass. Thank you @guyroznb and @szhorvat !

szhorvat:
The betweenness parts of the IGraph/M test suite pass now (other parts fail, probably because of the callback function changes, but that's something else). I didn't re-test the subset betweenness yet.

szhorvat:
Neither did I do any tests on weighted graphs. I didn't look at the tests you wrote for this. Did you test with units weights and check that the result is always the same as the unweighted one?

szhorvat:
Also, awesome addition! I wanted this functionality for  long time.

PR #1714, opened 2021-03-01
Add igraph_shortest_paths_johnson unit tests

GroteGnoom:
Part of #1592

I changed the output of NaN's and Infs when they're printed with %8g,
for better alignment. This also required fixing modularity_matrix
output.

Bugfix: If there are no edges, there are no negative weights. This
check was not correct before.

Bugfix: When selecting a subset of to and from vertices, the `to`
indices were incorrect when reweighting the results. It used the indices
of the matrix for the whole graph instead of the vertex subset.


src/paths/johnson.c
```diff
@@ -91,7 +94,7 @@ int igraph_shortest_paths_johnson(const igraph_t *graph,
     }
 
     /* If no negative weights, then we can run Dijkstra's algorithm */
-    if (no_of_edges > 0 && igraph_vector_min(weights) >= 0) {
+    if (no_of_edges == 0 || igraph_vector_min(weights) >= 0) {
```
szhorvat, line 97: Could we combine the `igraph_vector_min` call with NaN detection, and get rid of the preceding `igraph_vector_is_any_nan` call?  `igraph_vector_min` returns NaN if any elements are NaN.

As usual, we must continue to ensure that vector_min is not called for a zero-length vector.

szhorvat: I will do this and merge.

PR #1716, opened 2021-03-02
Add igraph_average_path_length_dijkstra unit tests

GroteGnoom:
Part of #1592

Micro bugfix: Now returns 0 instead of -0 unconnected pairs for null graph.


szhorvat:
Thanks again!

PR #1717, opened 2021-03-02
Add spmatrix clear col unit tests

GroteGnoom:
Part of #1592

(I still had this one lying around.)


src/core/spmatrix.c
```diff
@@ -581,14 +581,18 @@ int igraph_spmatrix_add_rows(igraph_spmatrix_t *m, long int n) {
 
 /**
  * \function igraph_spmatrix_clear_row
- * \brief Clears a row in the matrix (sets all of its elements to zero)
+ * \brief Clears a row in the matrix (sets all of its elements to zero).
  * \param m The matrix.
  * \param row The index of the row to be cleared.
+ * \return Error code. The current implementation always succeeds.
```
ntamas, line 587: THis is not true, you have just added a code path that returns `IGRAPH_EINVAL` :)

src/core/spmatrix.c
```diff
@@ -670,7 +674,7 @@ static int igraph_i_spmatrix_cleanup(igraph_spmatrix_t *m) {
 
 /**
  * \function igraph_spmatrix_clear_col
- * \brief Clears a column in the matrix (sets all of its elements to zero)
+ * \brief Clears a column in the matrix (sets all of its elements to zero).
  * \param m The matrix.
  * \param col The index of the column to be cleared.
  * \return Error code. The current implementation always succeeds.
```
ntamas, line 680: This is not true any more, the current implementation can return `IGRAPH_EINVAL` if the column index is invalid.

ntamas (changes requested):

ntamas:
Minor doc issues; I'm gonna fix these and merge.

PR #1718, opened 2021-03-02
Add lexicographical comparison of vectors

### LINKED ISSUE #1706: Lexicographic comparison of igraph_vector_t
Two `qsort`-compatible functions that can do lexicographic comparison of `igraph_vector_(whatever)_t` objects. One should do it the standard way, the other should effectively consider the reverse of each vector.

**Use cases for the feature**
Canonicalizing the output of some functions, canonicalizing results in unit tests.

Why the reverse lexicographic version as well? Some functions sort the output by the last element of vectors, which is useful to preserve (e.g. `get_all_shortest_paths_dijkstra`)

GroteGnoom:
Fixes #1706


szhorvat:
@ntamas Do we want this function to be named `_rev_cmp` instead of `_cmp_rev`? As I recall, most (if not all) comparison functions in igraph are named as `cmp`.  Personally, I don't have strong feeling about this.

src/core/vector.pmt
```diff
@@ -898,7 +898,86 @@ long int FUNCTION(igraph_vector, qsort_ind)(TYPE(igraph_vector) *v,
     return 0;
 }
 
-#endif
+/**
+ * \function igraph_vector_lex_cmp
+ * \brief Lexicographical comparison of two vectors.
+ *
```
szhorvat, line 904: In order to avoid any ambiguity, it will be useful to spell out that if all elements match, but the vectors have different sizes, then the shorter one comes first.

src/core/vector.pmt
```diff
@@ -898,7 +898,86 @@ long int FUNCTION(igraph_vector, qsort_ind)(TYPE(igraph_vector) *v,
...
+ * \param lhs The first vector.
+ * \param rhs The second vector.
+ * \return -1 if \p lhs is lexicographically smaller,
+ *         0 if \p lhs and \p rhs are equal, else 1.
+ *
```
szhorvat, line 909: Let's add an `\sa` to the reverse variant.

src/core/vector.pmt
```diff
@@ -898,7 +898,86 @@ long int FUNCTION(igraph_vector, qsort_ind)(TYPE(igraph_vector) *v,
...
+        return 0;
+    }
+    /* a is shorter, and equal to the first part of b */
+    return -1;
+}
+
+/**
+ * \function igraph_vector_lex_cmp_rev
+ * \brief Reverse lexicographical comparison of two vectors.
+ *
```
szhorvat, line 942: Let us spell out that this compares vectors starting with the last element. Someone may misinterpret "reverse" as sorting in reverse order.

src/core/vector.pmt
```diff
@@ -898,7 +898,86 @@ long int FUNCTION(igraph_vector, qsort_ind)(TYPE(igraph_vector) *v,
...
+ * \param lhs The first vector.
+ * \param rhs The second vector.
+ * \return -1 if \p lhs in reverse order is
+ *         lexicographically smaller than the reverse of \p rhs,
+ *         0 if \p lhs and \p rhs are equal, else 1.
+ *
```
szhorvat, line 948: Let's add an `\sa` to the forward variant.

ntamas:
> @ntamas Do we want this function to be named `_rev_cmp` instead of `_cmp_rev`? As I recall, most (if not all) comparison functions in igraph are named as `cmp`. Personally, I don't have strong feeling about this.

Well, you could also argue that all comparison functions in igraph that return things in reverse order are named as `cmp_rev` ;) (It just happens to be the case that there are no such functions yet). So, I don't mind either way. `cmp_rev` (as it is now) is okay for me.

szhorvat:
> all comparison functions in igraph that return things in reverse order

This does not sort in reverse order. It sorts lexicographically based on reverse vectors :-)

Let's pause here for a minute and before merging, let us do some research on the various orderings, their namings (are the namings consistent?) and their applications.

Here is one reference:

https://oeis.org/wiki/Orderings#Types_of_orderings

GroteGnoom:
If we call it colexicographic ordering, then at least it doesn't seem like it just reverses the order. That's also the name used on [wikipedia](https://en.wikipedia.org/wiki/Reverse_lexicographic_order#Colexicographic_order).

szhorvat:
I changed the example slightly to make it clear that shorter vectors come first only if all elements match, but not otherwise. In some orderings, shorter always comes first.

szhorvat:
Let's call if `igraph_vector_colex_cmp` then.

src/core/vector.pmt
```diff
@@ -898,7 +898,100 @@ long int FUNCTION(igraph_vector, qsort_ind)(TYPE(igraph_vector) *v,
...
+    if (i == sb) {
+        return 0;
+    }
+    /* a is shorter, and equal to the first part of b */
+    return -1;
+}
+
+/**
+ * \function igraph_vector_colex_cmp
+ * \brief Coexicographical comparison of two vectors.
```
vtraag, line 947: Small typo:
```suggestion
 * \brief Colexicographical comparison of two vectors.
```

vtraag:
Just my two cents, after the comment from @szhorvat in https://github.com/igraph/igraph/issues/1706#issuecomment-789003260: I think it would be OK if this is targeted at the `master` branch. The more limited semantic versioning that we agreed we would support is that minor release versions are API backwards-compatible, i.e. they can add, but not break API. We documented this explicitly in the [Contributing guidelines](https://github.com/igraph/igraph/blob/master/CONTRIBUTING.md#branching)

szhorvat (approved):

ntamas (approved):

ntamas:
Hmm, I have tried to change the base branch but it seems to have added lots of stuff to `master` that I definitely do _not_ want to be there (like the API-breaking changes in `igraph_diameter`). @GroteGnoom would you mind sending another PR that targets the `master` branch cleanly? I don't know how to sort this out using the web UI of Github.

szhorvat:
@ntamas How about squash-and-merge to develop then cherry pick the commit?

ntamas:
Hmmm okay, let's try and see

PR #1720, opened 2021-03-03
Add attribute_combination_remove unit tests, docs

GroteGnoom:
Part of #1592

There were no docs, so I added some. They're not amazingly thorough, but can be used as a start.

### DIFF
--- a/doc/attributes.xxml
+++ b/doc/attributes.xxml
@@ -17,12 +17,21 @@
 <!-- doxrox-include igraph_attribute_type_t -->
 </section>
 
+<section id="attribute-combinations">
+<title>Handling attribute combination lists</title>
+<!-- doxrox-include about_attribute_combination -->
+<!-- doxrox-include igraph_attribute_combination_init -->
+<!-- doxrox-include igraph_attribute_combination_add -->
+<!-- doxrox-include igraph_attribute_combination_remove -->
+<!-- doxrox-include igraph_attribute_combination_destroy -->
+<!-- doxrox-include igraph_attribute_combination_type_t -->
+</section>
+
 <section id="accessing-attributes-from-c">
 <title>Accessing attributes from C</title>
-
 <!-- doxrox-include cattributes -->
 
-<section id="query-attributes"><title>Query attributes</title>
+<section id="query-attributes"><title>Query C attributes</title>
 <!-- doxrox-include igraph_cattribute_list -->
 <!-- doxrox-include igraph_cattribute_has_attr -->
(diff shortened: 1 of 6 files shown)


doc/attributes.xxml
```diff
@@ -17,12 +17,21 @@
...
+<!-- doxrox-include igraph_attribute_combination_type_t -->
+</section>
+
 <section id="accessing-attributes-from-c">
 <title>Accessing attributes from C</title>
-
 <!-- doxrox-include cattributes -->
 
-<section id="query-attributes"><title>Query attributes</title>
+<section id="query-attributes"><title>Query C attributes</title>
```
ntamas, line 34: I would leave this as is; all the titles are in a section titled "Accessing attributes from C" so there is no need to emphasize again that we are dealing with C attributes.

doc/attributes.xxml
```diff
@@ -58,7 +67,7 @@
 </section>
 
 <section id="set-attributes">
-<title>Set attributes</title>
+<title>Set C attributes</title>
```
ntamas, line 70: See above (the other section title).

doc/attributes.xxml
```diff
@@ -91,7 +100,7 @@
 <!-- doxrox-include SETEASV -->
 </section>
 
-<section id="remove-attributes"><title>Remove attributes</title>
+<section id="remove-attributes"><title>Remove C attributes</title>
```
ntamas, line 103: See above (the other section title).

tests/unit/igraph_attribute_combination_remove.c
```diff
@@ -0,0 +1,70 @@
...
+void print_comb(igraph_attribute_combination_t *comb) {
+    int i;
+    igraph_attribute_combination_record_t *r;
+    for (i = 0; i < igraph_vector_ptr_size(&comb->list); i++) {
+        r = VECTOR(comb->list)[i];
+        if (r->name) {
+            printf("name: %s\n", r->name);
+        } else {
+            printf("name: NULL\n");
+        }
```
ntamas, line 31: Let's also print (and validate) the type here. It is not needed to convert the type into a string, just cast it to an integer and print that.

ntamas (changes requested):
I have clarified the documentation a bit and requested a minor change in the test so we can also validate the type of an attribute combination record. Otherwise it's all good, thanks!

ntamas (approved):

ntamas:
Thanks!

Issue #1479, opened 2020-10-03
Calling resize_min on an empty vector causes runtime error


dahlbaek:
- [x] This issue is for the C core of igraph.
- [x] This issue is a bug report or a feature request, not a support question.

I'm not sure if this is in fact the intended behaviour, but when I try to call resize_min on an empty vector, the result is a runtime error. I'm not that strong in C, but as far as I can tell it happens because realloc called with size 0 will deallocate the array powering the vector:

https://github.com/igraph/igraph/blob/master/src/vector.pmt#L946-L963

Would it be reasonable to resize the capacity to 1 in case the vector is empty? I.e. change this implementation to something like

```
int FUNCTION(igraph_vector, resize_min)(TYPE(igraph_vector)*v) {
    size_t size;
    size_t capacity;
    BASE *tmp;
    if (v->stor_end == v->end) {
        return 0;
    }

    size = capacity = (size_t) (v->end - v->stor_begin);
    if (capacity == 0) {
        capacity = 1;
    }
    tmp = igraph_Realloc(v->stor_begin, capacity, BASE);
    if (tmp == 0) {
        IGRAPH_ERROR("cannot resize vector", IGRAPH_ENOMEM);
    } else {
        v->stor_begin = tmp;
        v->end = tmp + size;
        v->stor_end = tmp + capacity;
    }

    return 0;
}
```

szhorvat:
Yes, I agree that realloc_min should work for zero-size vectors. I'll look into it.

It seems that we cannot count on any realloc behaviour for the zero-size case:

https://en.cppreference.com/w/c/memory/realloc

> If new_size is zero, the behavior is implementation defined (null pointer may be returned (**in which case the old memory block may or may not be freed**), or some non-null pointer may be returned that may not be used to access storage).

vtraag:
Similar problems may actually arise with simply allocating a zero-length array using `malloc`. This is supported as an extension by `gcc`, but I believe it is not part of any C standard (although I can't find an exact references at the moment). I believe we ran into this issue of zero-length arrays in some other issue as well, but I can't recall exactly where. @szhorvat, perhaps you should then also make sure the initial capacity is always at least 1? Perhaps there are also some other cases in `vector.pmt` for which a minimum capacity of 1 should be ensured.

szhorvat:
Yes, and not only vector, but some other data structures as well.

vtraag:
By the way, this is the solution taken by suitesparse:

```
[...]

PR #1724, opened 2021-03-04
Add clique_size_hist unit tests

GroteGnoom:
Part of #1592

Bugfix: igraph_vector_resize_min used a Realloc, which returns NULL
 when called with 0 size, which was then interpreted as an error.
 A free and a calloc does not have the same problem, because calloc
 can return a pointer which can be properly freed, even when called
 with size 0. Calloc can also return NULL when called with size 0, but
 apparently doesn't do this, because else every vector_init with 0
 members would also have problems.


vtraag:
This actually touches on a broader issue discussed in https://github.com/igraph/igraph/issues/1479, where it is noted that having arrays of size 0 are not always necessarily well-supported. We there discussed the possibility of fixing this in the `igraph_Calloc` and friends by ensuring a minimum allocation size of 1 (even though this need not be used in practice). Perhaps you could take a look at that problem?

szhorvat:
@vtraag I was just going to summon you with a link to #1479 :-)

@GroteGnoom For all of malloc/calloc/realloc, the behaviour with a zero size is implementation defined, and cannot be relied on.

szhorvat:
Another note: `free()` is always safe to use on `NULL`. We do rely on this in several cases, for example when freeing a `vector_ptr` when not all of its elements have been allocated yet (e.g. because we're aborting due to an error). This is why `igraph_Free` not only calls `free()` but also sets the pointer to `NULL`

GroteGnoom:
Allright, I'll first fix #1479, and then revert a bit here.

ntamas (approved):

PR #1726, opened 2021-03-07
Make zero allocs consistent

### LINKED ISSUE #1479: Calling resize_min on an empty vector causes runtime error

GroteGnoom:
Fixes #1479 

--------
I did check for code where size checks were now superfluous, and
There are more places that check for size, like this:
```
if (alloc_size <= 0) {
    alloc_size = 1;
}
```
But I didn't remove them when they also used alloc size in other
places, like:
```
g->stor_end = h_str_begin + alloc_size;
```
-----------

Avoiding double evaluation can still be combined with macro usage like this:

```
static void *igraph_i_calloc(size_t nitems, size_t size) {
    return calloc (nitems > 0 ? nitems : 1, size);
}

#define IGRAPH_CALLOC(n,t)    igraph_i_calloc( n, sizeof(t) )
```

So that's still an option, but I also wanted to you to be able to merge this if you don't want that.

----------------

https://github.com/igraph/igraph/issues/1479#issuecomment-791977594: 
> Could we then also remove IGRAPH_FREE and IGRAPH_CALLOC from the public headers and move them to an internal header?

Should this still be done? It makes sense to me, but I'm not sure how igraph internal headers work.

### DIFF
--- a/include/igraph_memory.h
+++ b/include/igraph_memory.h
@@ -29,9 +29,9 @@
 
 __BEGIN_DECLS
 
-#define igraph_Calloc(n,t)    (t*) calloc( (size_t)((n) + 1), sizeof(t) )
-#define igraph_Realloc(p,n,t) (t*) realloc((void*)(p), (size_t)(((n) + 1)*sizeof(t)))
-#define igraph_Free(p)        (free( (void *)(p) ), (p) = NULL)
+#define IGRAPH_CALLOC(n,t)    (t*) calloc( (n) > 0 ? (size_t)(n) : (size_t)1, sizeof(t) )
+#define IGRAPH_REALLOC(p,n,t) (t*) realloc((void*)(p), (n) > 0 ? (size_t)((n)*sizeof(t)) : (size_t)1)
+#define IGRAPH_FREE(p)        (free( (void *)(p) ), (p) = NULL)
 
 IGRAPH_EXPORT void igraph_free(void *p);
 IGRAPH_EXPORT void *igraph_malloc(size_t n);
--- a/include/igraph_memory.h
+++ b/include/igraph_memory.h
@@ -33,6 +33,10 @@ __BEGIN_DECLS
 #define IGRAPH_REALLOC(p,n,t) (t*) realloc((void*)(p), (n) > 0 ? (size_t)((n)*sizeof(t)) : (size_t)1)
 #define IGRAPH_FREE(p)        (free( (void *)(p) ), (p) = NULL)
 
+#define igraph_Calloc IGRAPH_CALLOC
+#define igraph_Realloc IGRAPH_REALLOC
+#define igraph_Free IGRAPH_FREE
+
 IGRAPH_EXPORT void igraph_free(void *p);
 IGRAPH_EXPORT void *igraph_malloc(size_t n);
 
--- a/include/igraph_memory.h
+++ b/include/igraph_memory.h
@@ -29,8 +29,8 @@
 
 __BEGIN_DECLS
 
-#define igraph_Calloc(n,t)    (t*) calloc( (size_t)(n), sizeof(t) )
-#define igraph_Realloc(p,n,t) (t*) realloc((void*)(p), (size_t)((n)*sizeof(t)))
+#define igraph_Calloc(n,t)    (t*) calloc( (size_t)((n) + 1), sizeof(t) )
+#define igraph_Realloc(p,n,t) (t*) realloc((void*)(p), (size_t)(((n) + 1)*sizeof(t)))
 #define igraph_Free(p)        (free( (void *)(p) ), (p) = NULL)
 
 IGRAPH_EXPORT void igraph_free(void *p);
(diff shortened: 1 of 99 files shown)


ntamas:
@szhorvat argued [here](https://github.com/igraph/igraph/issues/1479#issuecomment-792291934) that `IGRAPH_FREE` and `IGRAPH_CALLOC` should be kept public so I think we don't need to hide them now.

Nevertheless, internal headers are no big deal; basically all the headers in `include` eventually get installed on the end user's system so everything you put in there is public. Headers in subdirectories of `src` are not installed so they can be used for private stuff that we do not want to expose as part of the public API.

src/core/memory.c
```diff
@@ -69,5 +69,5 @@ void igraph_free(void *p) {
  */
 
 void *igraph_malloc(size_t n) {
-    return malloc(n);
+    return malloc(n + 1);
```
ntamas, line 72: Can't we make this `(n > 0) ? 1 : n`? I feel like we could be screwing things up in cases when the user attempts to allocate a chunk of memory that is exactly a power of two in order to make the area align nicely with page boundaries.

src/core/stack.pmt
```diff
@@ -192,7 +192,7 @@ int FUNCTION(igraph_stack, push)(TYPE(igraph_stack)* s, BASE elem) {
 
         BASE *bigger = NULL, *old = s->stor_begin;
 
-        bigger = igraph_Calloc(2 * FUNCTION(igraph_stack, size)(s) + 1, BASE);
+        bigger = IGRAPH_CALLOC(2 * FUNCTION(igraph_stack, size)(s) + 1, BASE);
```
ntamas, line 195: I think the `+ 1` is not necessary here any more.

src/core/memory.c
szhorvat: I would not change `igraph_malloc`. Its purpose is not internal use at all, but to allow users to ensure that malloc/free are always used as a matching pair.

szhorvat: I see three internal uses of `igraph_malloc` (likely added by myself in the past). I would change these to `IGRAPH_CALLOC`.

szhorvat:
We might also add `#define igraph_Calloc IGRAPH_CALLOC` and `#define igraph_Free IGRAPH_FREE` _just in case_ someone was using these ... (the Py and R interfaces certainly do, and some Linux distros still want to compile py-igraph with their version of the igraph library—so let's not break compatibility between 0.9.0 and 0.9.1)

include/igraph_memory.h
```diff
@@ -29,9 +29,13 @@
 
 __BEGIN_DECLS
 
-#define igraph_Calloc(n,t)    (t*) calloc( (size_t)(n), sizeof(t) )
-#define igraph_Realloc(p,n,t) (t*) realloc((void*)(p), (size_t)((n)*sizeof(t)))
-#define igraph_Free(p)        (free( (void *)(p) ), (p) = NULL)
+#define IGRAPH_CALLOC(n,t)    (t*) calloc( (n) > 0 ? (size_t)(n) : (size_t)1, sizeof(t) )
+#define IGRAPH_REALLOC(p,n,t) (t*) realloc((void*)(p), (n) > 0 ? (size_t)((n)*sizeof(t)) : (size_t)1)
```
ntamas, line 33: There is a slight discrepancy here; `IGRAPH_CALLOC()` and the earlier version of `igraph_Realloc()` ensured that the size of the allocated memory was a multiple of `sizeof(t)`. `IGRAPH_CALLOC()` ensures that in the new version as well, but `IGRAPH_REALLOC()` does not. I don't know whether that's a concern; probably not, but I'd like a second opinion here. (@szhorvat ?)

ntamas:
Looks good to me, let's see what @szhorvat says about that single comment above.

include/igraph_memory.h
szhorvat: I think  it shouldn't be a problem, after all this is memory that we (should!) never use.

I am not too experienced with alignment issues, but I don't see how this could cause trouble ...

PR #1728, opened 2021-03-07
Add pseudo_diameter function

### LINKED ISSUE #1223: Implement pseudo-diameter computation
Implement an algorithm to compute the pseudo-diameter of a graph. This is a means of obtaining a quick approximation (and lower bound) for the graph diameter.

https://en.wikipedia.org/wiki/Distance_(graph_theory)#Algorithm_for_finding_pseudo-peripheral_vertices

GroteGnoom:
Fixes #1223


src/paths/distances.c
```diff
@@ -212,3 +238,70 @@ int igraph_radius(const igraph_t *graph, igraph_real_t *radius,
...
+ * and lower bound of the diameter of a graph.
+ *
+ * </para><para>
+ * A pseudo-peripheral vertex is a vertex v, such that for every
+ * vertex u which is as far away from v as possible, v is also as
+ * far away from u as possible.
+ *
+ * \param graph The input graph, if it is directed, its edge directions
+ *   are ignored.
+ * \param p_diameter Pointer to a real variable, the result is stored
```
szhorvat, line 257: Let's call this simply `diameter`

src/paths/distances.c
```diff
@@ -212,3 +238,70 @@ int igraph_radius(const igraph_t *graph, igraph_real_t *radius,
...
+    igraph_integer_t u_vid;
+
+    if (no_of_nodes == 0) {
+        *p_diameter = IGRAPH_NAN;
+        return IGRAPH_SUCCESS;
+    }
+
+    RNG_BEGIN();
+    u_vid = RNG_INTEGER(0, no_of_nodes - 1);
+    RNG_END();
```
szhorvat, line 282: I'm not sure how I feel about having randomness in a function like this. I would make the starting vertex a parameter.

src/paths/distances.c
```diff
@@ -212,3 +238,70 @@ int igraph_radius(const igraph_t *graph, igraph_real_t *radius,
...
+ *   here.
+ * \return Error code.
+ *
+ * Time complexity: O(|V||E|)), where |V| is the number of
+ * vertices and |E| is the number of edges.
+ *
+ * \sa \ref igraph_eccentricity(), \ref igraph_diameter().
+ *
+ */
+int igraph_pseudo_diameter(const igraph_t *graph, igraph_real_t *p_diameter) {
```
szhorvat, line 267: It would be useful to also return the endpoints of the pseudo-diameter, just like `igraph_diameter` does.

src/paths/distances.c
```diff
@@ -80,6 +89,18 @@ static int igraph_i_eccentricity(const igraph_t *graph,
             long int dist = igraph_dqueue_long_pop(&q);
             int j, n;
 
+            if (ecc_vid) {
+                /* Return the vertex id of the vertex which has the lowest */
```
szhorvat, line 93: Let's add a note that this is used for the pseudo-diameter calculation.

src/paths/distances.c
```diff
@@ -212,3 +238,70 @@ int igraph_radius(const igraph_t *graph, igraph_real_t *radius,
...
 }
+
+/**
+ * \function igraph_pseudo_diameter
+ * \brief Approximation and lower bound of diameter.
+ *
+ * This algorithm finds a pseudo-peripheral vertex and returns its
+ * eccentricity. This value can be used as an approximation
+ * and lower bound of the diameter of a graph.
+ *
```
szhorvat, line 249: It should be noted that this function computes the pseudo-diameter per-component. This is another reason to allow choosing the starting vertex: it will use the component to which that vertex belongs.

It should not be difficult to detect if the graph is disconnected during the eccentricity calculation, but let's think about whether it is useful to detect that case at all.

GroteGnoom: My best guess is yes, people don't always know what kind of graph they're working with, and not realizing the graph is disconnected can lead you to the wrong interpretation of the results. Disconnectedness seems like something that is always relevant when running this function, and the main reason not to check is if you already know up front if the graph is or isn't disconnected. I'm not really a user though :slightly_smiling_face:

szhorvat:
It would be nice to think about whether it is possible to implement a directed generalization.

We could perhaps use the same general approach, but from each vertex, look for _both_ the outgoing and incoming longest shortest path. Then take the endpoint of the longer one of the two as the next vertex. A bit of care will be needed to correctly assign the `to` and `from` endpoints of the result.

This approach will correctly recover the diameter of a directed path graph or cycle graph.

src/paths/distances.c
```diff
@@ -30,14 +30,16 @@
 #include "igraph_vector.h"
 #include "igraph_interface.h"
 #include "igraph_adjlist.h"
+#include "igraph_random.h"
```
szhorvat, line 33: I think this is no longer needed.

szhorvat:
A second generalization that I would love to see implemented is support for weights. Perhaps this can be a separate function, for consistency with `igraph_diameter_dijkstra`. We might also want to look for ways to reduce code duplication and re-use parts of the existing diameter calculation code, and perhaps even share it with eccentricity. The basic building block is the single-source shortest path calculation, which _should_ be re-usable between these.

src/paths/distances.c
szhorvat: Similar to `igraph_diameter`, we can add an `unconn` parameter that controls what happens. If true, the calculation will be done for the component of the starting vertex If false, infinity should be returned for a disconnected graph.

szhorvat: Another view is that the pseudo-diameter gives a lower bound for the diameter: this returning a finite result is acceptable when the true result is infinity.

src/paths/distances.c
```diff
@@ -161,7 +186,8 @@ int igraph_eccentricity(const igraph_t *graph,
                         igraph_vs_t vids,
                         igraph_neimode_t mode) {
 
-    return igraph_i_eccentricity(graph, res, vids, mode, /*adjlist=*/ 0);
+    return igraph_i_eccentricity(graph, res, vids, mode, /*adjlist=*/ 0,
```
ntamas, line 190: If we agreed that we will spell out null pointers explicitly with `NULL`, let's change it for `adjlist=...` as well for sake of consistency.

src/paths/distances.c
ntamas: Agreed.

ntamas: But maybe we could make it so that if the starting vertex index is negative, then igraph picks a random start vertex.

GroteGnoom:
> We might also want to look for ways to reduce code duplication and re-use parts of the existing diameter calculation code, and perhaps even share it with eccentricity. The basic building block is the single-source shortest path calculation, which _should_ be re-usable between these, and perhaps between other functions as well.

I can let igraph_i_eccentricity return a vector of ecc_vids, which will contain the endpoints of the paths furthest from the starting point, with the vertex with lowest degree selected if there are multiple endpoints which are furthest. This can then be used in both igraph_pseudo_diameter and igraph_diameter. Is that what you had in mind?

szhorvat:
Let's get it working correctly first for all cases (directed/undirected, weighted/unweighted) and we can see about sharing code between the implementations later.

I'd suggest to update this implementation to handle directed graphs, then do a separate PR for weighted (separate function), then finally think about reducing code duplication.

szhorvat:
Within this PR, I would also add an `unconn` argument, just like in `igraph_diameter()`. If it has ever happened that we could not reach all vertices, then we know that the graph is unconnected. Then if `unconn == false`, we return `IGRAPH_INFINITY`. We can note in the documentation that in the directed case, this function may not detect that the graph is disconnected, thus the result is still to be taken merely as a lower bound for the true diameter.

GroteGnoom:
For the directed case, because I'm checking shortest *in* and *out* eccentricities, I think a non-strongly connected graph is always detected.

ntamas:
Looks good to me now, let's wait for @szhorvat to see if he is okay with it as well

src/paths/distances.c
```diff
@@ -80,6 +91,19 @@ static int igraph_i_eccentricity(const igraph_t *graph,
             long int dist = igraph_dqueue_long_pop(&q);
             int j, n;
 
+            if (vid_ecc) {
+                /* Return the vertex id of the vertex which has the lowest */
+                /* degree of the vertices most distant from the starting */
+                /* vertex. Assumes there is only 1 vid in vids. Used for */
+                /* pseudo_diameter calculations. */
+                IGRAPH_CHECK(igraph_degree(graph, &degrees, igraph_vss_1(act), mode, IGRAPH_NO_LOOPS));
+                degree = VECTOR(degrees)[0];
```
szhorvat, line 100: A simple and fast way to query the degree of just one vertex could be through the `adjlist` (the length of adjacency list vectors) instead of through `igraph_degree`

src/paths/distances.c
```diff
@@ -30,14 +30,17 @@
...
 
 #include "core/interruption.h"
 
 static int igraph_i_eccentricity(const igraph_t *graph,
                                  igraph_vector_t *res,
                                  igraph_vs_t vids,
                                  igraph_neimode_t mode,
-                                 const igraph_adjlist_t *adjlist) {
+                                 const igraph_adjlist_t *adjlist,
+                                 igraph_integer_t *vid_ecc,
```
szhorvat, line 42: Let's add a comment above the function explaining the meaning of non-obvious parameters. In this case: `vid_ecc`.

src/paths/distances.c
szhorvat: I can see now that this is not so easy due to an unfortunate mismatch between the vector type used by `igraph_neighbors` and adjlists: the former is `igraph_real_t` and the latter `igraph_integer_t`.

szhorvat: Nevertheless the code does retrieve the degree in both cases, into `n`.

GroteGnoom:
In case you were wondering why the output changed: When calculating degrees before, I ignored loops. Now loops are not ignored.

szhorvat:
I keep looking at this use of `igraph_neighbors` and I find it a bit unusual. I don't understand its purpose.  It seems to be an attempted optimization for the case when the eccentricity is computed only for part of the vertices.

But to compute the eccentricity of just one vertex, we must do a BFS from that vertex. That involves traversing all vertices, thus retrieving the neighbours of all vertices. Therefore, nothing would be saved by not using a complete adjlist (passed to the function as an argument, just like it is now, in order to avoid creating a new adjlist for each invocation).

Am I missing something?

If I am, even then the better solution would be to allow a lazy adjlist (which perhaps didn't exist when this function was originally written).

@ntamas @GroteGnoom what do you think?

> In case you were wondering why the output changed: When calculating degrees before, I ignored loops. Now loops are not ignored.

That is not nice indeed. If we were to go with the full adjlist solution, this could easily be taken care of with the `IGRAPH_NO_LOOPS` setting in the `adjlist_init`.

szhorvat:
Can you pls update the copyright header with the new version and "2011-2021"?

GroteGnoom:
> But to compute the eccentricity of just one vertex, we must do a BFS from that vertex. That involves traversing all vertices, thus retrieving the neighbours of all vertices. 

This does not hold for disconnected or not strongly connected graphs. That's when the amount of igraph_neighbours calls is less than the amount of nodes.

szhorvat:
I changed the eccentricity function to use a full adjlist, like this (on `master`):

```
int igraph_eccentricity(const igraph_t *graph,
                        igraph_vector_t *res,
                        igraph_vs_t vids,
                        igraph_neimode_t mode) {

    igraph_adjlist_t adjlist;

    IGRAPH_CHECK(igraph_adjlist_init(graph, &adjlist, mode, IGRAPH_NO_LOOPS, IGRAPH_MULTIPLE));
    IGRAPH_FINALLY(igraph_adjlist_destroy, &adjlist);

    IGRAPH_CHECK(igraph_i_eccentricity(graph, res, vids, mode, /*adjlist=*/ &adjlist));

    igraph_adjlist_destroy(&adjlist);
    IGRAPH_FINALLY_CLEAN(1);

    return IGRAPH_SUCCESS;
}
```

Then I did some benchmarking with the Mathematica interface. Note that there's a high overhead from converting the graph from Mathematica format to igraph format, and that the shortest timings may not be reliable. The graph was connected.

Old implementation timings:

image

Timings with a full adjlist:

image

It seems that there's a noticeable gain from using the full adjlist, except maybe when calculating for a single vertex. In that case it's unclear if the difference is real, and whether the extra-short timings are reliable.

szhorvat:
I made a few small changes.

The final thing to think about is which degree (in? out? total?) to use to choose one of several max eccentricity vertices in the directed case. This is a minor part of the heuristic and it probably does not matter much, but I would like to understand the implications a bit better (tomorrow, with a clearer mind).

szhorvat:
One more comment is that when setting the error handler to ignore errors, it is best to immediately restore it when the function exits. When _want_ the program to be aborted (and the test to fail) when an unexpected error comes up. See last commit.

szhorvat:
We can turn off the automated comments (this was the first one, and a bit of a surprise).

https://docs.codecov.io/docs/pull-request-comments

What's nice is that now we can see whether all newly introduced functionality is covered by tests. This is especially relevant for functions which have many branches, such as `igraph_is_perfect`.

szhorvat:
The igraph_sir PR led to a whopping 0.17% increase. There's a lot of code, so it's slow. https://codecov.io/gh/igraph/igraph/commit/02fd620c4d419dae3f76c7b79fdfd1cde8a0e007/

We should avoid the temptation to read too much into the numbers. E.g. most of the `gengraph` (`degseq_vl`) code is unused, therefore untested. It's pointless to bother with it.  It's just a tool to find what is actually worth testing better, but increasing the numbers for the number's sake should be avoided.

szhorvat:
If everyone is happy with this, and there are no comments on the tie-breaking in the directed case, I'll merge this later tonight.

Issue #1382, opened 2020-05-05
Crash in igraph_feedback_arc_set


szhorvat:
`igraph_feedback_arc_set` crashed with the `IGRAPH_FAS_EXACT_IP` method with the attached graph.

Reproduced on macOS with the Mathemtica interface (compiled with external GLPK 4.65) as well as with python-igraph 0.8.2 from anaconda. 

The attached file is an edge list of a directed graph, with vertices indexed from 0.

[offending_graph.txt](https://github.com/igraph/igraph/files/4580952/offending_graph.txt)

szhorvat:
On the command line, I see this:

```
glp_add_rows: nrs = 758841790; too many rows
Error detected in file api/prob1.c at line 259
```

This is all fine and reasonable (I never expected the algorithm to finish on this huge graph), but the error should have been caught by igraph, and the abort should have been prevented. We need to look into why this did not happen.

szhorvat:
To prevent this from happening, we need to install a custom error handler using `glp_error_hook()`. This error handler must not return. Instead, it should call `glp_free_env` and then perform a `longjmp`.

ntamas:
A similar issue in [pywr/pywr](https://github.com/pywr/pywr/issues/759); might be interesting to take a look at how they ended up handling this.

One potential issue that we might have is that `glp_free_env()` frees _all_ GLPK resources on its own. If the user is running multiple GLPK problems in parallel (is that possible?), we could be inadvertedly freeing resources related to all of them, not only ours.

PR #1729, opened 2021-03-07
Full interruption support for GLPK

### LINKED ISSUE #897: `igraph_feedback_arc_set` should be interruptable
`igraph_feedback_arc_set` with method `IGRAPH_FAS_EXACT_IP` is not currently interruptible. Making this function interruptible is highly preferable because it can be hard to tell in advance whether it will finish in a reasonable amount of time for any given (directed) graph.

szhorvat:
This fixes #897 and #1382

Interruption does not work yet with the vendored GLPK, and I have not yet had time to figure out if this is because this old GLPK does not properly support it, or because changes were made to it. I have already undone some changes which would have prevented it from working but it still does not work. 

Nevertheless, comments are already welcome.

I am not totally happy with putting `igraph_abort()` into an internal `error.h`, as it is not easy to include in vendored sources.


szhorvat:
OK, I found it. Now interruption works even with the vendored GLPK, but it's a bit slow. Unfortunately, interruption is tied to output printing, so there is an opportunity to interruption only as often as there is terminal output. I might have missed some places where output was disabled, which may be slowing down interruption. It is definitely slower with the vendored GLPK than an external GLPK 5.0.

szhorvat:
This is not going to work as-is. It is leaking memory horribly on every interruption.

Interruption is not always interactive. A realistic use case is to compute the feedback arc set of a bunch of graph, with a time limit set on each (interruption when reaching the time limit). In fact, I have done precisely this with SageMath.

I need to understand what `glp_free_env()` does precisely and when it is safe to call it. If we do call it, we must avoid calling `glp_delete_prob()` after it, as it would lead to an abort/crash due to GLPK's overzealous error checking (see the checks in `glp_free`). Normally, `glp_delete_prob` would be called by `IGRAPH_FINALLY_FREE`.

szhorvat:
Now we call `glp_free_env()`, no more memory leak on interruption.

src/internal/glpk_support.c
```diff
@@ -32,6 +32,35 @@
...
+
+    if (igraph_i_interruption_handler) {
+        if (!igraph_i_glpk_error_info.is_interrupted && igraph_allow_interruption(NULL) != IGRAPH_SUCCESS) {
+            /* If an interruption has already occurred, do not set another error. */
+            igraph_i_glpk_error_info.is_interrupted = 1;
+            glp_error("GLPK was interrupted.");
+        } else {
+            /* Copy the error messages into a buffer for later reporting */
+            const size_t n = sizeof(igraph_i_glpk_error_info.msg) / sizeof(char) - 1;
+            while (*s != '\0' && igraph_i_glpk_error_info.msg_ptr < igraph_i_glpk_error_info.msg + n) {
```
ntamas, line 48: This is essentially an `strncpy()`, isn't it?

szhorvat: Yes, it is, but it seemed that due to the need of keeping track of the end of the string, and maintaining the null terminator, `strncpy` would not have simplified the code.

vendor/glpk/glpenv03.c
```diff
@@ -45,9 +45,9 @@
 
 void glp_printf(const char *fmt, ...)
 {     va_list arg;
-      /* va_start(arg, fmt); */
-      /* xvprintf(fmt, arg); */
-      /* va_end(arg); */
+      va_start(arg, fmt);
+      xvprintf(fmt, arg);
```
ntamas, line 49: Would the R interface (or, well, more like the CRAN validation of the R interface) choke on the presence of a printf-like function?

szhorvat:
I needed to do another hack and add `glp_at_error()` to our vendored GLPK to be able to determine which messages are error messages, and therefore should be collected for reporting.

Once we update the vendored GLPK, this hack will no longer be needed. New GLPK have this function.

szhorvat:
I plan to merge this later today.

Issue #1122, opened 2018-10-24
Wishlist: Test if a graph is perfect


szhorvat:
Add a function

    int igraph_is_perfect(igraph_t *graph, igraph_bool_t *res)

to test if a graph is [perfect](https://en.wikipedia.org/wiki/Perfect_graph).

[Basically done in IGraph/M](https://github.com/szhorvat/IGraphM/blob/master/IGraphM/LibraryResources/Source/IG.h#L2760), but needs to be ported to the igraph C core. (It uses no Mathematica-specific features)

PR #1730, opened 2021-03-07
Perfect graph

guyroznb:
This PR answers the issue #1122 adding the function igraph_perfect (which is implemented in IGraph/M).


src/properties/perfect.c
```diff
@@ -0,0 +1,143 @@
...
+   GNU General Public License for more details.
+
+   You should have received a copy of the GNU General Public License
+   along with this program; if not, write to the Free Software
+   Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
+   02110-1301 USA
+
+*/
+
+#include "igraph_structural.h"
```
szhorvat, line 24: I don't immediately see which function this header is for.

src/properties/perfect.c
```diff
@@ -0,0 +1,143 @@
...
+            igraph_destroy(&cycle);
+            igraph_destroy(&comp_graph);
+            igraph_vector_destroy(&edge_list);
+            IGRAPH_FINALLY_CLEAN(1);
+            return IGRAPH_SUCCESS;
+        }
+        igraph_destroy(&cycle);
+        igraph_vector_destroy(&edge_list);
+        IGRAPH_FINALLY_CLEAN(1);
+    }
```
szhorvat, line 139: add an `IGRAPH_ALLOW_INTERRUPTION()` here

tests/unit/igraph_perfect.c
```diff
@@ -0,0 +1,73 @@
...
+    igraph_bool_t is_perfect;
+    int x;
+
+    printf("bipartite \n");
+    printf("==========================================================\n");
+    igraph_bipartite_game(&graph, NULL, IGRAPH_ERDOS_RENYI_GNM, 10, 10, 0, 20, 0, IGRAPH_ALL);
+    igraph_is_perfect(&graph, &is_perfect);
+    IGRAPH_ASSERT(is_perfect == 1);
+    igraph_destroy(&graph);
+    x = IGRAPH_FINALLY_STACK_SIZE();
```
szhorvat, line 39: What is the purpose of `x`?

tests/unit/igraph_perfect.c
```diff
@@ -0,0 +1,73 @@
...
+
+    printf("complanetry to star graph size 10 - chordal\n");
+    printf("==========================================================\n");
+    igraph_star(&graph, 10, IGRAPH_STAR_UNDIRECTED, 0);
+    igraph_complementer(&comp_graph, &graph, 0);
+    igraph_is_perfect(&comp_graph, &is_perfect);
+    IGRAPH_ASSERT(is_perfect == 1);
```
szhorvat, line 46: Please use `IGRAPH_ASSERT(is_perfect)`. It is good practice not to assume a specific representation for "true".

tests/unit/igraph_perfect.c
```diff
@@ -0,0 +1,73 @@
...
+    igraph_destroy(&graph);
+    igraph_destroy(&comp_graph);
+    x = IGRAPH_FINALLY_STACK_SIZE();
+
+    
+    printf("A cycle of size 5\n");
+    printf("==========================================================\n");
+    igraph_ring(&graph, 5, 0, 0, 1);
+    igraph_is_perfect(&graph, &is_perfect);
+    IGRAPH_ASSERT(is_perfect == 0);
```
szhorvat, line 56: `IGRAPH_ASSERT(!is_perfect)`

tests/unit/igraph_perfect.c
```diff
@@ -0,0 +1,73 @@
...
+
+    printf("bipartite \n");
+    printf("==========================================================\n");
+    igraph_bipartite_game(&graph, NULL, IGRAPH_ERDOS_RENYI_GNM, 10, 10, 0, 20, 0, IGRAPH_ALL);
+    igraph_is_perfect(&graph, &is_perfect);
+    IGRAPH_ASSERT(is_perfect == 1);
+    igraph_destroy(&graph);
+    x = IGRAPH_FINALLY_STACK_SIZE();
+
+    printf("complanetry to star graph size 10 - chordal\n");
```
szhorvat, line 41: spelling: "complement"

tests/unit/igraph_perfect.c
```diff
@@ -0,0 +1,73 @@
...
+#include <igraph.h>
+#include "igraph_structural.h"
+
+#include "test_utilities.inc"
+
+int main() {
+    igraph_t graph, comp_graph;
+    igraph_bool_t is_perfect;
+    int x;
+
```
szhorvat, line 32: Since the test uses random graphs, set a random seed here.

tests/unit/igraph_perfect.c
```diff
@@ -0,0 +1,73 @@
...
+   You should have received a copy of the GNU General Public License
+   along with this program; if not, write to the Free Software
+   Foundation, Inc.,  51 Franklin Street, Fifth Floor, Boston, MA
+   02110-1301 USA
+
+*/
+
+#include <igraph.h>
+#include "igraph_structural.h"
```
szhorvat, line 24: Why is this `#include` needed?

tests/unit/igraph_perfect.c
```diff
@@ -0,0 +1,73 @@
...
+   but WITHOUT ANY WARRANTY; without even the implied warranty of
+   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+   GNU General Public License for more details.
+
+   You should have received a copy of the GNU General Public License
+   along with this program; if not, write to the Free Software
+   Foundation, Inc.,  51 Franklin Street, Fifth Floor, Boston, MA
+   02110-1301 USA
+
+*/
```
szhorvat, line 21: Please use the new version of the header (see above), with date 2021 only.

szhorvat (changes requested):
I left a few comments.

Since the test does not print any *results*, I suggest not to print anything at all and remove the `.out` file. Should anything fail, `IGRAPH_ASSERT` will indicate the line number of the failure.

src/properties/perfect.c
```diff
@@ -0,0 +1,143 @@
...
+ *
+ * </para><para>
+ * A perfect graph is a graph in which the chromatic number of every induced
+ * subgraph equals the order of the largest clique of that subgraph.
+ * The chromatic number of a graph G is the smallest number of colors needed to 
+ * color the vertices of G so that no two adjacent vertices share the same color 
+ *
+ * </para><para>
+ * This implementation is based on the strong perfect graph theorem which was
+ * conjectured by Claude Berge and prooved by Maria Chudnovsky, Neil Robertson,
```
szhorvat, line 51: "proved"

src/properties/perfect.c
```diff
@@ -0,0 +1,143 @@
...
+    if (!perfect) {
+        return IGRAPH_SUCCESS;
+    }
+    
+    IGRAPH_CHECK(igraph_is_bipartite(graph, &is_bipartite, NULL));
+    IGRAPH_CHECK(igraph_is_chordal(graph, NULL, NULL, &is_chordal, NULL, NULL));
+    // chordal and bipartite graph types are perfect. 
+    // possibly more optimizations found here: http://www.or.uni-bonn.de/~hougardy/paper/ClassesOfPerfectGraphs.pdf
+
+    if ((is_chordal) || (is_bipartite)) {
```
szhorvat, line 76: There is no need to parenthesize single variable names.

szhorvat:
There are a few more issues that I did not originally notice. The IGraph/M implementation has a note at the top that:

> this function assumes the input to be undirected, simple, connected

The Mathematica code makes sure not to call the C++ code with any graph that violates these assumptions. These need to be check in this version.

 - This implementation does not in fact ignore edge directions. It would fail to work correctly for directed graphs as `igraph_subisomorphic_lad` _does_ consider edge directions and would error when trying to find an undirected subgraph in a  directed graph.  Thus: Check if the graph is directed and throw an error if it is.
 
 - Multi-edges: Similarly, these are not ignored, but cause an error. If the graph is not simple, the function should throw an error.
 - Connectedness: I do not recall why I required this. The Mathematica code decomponses the graph into connected components and tests each one separately. Possibly I did this for performance (avoid creating huge complement graphs), but we need to think it through if connected graphs can be admitted.

Please also add the usual edge cases to the tests: null graph, singleton graph, graph with at least two vertices but no edges.

src/properties/perfect.c
```diff
@@ -0,0 +1,143 @@
...
+
+#include "igraph_structural.h"
+#include "igraph_bipartite.h"
+#include "igraph_operators.h"
+#include "core/interruption.h"
+#include "igraph_constructors.h"
+#include "igraph_topology.h"
+#include "igraph_interface.h"
+
+#include <limits.h>
```
szhorvat, line 32: Do we need this?

ntamas:
Regarding code formatting: the root of the repository contains an `.astylerc` file, which can be used in conjunction with `astyle` to format a file in a way that is mostly consistent with the rest of the codebase. Since this PR contains a new file, the easiest way to fix its formatting would be to pipe it through `astyle` with the `.astylerc` configuration file:

```sh
$ astyle --project=.astylerc src/properties/perfect.c
```

src/properties/perfect.c
guyroznb: `igraph_girth`

guyroznb:
Thanks for the review.
@szhorvat The `igraph_subisomorphic_lad` through an error if the graph has multi edges. Is that enough?
About the Connectedness I don't think this is an issue to address. The `subisomorphic` works with unconnected graph. What do you think?
@ntamas when I execute the line you suggested at the igraph directory I get the following error: 
> Cannot open project option file .astylerc

Do you know what is the problem?

szhorvat:
> @szhorvat The `igraph_subisomorphic_lad` through an error if the graph has multi edges. Is that enough?

I would prefer to have a check for non-simple graphs, so that the function behaves consistently and rejects _all_ (not just some) non-simple graphs. First do a check for directed graphs (`igraph_is_directed`) and throw an error if the graph is directed. Then use `igraph_is_simple` to check for non-simple graphs.

> @ntamas when I execute the line you suggested at the igraph directory I get the following error:

Do you see a file named `.astylerc` in the root directory of the repository?  Note that it is a hidden file. Which operating system are you using?

I can help you with this re-formatting if for some reason it does not work. Just let me know when I should do it, so we do not edit the code at the same time and create conflicts.

> About the Connectedness I don't think this is an issue to address.

Yes, it seems that this is mainly a performance issue that we can ignore for now.

szhorvat:
> @szhorvat it would be helpfull if could do the re-formatting.

I did the reformatting now. Is it possible that you did not run the command suggested by Tamás from the root directory?

szhorvat:
Here are a few issues I noticed and corrected, which I am mentioning for future PRs you might submit:

 - Always base your PRs on the _latest_ `develop` branch. This one was based on a rather old version of that branch.
 - `IGRAPH_ALLOW_INTERRUPTION()` checks if an interruption was requested. It makes sense in a loop, not at the end of the function (where it would serve no purpose). It should be called at a point where a `return` from the function is not harmful and will not lead to memory leaks. 
 - When changing the error handler in order to test invalid input, always set it back immediately after the function was called. This way, if any later functions fail in the test program, we can notice. Do this even if the error test is the last one. Someone may add more tests later.
 - The guidelines for error messages are [here](https://github.com/igraph/igraph/wiki/Error-reporting-guidelines). We would like to have all of them have the same style, i.e. use sentence case.

szhorvat (approved):
Thanks for your work on this. It looks good, but let me sleep on it one more night, and let's wait for Tamás's review.

szhorvat:
I just noticed it's not included in the docs. Can you please include it in `coloring.xxml` @guyroznb ?

guyroznb:
Thanks for this @szhorvat.
I will add the docs.
I ran the command from the root directory i don't know why it didn't work...

ntamas:
I went through the code as well. I am not too familiar with perfect graphs, so it might not make sense (complexity-wise, or readability-wise), but how about reorganizing the code such that first we check things that we could check _without_ creating the complementer (bipartiteness, chordality and girth of the _original_ graph) in the hope that we can bail out early, and _then_ create the complementer only if we could not reach a conclusion just by looking at the original graph?

I'm just a bit worried about the case when the user throws a graph with 1000 vertices and only a few edges at the function; if the graph itself matches one of the easy branches, there is no need to create the complementer with tons of edges.

@szhorvat you are the expert on this, so I leave this decision up to you.

I would probably also create a helper function that takes a graph and checks the "easy routes" without creating the complementer, and returns true/false/inconclusive (and also the girth in an additional output argument). You could then just call this helper function first on the original graph, then on the complementer, and bail out early if any of the tests was conclusive. If neither are conclusive, you have the girth of both graphs and then you could proceed with the more complex subisomorphism tests.

szhorvat:
> but how about reorganizing the code such that first we check things that we could check _without_ creating the complementer (bipartiteness, chordality and girth of the _original_ graph) in the hope that we can bail out early, and _then_ create the complementer only if we could not reach a conclusion just by looking at the original graph?

Yes, let's do this.



> I'm just a bit worried about the case when the user throws a graph with 1000 vertices and only a few edges at the function; if the graph itself matches one of the easy branches, there is no need to create the complementer with tons of edges.

Yes, this is a concern. In fact, the original version of this function in IGraph/M first decomposes the graph into connected components, and runs the test separately on each component. I believe one might even use bi-connected components instead. This is to avoid creating the complement of a large sparse graph.

Doing this and creating a large number of `igraph_t` objects (even one-by-one) might be a bit messy though ... However, given how slow `igraph_subisomorphic_lad` is, it might be worth it, at least for graphs above a certain size.  If we go this route, it will be worthwhile doing some benchmarking. We would need to answer questions like:

 - Above what size is decomposition into connected components worth it?
 - Is biconnected components worth it?
 - etc.

----

Yet another early optimization we can add is checking the size of the graph: any graph with <= 4 vertices is perfect.

szhorvat:
@guyroznb Can you please `develop` into this branch once more? I did not do it myself to avoid and conflicts with new edits you may have on your side.

guyroznb:
I think that even if the graph is very sparse, we should first find if the complementary graph is bipartite or chordal and then find if a cycle is  sub - isomorphic to the graph (or its complementary). The reason is that the running time of the function `igraph_is_bipartite` is `O(|V|+|E|)` and the function `igraph_is_chordal ` is `O(V)`, while the running time of `igraph_subisomorphic_lad` is exponential. 

Let say that we have a large graph `G = (V,E)` (which isn't bipartite or chordal) that has a complementary G' with `O(|V|^2)` edges. The running time of executing `igraph_is_chordal ` on G' isn't affected by the number of edges the input graph has, while the running time of executing `igraph_is_bipartite` of G' would increase by a factor `|V| = n` relative to the running time of `igraph_is_bipartite(G)` (`O(n+n^2) = o(n^2)`). This is still preferable to running an exponential function (especially on large graphs) on the sparse graph `G`.

ntamas:
> I think that even if the graph is very sparse, we should first find if the complementary graph is bipartite or chordal and then find if a cycle is sub - isomorphic to the graph (or its complementary). 

Yes, absolutely, there's no debate about that. My preferred order for the checks would be:

* check vertex count < 4 (as proposed by @szhorvat)
* check bipartiteness
* check chordality
* check girth
* check bipartiteness of complementer
* check chordality of complementer
* check girth of complementer
* do all sorts of subisomorphism tests as a last resort

So we are essentially deferring the creation of the complementer graph to after the bipartiteness, chordality and girth tests, but _before_ we start doing subisomorphism tests.

szhorvat:
Can you please add a few more tests, to make sure we cover all branches? The coverage tool is nice to verify this

Some ideas:

 - A random tree (`igraph_tree_game()`) — perfect
 - The complement of the 7-cycle — not perfect
 - 4x4 grid graph — perfect
 - A couple more from `igraph_famous`, e.g. the Chvatal graph is not perfect

szhorvat:
Note: it's `<= 4` not `< 4`.

szhorvat:
One more thing I'd do is to have the opportunity to return from the function after each check.

E.g., now both the bipartite and chordality checks are run in full before checking their results.

Doing many checks will create a lot of repetitive code that frees data structures. The usual way I deal with this is to create multiple return points at the end of the function and jump to them with `goto`, e.g.

```
clean1:
        igraph_destroy(&comp_graph);
        IGRAPH_FINALLY_CLEAN(1);
        return IGRAPH_SUCCESS;

clean2:
        igraph_destroy(&cycle);
        igraph_destroy(&comp_graph);
        IGRAPH_FINALLY_CLEAN(2);
        return IGRAPH_SUCCESS;
```

then you just `goto clean1` or `goto clean2` as needed.

szhorvat:
I am happy to merge this if there are no additional comments.

szhorvat:
It is nice to see that Covecov will edit the original report ([above](https://github.com/igraph/igraph/pull/1730#issuecomment-795956133)) with the new data instead of posting more comments.

szhorvat:
It seems like you accidentally checked the vertex count, not the edge count, at the beginning of the function. I fixed this and retained an edge count check as well.

I will merge once the tests pass. Azure Pipelines is having problems today.

szhorvat:
Thanks for the contribution!

PR #1734, opened 2021-03-11
Add solve_lsap unit tests

GroteGnoom:
Part of #1592

I only noticed this was in /src/internal after adding tests.
It is not used internally. I therefore only did the minimal amount
of work to check that this works correctly in some cases, and disallow other
cases. If we want this to be part of public igraph, I can fix some
more things, like missing IGRAPH_CHECK calls, the malloc and calloc
calls, and the superfluous *n* parameter which can be taken from the matrix.

I think it's useful, and as it can be seen as a bipartite weighted graph problem I think it fits igraph, and it's worth the work to clean it up a bit.

### DIFF
--- a/src/internal/lsap.c
+++ b/src/internal/lsap.c
@@ -615,10 +615,51 @@ void preprocess(AP *p) {
     }
 }
 
+/**
+ * \function igraph_solve_lsap
+ * \brief Solve a balanced linear assignment problem.
+ *
+ * This functions solves a linear assinment problem using the Hungarian
+ * method. A number of tasks, an equal number of agents, and the cost
+ * of each agent to perform the tasks is given. This function then
+ * assigns one task to each agent in such a way that the total cost is
+ * minimized.
+ *
+ * </param><param>
+ * If the cost should be maximized instead of minimized, the cost matrix
+ * should be negated.
+ *
+ * </param><param>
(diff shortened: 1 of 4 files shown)


szhorvat:
This function is used by the R interface to implement [match_vertices](https://igraph.org/r/doc/match_vertices.html) which is a sort of "approximate isomorphism".  I would have liked to see this ported to the C core for a long time, but I have not had time to even review the math behind it.

gaborcsardi (approved):

ntamas (approved):

ntamas:
It seems like we have two implementations of the Hungarian algorithm now in igraph. One is in `internal/lsap.c`, and the other one is implicitly used by `igraph_maximum_bipartite_matching()`. I didn't know about this one. It seems like most of the code in `lsap.c` came from elsewhere; I tried to find the original implementation on the Internet by searching for stuff like `ap_hungarian()` but nothing useful came up. It would be great to know where the original code came from and whether there are unit tests for the original code that we could use. (But all of this is not urgent now).

ntamas:
Earlier versions of the `ap_assignment()` function appear in the Github history of the `clue` R package ([see here](https://github.com/cran/clue/commits/78f0849a6d816f27fbd70b1a8887b8259bfc4c86/src/assignment.c)) so this is probably the original source.

PR #1736, opened 2021-03-12
Add pseudo_diameter_dijkstra

### LINKED ISSUE #1733: Implement weighted pseudo-diameter computation
This is a feature request to implement a weighted pseudo-diameter computation as `igraph_pseudo_diameter_dijkstra`.

Ref: #1223 and #1728

GroteGnoom:
fixes #1733


szhorvat:
@vtraag @ntamas Since this PR already includes `igraph_i_eccentricity_dijkstra` anyway, would it be overkill to add a _public_ weighted eccentricity function as well? It would be minimal work. I would vote to do it.

szhorvat (changes requested):
For the weighted eccentricity calculation, I would prefer to avoid creating multiple inclists, and avoid using adjlists completely. Each call to `igraph_shortest_paths_dijkstra` creates a new inclist, which in my experience can have a noticeable performance impact.

I would suggest modelling this on `igraph_i_eccentricity`:

 - The function should take a pre-generated (perhaps lazy) inclist as an argument, and use it.
 - The function should contain a full-blown implementation of the Dijkstra algorithm. This can be mostly copied from the implementation of `igraph_diameter_dijkstra`.
 
Let us do this as a first step.

Later, we can:

 - Add a public weighted eccentricity function (perhaps in this same PR).
 - Once everything is done, it will be much easier to see where code duplication can be eliminated without sacrificing performance. We can do this later, not in this PR.

GroteGnoom:
The inclist does not have an option to ignore multiple edges, should I add an inclist function that removes them?

szhorvat:
> The inclist does not have an option to ignore multiple edges, should I add an inclist function that removes them?

I would say no.

 - The degree is used as a tie-breaker heuristic when there is more than one vertex at the same maximal distance. It seems to me that this is not a critical algorithm. Result would likely be good even if ties are broken randomly.

 - Especially in the weighted case, in most practical use (which I assume would use real-values weights), there are unlikely to be ties. Ties are much more likely with small integer weights or unit weights (i.e. unweighted).

 - There are currently no guarantees about the ordering of edges in an inclist, and we don't want to add guarantees so that we can experiment with different graph storage formats in the future. Removing parallel edges is not simple, and without guarantees on edge ordering, it is computationally expensive. Furthermore, if the edges have their own identities (as is the case in an inclist: two parallel edges have different IDs), they can have different properties, such as different weights. Thus, removing them is not something one should normally do. I would therefore not want to have such a function.

----

I did a brief search trying to find out the motivation behind the degree-based tie-breaking heuristic. I did not find anything. If you do, let us know. While it seems intuitive, it also seems of minor importance. I propose that we accept the small imperfection of not ignoring multi-edges when computing degrees.

szhorvat:
Just a heads up that this week is going to be super busy for all of us (I'm attending a conference). Reviews are likely to be slow, and some of us might not show up every day.

GroteGnoom:
I requested a review, not because I'm in a hurry, but I thought maybe this would remove the 'Changes requested', but that didn't work. 

Apparently I can resolve inline comments, but can't resolve requested changes.

I just wanted to indicate that I think I made the requested changes :slightly_smiling_face:

szhorvat:
Something's wrong, not with the Dijkstra version, but with the unweighted version. I am getting smaller pseudo-diameters than true diameters for some graphs. I am trying to debug it now.

szhorvat:
I'm sorry, I am being stupid. I need more sleep.

szhorvat:
There is a minor problem in that there may be a null-pointer dereference if `from` is given but `to` isn't. That is a ridiculous situation that's probably not worth bothering with ...

szhorvat (approved):

szhorvat:
Looks good to me.

The unweighted and weighted versions will not give the same result for the same starting vertex, and I didn't take the time to understand where the difference is. But I do not think that this is a problem, and I do not think that users should expect them to return the same result.

I'm feeling a bit foggy today, so I will leave merging to you @GroteGnoom @ntamas

PR #1739, opened 2021-03-15
Add motifs_randesu_estimate unit tests

GroteGnoom:
Part of #1592

### DIFF
--- a/src/misc/motifs.c
+++ b/src/misc/motifs.c
@@ -114,6 +114,10 @@ int igraph_motifs_randesu(const igraph_t *graph, igraph_vector_t *hist,
         IGRAPH_ERROR("Only 3 and 4 vertex motifs are implemented",
                      IGRAPH_EINVAL);
     }
+    if (igraph_vector_size(cut_prob) != size) {
+        IGRAPH_ERRORF("Cut probability vector size (%ld) must agree with motif size (%" IGRAPH_PRId ").",
+                      IGRAPH_EINVAL, igraph_vector_size(cut_prob), size);
+    }
     if (size == 3) {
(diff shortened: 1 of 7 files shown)


src/misc/motifs.c
```diff
@@ -424,11 +424,18 @@ int igraph_motifs_randesu_callback(const igraph_t *graph, int size,
 
 /**
  * \function igraph_motifs_randesu_estimate
- * \brief Estimate the total number of motifs in a graph
+ * \brief Estimate the total number of motifs in a graph.
+ *
+ * This function estimates the total number of connected induced
```
szhorvat, line 429: weakly connected

src/misc/motifs.c
```diff
@@ -479,9 +486,28 @@ int igraph_motifs_randesu_estimate(const igraph_t *graph, igraph_integer_t *est,
     long int sam;
     long int i;
 
+    if (no_of_nodes == 0) {
+        *est = 0;
+        return IGRAPH_SUCCESS;
+    }
```
szhorvat, line 492: Let's move this below the following checks, so that these checks would be performed consistently for graphs of any size (even 0).

szhorvat:
Thanks again! Just a few comments, I'll fix these myself and merge.

szhorvat:
@GroteGnoom Unfortunately I need to leave for today. If you have time, can you please update both this function, as well as other motif functions, and corresponding tests, to check that the cut probability vector size is the same as the motif size? If not, I'll do it in a couple of days.

GroteGnoom:
Allright, I will fix that, and I will also add the motifs_randesu_no tests in this PR.

szhorvat:
Thank you for fixing this!

PR #1745, opened 2021-03-26
use CMAKE_INSTALL_FULL_<dir> in igraph.pc

dotlambda:
`CMAKE_INSTALL_<dir>` cannot be assumed to be relative to `CMAKE_INSTALL_PREFIX`.
See https://github.com/NixOS/nixpkgs/pull/117512#issuecomment-808059541.

### DIFF
--- a/igraph.pc.in
+++ b/igraph.pc.in
@@ -1,7 +1,5 @@
-prefix=@CMAKE_INSTALL_PREFIX@
-exec_prefix=@CMAKE_INSTALL_PREFIX@
-libdir=${exec_prefix}/@CMAKE_INSTALL_LIBDIR@
-includedir=${prefix}/@CMAKE_INSTALL_INCLUDEDIR@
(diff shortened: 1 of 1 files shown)


ntamas:
Thanks a lot!

vtraag:
Just as an aside, following up from https://github.com/jtojnar/cmake-snips#assuming-cmake_install_dir-is-relative-path, the fix that @jtojnar explicitly refers to in https://github.com/fmtlib/fmt/pull/1657 leaves in the `prefix` and `exec_prefix`.

ntamas:
I have implemented the full solution from https://github.com/jtojnar/cmake-snips#assuming-cmake_install_dir-is-relative-path; seems to work fine with both absolute and relative paths now, and relative paths keep the `${prefix}` and `${exec_prefix}` variables in `igraph.pc` correctly.

PR #1750, opened 2021-04-06
Add is_mutual unit tests

GroteGnoom:
Part of #1592


ntamas:
Good to go if it passes CI.

PR #1752, opened 2021-04-07
Graph spanner

guyroznb:
This PR adds an algorithm for finding graph spanners. A spanner of a graph `G = (V,E)` with a stretch t is a subgraph `H = (V,Es)` such that `Es` is a subset of `E` and the distance between any pair of nodes in `H` is at most `t` times the distance is `G`. The returned graph is always a spanner of the given graph with the specified stretch.

The algorithm is based on the one introduced by baswana et al. : "A Simple and Linear Time Randomized Algorithm for Computing Sparse Spanners in Weighted Graphs". 
[https://onlinelibrary.wiley.com/doi/pdf/10.1002/rsa.20130?casa_token=Fesd03rNu1MAAAAA:9Aty-TF0vlQ4hFT3O3Bk-KU2NkdIKM7-dQ9qsxaNTxqH9JazjCtWAhqOTeqA9JwOr0_hyy6gjyDMag](url).

Thanks...


guyroznb:
I don't know why the documentation test failed. Can someone point me to the reason?

src/paths/sparsifier.c
```diff
@@ -0,0 +1,403 @@
...
+ *        the function would save the weights of the edges in the spanner
+ *        accourding to the edges IDs in the spanner graph.
+ * \param seed A pointer to unsigned long int or NULL. If not NULL would be
+ *        used a seed for the rng.
+ * 
+ * \return Error code:
+ *        \clist
+ *        \cli IGRAPH_ENOMEM
+ *           not enough memory for temporary data.
+ *
```
szhorvat, line 105: Put an `\endclist` here.

ntamas:
Hopefully [this paper](https://www.cse.iitd.ernet.in/~ssen/journals/randstruc.pdf) is the same as the one cited in the PR description, but it is not behind a paywall.

szhorvat:
I added a comment, but don't worry about the documentation, that can be fixed at the end.

src/paths/sparsifier.c
guyroznb: Thanks

doc/structural.xxml
```diff
@@ -13,6 +13,10 @@
 <!-- doxrox-include igraph_are_connected -->
 </section>
 
+<section id="sparsifiers"><title>sparsifiers</title>
```
szhorvat, line 16: Capitalize titles.

tests/unit/igraph_spanner.c
```diff
@@ -0,0 +1,147 @@
...
+    return IGRAPH_SUCCESS;
+}
+
+int main () {
+    igraph_t graph, spanner;
+    igraph_vector_t weights, spanner_weight;
+    igraph_bool_t check_spanner;
+    igraph_rng_t rng;
+    long int no_of_edges, no_of_edges_spanner;
+    IGRAPH_CHECK(igraph_rng_init(&rng, &igraph_rngtype_mt19937));
```
szhorvat, line 83: Please do not use `IGRAPH_CHECK` here. It should only be used in functions that return an igraph error code. In the future the type of igraph error codes is likely to change, and may not be compatible with a program exit code.

(All that said, use the default RNG unless there is a very good reason to create your own RNG.)

szhorvat:
Before we get started, a few small things:

 - Please do not use a private random number generator and remove the `seed` argument. Use the default RNG.  To generate a real, you can use `RNG_UNIF(min, max)`. There are several other similar `RNG_` functions. Sections of code that use the `RNG_` macros should be fenced in `RNG_BEGIN();` and `RNG_END();`. You will find many examples in the `games` directory, e.g. in `callaway_traits.c`.
 - I notice that the `stretch` argument is an `int`. In the public API, let us always use igraph-specific types, i.e. `igraph_integer_t`, `igraph_real_t`, etc. Can you explain why this needs to be an integer? Does this specific algorithm work only for integer stretch factors?
 - Do not use function names starting with `_`. Functions that are private to a certain file should be marked `static`, and can be named anything (but don't start with `_`). Functions that are internal, but not private to a file, should be named similarly to `igraph_i_myfun`.
 - Make tests deterministic: don't seed the RNG from `time(0)`. Is there any special reason you created a new RNG instead of using the default one?
 
These are just a few starting comments, not a real review. I need to hook the function up to IGraph/M and test it before starting a proper review.

szhorvat:
I see you changed the type of the stretch factor to real. Can you confirm then that the algorithm works for arbitrary non-integer values? I have not yet looked at the paper in detail. Does it say anything about this value being integer or real?

guyroznb:
No that was my mistake. In the paper the stretch factor `t` is defined by `for any t ≥ 1`

interfaces/functions.def
```diff
@@ -504,6 +504,13 @@ igraph_get_all_simple_paths:
         PP-R: get.all.simple.paths.pp
         IGNORE: RR
 
+igraph_spanner:
+        PARAMS: GRAPH graph, OUT GRAPH spanner, INT stretch, \
```
ntamas, line 508: This should be a `REAL` if the stretch factor can be an arbitrary number greater than 1.

tests/unit/igraph_spanner.c
ntamas: I've sorted this out.

ntamas:
Unfortunately the current implementation is at least O(n^2) due to the `igraph_vector_fill()` calls on vectors of length `no_of_nodes` (e.g., `lightest_neighbor` and `lightest_weight`) that are called in a loop that also has `no_of_nodes` iterations. We need to work on this to achieve the linear time complexity claimed by the original paper.

ntamas:
I am going through the paper and correlating the steps of the algorithm in the paper with the steps of the algorithm in the implementation. I have found no traces of an _absolute_ upper bound on the number of edges added per iteration, only a few lemmas that estimate the number of edges added in terms of expected value, so I have removed the size limit. Let me know if anyone disagrees with it.

ntamas:
Okay, I think it's good to go now. I'll wait for the CI tests to pass and then I'll merge. Thanks for your contribution!

guyroznb:
Hi, thanks @ntamas  for the review. I didn't have that much time recently so I wasn't able to do this myself

PR #1758, opened 2021-04-12
fix: improper executable bit

jgmbenoit:
Make non executable some source files, some data files, and some pictures.


ntamas (approved):
Thanks a lot!

PR #1763, opened 2021-04-15
Add k shortest paths

### LINKED ISSUE #1397: Feature request: K shortest paths
I would like to see "K shortest paths" functionality added to the core of igraph. This function would find the K shortest paths through a weighted graph using [Yen's algorithm](https://en.wikipedia.org/wiki/Yen%27s_algorithm).

The yenpathy package implements this, but as you can see in the issue tracker, there are several outstanding issues with the implementation that have not been addressed in several months (https://github.com/ecohealthalliance/yenpathy/issues). 

Thank you for your contributions to this community.

GroteGnoom:
Fixes #1397

Non-obvious things:
I use infinite weights to remove edges. This way I can use get_shortest_path_dijkstra directly.
I only give the path in edges. Vertices can be inferred from this afterwards.


szhorvat:
I'm going to add some comments on this, but for now they are mostly self-reminders. No need to act on them yet.

src/paths/shortest_paths.c
```diff
@@ -1275,3 +1276,326 @@ int igraph_diameter_dijkstra(const igraph_t *graph,
...
+    long int i;
+    igraph_vector_t weights_old;
+    igraph_vector_t edges_removed;
+    igraph_bool_t added_path, null_weights;
+    long int nr_edges = igraph_ecount(graph);
+    igraph_bool_t infinite_path, already_in_potential_paths;
+    igraph_vector_t *path_0;
+
+    igraph_vector_ptr_free_all(paths);
+    IGRAPH_VECTOR_PTR_SET_ITEM_DESTRUCTOR(paths, igraph_vector_destroy);
```
szhorvat, line 1384: Do not change the item destructor permanently.

src/paths/shortest_paths.c
```diff
@@ -1275,3 +1276,326 @@ int igraph_diameter_dijkstra(const igraph_t *graph,
...
+    IGRAPH_VECTOR_PTR_SET_ITEM_DESTRUCTOR(&paths_pot, igraph_vector_destroy);
+    IGRAPH_FINALLY(igraph_vector_ptr_destroy_all, &paths_pot);
+    IGRAPH_VECTOR_INIT_FINALLY(&path_spur, 0);
+    IGRAPH_VECTOR_INIT_FINALLY(&weights_old, 0);
+    IGRAPH_VECTOR_INIT_FINALLY(&edges_removed, 0);
+
+    /* If weights are NULL we use weights of 1, which then later can be replaced
+     * with infinite weights.
+     */
+    null_weights = !weights;
```
szhorvat, line 1433: `no_weights` to avoid confusion with 0 values.

ntamas:
Resolved merge conflicts after the recent merge of `feat/typed-vector-lists` into `develop`.

The PR should be updated so that the new function uses `igraph_vector_int_list_t` to return the paths instead of an `igraph_vector_ptr_t`. I'll do this in a separate round once I'm done with checking the other PRs for merge conflicts.

ntamas:
PR updated. One thing that I don't like about the current implementation is the fact that we are actively modifying the `weights` vector so it cannot be passed as a `const igraph_vector_t*` to the function, it must be non-const. The consequence is that I have to declare the parameter as `INOUT` in `functions.yaml`. Note that the function attempts to restore `weights` to its original state if it exist normally and not with an error code.

The alternatives would be to copy the `weights` vector before modifying it, and then working on the copy. In theory, the copying is O(m) so it does not affect the overall time complexity of the function, but it can still be problematic if the graph is large but the part actively modified during the traversal is small.

ntamas:
Conflicts resolved with `develop`. I guess that this function should be called `igraph_get_k_shortest_paths()` now that we have removed `igraph_shortest_paths()` to `igraph_distances()`, right?

src/paths/shortest_paths.c
ntamas: Not relevant any more as we've switched to `igraph_vector_int_list_t`.

src/paths/shortest_paths.c
ntamas: Renamed to `weights_missing` as `no_weights` can be mistaken for "number of weights".

ntamas:
I think this is now ready to merge if the CI passes.

src/paths/shortest_paths.c
```diff
@@ -1279,3 +1280,315 @@ igraph_error_t igraph_diameter_dijkstra(const igraph_t *graph,
...
+    } else {
+        result = n;
+    }
+
+    return result;
+}
+
+/**
+ * \function igraph_get_k_shortest_paths
+ * \brief K shortest paths (geodesics) between two vertices.
```
szhorvat, line 1340:
```suggestion
 * \brief k shortest paths between two vertices.
```

This function doesn't quite compute geodesics. Only the shortest path is a geodesic, not the 2nd shortest one.

src/paths/shortest_paths.c
```diff
@@ -1279,3 +1280,315 @@ igraph_error_t igraph_diameter_dijkstra(const igraph_t *graph,
...
+ * \param weights The edge weights of the graph. Can be \c NULL for an
+ *        unweighted graph. Infinite weights will be treated as missing
+ *        edges.
+ * \param paths Pointer to an initialized list of integer vectors, the result
+ *        will be stored here in \ref igraph_vector_int_t objects. Each vector
+ *        object contains the edges along the k-th shortest path between the
+ *         \p from and \p to vertex, where k is the pointer vector index.
+ * \param k The number of paths.
+ * \param from The id of the vertex from which the geodesics are
+ *        calculated.
```
szhorvat, line 1352:
```suggestion
 * \param from The ID of the vertex from which the paths are calculated.
```

src/paths/shortest_paths.c
```diff
@@ -1279,3 +1280,315 @@ igraph_error_t igraph_diameter_dijkstra(const igraph_t *graph,
...
+ * \param to The id of the vertex to which the geodesics are
+ *        calculated.
```
szhorvat, line 1354:
```suggestion
 * \param to The ID of the vertex to which the paths are calculated.
```

src/paths/shortest_paths.c
```diff
@@ -1279,3 +1280,315 @@ igraph_error_t igraph_diameter_dijkstra(const igraph_t *graph,
...
+ *        \cli IGRAPH_EINVVID
+ *           \p from or \p to is an invalid vertex id.
+ *        \cli IGRAPH_EINVMODE
+ *           Invalid mode argument.
+ *        \cli IGRAPH_EINVAL
+ *           Invalid argument.
+ *        \endclist
+ *
+ * Time complexity: K|V|(|V|log|V| + |E|), where |V| is the number of vertices,
+ *                  and |E| is the number of edges.
```
szhorvat, line 1379:
```suggestion
 * Time complexity:  k |V| (|V| log|V| + |E|), where |V| is the number of vertices,
 *                  and |E| is the number of edges.
```

szhorvat (changes requested):

tests/unit/igraph_get_k_shortest_paths.c
```diff
@@ -0,0 +1,112 @@
...
+    igraph_vector_int_list_t paths;
+
+    igraph_vector_int_list_init(&paths, 0);
+
+    igraph_small(&g_0, 0, 0, -1);
+    igraph_small(&g_1, 1, 0, -1);
+    igraph_small(&g_2, 2, 0, -1);
+    igraph_small(&g_2c, 2, 0, 0,1, -1);
+    igraph_small(&g_wiki, 6, 1, 0,1, 0,2, 1,3, 2,1, 2,3, 2,4, 3,4, 3,5, 4,5, -1);
+    igraph_small(&g_wiki_u, 6, 0, 0,1, 0,2, 1,3, 2,1, 2,3, 2,4, 3,4, 3,5, 4,5, -1);
```
szhorvat, line 52:
```suggestion
    igraph_small(&g_wiki, 6, IGRAPH_DIRECTED, 0,1, 0,2, 1,3, 2,1, 2,3, 2,4, 3,4, 3,5, 4,5, -1);
    igraph_small(&g_wiki_u, 6, IGRAPH_UNDIRECTED, 0,1, 0,2, 1,3, 2,1, 2,3, 2,4, 3,4, 3,5, 4,5, -1);
```

Just for my sanity.

szhorvat, line 52:
```suggestion
    igraph_small(&g_0, 0, IGRAPH_UNDIRECTED, -1);
    igraph_small(&g_1, 1, IGRAPH_UNDIRECTED, -1);
    igraph_small(&g_2, 2, IGRAPH_UNDIRECTED, -1);
    igraph_small(&g_2c, 2, IGRAPH_UNDIRECTED, 0,1, -1);
    igraph_small(&g_wiki, 6, IGRAPH_DIRECTED, 0,1, 0,2, 1,3, 2,1, 2,3, 2,4, 3,4, 3,5, 4,5, -1);
    igraph_small(&g_wiki_u, 6, IGRAPH_UNDIRECTED, 0,1, 0,2, 1,3, 2,1, 2,3, 2,4, 3,4, 3,5, 4,5, -1);
```

Just for my sanity, there were too many numbers.

src/paths/shortest_paths.c
```diff
@@ -1279,3 +1280,315 @@ igraph_error_t igraph_diameter_dijkstra(const igraph_t *graph,
...
+ * \brief K shortest paths (geodesics) between two vertices.
+ *
+ * \param graph The graph object.
+ * \param weights The edge weights of the graph. Can be \c NULL for an
+ *        unweighted graph. Infinite weights will be treated as missing
+ *        edges.
+ * \param paths Pointer to an initialized list of integer vectors, the result
+ *        will be stored here in \ref igraph_vector_int_t objects. Each vector
+ *        object contains the edges along the k-th shortest path between the
+ *         \p from and \p to vertex, where k is the pointer vector index.
```
szhorvat, line 1349:
```suggestion
 *        object contains the edges along the <code>k</code>th shortest path between
 *         \p from and \p to, where \c k is the vector list index.
```

szhorvat:
Are the paths returned in reverse? I am looking a the undirected Wikipedia example. The first result is `7, 4, 1`. But this is the path from `5` to `0`, not from `0` to `5`. It is reversed.

For reference, the graph is the following, with edge and vertex IDs marked (0-based):

image

szhorvat:
Sorry, the function call was indeed from `5` to `0`, so it's fine.

szhorvat:
Could we add some non-simple graphs to the tests?

image

image

CHANGELOG.md
```diff
@@ -667,8 +667,8 @@ Some of the highlights are:
  - `igraph_roots_for_tree_layout()` computes a set of roots suitable for a nice tree layout.
  - `igraph_fundamental_cycles()` computes a fundamental cycle basis (experimental).
  - `igraph_minimum_cycle_basis()` computes an unweighted minimum cycle basis (experimental).
- - `igraph_strvector_merge()` moves all strings from one string vectors to the end of another without
-   re-allocating them.
+ - `igraph_strvector_merge()` moves all strings from one string vectors to the end of another without re-allocating them.
+ - `igraph_get_k_shortest_paths()` to find at most k alternative shortest paths between a source and a target vertex (#1763, thanks to @GroteGnoom)
```
szhorvat, line 671: This does not find `k` "alternative" shortestest paths, but simple _the `k` shortest paths_. They are not alternatives. They are the shortest, the 2nd shortest, the 3rd shortest, and so on.

src/paths/shortest_paths.c
```diff
@@ -1279,3 +1280,315 @@ igraph_error_t igraph_diameter_dijkstra(const igraph_t *graph,
...
+ *        calculation in directed graphs. Possible values:
+ *        \clist
+ *        \cli IGRAPH_OUT
+ *          The shortest outgoing paths are calculated.
+ *        \cli IGRAPH_IN
+ *          The shortest incoming paths are calculated.
+ *        \cli IGRAPH_ALL
+ *          The directed graph is considered as an
+ *          undirected one for the computation.
+ *        \endclist
```
szhorvat, line 1365:
```suggestion
 * \param mode The type of paths to be used for the
 *        calculation in directed graphs. Possible values:
 *        \clist
 *        \cli IGRAPH_OUT
 *          The outgoing paths of \p from are calculated.
 *        \cli IGRAPH_IN
 *          The incoming paths of \p from are calculated.
 *        \cli IGRAPH_ALL
 *          The directed graph is considered as an
 *          undirected one for the computation.
 *        \endclist
```

szhorvat:
Are we sure we don't want to add an output argument to return the vertex paths? All other path functions have this, except `igraph_get_all_simple_paths`, which needs a complete refactoring anyway.

Note that the vertex paths can't be computed from the edge paths without also having at least the start vertex.

It would be a useful convenience to be able to return the vertex paths.

CHANGELOG.md
szhorvat:
```suggestion
 - `igraph_get_k_shortest_paths()` finds the k shortest paths between a source and a target vertex (#1763, thanks to @GroteGnoom)
```

szhorvat:
```suggestion
 - `igraph_get_k_shortest_paths()` finds the k shortest paths between a source and a target vertex (#1763, thanks to @GroteGnoom)
```

doc/structural.xxml
```diff
@@ -24,6 +24,7 @@
 <!-- doxrox-include igraph_distances_johnson -->
 <!-- doxrox-include igraph_get_shortest_paths -->
 <!-- doxrox-include igraph_get_shortest_path -->
+<!-- doxrox-include igraph_get_k_shortest_paths -->
```
szhorvat, line 27: Can you move this to just before `get_all_simple_paths()`? It should not interrupt a block of closely related functions.

szhorvat, line 27: Can you move this to just before `get_all_simple_paths()`? It should not interrupt a block of closely related functions.

src/paths/shortest_paths.c
GroteGnoom: Isn't geodesic just a synonym for shortest path? The second shortest path is not a geodesic in the same way as the second shortest path is not a shortest path.

GroteGnoom: I do think it's fine without :)

szhorvat (approved):
@GroteGnoom Thanks for much for implementing this! This can be merged as far as I'm concerned.

I would still like to add an output argument for getting the vertex path, but I suggest doing that later, so this PR doesn't keep getting out of sync with develop.

PR #1771, opened 2021-04-21
Fix triangles_template1.h cleanup

GroteGnoom:
This caused igraph_transitivity_barrat to sometimes reference undefined
memory during an IGRAPH_FINALLY_FREE.

### DIFF
--- a/src/properties/triangles_template1.h
+++ b/src/properties/triangles_template1.h
@@ -38,6 +38,8 @@ nodes_to_calc = IGRAPH_VIT_SIZE(vit);
 
 if (nodes_to_calc == 0) {
     igraph_vector_clear(res);
+    igraph_vit_destroy(&vit);
+    IGRAPH_FINALLY_CLEAN(1);
     return IGRAPH_SUCCESS;
 }
 


ntamas (approved):
Nice catch, thanks!

GroteGnoom:
> Nice catch, thanks!

Thanks! This was one of the most confusing bugs I ever encountered.

I wonder why VERIFY_FINALLY_STACK didn't catch it, but did catch it when calling it after each igraph_transitivity_barrat function call. Maybe because when ERRORF is called, this eventually calls IGRAPH_FINALLY_FREE? If that's the case, VERIFY_FINALLY_STACK should be called before any errors in tests.

Also: can I always merge master into develop?

ntamas:
> Maybe because when ERRORF is called, this eventually calls `IGRAPH_FINALLY_FREE`?

Exactly, `IGRAPH_ERRORF` frees everything that is in the "finally" stack and then clears the stack. It does not care (it does not even notice) if it frees something that should have been freed by something else earlier.

> Can I always merge master into develop?

Yes, that should be safe. Only bugfixes, documentation improvements and other non-breaking stuff go into `master`, and sooner or later it will always get merged into `develop` so feel free to do it any time the `develop` branch is behind `master`. Note that eventually it might become harder to keep on doing it regularly; currently the `develop`  branch does not include too many breaking changes yet, but in the near future we want to replace `int`  return values with `igraph_error_t` (which will be a typedef to `int`), which will probably result in merge conflicts all the time even though semantically the two types will be the same.

PR #1774, opened 2021-04-22
Add i neighbors

### LINKED ISSUE #1737: neighbors() and incident() should allow controlling loop inclusion
Similar to `adjlist_init()` and `inclist_init()`, `neighbors()` and `incident()` should allow controlling whether self-loops are included: not at all; once; or twice.

**Use cases for the feature**
The use case is the same as for `adjlist_init` and `inclist_init`. For many graph algorithms, it is necessary to have this control. `neighbors()` and `incident()` are sometimes used in place of `adjlist_init()` or `inclist_init()` because there is no need to keep the entire adjlist/inclist in memory. Recently, while implementing cycle basis calculations, I found myself using an inclist instead of `incident()` just because `icident()` did not have the necessary `IGRAPH_LOOPS_ONCE`. I suspect that the VF2 algorithm could be adapted to loopy graphs relatively easily if we had this feature. The need keeps coming up repeatedly.

I consider this a high-priority issue, but I don't mind if this is implemented in` neighbors()` itself or an additional function.

GroteGnoom:
Fixes #1737

Some code is copied to not slow down the original code: First it is checked if the original situation is used, and if not we go to a different loop with internal branches. I haven't checked if compilers can optimize this all out by themselves.


src/graph/neighbors.h
```diff
@@ -0,0 +1,36 @@
...
+   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+   GNU General Public License for more details.
+
+   You should have received a copy of the GNU General Public License
+   along with this program.  If not, see <https://www.gnu.org/licenses/>.
+*/
+
+#ifndef IGRAPH_NEIGHBORS_H
+#define IGRAPH_NEIGHBORS_H
+
```
ntamas, line 21: Ideally, the header should stand on its own so we should `#include` all the headers here that define `igraph_t`, `igraph_vector_t`, `igraph_integer_t`, `igraph_neimode_t` and all the other igraph stuff that  you use in the header.

src/graph/type_indexededgelist.c
```diff
@@ -736,25 +737,48 @@ igraph_integer_t igraph_ecount(const igraph_t *graph) {
...
+            mode != IGRAPH_ALL) {
+        IGRAPH_ERROR("Mode should be either IGRAPH_OUT, IGRAPH_IN or IGRAPH_ALL.", IGRAPH_EINVMODE);
     }
 
     if (!igraph_is_directed(graph)) {
         mode = IGRAPH_ALL;
     }
 
+    if (mode != IGRAPH_ALL && loops == IGRAPH_LOOPS_TWICE) {
+        IGRAPH_ERROR("For a directected graph (with directions not ignored), "
```
ntamas, line 779: Typo here (`directected`)

src/graph/type_indexededgelist.c
```diff
@@ -783,61 +807,126 @@ int igraph_neighbors(const igraph_t *graph, igraph_vector_t *neis, igraph_intege
...
+                    i1++;
+                    i2++;
+                    IGRAPH_I_NEIGHBORS_SKIP(n1, 2);
+                    VECTOR(*neis)[idx++] = n1;
+                    if ((loops == IGRAPH_LOOPS_ONCE && n1 == pnode && last_added == pnode) ||
+                         (multiple == IGRAPH_NO_MULTIPLE)) {
+                         length--;
+                         continue;
+                    }
+                    VECTOR(*neis)[idx++] = n2;
```
ntamas, line 906: Not sure if you see this; Codecov indicates that this branch was not covered by tests. It would be great if you could add a test case to cover this  branch.

ntamas (changes requested):
Looks good to me at first glance; Codecov indicates that some of the newly added code branches are not covered by tests, though. Since this is a very important part of the library, it would be great if we could cover the entire function; you can check the coverage [here](https://app.codecov.io/gh/igraph/igraph/compare/1774/tree/src/graph/type_indexededgelist.c) (let me know if you don't have access to this).

A few more comments:

* I can see that there's a lot of code duplication now and I understand that it makes sense to do so due to performance concerns (otherwise we would have to check whether we need to call `IGRAPH_I_NEIGHBORS_SKIP` or not in every loop iteration. However, if we simply stored the result of the condition in an `igraph_bool_t` (like `igraph_bool_t should_filter_duplicates`), the compiler would probably be smart enough to store the result of the condition in a register so using it would have almost zero runtime cost, and we could get rid of the code duplication by re-writing `IGRAPH_I_NEIGHBORS_SKIP` to test this boolean before doing anything. The branches where skipping is not needed could simply set the `should_filter_duplicates` flag to `false`.

* Please `#undef IGRAPH_I_NEIGHBORS_SKIP` at the end of the function. Actually, by doing so, you could also shorten its name to `DEDUPLICATE_IF_NEEDED` or something like that; since it is guaranteed not to leak to other compilation units, you can get rid of the `IGRAPH_I` prefix.

szhorvat:
> you can check the coverage [here](https://app.codecov.io/gh/igraph/igraph/compare/1774/tree/src/graph/type_indexededgelist.c) (let me know if you don't have access to this).

I think you need to log in to Covecov with your Github account.

src/graph/type_indexededgelist.c
GroteGnoom: Yes, I see them :slightly_smiling_face: A great feature, and I'll try to fix them.

ntamas (approved):

ntamas:
Thanks again!

PR #1777, opened 2021-04-27
Add igraph_subcomponent unit tests

GroteGnoom:
Part of #1592

Everything seems to be in order.


ntamas:
Can you sort the vectors before being printed to stdout? The order of traversal in `igraph_subcomponent()` is entirely implementation-dependent, and I don't want the test case to break in the future if we change the implementation. All that matters is that all the nodes in the same component as the seed vertex should be in the output, it does not matter what the order is.

GroteGnoom:
Of course! It's fixed now.

PR #1782, opened 2021-04-29
Add igraph_get_incidence unit tests

GroteGnoom:
Part of #1592

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 
 ### Fixed
 
+ - `igraph_get_incidence()` no longer reads and writes out of bounds when given a non-bipartite graph, but gives an error.
  - `igraph_dyad_census()` no longer reports an overflow on singleton graphs, and handles loops and multigraphs correctly. Undirected graphs are handled consistently and will no longer give a warning.
  - CMake package files did not respect `CMAKE_INSTALL_LIBDIR`. This only affected Linux distributions which install into `lib64` or other locations instead of `lib`.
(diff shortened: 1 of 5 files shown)


szhorvat:
I would opt for issuing a warning instead of an error, and simply ignoring edges that connect vertices within the same partition.

szhorvat:
(I fixed the doc failure on master now.)

ntamas:
> I would opt for issuing a warning instead of an error

In that case, we should probably count how many such edges there are, and just report _one_ warning at the end of the routine with the number of non-bipartite edges.

szhorvat:
Yes, that, or if we want to keep it simple, then we can also omit the number, and just use a flag to determine if the warning was already reported. Let's get this in before 0.9.3.

szhorvat:
Another thought: do not say "not bipartite", but say "n edges running within partitions were ignored" or something similar.  The graph may be bipartite, but the supplied types vector may not be consistent with the two partitions.

ntamas:
Fixed wording; "independent sets" has another meaning in graph theory so I've removed that part.

GroteGnoom:
I thought I was using this definition: https://en.wikipedia.org/wiki/Independent_set_(graph_theory)

ntamas:
Indeed that's the definition I was thinking about, you are right. I can't quite put my finger on what bothered me with the original message; somehow it felt confusing. Sorry :)

szhorvat:
I was slightly confused too at first, but the usage was perfectly correct.

PR #1783, opened 2021-05-02
Remove unistd replacement for MSVC

szhorvat:


szhorvat:
This was an experiment to get rid of the empty `msvc/include/unistd.h`.  On Windows, `flex` can take the `--wincompat` option to avoid using `unistd.h`.  But now I realize that the parser sources are usually generated on some Unix-like system, so release tarball would still need `unistd.h`.

szhorvat:
See #1784 for an alternative solution.

PR #1788, opened 2021-05-10
Add igraph_sbm_game unit tests

GroteGnoom:
Part of #1592

The output graph should be uninitialized, which is unusual, but not
necessarily wrong, and changing it would change the interface.

### DIFF
--- a/src/games/sbm.c
+++ b/src/games/sbm.c
@@ -103,9 +103,9 @@ int igraph_sbm_game(igraph_t *graph, igraph_integer_t n,
     }
 
     if (igraph_vector_int_size(block_sizes) != no_blocks) {
-        IGRAPH_ERRORF("Block size vector length (%ld) not equal to number of rows "
-                      "in preference matrix (%ld).", IGRAPH_EINVAL, igraph_vector_int_size(block_sizes),
-                      no_blocks);
+        IGRAPH_ERRORF("Block size vector length (%ld) does not agree with "
+                      "preference matrix size (%ld).", IGRAPH_EINVAL, 
+                      igraph_vector_int_size(block_sizes), no_blocks);
     }
 
     if (no_blocks > 0) {


szhorvat:
> The output graph should be uninitialized, which is unusual, but not
> necessarily wrong, and changing it would change the interface.

For graph constructors, this is actually the standard behaviour.

When a function returns a result through a vector or other similar data structure, it requires an already initialized vector as input. This is for re-usability: it makes it possible to repeatedly use a vector without re-allocating it, a common use case.

However, `igraph_sbm_game` is a constructor: it creates an entirely new graph. It is comparable not to function that return a vector, but to functions like `vector_init`.

szhorvat (approved):

szhorvat:
This looks perfectly fine to me, but I am feeling a bit tired today so I would like to leave it to others to merge.

szhorvat:
Continuing the line of thought about functions that _construct_ new things vs functions which _update_ existing things vs functions that _return_ results, personally I do find it irritating that some of the vector constructors are not available as updating functions.

https://igraph.org/c/doc/igraph-Data-structures.html#vector-constructors-and-destructors

In particular, I always wanted `vector_init_seq` to work on an already created vector.

I also find it irritating that it is sometimes not completely obvious which functions is a constructor and which is an updater. Most vector constructors have `init` in their names, but not all. Consider `vector_init_copy`  vs `vector_copy`: the difference couldn't be guessed from the naming. Both are constructors, but the first one takes a raw array and the second one an existing vector. I was confused by this in the past, assuming that `vector_copy` is not a constructor.

I think it would be nice to make naming more consistent for 1.0, or at least consider it ... all data structure constructors (except for graphs) could have `init` in the name. I wonder what others think.

szhorvat:
A final comment @GroteGnoom : For a while I wanted to have a function which is just like `sbm_game`, but asymmetric. This would be completely analogous to `igraph_asymmetric_preference_game`, but the out- and in-types of each node would be given by the user instead of generated randomly (as `igraph_asymmetric_preference_game` does).  Do you want to implement this? It's not a big deal really and I would do it at some point myself. I am mentioning it only in case you are interested.

ntamas:
> I think it would be nice to make naming more consistent for 1.0, or at least consider it ... all data structure constructors (except for graphs) could have init in the name. I wonder what others think.

That would be great and it's probably worth collecting all sorts of these API-breaking things somewhere that would improve consistency throughout the library so we can break all of them at  the same time (and hopefully never touch them again). Yes, `_copy()` is a constructor so it would be phrased better as`_init_with_copy()`. I think `_copyto()` or `_update()` are the ones that work on existing vectors, and the only difference between them is the order of arguments (`_copyto()` works by copying the first arg to the second, `_update()` updates the first _from_ the second).

ntamas (approved):

ntamas:
Thanks!

GroteGnoom:
> A final comment @GroteGnoom : For a while I wanted to have a function which is just like `sbm_game`, but asymmetric. This would be completely analogous to `igraph_asymmetric_preference_game`, but the out- and in-types of each node would be given by the user instead of generated randomly (as `igraph_asymmetric_preference_game` does). Do you want to implement this? It's not a big deal really and I would do it at some point myself. I am mentioning it only in case you are interested.

We can just add it as an issue for now, and maybe I'll look into it later :slightly_smiling_face:

Issue #907, opened 2015-12-10
igraph_transitivity_local_undirected may return wrong result for directed or multigraphs


szhorvat:
`igraph_transitivity_local_undirected` sometimes returns wrong results for multigraphs.  I suspect that this is because it counts the number of neighbours of a node as the degree of that node.  This is not true when multiple parallel edges are present.

Note that I only tested this for the case when the local clustering coefficient is computed for _all_ nodes.  The function seems to have several different code paths; I don't know if all are affected.

A more serious problem is that mutual edges in directed graphs are treated as two parallel undirected edges.  Thus the function also returns wrong results for directed graphs.  This is a more serious problem as people are more likely to apply the function to a directed graphs.

The documentation states that for directed graphs edge directions are ignored, so a fix should probably ensure that this is really the case.

Example code in R:

```
> g<-make_graph(c(1,2, 2,1, 2,3, 3,1))
> transitivity(g,"local")
[1] 0.3333333 0.3333333 1.0000000

> transitivity(as.undirected(g),"local")
[1] 1 1 1
```

Link to mailing list thread: http://lists.nongnu.org/archive/html/igraph-help/2015-12/msg00010.html

yfiua:
The problem is still not solved. See https://github.com/igraph/igraph/issues/953
The result is to me so wrong, and I cannot imagine how this could not be done properly, because this is sort of a homework question for students.

gaborcsardi:
As the name suggests, you are not expected to use this with directed or multigraphs.

vtraag:
@szhorvat do you have a particular definition of transitivity in mind for graphs with parallel edges? I'm thinking about how to define it properly, but I don't have an immediate clear idea.

To me it seems most straightforward to ensure that the graph is simple and undirected.

szhorvat:
For directed graphs, there are a few generalizations that I've seen.

For multigraphs, I would say the concept doesn't apply ... (One can always interpret edge multiplicities as edge weights, and look for weighted generalizations of the concept, but I would call that a change of concept ...)

----

Getting back to this specific function: What I would expect from this function is to ignore edge multiplicities and edge directions. Unfortunately, dealing with multigraphs sometimes has performance penalties. That is why I think it would be nice to cache whether a graph is a multigraph (to make the check effectively O(1)). However, that's a large, complicated and difficult project ...

vtraag:
[...]

Issue #782, opened 2014-12-10
Make local.scan() work on multigraphs


gaborcsardi:
From https://github.com/igraph/xdata-igraph/issues/53

szhorvat:
We need to at least check if this is still a problem or already solved.

I believe it refers to https://igraph.org/c/doc/igraph-Structural.html#idm209411768240

See also https://github.com/igraph/igraph/wiki/Checklist-for-new-(and-old)-functions

szhorvat:
@GroteGnoom Is this already fixed?

GroteGnoom:
`igraph_local_scan_k_ecount`, `igraph_local_scan_k_ecount_them` and  `igraph_local_scan_subset_ecount` are tested for this, and they call all the other functions in src/misc/scan.c, so yes.

PR #1790, opened 2021-05-10
Support multigraphs in transitivity calculations

szhorvat:
This PR ensures that non-weighted transitivity calculations work consistently. Multi-edges and self-loops are simply ignored.

The source of the problem was that in triangles_template the degree vector and the adjacency list were inconsistent: one ignored self-loops/multi-edges and the other did not.

The original quick-fix was to disallow multigraphs and directed graphs. The problem with directed graphs is that reciprocal edges are effectively parallel edges once directions are ignored, however, we do not have a function to test for such graphs and disallow them selectively. See #907. The original quick-fix did not fully update the documentation, which still claimed that edge directions are ignored.

I need a review on this because:

1. The affected functions have multiple code paths, apparently used for improved performance. Please double check that I didn't miss any.

2. triangle_template is used by many functions. Hopefully I didn't break any. The only suspect one is `igraph_local_scan_1_ecount`. However, local scan statistics were never put in order for multigraphs anyway? See #782 

3. `igraph_transitivity_undirected` and `igraph_transitivity_avglocal_undirected` still use inconsistent degree vectors. However, the only use seems to be to process nodes in order of decreasing degree. I do not see how the order makes a difference. Am I missing something? One guess is that it is for performance, but I cannot measure it using `benchmark_igraph_transitivity`. See https://igraph.discourse.group/t/transitivity-calculation-and-node-ordering/749/2

This work was motivated by: https://igraph.discourse.group/t/local-transitivity-and-barrat-transitivity-for-directed-graph/747/7


szhorvat:
If I disable ordering vertices by degree in the **average local transitivity** and the **global transitivity** calculation, simply by constructing `order` using `vector_init_seq`, then I get:

```
|> Benchmark file: ../tests/benchmarks/igraph_transitivity.c
|  1 Local transitivity, all vertices method, GNM                                  0.889s  0.878s   0.01s
|  2 Local transitivity, subset method, GNM                                         6.77s   6.75s  0.011s
|  3 Average local transitivity GNM                                                  1.9s   1.89s  0.005s
|  4 Global transitivity GNM                                                        2.31s   2.29s   0.01s

|  5 Local transitivity, all vertices method, Barabasi                              1.64s   1.63s  0.003s
|  6 Local transitivity, subset method, Barabasi                                    2.91s    2.9s  0.005s
|  7 Average local transitivity, Barabasi                                           3.01s      3s  0.004s
|  8 Global transitivity, Barabasi                                                  4.07s   4.06s  0.005s

|  9 Local transitivity, all vertices method, small GNM                            0.174s  0.173s      0s
| 10 Local transitivity, subset method, small GNM                                  0.312s  0.312s      0s
| 11 Average local transitivity, small GNM                                         0.176s  0.175s      0s
| 12 Global transitivity, small GNM                                                0.155s  0.155s      0s

| 13 Local transitivity, all vertices method, tiny GNM                             0.147s  0.147s      0s
| 14 Local transitivity, subset method, tiny GNM                                   0.305s  0.304s      0s
| 15 Average local transitivity, tiny GNM                                          0.153s  0.153s      0s
| 16 Global transitivity, tiny GNM                                                 0.142s  0.142s      0s
```

The most striking differences are in benchmark 3, where avoiding ordering provided a large speeding, and in benchmarks 7 and 8, where it caused a smaller slowdown.

----

Here is a second run _with ordering_ for comparison with the previous post:

```
|> Benchmark file: ../tests/benchmarks/igraph_transitivity.c
|  1 Local transitivity, all vertices method, GNM                                  0.858s  0.849s  0.007s
|  2 Local transitivity, subset method, GNM                                         6.93s   6.91s   0.01s
|  3 Average local transitivity GNM                                                 4.51s    4.5s  0.008s
|  4 Global transitivity GNM                                                        2.11s    2.1s  0.006s

|  5 Local transitivity, all vertices method, Barabasi                              1.52s   1.51s  0.006s
|  6 Local transitivity, subset method, Barabasi                                    3.07s   3.06s  0.007s
|  7 Average local transitivity, Barabasi                                           2.48s   2.48s  0.004s
|  8 Global transitivity, Barabasi                                                  3.14s   3.12s  0.009s

|  9 Local transitivity, all vertices method, small GNM                            0.179s  0.178s  0.001s
| 10 Local transitivity, subset method, small GNM                                  0.308s  0.307s      0s
| 11 Average local transitivity, small GNM                                           0.2s    0.2s      0s
| 12 Global transitivity, small GNM                                                0.164s  0.164s      0s

| 13 Local transitivity, all vertices method, tiny GNM                             0.147s  0.147s      0s
| 14 Local transitivity, subset method, tiny GNM                                   0.304s  0.303s      0s
| 15 Average local transitivity, tiny GNM                                          0.179s  0.179s      0s
| 16 Global transitivity, tiny GNM                                                 0.143s  0.142s      0s
```

src/properties/triangles.c
```diff
@@ -260,7 +262,7 @@ int igraph_transitivity_local_undirected2(const igraph_t *graph,
 
     neis = IGRAPH_CALLOC(no_of_nodes, long int);
     if (neis == 0) {
-        IGRAPH_ERROR("local transitivity calculation failed", IGRAPH_ENOMEM);
+        IGRAPH_ERROR("Insufficient memory for local transitivity calculation.", IGRAPH_ENOMEM);
```
ntamas, line 265: I would not change this; the default error handler shows an explanation of `IGRAPH_ENOMEM` by default and it already includes the fact that it is an out-of-memory condition.

src/properties/triangles.c
```diff
@@ -668,8 +664,8 @@ int igraph_transitivity_undirected(const igraph_t *graph,
     IGRAPH_FINALLY(igraph_adjlist_destroy, &allneis);
 
     neis = IGRAPH_CALLOC(no_of_nodes, long int);
-    if (neis == 0) {
-        IGRAPH_ERROR("undirected transitivity failed", IGRAPH_ENOMEM);
+    if (! neis) {
+        IGRAPH_ERROR("Insufficient memory for undirected global transitivity.", IGRAPH_ENOMEM);
```
ntamas, line 668: Same here ("Not enough memory" is already appended to the string in the error handler).

ntamas (approved):
Looks good to me. I'm not going to merge until we understand the performance anomaly that you mentioned.

szhorvat:
> So a hasty conclusion would be that we should replace igraph_transitivity_avglocal_undirected by a call to igraph_transitivity_local_undirected and then calculating the average from the resulting vector? This would make things clearer and faster (if taking the average doesn't take ages).

Yes, let's do this for the moment. I recall that in the past I have avoided `igraph_transitivity_avglocal_undirected` in favour of `igraph_transitivity_local_undirected` when doing performance-sensitive calculations for my own research.  That is ridiculous.  

We can look into improving this after the integer transition.

src/properties/triangles.c
ntamas: Okay, nevermind. I see a tendency in recent PRs where the error messages are designed to be self-explanatory now such that the error code is only an "auxiliary" bit of information but not an essential one. All this points in the direction that in the near future the higher level interfaces do not have to add `igraph_strerror(error_code)` at the end of the error message to ensure that the user gets the full picture. I'm okay with the current version of the error message in this PR.

szhorvat: OK, I'll leave it then. I do not have strong feelings about this topic. I really just wanted to capitalize it and add a fullstop, but then I ended up changing the message too.

ntamas:
I'll change `igraph_transitivity_avglocal_undirected()` to use `igraph_transitivity_local_undirected()` behind the scenes. This will come in a separate commit.

szhorvat:
> I'll change `igraph_transitivity_avglocal_undirected()` to use `igraph_transitivity_local_undirected()` behind the scenes. This will come in a separate commit.

Some things to pay attention to:

 - Do not change the behaviour with the null graph
 - Correct handling of NaNs based on `igraph_transitivity_mode_t mode`
 - Is there any other way we can get a NaN in the result from `igraph_transitivity_local_undirected()` than through `mode`?  That _should_ not be the case, so it _should_ be sufficient to ignore NaNs during the averaging.

ntamas:
> Do not change the behaviour with the null graph

I'll add a unit test for that

> Correct handling of NaNs based on `igraph_transitivity_mode_t mode`

I think we can safely forward `mode` to `igraph_transitivity_local_undirected()` as is, and ignore NaNs in the result vector when taking the average. This also answers your third point.

PR #1793, opened 2021-05-19
Add igraph_local_scan_k_ecount unit tests

GroteGnoom:
Part of #1592

Stopped using the trangles_template with TRIEDGES, because it handled
loops incorrectly. Maybe it was faster for graphs without loops than the
general function which is now used. If that's an important issue we could
benchmark it, and handle `k=1` `undirected` as a special case again.

### DIFF
--- a/src/misc/scan.c
+++ b/src/misc/scan.c
@@ -158,6 +158,9 @@ static int igraph_i_local_scan_1_directed(const igraph_t *graph,
         for (i = 0; i < edgeslen1; i++) {
             int e2 = VECTOR(*edges1)[i];
             int nei = IGRAPH_OTHER(graph, e2, node);
+            if (nei == node) {
+                break;
+            }
             igraph_vector_int_t *edges2 = igraph_inclist_get(&incs, nei);
             int j, edgeslen2 = igraph_vector_int_size(edges2);
             for (j = 0; j < edgeslen2; j++) {
@@ -372,11 +375,7 @@ int igraph_local_scan_1_ecount(const igraph_t *graph, igraph_vector_t *res,
         if (weights) {
             return igraph_i_local_scan_1_sumweights(graph, res, weights);
         } else {
-
-#define TRIEDGES
-#include "properties/triangles_template.h"
-#undef TRIEDGES
-
+            return igraph_local_scan_k_ecount(graph, 1, res, weights, mode);
         }
     }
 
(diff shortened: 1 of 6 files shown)


GroteGnoom:
I will add some tests to improve coverage.

szhorvat:
Would it handle loops incorrectly with the changes in #1790 as well?

GroteGnoom:
Yes, I think so, but in a different way. Instead of `IGRAPH_NO_LOOPS`, I'd want a `IGRAPH_LOOPS_ONCE`, but I think even that wouldn't solve both problems I saw.

Problem 1 was that a loop at the starting vertex was counted twice.
Problem 2 was that a loop at an adjacent vertex was not counted at all.

szhorvat:
`TRIEDGES` is only used by this single function, so it seems safe to completely remove from `triangles_template.h` as well.

I have never looked a the local scan statistics carefully, and don't have the time to do so now. @GroteGnoom, if you are confident about the change, I suggest we go ahead. @ntamas any comments, have you ever looked into this functionality area?

src/misc/scan.c
```diff
@@ -579,12 +578,9 @@ int igraph_local_scan_1_ecount_them(const igraph_t *us, const igraph_t *them,
 
 /**
  * \function igraph_local_scan_k_ecount
- * Local scan-statistics, general function, edge count and sum of weights
- *
- * Count the number of edges or the sum the edge weights in the
- * k-neighborhood of vertices.
```
szhorvat, line 585: Was this comment incorrect? I notice you removed any mention of weights.

GroteGnoom: I found the brief description confusing, so I changed it, but then it's equal to the other description, so I removed that. I think this sentence captures what the function does. Maybe the new version should say "edge weights" instead of just "weights".

The weights are still mentioned :slightly_smiling_face:

ntamas: I think the new version is okay.

GroteGnoom:
> I have never looked a the local scan statistics carefully, and don't have the time to do so now. @GroteGnoom, if you are confident about the change, I suggest we go ahead. @ntamas any comments, have you ever looked into this functionality area?

It now does what the documentation says, and the k=1 case is now consistent with the k=2 case, so it also seems to do what was intended when the general case was implemented. So I think it's correct. Else the documentation would be wrong and the k>1 case would accidentally implement this incorrect idea.

ntamas:
I have checked the results manually and I think they are logical the way they are now with this PR.  Thanks!

PR #1794, opened 2021-05-19
Add stack_ptr_destroy_all unit tests

GroteGnoom:
Part of #1592

This just lets the memory checker do all the work.

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -86,6 +86,7 @@ add_legacy_tests(
   igraph_sparsemat_view
   igraph_sparsemat_which_minmax
   igraph_spmatrix_add_col_values
+  igraph_stack_ptr_destroy_all
   igraph_vector_floor
   igraph_vector_lex_cmp
   matrix
--- a/tests/unit/igraph_stack_ptr_destroy_all.c
+++ b/tests/unit/igraph_stack_ptr_destroy_all.c
@@ -1,0 +1,33 @@
+/*
+   IGraph library.
+   Copyright (C) 2021  The igraph development team <igraph@igraph.org>
+
+   This program is free software; you can redistribute it and/or modify
+   it under the terms of the GNU General Public License as published by
+   the Free Software Foundation; either version 2 of the License, or
+   (at your option) any later version.
+
+   This program is distributed in the hope that it will be useful,
+   but WITHOUT ANY WARRANTY; without even the implied warranty of
+   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+   GNU General Public License for more details.
+
+   You should have received a copy of the GNU General Public License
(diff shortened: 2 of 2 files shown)


szhorvat:
I would wait with this a bit, as the whole item destructor implementation is a bit of a mess. It'd take a while to write up how I wanted to go forward with it, and unfortunately I can't do it right now, so I ask for a bit of time.

szhorvat:
@GroteGnoom @ntamas @vtraag What do we do with these functions, and with `igraph_stack_ptr_t` in general? It seems to me that `igraph_stack_ptr_t` is not used anywhere at the moment. I suggest completely removing support for this type.

It seems to me that more generally, `stack_t` objects don't do anything different than `vector_t` objects, they just happen to have a name that communicates intent better. Thus, should we ever need something like a `stack_ptr_t`, we can use a `vector_ptr_t` or a `vector_list_t` instead.

----

We should also re-examine the non-ptr version of stacks (such as `igraph_stack_t`, `igraph_stack_int_t`, etc.). Technically, these can be removed as well. 

Arguments for keeping them:

 - Personally, I find value in the naming of these types. If they are called "stack", then it is clear that they are used for.
 - If we keep stacks and vectors separate, we keep the door open for optimizing stacks for their specific use case in the future.

Arguments for removing them, or for aliasing them to vectors:

 - Currently, the `stack_t` implementation duplicates the `vector_t` implementation. `vector_t` has received a number of fixes recently, ranging from overflow-protection to making sure that `reserve` never shrinks the capacity. Now these fixes have to be ported over to stacks, which is extra work. If stacks and vectors shared implementation, this would not be a problem anymore.

szhorvat:
Personally, I would do this:

 - Remove `igraph_stack_ptr_t` completely.
 - Keep `igraph_stack_t` and make the necessary fixes. There are not _that_ many of them. While aliasing stacks to vectors would make some things easier, it prevents us from optimizing each to a specific use case without breaking ABI.

ntamas:
Agreed. `igraph_stack_ptr_t` is not needed at the moment. For the remaining stack types, let's keep the implementation and port the fixes over from the vector type.

szhorvat:
Fixes have now been ported to `stack.pmt`.

I have not yet removed `stack_ptr_t`, waiting for feedback from others as well.

GroteGnoom:
I would remove igraph_stack_t (after deprecation maybe?) It's not used at all, right? I don't like keeping things around just in case it would make sense to have it in the future, especially if it's just a vector.

szhorvat:
I removed `igraph_stack_ptr_t` now. It was not used anywhere.

`igraph_stack_t` (with all typed versions) is still there, and updated to match `vector_t`. I added `igraph_stack_capacity()` to make it easier to mirror `igraph_vector_t` functionality.

szhorvat:
Closing this PR then (we can continue the discussion though).

PR #1798, opened 2021-06-01
fix: build 32-bits CXSparse properly

GroteGnoom:
I know very little about CXParse, so I just checked what was needed to pass the tests, and if there were any other `cs_dl`'s present in sparsemat.c. I'm also not sure about the right place for every macro.

### DIFF
--- a/include/igraph_sparsemat.h
+++ b/include/igraph_sparsemat.h
@@ -29,11 +29,14 @@
 #include "igraph_vector.h"
 #include "igraph_datatype.h"
 #include "igraph_arpack.h"
+#include "igraph_config.h"
 
 #include <stdio.h>
 
 __BEGIN_DECLS
 
+#if IGRAPH_INTEGER_SIZE == 64
+#define CS_LONG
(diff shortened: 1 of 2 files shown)


szhorvat:
My preference in dealing with CXSparse would be to:

 1. Bundle a modified CXSparse with igraph (this is not a lot of work at all)
 2. Always use the `dl` API and define `cs_long_t` to be `igraph_integer_t`. This modification of `cs_long_t` would be the only thing changed in CXSparse, thus updates to new versions would be easy. This can either be done to `cs_long_t` directly, or to `SuiteSparse_long` in `SuiteSparse_config.h`. I vote for the latter, as this will make it possible to keep the CXSparse files 100% unmodified compared to upstream. Only the SuiteSparse configuration would be modified.

But I have not been able to follow the int64 transition closely enough in the past couple of weeks, so I may have missed something.

ntamas:
I'll merge this and use this as a starting point for further development along the plan outlined by @szhorvat

PR #1801, opened 2021-06-25
Change community membership and csize to int vecs

GroteGnoom:
I was working on changing `long int`s in flow to igraph_integer_t when i noticed some casts which seemed unnecessary. They were caused by membership and community size vectors using doubles instead of integers. So I changed them to integer vectors.

I didn't check all of the documentation, or the interface with R, but I wanted to see if people agree with the general plan. If so, I will continue working on this and fix the errors.

### DIFF
--- a/examples/simple/igraph_community_leiden.c
+++ b/examples/simple/igraph_community_leiden.c
@@ -25,7 +25,8 @@
 
 int main() {
     igraph_t graph;
-    igraph_vector_t membership, degree;
+    igraph_vector_int_t membership;
+    igraph_vector_t degree;
     igraph_integer_t nb_clusters;
     igraph_real_t quality;
 
@@ -39,12 +40,12 @@ int main() {
                  0, 5, -1);
 
     /* Perform Leiden algorithm using CPM */
-    igraph_vector_init(&membership, igraph_vcount(&graph));
+    igraph_vector_int_init(&membership, igraph_vcount(&graph));
     igraph_community_leiden(&graph, NULL, NULL, 0.05, 0.01, 0, &membership, &nb_clusters, &quality);
 
     printf("Leiden found %" IGRAPH_PRId " clusters using CPM (resolution parameter 0.05), quality is %.4f.\n", nb_clusters, quality);
     printf("Membership: ");
-    igraph_vector_print(&membership);
+    igraph_vector_int_print(&membership);
     printf("\n");
 
     /* Start from existing membership to improve it further */
@@ -52,7 +53,7 @@ int main() {
 
     printf("Iterated Leiden, using CPM (resolution parameter 0.05), quality is %.4f.\n", quality);
     printf("Membership: ");
-    igraph_vector_print(&membership);
+    igraph_vector_int_print(&membership);
     printf("\n");
 
     /* Initialize degree vector to use for optimizing modularity */
@@ -64,11 +65,11 @@ int main() {
 
     printf("Leiden found %" IGRAPH_PRId " clusters using modularity, quality is %.4f.\n", nb_clusters, quality);
     printf("Membership: ");
-    igraph_vector_print(&membership);
+    igraph_vector_int_print(&membership);
     printf("\n");
 
     igraph_vector_destroy(&degree);
-    igraph_vector_destroy(&membership);
+    igraph_vector_int_destroy(&membership);
     igraph_destroy(&graph);
 
     return 0;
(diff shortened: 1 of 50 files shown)


examples/simple/igraph_community_leiden.c
```diff
@@ -25,7 +25,7 @@
 
 int main() {
     igraph_t graph;
-    igraph_vector_t membership, degree;
+    igraph_vector_int_t membership, degree;
```
vtraag, line 28: I see that the `degree` is now also changed to an `igraph_vector_int_t`, but the initialization is still done using `igraph_vector_init` and `igraph_degree` still uses a `igraph_vector_t` instead of an `igraph_vector_int_t`. Presumably best to leave this as an `igraph_vector_t` and change it when we change `igraph_degree` to us an `igraph_vectort_int_t` (if we so decide)?

vtraag (commented):
Great idea! I think it makes a lot of sense to start using `igraph_vector_int_t` for those instead of `igraph_vector_t`! I'll have to look through it in more detail, but most of it seems reasonably clear to me.

One thing that might be a problem at some points is that the graph object is of course still using `double`s. Perhaps there are some methods that would actually expect a vector of `double`s instead of `igraph_integer_t`s, but I haven't yet looked at that.

I did note one thing regarding a degree vector, which you also changed to `igraph_vector_int_t`.

examples/simple/igraph_community_leiden.c
GroteGnoom: Yes, I was just a bit sloppy there. I will change it back and maybe handle degrees later.

GroteGnoom:
VECTOR_INT_OR_0 does not seem to exist in interfaces. Should I create it in interfaces/shell/types.def? The R interface is not my strong suit.

I'm trying to convert from VECTOR_OR_0.

ntamas:
> `VECTOR_INT_OR_0` does not seem to exist in interfaces. Should I create it in `interfaces/shell/types.def`?

No, just start using `VECTOR_INT_OR_0` in `interfaces/functions.def`.

To put things in context:

* `interfaces/functions.def` is a sort-of-machine-readable specification of the types of arguments that igraph functions accept, with suggested default values. Right now it is _only_ used by the R interface to generate parts of the glue code between igraph and R. Its syntax is quite ad-hoc; it is meant to be parsed with `tools/stimulus.py`, which then generates code for the R interface.
* `interfaces/shell` is sort-of-defunct; it used to be an experiment to provide some kind of a command-line interface to igraph. I'm keeping it around because it could be a useful testbed if we ever decide to resurrect the code generation approach.

In fact, there are plans to update the Python interface in the future so that we could generate most of its code using either `tools/stimulus.py`, or `pybind11`, or SWIG, or Cython. At some point (if we obtain funding for it), we will need to evaluate which of these solutions would be best for the Python interface. Until then I don't think we should put too much effort in maintaining `interfaces/functions.def` -- I think it's out-of-sync with parts of the library already, and we would know it soon enough once we start working with it.

include/igraph_community.h
```diff
@@ -92,13 +92,13 @@ IGRAPH_EXPORT igraph_error_t igraph_community_walktrap(const igraph_t *graph,
                                             int steps,
                                             igraph_matrix_t *merges,
```
ntamas, line 93: This will have to become an `igraph_matrix_int_t` as well in the future, but it's not necessary to do it now.

include/igraph_community.h
```diff
@@ -107,7 +107,7 @@ IGRAPH_EXPORT igraph_error_t igraph_community_edge_betweenness(const igraph_t *g
                                                     igraph_matrix_t *merges,
                                                     igraph_vector_t *bridges,
```
ntamas, line 108: `merges` has to become an `igraph_matrix_int_t` in the future; `bridges` needs to be an `igraph_vector_int_t`; again, it's not necessary to do them now, just let's be aware of it.

ntamas (changes requested):
Can you add an entry to `CHANGELOG.md` that explains that all arguments where igraph used to return membership vectors as `igraph_vector_t` are now typed as `igraph_vector_int_t`?

vtraag:
For a brief introduction to Stimulus, see https://github.com/igraph/igraph/wiki/Code-generation-using-Stimulus. That document should still be finished and expanded, I hadn't yet gotten around to do that.

> just start using VECTOR_INT_OR_0 in interfaces/functions.def.

Yep, that would be easiest. It would mean that we would have to implement that type in the R interface, but we'll have to do that when we come to that.

ntamas (approved):

PR #1802, opened 2021-07-03
refactor: working to use integers for degrees

GroteGnoom:
I did some crude changes to try to make everything work. Mostly copying vectors from one type to another. Tests are compiling but failing. I'll stop working on this for today, so please feel free to edit, but comment first if you're starting so we won't interfere with each other.

### DIFF
--- a/examples/simple/assortativity.c
+++ b/examples/simple/assortativity.c
@@ -229,6 +229,7 @@ int main() {
     printf("%.5f\n", res);
 
     igraph_destroy(&g);
+    igraph_vector_destroy(&types);
 
     return 0;
 }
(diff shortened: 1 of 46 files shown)


ntamas:
I'm working on this now, FYI.

ntamas:
Fixed a few test failures; 20/46 files reviewed. I'll take a break now and maybe return to this PR in a few hours.

ntamas:
35/46 files reviewed and the tests should be fixed by now. It's probably worth going through the changed files again to see if there are any places where we could get rid of the copying of the degrees from an `igraph_vector_int_t` to an `igraph_vector_t`; most of the time the same thing could be achieved with just calling `igraph_strength()` with a null weight vector (I just came to realize this today). That's all for me today, feel free to pick it up later. I'll ping if I start working on it again.

ntamas:
I'll work again on this a bit now.

ntamas:
Thanks a lot!

Issue #1803, opened 2021-07-03
Steiner Tree implementation


Nachiket18:
**What is the feature or improvement you would like to see?**
Steiner Tree problem is one of the 21 NP-Complete problem suggested by Richard Karp. For a given graph G(V,E), Steiner tree is a tree which has minimum length and connects subset of vertices in the graph. If the subset of vertices is all vertices in the graph then problem is reduced to minimum spanning tree that can be solved in polynomial time using various algorithms. In this implementation we need to implement FPT (Fixed Parameter Tractable) algorithm as it's NP problem. Further implementations can have approximate algorithms.

**Use cases for the feature**
A simple use case can be a road graph where we have to find our minimum length route between multiple cities.

**References**
1. https://en.wikipedia.org/wiki/Steiner_tree_problem
2. Dreyfus-Wagner algorithm (https://onlinelibrary.wiley.com/doi/abs/10.1002/net.3230010302)
3. Technical blog - https://medium.com/dataengineering-and-algorithms/steiner-problem-in-graphs-9d63ab6214de

Nachiket18:
The algorithm implementation needs use of data structure for mapping subsets to a number. I noticed that igraph has hashtable designed for string, can similar data structure be created for 'int' or there's any known way to implement hashtable for 'int' data type.

PS. Implementation can be done using 'map' as well.

ntamas:
Well, the "hashtable" is not really a hash table anyway -- I guess it is implemented with a trie and a few vectors, and there's no int variant for it. Can you please elaborate a bit on why you need a hashtable for ints, i.e. what sort of subsets you would like to map to numbers? Maybe there's a way around that with some alternative data structure that igraph already provides. If there isn't, the easiest is probably to allow C++ for this part of the library and to use an `std::map<>`

Nachiket18:
Thanks for coming back on this. Yes, I went through hashtable.c code and it's implemented using tries and vectors. So, what our algorithm is trying to do is , we take terminals (known as steiner terminals that are subset of vertices in graph G) from user and we want the minimum path tree between these vertices. In order to do that algorithm creates subsets of the steiner terminals of length 2,3, and so on. We then map these subsets to integers. During further process when we find distances from these subsets of terminal vertices to other vertices, we store them into dynamic programming table where the mapping for subset to 'int' which was previously done is used.

If there's a way to do this using igraph data structures, that would be useful. I would think in that direction. If not then, I guess I can switch the std:map<>.

[...]

PR #1809, opened 2021-08-24
Steiner Tree calculation

Nachiket18:
The functionality that is being developed is calculating the steiner tree in a graph. Steiner Tree is created on basis of subset of terminals in the graphs with the condition that length of the tree should be minimum.

We are following the approach suggested in this paper - https://arxiv.org/abs/1710.00668

Also, we are taking reference of an FPT algorithm known as Dreyfus- Wagner algorithm - https://onlinelibrary.wiley.com/doi/10.1002/net.3230010302

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -434,3 +434,16 @@
  - igraph now uses the high-performance [Cliquer library](https://users.aalto.fi/~pat/cliquer.html) to find (non-maximal) cliques
  - Provide proper support for Windows, using `__declspec(dllexport)` and `__declspec(dllimport)` for `DLL`s and static usage by using `#define IGRAPH_STATIC 1`.
  - Provided integer versions of `dqueue` and `stack` data types.
+
+[Unreleased]: https://github.com/igraph/igraph/compare/0.9.4..HEAD
+[0.9.4]: https://github.com/igraph/igraph/compare/0.9.3...0.9.4
+[0.9.3]: https://github.com/igraph/igraph/compare/0.9.2...0.9.3
+[0.9.2]: https://github.com/igraph/igraph/compare/0.9.1...0.9.2
+[0.9.1]: https://github.com/igraph/igraph/compare/0.9.0...0.9.1
+[0.9.0]: https://github.com/igraph/igraph/compare/0.8.5...0.9.0
+[0.8.5]: https://github.com/igraph/igraph/compare/0.8.4...0.8.5
+[0.8.4]: https://github.com/igraph/igraph/compare/0.8.3...0.8.4
+[0.8.3]: https://github.com/igraph/igraph/compare/0.8.2...0.8.3
+[0.8.2]: https://github.com/igraph/igraph/compare/0.8.1...0.8.2
+[0.8.1]: https://github.com/igraph/igraph/compare/0.8.0...0.8.1
+[0.8.0]: https://github.com/igraph/igraph/releases/tag/0.8.0
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # igraph C library changelog
 
+## [Unreleased]
+
 ## [0.9.4] - 2021-05-31
 
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # igraph C library changelog
 
-## [Unreleased]
+## [0.9.4] - 2021-05-31
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Changed
 
  - Unweighted transitivity (i.e. clustering coefficient) calculations now ignore multi-edges and edge directions instead of rejecting multigraphs and directed graphs.
+ - `igraph_transitivity_barrat()` now returns an error code if the input graph has multiple edges (which is not handled correctly by the implementation yet).
 
 ### Fixed
 
--- a/examples/simple/igraph_floyd_warshall_test.c
+++ b/examples/simple/igraph_floyd_warshall_test.c
@@ -1,0 +1,39 @@
+
+
+
+
+
+int main() {
+
+    igraph_t g;
+    igraph_vector_t weights;
+    igraph_real_t weights_data_0[] = { 0, 2, 1, 0, 5, 2, 1, 1, 0, 2, 2, 8, 1, 1, 3, 1, 1, 4, 2, 1 };
+    igraph_real_t weights_data_1[] = { 6, 7, 8, -4, -2, -3, 9, 2, 7 };
+    igraph_real_t weights_data_2[] = { 6, 7, 2, -4, -2, -3, 9, 2, 7 };
+    igraph_matrix_t res;
+
+    /* Graph with only positive weights */
+    igraph_small(&g, 10, IGRAPH_DIRECTED,
+                 0, 1, 0, 2, 0, 3,    1, 2, 1, 4, 1, 5,
+                 2, 3, 2, 6,         3, 2, 3, 6,
+                 4, 5, 4, 7,         5, 6, 5, 8, 5, 9,
+                 7, 5, 7, 8,         8, 9,
+                 5, 2,
+                 2, 1,
+                 -1);
+    
+    igraph_vector_view(&weights, weights_data_0,
+                       sizeof(weights_data_0) / sizeof(igraph_real_t));
+    igraph_matrix_init(&res, 0, 0);
+    
+    igraph_shortest_paths_floyd_warshall(&g, &res,&weights, IGRAPH_OUT);
+    print_matrix(&res);
+
+    igraph_matrix_destroy(&res);
+    igraph_destroy(&g);
+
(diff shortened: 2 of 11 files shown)


szhorvat:
I won't have time to look at this for at least a couple of weeks (moving house and other things), so I'll just add a few notes:

 - At this point, you should start basing the PR on the `develop` branch. We made several breaking changes on the `develop` branch, mainly having to do with consistently using `igraph_integer_t` instead of `igraph_real_t` for all indices and `igraph_error_t` instead of `int` for error codes.
 - Add the new file to the build system so that it can start to be tested by CI. Take a look at this wiki page if you have not yet seen it: https://github.com/igraph/igraph/wiki/Contributing-your-first-function-to-igraph Basically, you need to add it to `src/CMakeLists.txt`. Then make sure the build completes.
 - First make sure your code _builds_ correctly (even if it doesn't work correctly). Hint: As it currently stands, it won't build because you used the `.c` extension for C++ code. Use `.cpp`.
 - You may want to add a skeleton test soon. See the wiki again.

szhorvat:
Don't forget to add the new files to the build system, so that the CI tests can run!

I'm travelling for a while, and could take a first look next week.

szhorvat:
Can you please add the new files to the build system (the `CMakeLists.txt` file), just to see if everything compiles as is?

The first step would be to make sure the code compiles. Second, write basic tests and make sure they pass. We can worry about memory management later.

The reason for the test failures (even without adding the new code) is that the function prototypes added to the headers lack semicolons.

Nachiket18:
> function prototypes

Thanks for pointing in correct direction.

Can you please add the new files to the build system (the CMakeLists.txt file), just to see if everything compiles as is?

-- I will do this today

The first step would be to make sure the code compiles. Second, write basic tests and make sure they pass. We can worry about memory management later.
-- Sure.

The reason for the test failures (even without adding the new code) is that the function prototypes added to the headers lack semicolons.

-- I corrected that part

szhorvat:
Very good. The next step, before any other changes, is to make sure that the code compiles.

The first problem is what I mentioned earlier: a file containing C++ code has the `.c` extension. It should be `.cpp`.

szhorvat:
Here's my suggestion for going forward with this PR:

The key is to go in small steps, and ask questions often, so we can follow what you are doing, and avoid spending too much time in an unproductive side alley.

Let us make the first goal just to get the Floyd-Warshall implementation right, and break it down like this:

 - **Ensure that the code compiles.** Obviously, one can't test code that doesn't compile, so we can't work on it. Given that you already have a lot of code written, I am wondering if you compiled it in a different manner (separately from igraph's build system) on your computer, so you could test it? If so, avoid working on two parallel versions of the project: let's make sure it compiles _within_ igraph's build system.
 - **Write a small test program**, and make sure that your function produces the expected result. This will be easy since igraph already contains other shortest path algorithms that you can compare to.
 - At this point, **ask us to have a look at the code, and we will agree on next steps**. These next steps will include making sure that the code plays well with the rest of igraph, does memory management correctly, handles edge cases, etc. We will look at [this list](https://github.com/igraph/igraph/wiki/Checklist-for-new-%28and-old%29-functions) as well.

In short: 

First do just the Floyd-Warshall (leave the Steiner tree for later). It will be the vehicle to help you get comfortable with igraph. It is important that you should ask for feedback (and have a discussion with someone on the igraph team) as soon as it is producing the correct result, even before the code is perfect. It will be helpful if you also include a list of necessary improvements (e.g. "memory management needs to be finalized") before having this discussion, but let's talk as soon as the function is working and you have a test program to verify that it produces the correct results.

Nachiket18:
> Here's my suggestion for going forward with this PR:
> 
> The key is to go in small steps, and ask questions often, so we can follow what you are doing, and avoid spending too much time in an unproductive side alley.
> 
> Let us make the first goal just to get the Floyd-Warshall implementation right, and break it down like this:
> 
> * **Ensure that the code compiles.** Obviously, one can't test code that doesn't compile, so we can't work on it. Given that you already have a lot of code written, I am wondering if you compiled it in a different manner (separately from igraph's build system) on your computer, so you could test it? If so, avoid working on two parallel versions of the project: let's make sure it compiles _within_ igraph's build system.
> * **Write a small test program**, and make sure that your function produces the expected result. This will be easy since igraph already contains other shortest path algorithms that you can compare to.
> * At this point, **ask us to have a look at the code, and we will agree on next steps**. These next steps will include making sure that the code plays well with the rest of igraph, does memory management correctly, handles edge cases, etc. We will look at [this list](https://github.com/igraph/igraph/wiki/Checklist-for-new-%28and-old%29-functions) as well.
> 
> In short:
> 
> First do just the Floyd-Warshall (leave the Steiner tree for later). It will be the vehicle to help you get comfortable with igraph. It is important that you should ask for feedback (and have a discussion with someone on the igraph team) as soon as it is producing the correct result, even before the code is perfect. It will be helpful if you also include a list of necessary improvements (e.g. "memory management needs to be finalized") before having this discussion, but let's talk as soon as the function is working and you have a test program to verify that it produces the correct results.

Thanks for this and I will follow this to the t. I will start discussion with my collaborator Yash and start working on test cases for Floyd-Warshall algorithm today itself.

ntamas:
Unit tests should be added to `tests/unit` instead of `examples/simple`; also consider _validating_ the results in the test instead of simply printing them to stdout. The `IGRAPH_ASSERT()` macro is the preferred way of doing this. `test_utilities.inc` contains handy helper functions that are typically used in unit tests; take a look at, e.g., `tests/unit/vector.c` for an example.

If you print stuff to stdout during the test, you should also provide an _expected_ output next to your test in a file whose name is identical to the test, except its extension, which should be `.out`. Also, add your unit test to the appropriate part of `tests/CMakeLists.txt` so it gets executed as part of the normal igraph test cycle.

szhorvat:
Since you already put quite a bit of work into this, it would be great to see it finished! 🙂

Nachiket18:
Yes. I'm trying to connect with my collaborator and get things moving. I'm also interested in this project and want tom to implement the approximation algorithm as well. I have scheduled a few hours for tomorrow and will try to get an update by mid-next week.

szhorvat:
If you feel overwhelmed by some of the technical peculiarities of igraph, consider just contributing the all-pairs shortest path function you already have in here. You can do that in a separate PR. This will let you familiarize with the process a contribution without having to also focus on a complicated algorithm. Once that is done, you can come back to the Steiner tree.

Nachiket18:
@szhorvat Thanks for the suggestion and yes, although we progressed with creating .out files testing in igraph is causing little issue for us. So the semester finals are this week and would end next Wednesday, Post that in winter break we have decided to provide dedicated time for this project to complete. I will keep you updated on weekly basis in winter break on the progress of the work.

Sorry to keep you and other maintainers waiting for this pull request.

Nachiket18:
Here are the few updates - 
1. There was some unexpected delay with my collaborator so I decided to use the Johnson's algorithm for all pairs shortest path instead of Floyd-warshall implementation to speed up the process.
2. I was able to resolve several compiler time issues for steiner.cpp code but one major issue remains. The issue is appearing in in the function fetchIndexofMapofSets where were are trying to iterate over a map <igraph_vector_t , int > variable and compare the key with one coming as input the get it's value which is integer. (This is necessary because we are creating subsets of vertices and capturing the distance value in dynamic programming table. So the subsets are mapped to an integer and stored in the dynamic programming table). I would need thoughts about how to handle this issue. Should I implement the comparison feature for <igraph_vector_t , int > or any other better approach is to be followed.

szhorvat:
Do your subsets ever overlap with each other? If not, you can use the suggestion here:

https://github.com/igraph/igraph/issues/1803#issuecomment-883148067

Otherwise, it might be simplest to use an `std::vector` for the internals of this function.

szhorvat:
At this point, I very strongly suggest that you rebase the PR onto the `develop` branch, or for simplicity, start a new PR where you manually move over the changes to a branch based on `develop`. A lot of changes have happened since this PR was proposed. The `develop` branch now consistently uses integers for vertex and edge IDs, as well as for all kinds of indexing. We no longer use `long int`. Instead, all integers are consistently `igraph_integer_t`, which may be 32- or 64-bit depending on the platform and configuration.

Since this PR won't be merged into `master`, and will need to be ported to `develop` anyway, it's better to start doing that now.

Nachiket18:
> 

Thanks. The subsets are overlapping so I used `std::set` for subsets and rectified all the compile time errors and added a test case for the function that was developed.

szhorvat:
@Nachiket18 Can you please merge the latest `develop` branch into this branch, and resolve all merge conflicts?

@ntamas Do you see a reason for why CI is not running for this PR?

ntamas:
Idk; let's try to resolve the conflicts and see if that changes things.

szhorvat:
It might be easier for @Nachiket18 to create a new PR which branches off directly from `develop`, and manually transfer the changes that are present here. It would be great to finish this, since a lot of work went into it already.

Nachiket18:
@szhorvat Thanks. I have created a new PR  #1939 based on develop branch. Should I close this PR with comments?

PR #1811, opened 2021-09-14
Let rewire_directed_edges generate multi-edges for undirected graphs

### LINKED ISSUE #1527: rewire_directed_edges should not create multi-edges
`igraph_rewire_directed_edges` may create multi-edges in the `IN` and `OUT` modes, but not in the `ALL` mode.

This is inconsistent. Either it should not create multi-edges, or there should be an argument which controls whether it is allowed to create multi-edges (just like `igraph_rewire_edges`).

GroteGnoom:
Fixes #1527


ntamas:
This is great, thanks! The only problem is that it is based on `develop` and we want to get it into 0.9.5 so it should be based on `master`. Rebasing (as above) didn't work so I'll just squash & merge it into `develop` and then cherry-pick the commit back to `master`.

szhorvat:
I would not mess with this in 0.9.5.

ntamas:
Why not? It seems like a bugfix to me.

szhorvat:
OK, I looked at it, and it's fine as it is. 

The real solution is not to allow multi-edges for the undirected case, but to make it possible (through an option) to disallow multi-edges in the directed case. That is something to tackle later, probably for post-0.10 (0.11 or 1.0).

Issue #1491, opened 2020-10-23
CMake with external BLAS/LAPACK libraries on Windows


vtraag:
I thought it would be more convenient to keep track of the separate issues for the CMake build, and only keep the overview of all CMake related issues in #1490. This issue is to keep track of the CMake build using external BLAS/LAPACK libraries.

It seems that a more recent release (0.3.10) of OpenBLAS has solved the issues that I encountered with an earlier version (hanging on `openblas.dll!blas_lock` as reported earlier in https://github.com/igraph/igraph/issues/1490#issuecomment-715383082). It now builds correctly and no tests seem to hang anymore. The current configuration is built using `vcpkg` as a shared library. Since `LAPACK` is built using `gfortran` from `mingw`, it can only be used as a shared library, because according to the [`portfile.cmake`](https://github.com/microsoft/vcpkg/blob/70f380e802852fc09c929dead9d8b8659d394181/ports/lapack-reference/portfile.cmake#L53) for `LAPACK`

```
# If openblas has been built with static crt linkage we cannot use it with gfortran!
```

The alternative is to build with Anaconda. The benefit of this is that these are pre-compiled binaries, which should make it easier to link against these libraries. This is also important in the CI, since installing the packages with `vcpkg` takes over half an hour, because they are all built from source. Possibly, some caching might be done, but this might still be more complicated. Note that a more recent version of OpenBLAS (0.3.11) is not recommended. Unfortunately, this version did seem to make its way into `conda-forge` in Anaconda, so it might be necessary to explicitly request version 0.3.10.

However, there are unfortunately three tests that fail with these external libraries:

```
99% tests passed, 3 tests failed out of 234

Total Test time (real) =  76.32 sec

The following tests FAILED:
         10 - igraph_sparsemat5 (Failed)
        197 - igraph_arpack_rnsolve (Failed)
        218 - igraph_scg_grouping (Failed)
```

Diffs:
<details>
<summary><code>10 - igraph_sparsemat5</code></summary>

```
Comparing files C:\USERS\TRAAGVA1\SRC\IGRAPH\EXAMPLES\SIMPLE\igraph_sparsemat5.out and C:\USERS\TRAAGVA1\BUILD\IGRAPH-SHARED\TESTS\TEST_IGRAPH_SPARSEMAT5.OUT
***** C:\USERS\TRAAGVA1\SRC\IGRAPH\EXAMPLES\SIMPLE\igraph_sparsemat5.out
2.61264 0
...
--
[...]

PR #1813, opened 2021-09-22
Use matching vectors for hub and authority

### LINKED ISSUE #1424: Hub and authority scores are not computed as a matching pair
Let us denote hub and authority scores by `h`, `a` and the adjacency matrix and its transpose by `A`, `At`.

Then `h` and `a` are defined as

   - `A.At.h = alpha h`
   - `At.A.a = beta a`

I.e. they're the leading eigenvectors of `A.At` and `At.A`, respectively. Warning: If you are looking at the book MEJ Newman: Networks, 2nd ed., the definitions are interchanged there! (It's a mistake in the text.)

This is precisely how igraph computes them, using ARPACK.

However, there is an additional expectation that `a` and `h` "match" each other in that `At.h ~ a` and `A.a ~ h`, where I used `~` to denote proportionality.

If the leading eigenvalue is degenerate, and thus there are many possible solutions for `a` and `h`, then the results returned by igraph may not match, i.e. `At.h` may not be proportional to `a` and `A.a` may not be proportional to `h`.

@vtraag Do you have practical experience with these scores? Are they still useful for anything if degeneracy arises? Still, it would make sense for igraph to guarantee returning a matching pair.

Small example graph with degeneracy and its edge list:

image

    {{1, 2}, {2, 1}, {2, 3}, {3, 2}, {3, 4}, {4, 1}}

### LINKED ISSUE #1591: Hub and authority score calculation has no tests
`igraph_hub_score` and `igraph_authority_score` are not exercised anywhere within the tests (either directly or indirectly).

Fix #1424 before fixing this.

GroteGnoom:
Also adds tests.

Fixes #1424
Fixes #1591

This implementation is not the most efficient, but I tried to produce something correct to see if this produces the expected results.

I changed the Kleinberg functionality for edgeless graphs to be consistent with other results, because I was surprised by it.

In practice the hub and authority eigenvalues seem to be the same, I can prove it for a 2x2 matrix, but I didn't prove it for larger matrices. I guess the proof could be one-liner, so I hope someone who is a bit better at linear algebra can confirm this.

Edit: Thinking that the degenerate example would work consistently on different systems was a bit too optimistic.


szhorvat:
A comment and some questions on:

> - `igraph_hub_and_authority_scores()` calculates the hub and authority scores of a graph as a matching pair.

Is this achieved with the following line?

   igraph_vector_fill(vector, 1.0 / no_of_nodes);

----

In most cases, ARPACK appears to return eigenvectors which are normalized so that their Euclidean norm is 1. A vector consisting of `1.0 / no_of_nodes` is not normalized this way. For that we need `1.0 / sqrt(no_of_nodes)`.

However, I believe (have to check!) that ARPACK does not _guarantee_ returning normalized results. Therefore, igraph should not guaratee this either. The hub and authority scores functions have the `scale` feature, which is a kind of normalization: it normalizes so that the largest value is 1. When `scale=false`, we should make no promises about normalization.

In fact, there is a test which tends to fail on Windows when using external BLAS/LAPACK (discussed in #1491, I think). The result is not in fact incorrect, just not normalized as usual. It was never clear if this should be considered a bug or not.

My suggestion is:

 - We can include the normalization fix if you like, but use `1.0 / sqrt(no_of_nodes)`.
 - Do not mention the fix in the changelog. Mentioning it makes it look like igraph makes guarantees about the normalization, but it does not.
 - Make it explicit in the docs that nothing is guaranteed about normalization.

@GroteGnoom Am I misunderstanding anything?

szhorvat:
I'll have to look carefully at the matrix multiplication you do to obtain one score from the other. I just wanted to note that I have the impression that it will not work correctly for an undirected graph because of the use of `IGRAPH_TO`.

Could we not just re-use the same `igraph_i_kleinberg_unweighted` and `igraph_i_kleinberg_weighted` functions to do the multiplication? That's what these functions do.

szhorvat:
Sorry, to be clear: these scores make little sense for undirected graphs, but currently they _do_ work and should return the same result as for a directed graph with all-reciprocal edges. We should not break this.

GroteGnoom:
> @GroteGnoom Am I misunderstanding anything?

No, I forgot the sqrt, and didn't realize normalization isn't guaranteed. I don't want to change things without a changelog, even if we didn't guarantee a specific outcome. If the vectors returned by ARPACK are not always normalized in practice, then it seems better to just keep the old behavior.

szhorvat:
Is it ready / shall I take another look?

GroteGnoom:
Oh, I completely forgot about it after sending it through the CI, yes please, I'm definitely interested in your comments for this new version.

szhorvat:
One small comment:

I was thinking about the long name `hub_and_authority_scores` and noticed that both networkx and Mathematica call it "HITS", which is more concise. I was wondering if we should consider this as well. However, both Wikipedia and Newman's book seem to use "HITS" to refer not to the scores, but to the iterative algorithm to compute them. We don't use this algorithm, so perhaps we should not use the name "HITS" either.

GroteGnoom:
Maybe we should just remove the "scores". That should help, and I don't think it makes it any less clear.

szhorvat:
I think it's fine as it is, was just thinking aloud.

src/centrality/hub_authority.c
```diff
@@ -119,31 +137,77 @@ static igraph_error_t igraph_i_kleinberg_weighted(igraph_real_t *to,
...
     if (igraph_ecount(graph) == 0 || no_of_nodes == 1) {
         /* special case: empty graph or single vertex */
         if (value) {
             *value = igraph_ecount(graph) ? 1.0 : IGRAPH_NAN;
         }
-        if (vector) {
-            igraph_vector_resize(vector, no_of_nodes);
-            igraph_vector_fill(vector, 1);
+        if (hub_vector) {
+            igraph_vector_resize(hub_vector, no_of_nodes);
```
ntamas, line 205: We probably need an `IGRAPH_CHECK()` around here

src/centrality/hub_authority.c
```diff
@@ -119,31 +137,77 @@ static igraph_error_t igraph_i_kleinberg_weighted(igraph_real_t *to,
...
+            igraph_vector_fill(hub_vector, 1.0);
+        }
+        if (authority_vector) {
+            igraph_vector_resize(authority_vector, no_of_nodes);
```
ntamas, line 209: We probably need an `IGRAPH_CHECK()` around here

src/centrality/hub_authority.c
```diff
@@ -162,9 +231,13 @@ static igraph_error_t igraph_i_kleinberg(const igraph_t *graph, igraph_vector_t
             if (value) {
                 *value = IGRAPH_NAN;
             }
-            if (vector) {
-                igraph_vector_resize(vector, no_of_nodes);
-                igraph_vector_fill(vector, 1);
+            if (hub_vector) {
+                igraph_vector_resize(hub_vector, no_of_nodes);
```
ntamas, line 235: We probably need an `IGRAPH_CHECK()` around here

src/centrality/hub_authority.c
```diff
@@ -162,9 +231,13 @@ static igraph_error_t igraph_i_kleinberg(const igraph_t *graph, igraph_vector_t
...
+                igraph_vector_fill(hub_vector, 1);
+            }
+            if (authority_vector) {
+                igraph_vector_resize(authority_vector, no_of_nodes);
```
ntamas, line 239: We probably need an `IGRAPH_CHECK()` around here

ntamas (changes requested):
I'll fix the `iGRAPH_CHECK()` issues that I commented about and then I'll merge, thanks!

PR #1815, opened 2021-10-10
Add random walk unit tests

GroteGnoom:
Part of #1592

Look at the separate commits to clear up the test files. There was already a test called igraph_random_walk, but it actually tested random_edge_walk.

It would be nice to cover everything, but https://app.codecov.io/gh/igraph/igraph/blob/develop/src/paths/random_walk.c
gives me:
"Well this is embarassing, looks like we had an error."

### DIFF
--- a/src/paths/random_walk.c
+++ b/src/paths/random_walk.c
@@ -31,7 +31,7 @@
 
 /**
  * \function igraph_random_walk
- * Perform a random walk on a graph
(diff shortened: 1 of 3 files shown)


GroteGnoom:
> > "Well this is embarassing, looks like we had an error."
> 
> Do you mean that this is an error message from CodeCov?

Yes.

PR #1822, opened 2021-10-25
fix: correct/ignore some warnings when building with clang-13.

### LINKED ISSUE #1821: Build with clang-13 fails
**Describe the bug**
Clang 13 was released about a month ago, but it triggers some errors, because some warnings are treated as errors.

```

src/igraph/src/community/spinglass/pottsmodel_2.cpp:123:12: error: variable 'av_k_squared' set but not used [-Werror,-Wunused-but-set-variable]
    double av_k_squared = 0.0;
           ^
src/igraph/src/community/spinglass/pottsmodel_2.cpp:124:12: error: variable 'av_k' set but not used [-Werror,-Wunused-but-set-variable]
    double av_k_squared = 0.0;
           ^

src/igraph/src/misc/matching.c:555:19: error: variable 'dual' set but not used [-Werror,-Wunused-but-set-variable]
    igraph_real_t dual;               /* solution of the dual problem */
                  ^

src/igraph/src/misc/motifs.c:201:14: error: variable 'motifs' set but not used [-Werror,-Wunused-but-set-variable]
    long int motifs = 0;
             ^
src/igraph/src/internal/qsort.c:128:9: error: performing pointer subtraction with a null pointer has undefined behavior [-Werror,-Wnull-pointer-subtraction]
loop:   SWAPINIT(a, es);
        ^~~~~~~~~~~~~~~
src/igraph/src/internal/qsort.c:77:47: note: expanded from macro 'SWAPINIT'
#define SWAPINIT(a, es) swaptype = ((char *)a - (char *)0) % sizeof(long) || \
                                              ^ ~~~~~~~~~
```  

**To reproduce**
After installing `clang-13` and making sure it is on the PATH somewhere:
```
cmake .. -DCMAKE_C_COMPILER=clang-13 -DCMAKE_CXX_COMPILER=clang-13
cmake --build . --target igraph
```

**Version information**
Latest `master` branch.

vtraag:
This fixes #1821. It ignores some warnings explicitly for `clang`, and simply removes/comments out some unused variables.

### DIFF
--- a/etc/cmake/compilers.cmake
+++ b/etc/cmake/compilers.cmake
@@ -31,7 +31,9 @@ macro(use_all_warnings TARGET_NAME)
       # GCC-style compilers:
       $<$<C_COMPILER_ID:GCC,Clang,AppleClang,Intel>:
         $<$<BOOL:${IGRAPH_WARNINGS_AS_ERRORS}>:-Werror>
-        -Wall -Wextra -pedantic -Wno-unused-function -Wno-unused-parameter -Wno-sign-compare
+        -Wall -Wextra -pedantic
+        -Wno-unused-function -Wno-unused-parameter -Wno-sign-compare
+        -Wno-unknown-warning-option
       >
       $<$<BOOL:${COMPILER_SUPPORTS_NO_VARARGS_FLAG}>:-Wno-varargs>
       # Intel compiler:


szhorvat:
`-Wno-unknown-warning-option` is not supported by the Intel compiler. I'll try to fix this.

szhorvat:
Before fixing this minor issue for the Intel compiler, let's deal see #1824. Compilation doesn't fail with Intel, it just outputs a large number of complaints about the unknown option. So it's not a big deal if it doesn't get fixed for 0.9.5.

PR #1823, opened 2021-10-30
Add benchmarks to optimize weighted betweenness

GroteGnoom:
A discussion on weighted betweenness optimizations started from here:
https://igraph.discourse.group/t/calculation-too-slow/920

This PR tries to figure out if we can optimize weighted betweenness.

The first commit adds an inclist vs adjlist benchmark.

This is a benchmark of low-level functionality that is often used.
It is used in betweenness calculations in particular, and could be the major contribution
to the runtime difference between weighted and unweighted betweenness calculations.

results:
|  1 go over vertices (multiple times) using adjlist                               0.898s  0.674s  0.223s
|  2 go over vertices (multiple times) using inclist                                1.63s    1.4s  0.235s

This draft is supposed to be a starting point to check for other points of optimization.


szhorvat:
> | 3 go over vertices (multiple times) using adjlist 0.046s 0.046s 0s
> | 4 go over vertices (multiple times) using inclist 0.414s 0.414s 0s

Well, now that's dramatic!

vtraag:
This is very interesting to see!

However, the two don't do comparable things: the one loops over both edge indices and node indices, and the other only over node indices. The real comparison is: can we iterate over both edges and nodes faster using `adjlist` and `inclist` combined, or by using `IGRAPH_OTHER`?

One thing that would also be interesting is to profile this benchmark code. Presumably,  most time is spent indeed in `IGRAPH_OTHER`, but it would be interesting to see how much.

GroteGnoom:
> The real comparison is: can we iterate over both edges and nodes faster using `adjlist` and `inclist` combined, or by using `IGRAPH_OTHER`?

In this case the initialization of the adjlist takes longer than looping over the vertices with an inclist, so it seems to be impossible to speed up the process by also generating an adjlist. But because betweenness scales with |V||E|, and initialization with |V| + |E|, there should be some point where it is useful to also use an adjlist, even when you've already initialized an inclist. I'm not sure what to benchmark next to check this in realistic cases.

szhorvat:
> However, the two don't do comparable things: the one loops over both edge indices and node indices, and the other only over node indices.

I would call these comparable. Conceptually, both iterate over all neighbours of all nodes. The number of iterations is the same in both cases, except one looks at `(node_id)` while the other at `(edge_id, node_id)` pairs. That does not explain a 10x performance difference.

In order to keep benchmarked things as separate as possible, we could add a third program, which iterates only over edges, but skips retrieving the node at the other end of the edge.  I would expect this to run in time comparable to "3 go over vertices (multiple times) using adjlist"

I do wonder if the time spent in `IGRAPH_OTHER` can really be accurately measured by a profiler, if the cause of the slowdown is indeed a failure of branch prediction (i.e. inner workings of the CPU). If there are still doubts, we could add yet another test where `IGRAPH_OTHER` is replaced by `IGRAPH_FROM` or `IGRAPH_TO` (whichever). This will eliminate the conditional (which is contained in `IGRAPH_OTHER`) and can tell us whether the problem is indeed with branch prediction.  It is also a useful benchmark because it's what's being done in the directed case (we usually use `IGRAPH_TO` for directed and `IGRAPH_OTHER` for undirected).

szhorvat:
Another note is that this is not just about betweenness. The same kind of iteration over adjlist/inclist happens in all sort of graph traversals, including: DFS/BFS, shortest paths, closeness, betweenness, random walks, etc.

szhorvat:
> In this case the initialization of the adjlist takes longer

And yes, this is a good point. Using both an adlist and inclist is probably not going to help. But we can think about ways to encode which is the "other" end of edges in the inclist, in some way. First let's figure out where the problem is, though.

szhorvat:
So `IGRAPH_TO` is not actually faster. This is not what I expected. Could it be a non-locality issue, i.e. that determining the vertex requires fetching data from a far-away memory location?

I would do the following now, even if some of it looks redundant:

 - Add a benchmark that iterates over incident edges only, but does not get the vertex. Is it still slow? I see no reason why it would be.
 - What Vincent requested: iterate both over vertices and edges, but get the vertex from the adjlist instead of the inclist. This would be fully comparable with benchmarks 4-5.

Something not quite related, but still interesting:

 - Does the adjlist/inclist initialization performance change at all when NOT removing self-loops and multi-edges? I.e. use LOOPS_TWICE and MULTIPLE.

vtraag:
It is very interesting to see these results for such a fundamental part of the library!

> Could it be a non-locality issue, i.e. that determining the vertex requires fetching data from a far-away memory location?

I was wondering the same indeed. Could we perhaps test with a graph that has edges ordered in exactly the same way as the order in which the inclist iterates over? That might make caching much more efficient. Or is this already the case here?

GroteGnoom:
The speedup when not excluding loops for the inclist is pretty amazing. This also happens when I test that first, so I don't think it's caching related.

I'm not sure yet how to improve caching on edge retrieval. Maybe I can make it worse to check if that has an effect.

szhorvat:
This is getting quite interesting.

It would be interesting to have two other cases to compare to:

 - A random graph. The difference compared to the complete graph is that edges are ordered unpredictably.
 - A small graph (either random or complete). In the `igraph_pagerank.c` benchmark I used a helper macro called `REPEAT` to run the same operation many times, and thus be able to benchmark things that take a short amount of time. Performance is usually different on small graphs compared to large ones because of cache effects.

Maybe come up with another (even empty) benchmark for number 10, then add the other two cases (random graph, small graph) afterwards? This way the numberings will match between the three cases, making comparisons easy.

szhorvat:
> The speedup when not excluding loops for the inclist is pretty amazing.

At present, loop removal performs the same type of iteration over the inclist that we are benchmarking here. Notably, it looks back into the `igraph_t` using an `IGRAPH_TO` and `IGRAPH_FROM`. It may actually be the same effect.




> This also happens when I test that first, so I don't think it's caching related.

Regarding the timings of the first benchmark:

I forgot to warn you to build igraph with the vendored LAPACK/BLAS/ARPACK whenever benchmarking. I noticed that when linking to these external libraries, the timings of the first benchmark are sometimes off. I think this is because the external versions of these libraries are often multithreaded, and will execute some initialization code that spins up some threads. The timing from that shows up in the benchmark.  Whatever the reason, I suggest using the vendored LAPACK/BLAS/ARPACK and disabling OpenMP, whenever benchmarking things that do not depend on these.

GroteGnoom:
> Maybe come up with another (even empty) benchmark for number 10, then add the other two cases (random graph, small graph) afterwards? This way the numberings will match between the three cases, making comparisons easy.

Is that numbering system required? Or can I just call them 1-1 and 2-1 for example?

szhorvat:
> Is that numbering system required? Or can I just call them 1-1 and 2-1 for example?

Whatever you prefer!

szhorvat:
So here the "random graph" is actually a complete graph as well, just with randomized edge ordering. This is not what I was thinking of, but this may actually be good (as it's more comparable).

It looks like all the timing are essentially the same, except for 4, which you mention is very unpredictable. I'll venture a guess that what we see here is a failure of branch prediction (due to the randomization). So `IGRAPH_OTHER` _is_ in fact more costly than `IGRAPH_TO`, but still: both are pretty slow to use.

GroteGnoom:
> It looks like all the timings are essentially the same, except for 4, 

rg - 3 is also way slower (and more varying) than fg-3. This is only using the adjlist.

vtraag:
I have been thinking about this a bit. 

Ideally, we wouldn't even need an `adjlist` or `inclist` because the information is essentially already contained in `igraph_t`. However, iteration over `igraph_t` is slow, in part because it needs to get information from the `igraph_t->from` and `igraph_t->to` vectors based on the `igraph_t->oi` and `igraph_t->ii` vectors. This requires to get non-local information from the `from` and `to` vector, preventing the CPU to use the cache efficiently, as it essentially may need to get information from pretty random parts across these vectors. One of the most typical operations across `igraph` is to iterate over the neighbours and/or incident edges of a node. If somehow, the `igraph_t->from` and `igraph_t->to` vectors are aligned with the `igraph_t->oi` and `igraph_t->ii` orders, this would greatly benefit the use of caching and should make it quite a bit faster, in line with what we observe here. Unfortunately, we of course cannot align them both, since the sorting order will inherently always be different for incoming and outgoing edges.

Here's one possible approach to deal with this. How about we simply keep track of the "other" node in a separate vector? That is, we would essentially add two vectors for the "other" incoming nodes and the "other" outgoing nodes, i.e. the neighbours. This of course adds additional memory storage of 2m, with m being the number of edges, at the benefit of running faster. If we want to save on memory, we could in principle get rid of the `from` and `to` vectors, since they are then in principle redundant. However, this would make it impossible to quickly retrieve the endpoints for a certain edge id.

A less invasive proposal, just to deal with the simultaneous iteration over incident edges and neigbhours: should we perhaps simply introduce an `igraph_adjinclist_t` type that contains both vectors of neighbours and of incident edges?

> This also happens when I test that first, so I don't think it's caching related.

I would just like to clarify something here about the caching to make sure that we are on the same page here. The point about caching is that memory gets read into the CPU cache. Whenever a piece of memory is required it fetches a larger part of the memory, not just the single index that is accessed. That means that if we access first `to[30]` and then `to[31]` it will be faster than if we first access `to[30]` and then `to[3944]`: accessing `30` and `31` requires only a single fetch from main memory (or higher level caches), while accessing `30` and `3944` requires two fetches from main memory (or higher level caches), which is slower. For example, this is also why looping in the "incorrect" order of a matrix you a take performance hit: https://stackoverflow.com/questions/9936132/why-does-the-order-of-the-loops-affect-performance-when-iterating-over-a-2d-arra The graphs tested here are too large to fit into any reasonable cache, so probably there will often be fetches from main memory.

ntamas:
> Here's one possible approach to deal with this. How about we simply keep track of the "other" node in a separate vector? 

Do you mean that we should do this "locally" in the weighted betweenness algorithm only, or that we should change `igraph_t` to have this information permanently?

> should we perhaps simply introduce an `igraph_adjinclist_t` type that contains both vectors of neighbours and of incident edges?

That's an easy thing to try and it would be interesting to see whether it improves things.

szhorvat:
> > should we perhaps simply introduce an `igraph_adjinclist_t` type that contains both vectors of neighbours and of incident edges?
> 
> That's an easy thing to try and it would be interesting to see whether it improves things.

The concern here is memory usage (which would be doubled). Would we ever want to use a plain inclist instead of an incadjlist? I don't think so, because the performance issues apply in almost every case. Therefore, it would be nice to think about a solution that is more memory friendly. 

A possible alternative to an "incadjlist" would be a augmented inclist where for each edge we store a boolean value indicating which end of the edge is the "other" one. We can store the booleans as a `char`, which takes 1/8th of the memory of an `igraph_integer_t` on the most common platforms (64-bit).

From the benchmarking, it appears that there may be two reasons for the performance problems:

 1. Non-locality. This would be resolved by such an augmented inclist. So far, it appears that this is by far the biggest issue.
 2. Branch prediction failure. This would NOT be resolved by the augmented inclist, as we'd still have conditionals, but so far it appears that this is a minor effect anyway.

vtraag:
> > Here's one possible approach to deal with this. How about we simply keep track of the "other" node in a separate vector?
> 
> Do you mean that we should do this "locally" in the weighted betweenness algorithm only, or that we should change `igraph_t` to have this information permanently?

I was thinking of doing this more permanently. We now translate all the time the actual `igraph_t` into an `adjlist` or `inclist`, which wastes resources, both computationally and memory wise. If we can find a representation in which we can avoid doing this, that would be great.

Of course, in the meantime, it might be worthwhile to do this locally for some algorithm, but I think it would then be better to do this in the form an explicit structure, such as the `adjinclist`.

> A possible alternative to an "incadjlist" would be a augmented inclist where for each edge we store a boolean value indicating which end of the edge is the "other" one. We can store the booleans as a `char`, which takes 1/8th of the memory of an `igraph_integer_t` on the most common platforms (64-bit). On 32-bit it's still only a 25% increase in memory use instead of a 100% one.

I am not sure if this is actually solves the non-locality problem. The problem is that we need to fetch the actual endpoints of an edge from the vectors `from` and `to`. Storing such a boolean improves things in the sense that we only need to access either the `from` or the `to` vectors, and not both. That is, we still need to access something like `to[30]` and `to[3944]` instead of `to[30]` and `to[31]`. Or am I perhaps misunderstanding your proposal?

> The concern here is memory usage (which would be doubled). Would we ever want to use a plain inclist instead of an incadjlist? I don't think so, because the performance issues apply in almost every case, not just for betweeness. Therefore, it would be nice to think about a solution that is more memory friendly.

Yes, I agree, it takes up more memory. However, don't forget we are *already* taking up more memory because the `inclist` is basically a copy of the `igraph_t`, although `inclist` is only about 25-50% of the memory of `igraph_t` (depending on how it is constructed). Suppose we add the "other" vectors that I proposed to `igraph_t`, but we no longer construct an `inclist`. I believe the memory requirements would then be roughly comparable. Of course, avoiding the use of an `inclist` all together would require a major overhaul of the whole library, which is a bit much at the moment. At any rate, I think we should not be concerned too much with memory specifically of such an `adjinclist` or `inclist`, but look at it in combination with `igraph_t`. Finally, although memory concerns are obviously valid, I think that memory is less and less of a problem.

One practical way to go about this is to see if we can test a few solutions in practice for betweenness, and then try to see which one fits best and can potentially be easily generalised to the rest of the library.

szhorvat:
> Or am I perhaps misunderstanding your proposal?

No, it was my mistake.

GroteGnoom:
It seems like an incadjlist could be useful compared to an inclist when looping over vertices and their edges.

vtraag:
Great, thanks for the implementation @GroteGnoom ! 

I think that we might think about using `incadjlist`, using the `igraph_t` internal directly would not be a good design if we want to make it possible to think about alternative graph structures at one point. The current implementation that you made uses interleaved edge indices and vertex indices. Perhaps it would be easier to use simply two separate vectors, which might make the function easier to use in practice? 

Perhaps we could think about making just a single type instead of having three types `inclist`, `adjlist` and `incadjlist` and their lazy alternatives? Upon construction of the type, a caller could then indicate whether you want neighbours, incident edges, or both. That way, we can keep only a single type and get rid of the two separate `adjlist` and `inclist` types. The API would then be something like:

```c
igraph_adjlist(igraph_t *graph, const igraph_adjlist_t *adjlist,
               igraph_neimode_t mode, igraph_bool_t duplicate,
               igraph_bool_t list_neighbors, igraph_bool_t list_incident_edges);

igraph_adjlist_get_neighbors(igraph_adjlist_t *adjlist);
igraph_adjlist_get_incident_edges(igraph_adjlist_t *adjlist);
```

I'm not sure however whether it wouldn't be a better idea to think about this in the context of iterators. Instead of coming up with some special type that allows fast iteration over neighbours / incident edges, we might think about how such iterators could be best implemented, where this type of iterator would be only one of such type of iterators.

szhorvat:
Can we merge this? It's still possible to keep adding to it later.

GroteGnoom:
Sure!

vtraag:
Is it possible to merge and continue the conversation here and merge again later with additional commits? I think we then have to open a new PR, don't we?

Issue #1824, opened 2021-10-31
Does the qsort code need a revamp?


szhorvat:
Clang started issuing warnings about our qsort implementation (specifically the SWAPINIT macro) since Clang 13. This was solved by disabling the warning with an ugly Clang-specific pragma. See #1822. This in turn caused other annoyances, see [here](https://github.com/igraph/igraph/pull/1822#issuecomment-955665148).

However, I wondered if there was a merit to this warning, so I looked how others solved this issue. This qsort code comes from BSD and a variant of it survives in FreeBSD. Here's the history of the file:

https://svnweb.freebsd.org/base/head/lib/libc/stdlib/qsort.c

Notice that they removed SWAPINIT entirely, with this note:

> libc qsort(3): stop aliasing.
>
> Qsort swap code aliases the sorted array elements to ints and longs in
order to do swap by machine words.  Unfortunately this breaks with the
full code optimization, e.g. LTO.
>
> See https://gcc.gnu.org/bugzilla/show_bug.cgi?id=83201 which seems to
reference code directly copied from libc/stdlib/qsort.c.

I am wondering if we shouldn't update our code to match with FreeBSD's, and avoid the problems mentioned there. It sounds like this code is dangerous, and won't work as intended when using certain optimization options in GCC. It's not clear to me if the fix they used in FreeBSD will reduce performance. But if it was good enough for FreeBSD, it should work for us too?

I don't have the time to look into this anymore today or next week, so I opened this issue. Could you please look into it and make a decision about what to do, before 0.9.5 comes out, @ntamas and @vtraag ?

vtraag:
I've tried to look into this, and I believe it is not too difficult to update `qsort`. See PR #1825 for specifics.

ntamas:
Let's continue in #1825 so we don't have to look at two threads at the same time :)

PR #1825, opened 2021-11-02
Updates `qsort` to a more recent version

vtraag:
As suggested by @szhorvat in #1824 we should update `qsort`. This PR addresses this and updates `qsort` to a more recent version as available from FreeBSD. To be specific, the source was pulled from  the most recent commit https://cgit.freebsd.org/src/commit/a670e1c13a522df4fb8c63bb023b88b1d65de797.

I have made the minimal necessary changes to make it easier to include any possible subsequent updates of `qsort.c`. At the moment, the code already works with a "local" version of `qsort` making it easy to do the necessary changes. For future reference and to be complete: these are the changes with respect to the original file from FreeBSD:

```diff
31a32,43
> #include "igraph_qsort.h"
> 
> #ifdef _MSC_VER
>     /* MSVC does not have inline when compiling C source files */
>     #define inline __inline
>     #define __unused
> #endif
> 
> #ifndef __unused
>   #define __unused    __attribute__ ((unused))
> #endif
> 
35,36d46
< #include <sys/cdefs.h>
< __FBSDID("$FreeBSD$");
38,39d47
< #include <errno.h>
< #include <stdint.h>
41,42d48
< #include <string.h>
< #include "libc_private.h"
212c218
< qsort_r(void *a, size_t n, size_t es, void *thunk, cmp_t *cmp)
---
> igraph_qsort_r(void *a, size_t n, size_t es, void *thunk, cmp_t *cmp)
218c224
< qsort_s(void *a, rsize_t n, rsize_t es, cmp_t *cmp, void *thunk)
---
> igraph_qsort_s(void *a, rsize_t n, rsize_t es, cmp_t *cmp, void *thunk)
244c250
< qsort(void *a, size_t n, size_t es, cmp_t *cmp)
---
> igraph_qsort(void *a, size_t n, size_t es, cmp_t *cmp)
```

### DIFF
--- a/src/internal/hacks.h
+++ b/src/internal/hacks.h
@@ -56,11 +56,6 @@ __BEGIN_DECLS
     #endif
 #endif
 
-#if defined(_MSC_VER)
-#include <BaseTsd.h>
-typedef SSIZE_T ssize_t;
-#endif
-
 __END_DECLS
 
 #endif
--- a/src/internal/hacks.h
+++ b/src/internal/hacks.h
@@ -56,6 +56,11 @@ __BEGIN_DECLS
     #endif
 #endif
 
+#if defined(_MSC_VER)
+#include <BaseTsd.h>
+typedef SSIZE_T ssize_t;
+#endif
+
 __END_DECLS
 
 #endif
--- a/src/internal/qsort.c
+++ b/src/internal/qsort.c
@@ -29,11 +29,7 @@
  * SUCH DAMAGE.
  */
 
-#include <sys/types.h> /* ssize_t on *nix platforms */
-
 #include "igraph_qsort.h"
-
-#include "internal/hacks.h" /* ssize_t on Windows */
 
 #ifdef _MSC_VER
     /* MSVC does not have inline when compiling C source files */
@@ -186,7 +182,7 @@ loop:
 	 * expression, to avoid sign ambiguity in the implied comparison.  es
 	 * is safely within [0, SSIZE_MAX].
 	 */
-	d1 = MIN(pd - pc, pn - pd - (ssize_t)es);
+	d1 = MIN(pd - pc, pn - pd - (ptrdiff_t)es);
 	vecswap(pb, pn - d1, d1);
 
 	d1 = pb - pa;
--- a/src/internal/qsort.c
+++ b/src/internal/qsort.c
@@ -29,7 +29,11 @@
  * SUCH DAMAGE.
  */
 
+#include <sys/types.h> /* ssize_t on *nix platforms */
+
 #include "igraph_qsort.h"
+
+#include "internal/hacks.h" /* ssize_t on Windows */
 
 #ifdef _MSC_VER
     /* MSVC does not have inline when compiling C source files */
--- a/src/internal/qsort.c
+++ b/src/internal/qsort.c
@@ -29,6 +29,14 @@
  * SUCH DAMAGE.
  */
 
+/* This file originates from the following URL:
+ *
+ * https://cgit.freebsd.org/src/commit/lib/libc/stdlib/qsort.c?id=7f8f79a5c444a565a32b0c6578b07f8d496f6c49
+ *
+ * Create a diff against the revision given above to see what we have changed
+ * to facilitate inclusion into igraph
+ */
+
 #include "igraph_qsort.h"
 
 #ifdef _MSC_VER
--- a/src/internal/qsort.c
+++ b/src/internal/qsort.c
@@ -1,6 +1,8 @@
 /*-
+ * SPDX-License-Identifier: BSD-3-Clause
+ *
  * Copyright (c) 1992, 1993
- *  The Regents of the University of California.  All rights reserved.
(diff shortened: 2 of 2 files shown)


szhorvat:
I suggest not using the latest commit, but the one before that. The latest commit added `qsort_s`, which we don't have (or need?) in igraph.

szhorvat:
@ntamas Instead of providing the non-standard `ssize_t`, would it not be better to just use the standard and essentially equivalent `ptrdiff_t`?

Also, please don't make the totally unnecessary `qsort_s` stuff. If you decide to include this code, at least it should be included in its entirety, not just partially, as it is now.

vtraag:
> Also, please don't make the totally unnecessary `qsort_s` stuff. If you decide to include this code, at least it should be included in its entirety, not just partially, as it is now.

I would preferably make minimal changes. We can pull in the additional `qsort_s` stuff, but why do you have a problem keeping this stuff be in `qsort.c`? It's all conditional anyway, so there's no harm in letting it be in there.

ntamas:
> Instead of providing the non-standard `ssize_t`, would it not be better to just use the standard and essentially equivalent `ptrdiff_t`?

Some arcane magic seems to be going on at the point where `ssize_t` is used so I thought it's easier to leave that alone and just work around it on Windows. I mean, there must be a reason why they did not use the standard `ptrdiff_t` type and opted for the non-standard (POSIX-only) `ssize_t`.

szhorvat:
To avoid confusion, I wanted to note that originally I was looking here, at their SVN:

https://svnweb.freebsd.org/base/head/lib/libc/stdlib/qsort.c

Their git repo has newer code, with two more commits. The last commit introduced a use of `ssize_t`:

https://cgit.freebsd.org/src/log/lib/libc/stdlib/qsort.c (click "tree" to see the file)

vtraag:
> To avoid confusion, I wanted to note that originally I was looking here, at their SVN:
> 
> https://svnweb.freebsd.org/base/head/lib/libc/stdlib/qsort.c

Yes, I understood you were looking here.

> Their git repo has newer code, with two more commits. The last commit introduced a use of `ssize_t`:
> 
> https://cgit.freebsd.org/src/log/lib/libc/stdlib/qsort.c (click "tree" to see the file)

Yes, I thought it would be best to use the most recently available version.

The reasoning that is provided in those latest commits indeed suggest that it makes sense to use `ptrdiff_t` instead of `ssize_t` and keep it clean. I'm not sure why they would opt to use `ssize_t` instead, possibly because `SSIZE_MAX` could (theoretically) be larger than `PTRDIFF_MAX` because it only needs to support `-1` instead of the whole negative range. I'm not sure how they would do this in practice though.

ntamas:
@szhorvat now that CI has passed, are you okay with keeping things this way and merging it (with `ssize_t` hacks and the part that relates to `qsort_c` and that is protected by a `#define` now, or do you insist on removing the unused parts? I'm in favour of keeping it as it is easier to diff this against the FreeBSD commit to see what we've changed if we need to update it again later.

szhorvat:
I don't have a confident answer about `ssize_t` vs `prtdiff_t`, but I am a bit uncomfortable with the reasoning that we should not change this because we don't fully understand it ("magic"), because:

 - Do we understand how `SSIZE_T` on Windows relates to the POSIX `ssize_t`?
 - FreeBSD code may be assuming that types/ranges have certain properties that standards don't guarantee. They only need to care about one system: FreeBSD.
 - According to [this](https://stackoverflow.com/a/10884990/695132), `ssize_t` is only required to hold -1, but not any other negative values. Could this be a problem on some other, more exotic system?
 - The `ssize_t` value is being operated together with `ptrdiff_t` values. What is the result? It appears to me that the assumption is that the result is also a `ptrdiff_t`, so is there any reason not to just stick to that single standard type?
 - [Here's an example](https://stackoverflow.com/a/8649077/695132) where `ptrdiff_t` and `ssize_t` differ, but `ptrdiff_t` is the larger type—so it seems that `ptrdiff_t` should be safe there? Is there a case when `ptdiff_t` is smaller? That shouldn't be possible.

vtraag:
In principle, I'm in favour of keeping changes to a minimum, so that it's indeed easier to update. However, since `ssize_t` is used in only a single location, I don't think it's a problem to change that to `ptrdiff_t` instead. I would be in favour of doing so, especially with an eye on Windows support.

Providing comments in the file itself should also be fine. I provided the exact reference to the source in the commit message, which I thought should be sufficient. But if you prefer in the file, that's also OK with me.

szhorvat (approved):

ntamas:
I think we can keep it -- it would allow us to add new warning options later to the compiler definitions without worrying about versions of the compiler that do _not_ support the option; as long as the compiler supports `-Wno-unknown-warning-option`, it will not warn about the new warning options that it does not know and will not turn them into errors. Any objections?

szhorvat:
No objections, but then I'm wondering if we still need to make `-Wno-varargs` conditional:

https://github.com/igraph/igraph/blob/master/etc/cmake/compilers.cmake#L38

PR #1833, opened 2021-11-06
Make igraph_random_sample_real internal

### LINKED ISSUE #1829: igraph_random_sample_real() needs docs
`igraph_random_sample_real()` was added by @GroteGnoom as a public function on the `develop` branch, but I don't know the reason for it. Can you please confirm that we need it as a public function, and if yes, can add documentation?

GroteGnoom:
Fixes #1829

### DIFF
--- a/src/random/random_internal.h
+++ b/src/random/random_internal.h
@@ -27,7 +27,7 @@
 
 __BEGIN_DECLS
 
-IGRAPH_PRIVATE_EXPORT igraph_error_t igraph_random_sample_real(
+igraph_error_t igraph_random_sample_real(
         igraph_vector_t *res, igraph_real_t l, igraph_real_t h,
         igraph_integer_t length);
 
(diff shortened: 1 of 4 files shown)


src/random/random_internal.h
```diff
@@ -0,0 +1,36 @@
...
+*/
+
+#ifndef IGRAPH_RANDOM_INTERNAL_H
+#define IGRAPH_RANDOM_INTERNAL_H
+
+#include "igraph_types.h"
+
+__BEGIN_DECLS
+
+IGRAPH_PRIVATE_EXPORT igraph_error_t igraph_random_sample_real(
```
szhorvat, line 30: `IGRAPH_PRIVATE_EXPORT` is used when we need to have some internal functions exported from shared libraries so that they can be used in test programs. Functions which are truly purely internal to igraph (i.e. not even used from the tests) should not be marked for exporting at all.

szhorvat:
One more thing I just noticed: `random_internal.h` should be included in the file where `igraph_random_sample_real()` is defined. Why? Because the definition and the prototype disagree (e.g. because we accidentally only updated one of them), the compiler will be able to show a warning.

szhorvat (approved):

PR #1836, opened 2021-11-08
Re-work automatic root selection for Reingold-Tilford layout

szhorvat:
This PR improved automatic root selection for the Reingold-Tilford layout. The documentation and tests are not updated yet, but I'd like a review before I proceed. One test fails because the root selection has changed.

This is how the previous version worked:

 - Undirected: Choose the largest degree node from each connected component
 - Directed: Do a topological sort (only works if the graph is a DAG) and choose the first node in topological order from each weakly connected component.

Problems with the directed root selection:
 - It fails if the graph is not a DAG.
 - Even for a DAG, the root list would only be complete if it includes all source vertices (zero in-degree vertices). There may be more than one source vertex per _weakly_ connected component, thus not all were selected.

This is how the new version works:

 - Directed: We break the graph into _strongly_ connected components and find those components that have no incoming connections. One root will be chosen from each such component based on some "prioritization". If the graph is a DAG, this will simply select source vertices, regardless of prioritization (as each source vertex is a strongly connected component on its own).
 - Undirected: Select one root from each connected component based on the "prioritization".

I implemented two choices for prioritization:
 - Take the maximum degree (or out-degree) vertex. This is as before, and very fast.
 - Take a lowest eccentricity vertex. This is also called the [graph center](https://en.wikipedia.org/wiki/Graph_center). This minimizes the number of levels in the tree drawing, and often leads to nice outcomes. The drawback is that it is slower than the Reingold-Tilford layout itself.

To balance performance with quality, I use eccentricity for less than 500 vertices and max degree otherwise.

### DIFF
--- a/src/layout/reingold_tilford.c
+++ b/src/layout/reingold_tilford.c
@@ -646,6 +646,8 @@ int igraph_i_layout_reingold_tilford_select_roots(
  *   then the roots will be selected automatically. Currently, automatic root
  *   selection prefers low ecccentricity vertices in graphs with fewer than
  *   500 vertices, and high degree vertices (acording to \p mode) in larger graphs.
+ *   The root selecton heuristic may change without notice. To ensure a consistent
+ *   output, please specify the roots manually.
  * \param rootlevel This argument can be useful when drawing forests which are
  *   not trees (i.e. they are unconnected and have tree components). It specifies
  *   the level of the root vertices for every tree in the forest. It is only
@@ -910,6 +912,8 @@ int igraph_layout_reingold_tilford(const igraph_t *graph,
  *   then the roots will be selected automatically. Currently, automatic root
  *   selection prefers low ecccentricity vertices in graphs with fewer than
  *   500 vertices, and high degree vertices (acording to \p mode) in larger graphs.
+ *   The root selecton heuristic may change without notice. To ensure a consistent
+ *   output, please specify the roots manually.
  * \param rootlevel This argument can be useful when drawing forests which are
  *   not trees (i.e. they are unconnected and have tree components). It specifies
  *   the level of the root vertices for every tree in the forest. It is only
--- a/src/layout/reingold_tilford.c
+++ b/src/layout/reingold_tilford.c
@@ -31,7 +31,6 @@
 #include "igraph_paths.h"
 #include "igraph_progress.h"
 #include "igraph_structural.h"
-#include "igraph_topology.h"
 
 #include "core/math.h"
 
--- a/src/layout/reingold_tilford.c
+++ b/src/layout/reingold_tilford.c
@@ -28,6 +28,7 @@
 #include "igraph_dqueue.h"
 #include "igraph_interface.h"
 #include "igraph_memory.h"
+#include "igraph_paths.h"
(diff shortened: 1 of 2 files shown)


szhorvat:
To illustrate why eccentricity is a nice way to choose roots, consider a Bethe lattice like this:

image

This is a nice drawing. But what drawing will igraph choose? Notice that all non-leaf-nodes have degree 3. Thus, if we select a maximum degree node, we might choose any of them, and we might end up with a drawing like this one:

image

Using eccentricity will select the "central" node and produce a nice drawing:

image

szhorvat:
Final notes:

 - I separated root selection into its own function. I would like to make this a public function for version 0.10 (or even 0.9.5 if you prefer @ntamas?), with an explicit flag for using degree or eccentricity. Now this is a boolean flag, but we should make it an enum for the public version, to allow for future flexibility. 

 - I would like to say in the documentation that the automatic root selection heuristic may change without notice. This will allow us to make improvements in the future. Those who want a consistent drawing should specify the roots explicitly. I'm a bt undecided about this. Opinions @ntamas ?

szhorvat:
### More example layouts

----

Input graph:

image

Output:

image

Notice how 5, 6, 7 are all on a vertical line. Without curved edges, edges would overlap. This shows why this layout is not really suitable for anything else than trees.

----

A bunch of undirected and directed trees:

image

image

Same trees as above in undirected mode:

image

----

A bunch of very sparse DAGs. Again, this shows why this layout is not ideal for anything but a tree, not even for DAGs. Use Sugiyama for DAGs.

image

ntamas:
One question only after having reviewed the changes: could we use coreness instead of eccentricity as an approximation? In the case of the Bethe lattice it would give identical results, but coreness is O(m) only.

szhorvat:
> One question only after having reviewed the changes: could we use coreness instead of eccentricity as an approximation? In the case of the Bethe lattice it would give identical results, but coreness is O(m) only.

The coreness is actually 1 for each vertex of the Bethe lattice, so this does not allow for distinguishing between them.

----

In the meantime, I realized that there is a much simpler way to think about root selection. A "good" set of roots is a minimal set of vertices so that the entire graph is reachable from them. That's all that's needed: reachability.  The rest is all about prioritization.

I updated the docs and tests, and need to leave for a couple of hours now.

ntamas:
> The coreness is actually 1 for each vertex of the Bethe lattice

Oh sure, right; I was thinking about the iterative algorithm used to calculate coreness (i.e. mark all vertices with degree = 1 as being in core 1, remove them, rinse and repeat), but I didn't realize that you need to keep on going until all the remaining vertices have at least degree 2. Okay, let's leave it as is then, seems okay to me.

I won't merge it yet in case you still want to work on it when you come back, so give me the green light to merge when you are back.

szhorvat:
I think it's good to go if you didn't notice any problems (or documentation typos).

szhorvat:
Ah, one more thing:

Can you please FIRST merge `master` into `develop` (if needed), and only THEN merge this into `master`?

The the next master -> develop merge will be easier. It will require updting to the integer-related changed

ntamas:
Thanks a lot!

Issue #1819, opened 2021-10-22
Wishlist: Generalized Petersen graph


szhorvat:
_This is a good starting point for new igraph contributors._

**What is the feature or improvement you would like to see?**

A graph generator for the generalized Petersen graph.

The construction for `G(n,k)` is as follows:

 - Create a cycle graph on `n` vertices. Name these vertices as `u_0, ..., u_(n-1)`.
 - Create `n` more vertices, `v_0, ..., v_(n-1)` and connect each `u_i` to its corresponding `v_i`.
 - Connect each `v_i` to `v_(i + k mod n)`.

The standard Petersen graph is then `G(5,2)`.

**Use cases for the feature**

This is a well-known graph construction in graph theory.

**References**

 - https://en.wikipedia.org/wiki/Generalized_Petersen_graph

alexsyou:
Hello! I was wondering if this feature is already implemented, and if not could I work on this feature? It would be my first issue but I think I could complete it after looking at the code.

szhorvat:
@fazal452 Are you still working on this / are you planning to complete it?

szhorvat:
@alexsyou Since now it's the weekend, let us wait until Tuesday and see if @fazal452 responds. If not, you can go ahead.

fazal452:
Hey sorry for the late reply. Unfortunately I was not able to complete this project due to my preparation for finals. In the time I allocated I wasn’t able to understand the program well enough.

@alexsyou Feel free to takeover and Good Luck!!

PR #1844, opened 2021-11-15
Generalized Petersen constructor

alexsyou:
This pull request completes #1819.

### DIFF
--- a/src/constructors/generalized-petersen.c
+++ b/src/constructors/generalized-petersen.c
@@ -1,0 +1,82 @@
+/*
+   IGraph library.
+   Copyright (C) 2021  The igraph development team <igraph@igraph.org>
+
+   This program is free software; you can redistribute it and/or modify
+   it under the terms of the GNU General Public License as published by
+   the Free Software Foundation; either version 2 of the License, or
+   (at your option) any later version.
+
+   This program is distributed in the hope that it will be useful,
+   but WITHOUT ANY WARRANTY; without even the implied warranty of
+   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+   GNU General Public License for more details.
+
+   You should have received a copy of the GNU General Public License
+   along with this program.  If not, see <https://www.gnu.org/licenses/>.
+*/
+
+#include "igraph_constructors.h"
+
+#include "igraph_interface.h"
+
+
+/**
+ * \function igraph_generalized_petersen
+ * \brief Generate the Generalized Petersen graph.
+ * 
+ * The generalized Petersen graph is a graph consisting of an inner 
+ * star graph and an outer cycle graph, each with \c n vertices. The
+ * cycle graph is defined by vertex \c i connecting to vertex \c i+i .
+ * The star graph is defined by vertex \c j connecting to vertex \c j+k .
+ * 
+ * </para><para>
+ * The generalized Petersen graph will have \c 3n edges and \c 2n vertices.
+ * 
+ * </para><para>
+ * Generalized Petersen graphs have some interesting properties, please see 
+ * another source, e.g. Wikipedia for details.
+ * 
+ * \param graph Pointer to an uninitialized graph object, the result will
+ * be stored here.
+ * \param n Integer, \c n is the number of vertices in the star and cycle graphs.
+ * \param k Integer, \c k is the shift for the star graph.
+ * \return Error code.
+ * 
+ * \sa \ref igraph_famous() for the original Petersen graph.
+ * 
+ * Time complexity: O(|V|), the number of vertices in the graph.
+ */
+igraph_error_t igraph_generalized_petersen(igraph_t *graph, igraph_integer_t n, igraph_integer_t k) {
+    /* This is a generalized Petersen graph constructor*/
+    igraph_vector_int_t edges;
+    igraph_integer_t v = 2*n;
+    igraph_integer_t i;
+
+    if (n < 3) {
+        IGRAPH_ERROR("n must be greater than or equal to 3", IGRAPH_EINVAL);
+    }
+
+    if (k < 1 || k > (n-1)) {
+        IGRAPH_ERROR("k must be positive and less than n", IGRAPH_EINVAL);
+    }
+
+    IGRAPH_VECTOR_INT_INIT_FINALLY(&edges, 0);
+    IGRAPH_CHECK(igraph_vector_int_reserve(&edges, 3*n));
+
+    for (i = 0; i < n; i++) {
+        igraph_vector_int_push_back(&edges, i);
+        igraph_vector_int_push_back(&edges, (i + 1) % n);
+        igraph_vector_int_push_back(&edges, i);
+        igraph_vector_int_push_back(&edges, i + n);
+        igraph_vector_int_push_back(&edges, i + n);
+        igraph_vector_int_push_back(&edges, ((i + k) % n) + n);
+    }
+
+    IGRAPH_CHECK(igraph_create(graph, &edges, v, IGRAPH_UNDIRECTED));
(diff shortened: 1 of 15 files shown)


src/constructors/generalized-petersen.c
```diff
@@ -0,0 +1,45 @@
...
+   You should have received a copy of the GNU General Public License
+   along with this program.  If not, see <https://www.gnu.org/licenses/>.
+*/
+
+#include "igraph_constructors.h"
+
+#include "igraph_interface.h"
+
+
+int igraph_generalized_petersen(igraph_t *igraph, igraph_integer_t n, igraph_integer_t k) {
```
szhorvat, line 24: Looks good so far! Two minor things are to change the return type to `igraph_error_t` (this is new on the `develop` branch) and to name the argument `graph` instead of `igraph`. Let us know if you have any questions as you go.

szhorvat:
In case it helps with writing a first test, this is the edge list of G(5,2), as generated with Mathematica. You can paste this directly into an `igraph_small` in the test program, then use an `igraph_isomorphic` for comparison.

```
{0, 2, 0, 3, 0, 5, 1, 3, 1, 4, 1, 6, 2, 4, 2, 7, 3, 8, 4, 9, 5, 6, 5, 9, 6, 7, 7, 8, 8, 9}
```

src/constructors/generalized-petersen.c
alexsyou: Got it, thanks! I've fixed these on the latest commit.

ntamas:
Thanks! You'll also need to add the new function to one of the headers in `include/`; `include/igraph_constructors.h` seems to be the most natural choice.

alexsyou:
In [https://github.com/igraph/igraph/wiki/Contributing-your-first-function-to-igraph](url), Step 7 it describes adding the function name to `interfaces/functions.def`. I cannot find that particular file, so is there somewhere else the function name should be added?

ntamas:
Sorry, this file is now called `functions.yaml` in the `develop` branch and it is a standard YAML-formatted file. I'll update the wiki page.

szhorvat:
Updating this file is a minor thing that does not affect the functionality of the C library, and I think we can leave it until the end.

szhorvat:
Let us know when this is ready for review.

szhorvat:
Let me also suggest a few more test cases. `G(10,2)` should be isomorphic to `igraph_famous("Dodecahedral")`. Since I forgot that `5,2` is in `igraph_famous`, let me give you another edge list for `4,2` instead:

```
{0, 2, 0, 4, 1, 3, 1, 5, 2, 6, 3, 7, 4, 5, 4, 7, 5, 6, 6, 7}
```

Why do we want to restrict `k` to at most `(n-1)/2`? For even `n`, `k = n/2` also seems useful.

We can consider if we should have a limit on `k` at all. Perhaps it's good to have a limit because when `k` is a multiple of `n`, we would get self-loops, which is unexpected. Other large values of `k` are harmless but redundant. Because of the possibility of self-loops, I agree that it's good to have an upper limit.

alexsyou:
Thank you for the extra test cases, I was lacking in ideas for them!


`k=(n-1)/2` was an accident, it should be `k=n/2` for both even and odd `n`. The limit on `k` was placed solely because the sources I found ([1](https://en.wikipedia.org/wiki/Generalized_Petersen_graph), [2](https://mathworld.wolfram.com/GeneralizedPetersenGraph.html)) placed constraints on `k`. However, looking at Watkins's original paper ([3](https://doi.org/10.1016%2FS0021-9800%2869%2980116-X)), `k` is constrained by `1` and `n-1`. I think we should constrain it by `n-1`.

alexsyou:
When I test isomorphism for `igraph_famous("dodecahedron")`, I get an error that the Dodecahedron graph is a multigraph, and therefore I cannot check for isomorphism. However, the Dodecahedron graph is not a multigraph. 

Error message: `Error at src/isomorphism/queries.c:110 : Isomorphism testing is not implemented for multigraphs - Unimplemented function call.`

ntamas:
I strongly suspect that the new `igraph_generalized_petersen()` graph constructed a multigraph somehow and that's why you ended up with the error. Check the edge list of the generated graph (you can use `print_graph()` from `tests/unit/test_utilities.inc` for that).

szhorvat:
Here's a tip: 

When there's a test failure, look at the `linux_static_vendored` test on Azure. It will often show a stack trace.

https://dev.azure.com/igraph-team/igraph/_build/results?buildId=1713&view=logs&j=4c3a8381-b9de-5463-904f-918810769be8&t=a906eab8-d2a3-5e54-d199-606dbbfb0805&l=235

We should document this in the wiki. It tells us that it's the second test case that fails, not the third.

szhorvat:
So it looks like we discovered what is wrong with allowing `k` up to `n-1`: if `2k = n` then we'll get multi edges from `i -- (i+k) % n`connections. In fact, Watkins explicitly disallows `2k = n`.

This was my mistake when giving you the example, as I just naively generated a graph with Mathematica and took the edge list, without looking carefully.

Now we need to decide whether it is worth allowing this special case for generality, or if it's better to exclude it to make sure that the resulting graph really has all the properties that are described in the literature for "generalized Petersen graphs".

I'll have a look at what other systems do with this case.

szhorvat:
- Maple: requires 0 < k < n/2
 - networkx: not available
 - Mathematica: requires n >= 3, k > 0, but has no upper limit on k. If n = 2k, it simplifies multi-edges. If k = m*n, it keeps self-loops. I don't like this.
 - [Sage](https://doc.sagemath.org/html/en/reference/graphs/sage/graphs/graph_generators.html#sage.graphs.graph_generators.GraphGenerators.GeneralizedPetersenGraph): limits to k < n/2
 - Magma: doesn't implement it

I would go with limiting to 0 < k < n/2

Sorry about the confusion.

szhorvat:
Here's a replacement test case for 9,3:

```
0, 3, 0, 6, 0, 9, 1, 4, 1, 7, 1, 10, 2, 5, 2, 8, 2, 11, 3, 6, 3, 12, 4, 7, 4, 13, 5, 8, 5, 14, 6, 15, 7, 16, 8, 17, 9, 10, 9, 17, 10, 11, 11, 12, 12, 13, 13, 14, 14, 15, 15, 16, 16, 17
```

On a related note, in the docs you write that there is an inner "star graph". I see why, since it has the shape of a star. But "star graph" means something different: https://en.wikipedia.org/wiki/Star_(graph_theory)

The inner graph is in fact either a single cycle as well, or as many cycles as the greatest common divisor of n and k. In the above example that's 3.

szhorvat:
Added more tips here: https://github.com/igraph/igraph/wiki/How-to-write-unit-tests%3F#debugging-tips

szhorvat:
I'll make a few small nitpick edits I and will merge this tomorrow.

Thanks for all your work on this @alexsyou !

szhorvat:
Sorry, I had to leave right after I broke things. I want to clarify the docs a bit and will merge after that, probably tomorrow evening.

alexsyou:
Thanks for guiding me through this process!

szhorvat:
Thanks for your help with this @alexsyou ! If you are looking to do something similar in the future, you can give circulant graphs a try:

https://mathworld.wolfram.com/CirculantGraph.html
https://en.wikipedia.org/wiki/Circulant_graph

szhorvat:
@alexsyou The circulant graphs work has been taken by another beginner contributor. If you are interested in further contributions, let us know, and we can find something more interesting for you. Simple tasks like the generalized Petersen and circulant constructors are good for learning the very basics of igraph. Now that you are past that, you are ready for more complex tasks.

PR #1853, opened 2021-11-20
fix: community_label_propagation should not return unlabelled nodes

### LINKED ISSUE #1852: cluster_label_prop() crashes with initial labels
Hello,

The function cluster_label_prop() lets R crash when using an initial label (ok without initial state): I have a graph g of twitter retweets and 

```
V(g)$label = -1
V(g)[name=="XXX"] = 1
V(g)[name=="YYY"] = 2 etc.
cluster_label_prop(as.undirected(g), initial = V(g)$label) # -> crash
```

This does not depend on the size of g.

szhorvat:
Fixes #1852.

In summary, the problem was that if:
 - an initial labelling was given.
 - some nodes were unlabelled, and not reachable from any labelled nodes.
then `-1` values (representing no labelling) were returned in the `membership` vector.

This causes trouble for the end user because it's unexpected, and causes trouble within the function because the -1 values are sent for modularity calculation, which then errors.

The fix used here is the following:

After label propagation is done, find the first unlabelled node. Assign a new label to it, and propagate the same label to all other unlabelled nodes reachable from it. In the directed case, this is a bit messy because the result depends on the node ordering. We randomize the node ordering to be consistent with what is done within the main label propagation algorithm. The undirected case is simpler: the only way we can end up with unlabelled nodes at the end is if an entire connected component was unlabelled from the beginning. Then we just label by connected components.


szhorvat:
@ntamas If you're happy with this, I'll also remove the `At least one vertex must be labeled in the initial labeling.` error, as with this fix it's not really needed. Also, it's not really meaningful. The true requirement is that "all unlabelled vertices should be reachable from some labelled vertex", which reduces to "at least one vertex must be labelled in each connected component" in the undirected case.

ntamas:
Okay, I read #1852 first and then this one, and I noticed that the questions I raised in #1852 were decided one way or another in this PR, and that the case of directed graphs was also handled. I'll review this PR tomorrow and then get back to you.

szhorvat:
Let me answer here.

> However, I wonder whether the vertices that remain unlabelled at the end may be of interest to anyone.

So the question is: should we fix this at all? Reasons to fix it:

 - It is fairly trivial which vertices remain unlabelled. It is those that are not reachable from labelled ones. This is almost easy to compute with `subcomponent()`. See also #1854. You already mention this yourself.
 - If we allow unlabelled vertices to pass through, then we should not call the output a "membership vector", otherwise the expectation is that it is usable wherever membership vectors are appropriate. See how calling `modularity()` failed here. People usually want a usable "community/cluster object" in high-level interfaces, and such an incomplete membership vector will be difficult to tack onto existing community representations without breaking stuff.

---

> and this may potentially change a near-linear time algorithm to the complexity of the connected component search

Finding connected components is also linear time, but it may still have a measurable impact on the performance of this function. This is why in this PR I didn't run a full decomposition. I just traversed the unlabelled parts of the graph, to waste as little CPU time as possible.

---

> So, the way I see it, we have three options:

I prefer option 1. It is nice because:

 - In the undirected case, it is equivalent to adding precisely one label to each connected component and re-running label propagation. The end result does not depend on which node gets labelled.
 - The function will return a standard community structure which is usable with all compatible functions (`modularity`, `compare_communities`, plotting functions, etc.)
 - There is no real performance penalty. I believe that with this PR, the performance will not be worse compared to as if each component had labelled nodes.

However, there is a problem: In the directed case, the behaviour is not as nice. Just adding a label to an unlabelled node and re-running the full label propagation would not necessarily give the same result as my PR does. Unlike in the undirected case, it does matter _which_ node gets the label. On the other hand, the documentation already says that "The input graph, should be undirected to make sense." Does label propagation, _as implemented here_, make sense at all for the directed case? There are papers on this [from as recently as 2019](https://www.sciencedirect.com/science/article/abs/pii/S0375960118311939), which suggests that this is not a simple question. Do we have time to dig into this now? (I don't.)

Option 2:

I don't like this because it produces a result that is not usable as a standard community structure with other parts of the library. I don't know how I would tack this into the Mathematica interface ... 

Option 3:

I don't like this because it is limiting possibilities for no good reason. I would rather use the current PR and issue a _warning_ (not an error). 

---

> So maybe we should decompose the graph into connected components first, run the algorithm component-by-component, and then unify the result into a consistent membership vector in the end?

This is not necessary because the the algorithm already effectively handles components separately.

---

**Summary**

I think that the solution in this PR is very nice for undirected graphs.

But I worry about directed graphs. This solution is not well-founded for that case, but I also don't know what a well-founded solution would be, even ignoring performance implications. Even if we were to choose some nodes to auto-label, and re-run the whole algorithm, I don't know which nodes we should choose. 

In fact, I don't know if this algorithm leads to meaningful results at all in the directed case.

szhorvat:
Assuming that we go ahead with this PR, three questions:

 1. Do we remove the error when no labels are given? (I propose yes.)
 2. Do we add a warning for the directed case? (I propose yes, but I do think it's a bit strange to have a feature but also warning about using it ...)
 3. Do we add a warning when unlabelled nodes remain? (I propose no.)

ntamas:
Okay, read the PR and the discussions once again; I think we should:

* go ahead with the PR
* I wouldn't add a warning to the directed case, it's already in the docs and as you mentioned, it would be strange to print a warning explicitly
* no warning for unlabeled nodes -- if I understand correctly, this PR will ensure that each node would get some label in the end (and it is also deterministic for the undirected case, which is what we are primarily interested in)

I'd probably expand the docs and explain that new labels may be created by the algorithm if there are unlabeled nodes at the end. We should ensure that these are easy to recognize; for instance, they should always be larger than `max(initial_labels)`. End users who care about these labels should then check the membership vector and look for unexpectedly large values.

src/community/label_propagation.c
```diff
@@ -274,6 +276,13 @@ int igraph_community_label_propagation(const igraph_t *graph,
...
+    if (weights) {
+        igraph_inclist_destroy(&il);
+    } else {
+        igraph_adjlist_destroy(&al);
+    }
+    IGRAPH_FINALLY_CLEAN(1);
+
     /* Shift back the membership vector, permute labels in increasing order */
     /* We recycle label_counters here :) */
     igraph_vector_fill(&label_counters, -1);
```
ntamas, line 288: `unlabelled_left` is initialized up at the top of the function but it is not used up to this point. I'd leave the _declaration_ up at the top and move the _initialization_ here so it stays closer to the for loop where its "real" value is being computed. (In other words, the `for` loop has a pre-condition that `unlabelled_left` is false, and it is easier to see the pre-condition if it is declared here and not at the top).

ntamas:
When reading this PR, I realized that we do a re-labeling of nodes at the end such that there are no "empty" clusters. I don't know whether this is good or bad; I could argue for both

* On one hand, this makes the results easier to interpret for the naked eye.
* Also, if there are, say, 200 nodes in the graph but only two significant clusters, it is quite random which of the original labels end up dominating the graph; all that matters is that there are two labels in the end, so it makes sense to return a membership vector that looks like `[lots-of-zeros, lots-of-ones]` instead of `[lots-of-X, lots-of-Y]` where X and Y are arbitrary integers. This also helps in finding consensus clusterings when the algorithm is executed many times and the results are averaged.
* On the other hand, this is problematic if there are initial labels and they have some kind of semantic meaning to the user who calls the function; e.g., the user may have six categories and decide to use categories 4, 5 and 6 as initial labels, intentionally leaving categories 1, 2 and 3 unused in the initial labeling. If this happens, we re-write the labels at the end so the final membership vector ends up containing labels 1, 2 and 3 instead of 4, 5 and 6, and the user needs to figure out which of the original labels that they assigned ended up at which final label.

We shouldn't do anything about this in this PR, but maybe it's a good candidate for an API-breaking change in 0.10 where we could introduce an additional argument that specifies whether to "canonicalize" the membership vector at the end? Or maybe we shouldn't canonicalize at all and just provide a function that reassigns items in an arbitrary membership vector such that there are no empty clusters any more?

ntamas:
Waiting for CI to pass and then I'll merge. Thanks a lot!

szhorvat:
> When reading this PR, I realized that we do a re-labeling of nodes at the end such that there are no "empty" clusters. I don't know whether this is good or bad; I could argue for both

I noticed the same and I also wondered if keeping the original labels would have been good ...

Do any `membership`-handling functions expect that cluster labelling is contiguous? Hopefully not.

Thanks for merging!

szhorvat:
> where we could introduce an additional argument that specifies whether to "canonicalize" the membership vector at the end?

We could even have a separate function for canonicalizing membership vectors.

ntamas:
> Do any membership-handling functions expect that cluster labelling is contiguous?

I don't think so; however, they could do something like allocating a vector whose length is equal to the maximum cluster index plus one, in which case one could run into problems if arbitrary integers are used. In such cases it would be advisable to canonicalize the membership vector first.

Actually, we _do_ have such a function: `igraph_reindex_membership()`.

lovre:
Regarding applying label propagation to directed networks, in our chapter, we have written:

"...there seem to be no straightforward extension to networks with directed arcs. The reason for this is that propagating the labels exclusively in the direction of arcs enables exchange of labels only between mutually reachable nodes forming a strongly connected component. Since any directed network is a directed acyclic graph on its strongly connected components, the labels can propagate between the nodes of different strongly connected components merely in one direction. Therefore, one usually disregards the directions of arcs when applying the label propagation method to directed networks except in the case when most arcs are reciprocal."

As far as I know, there is no straightforward extension of label propagation to directed networks. Any such paper either changes the optimization function (e.g. to modularity) or the definition of community (e.g. flow communities).

vtraag:
Thanks for sharing @lovre! Given what he writes, wouldn't it make more sense then perhaps to limit the label propagation to undirected graphs only, @ntamas and @szhorvat ?

szhorvat:
I propose this:

 - In 0.9, there should be a warning when the input is directed. "The result is not meaningful for directed input."
 - In 0.10, we can ignore edge directions, and show a different warning: "Edge directions are ignored." We might remove that warning in some future version.

My only concern is that the functionality _is_ already implemented, the mechanism is straightforward and easy to understand (even if its outcome isn't always useful). Why not let people experiment if they want to? @lovre, do you think there's value in keeping it in some form?

@lovre Since you are deeply familiar with the method, I am wondering if you'd have the time to improve our documentation a bit. Ideally, there would be a bit of high-level guidance in the docs about when it is or isn't reasonable to use a given method. I like documentation that is also educational. If you could provide a short paragraph here on GitHub, I'd be grateful (and I'd integrate it into the docs). Of course, I understand if you don't have the time for this.

szhorvat:
Here's a small experiment where I generated a graph from an SBM and took the largest strongly connected component.  When ignoring edge directions, I typically get results like these:

image
image
image

When directions into account,

image
image
image

lovre:
@szhorvat, it probably makes sense to allow the user to propagate the labels only in the direction of the edges. But, if I would be implementing this, I would only include this as an option/parameter that the user must set explicitly. Otherwise, I would ignore the directions and output a warning that the input graph is treated as undirected. This is how I have always implemented it in the past. In other words, I do not even understand what kind of clusters one should expect under "directed" propagation. I believe that the experiments above confirm this.

Let me keep this message in my inbox and I will try to get back to it in December. I am currently spending more than 15 hours per week in the classroom and I have very limited time for anything else. I expect that we will be in lockdown starting with December, when I will have more time.

ntamas:
> In 0.9, there should be a warning when the input is directed. "The result is not meaningful for directed input."

That's probably too harsh; I mean, if the graph is strongly connected, then the result _could_ be meaningful, or at least it's not less meaningful than trying to apply any other community detection algorithm on a directed graph. In my head, the very notion of communities in a directed graph is not as clear-cut as in undirected graphs so some ambiguity is expected. It is the responsibility of the user to interpret the results, but it is nevertheless a good idea to provide a parameter that allows the user to decide whether edge directions should be taken into account or not. Most other igraph functions do this as well, either with an `igraph_bool_t directed` argument or an `igraph_neimode_t mode` argument (which also allows to consider the edges in reverse).

szhorvat:
Writing from a train so I'm brief...

Let's do a mode argument for 0.10, and add a note to the docs that high
level interfaces are recommended to use ALL by default unless explicitly
setting IN or OUT.

Let's leave 0.9 as it is now on master.

Agreed?

--
sent from a phone

ntamas:
I'll do it.

Issue #1851, opened 2021-11-20
Circulant Graphs


Gomango999:
**What is the feature or improvement you would like to see?**

A circulant graph generator, as mentioned in [#1844](https://github.com/igraph/igraph/pull/1844#issuecomment-972959559)

A circulant graph is a graph such that if the n vertices are labelled from 0 to n-1, and two vertices numbered x and x+d mod n are adjacent, then all vertices numbered z and z+d mod n are adjacent.

The function `G(n, l)` will accept `n`, the number of vertices, and `l`, a list of distances to generate the graph. We will then
- Create `n` nodes `v_0, v_1, ..., v_{n-1}`
- For each node `v_i` and for each distance `d_j`, connect `v_i` to `v_(i+d_j mod n)`

**Use cases for the feature**

From Wikipedia: Circulant graphs can be used to generate Paley graphs, which have applications in number theory, as well as Möbius ladders, which can be used to model molecular structures. There is also a larger list of other classes of graphs based on Circulant graphs which can be found on Wolfram Alpha.

**References**

https://mathworld.wolfram.com/CirculantGraph.html
https://en.wikipedia.org/wiki/Circulant_graph

Gomango999:
I'd also be happy to take on this project as a (hopefully much easier!) first contribution

GroteGnoom:
Awesome that you also made the issue, that makes things a little clearer. Looking forward to your work 🙂

ntamas:
> This way you can quickly test your own code, and if you want a full test you can still do `cmake --build . --target check`

Another neat trick is to install `ccache` and `ninja`, which makes even full rebuilds blazing fast. `ccache` caches compiled object files, while `ninja` is simply much faster than `make`. You need to tell CMake explicitly that you want to use Ninja (`mkdir build && cmake -GNinja ..`), but `ccache` works out of the box as far as I know.

ntamas:
@Gomango999 You can look at https://github.com/igraph/igraph/pull/1844 for inspiration about how a PR of a new graph constructor function looks like.

szhorvat:
Great job on opening the issue! We can point newcomers to it as an example of how to get started.

szhorvat:
> Do you want an option to make it either a directed or undirected graph?

It would be nice if you guys could do a little research on what is being written about directed circulant graphs (if anything), and how they're typically defined.

[...]

Issue #1850, opened 2021-11-17
glpk_error test sometimes segfaults


szhorvat:
**Describe the bug**
The `glpk_error` test sometimes segfaults. Example:

https://github.com/igraph/igraph/runs/4125448797?check_suite_focus=true

```
 399/399 Test #394: test::glpk_error ..................................***Exception: SegFault  0.36 sec
```

**To reproduce**

I cannot reproduce it yet. The failure is very rare.

This test is meant to check that error handling and interruption work with GLPK. One of the tests will wait for a set amount of time (1 second) before interrupting. This shows why reproduction is tricky (timing dependent). I wrote this test, but did not have a better way to test interruption.

It should be verified if these failures started after the GLPK 5.0 or were present before.

**Version information**

Present on `master` (as per linked example). Should be checked if also present on `develop`.

szhorvat:
This may or may not have to do with broken `setjmp`/`longjmp` in MinGW, as mentioned here: https://stackoverflow.com/q/53709069/695132 We use setjmp/longjmp to allow interruption in GLPK.

szhorvat:
This is reproducible locally, but requires a very large number of trials.

When the crash happened for me, it was indeed at the point of the `longjmp`, as confirmed with some simple printf debugging. 

The `glpk_error` test has two sections:

 - Simple deterministic error reporting
 - Timing-based interruption. Since this uses an interrupt handler that signals the interrupt after 1 second, it is not deterministic.

I always assumed that the crash was random because it occurred in the non-deterministic part. However, when I managed to trigger it locally, it actually happened in the first part, which triggers an error completely deterministically.

szhorvat:
To be clear, I had a pair of `fprintf(stderr, "..."); fflush(stderr);` both directly before the `longjmp()` call and one directly at the receiving `setjmp()`. This confirms that the crash happens during the jump.

Why would it crash?

There are many restrictions by the standard on how [setjmp](https://en.cppreference.com/w/c/program/setjmp)/[longjmp](https://en.cppreference.com/w/c/program/longjmp) can be used. I checked them again and I cannot see what we are doing wrong.

It seems likely that this is a problem with MinGW and that there is not much we can do about it. There are similar complaints around the internet, e.g. [here](https://stackoverflow.com/q/537090690)

ntamas:
The last link you posted goes to a 404 page.

szhorvat:
[...]

PR #1856, opened 2021-11-24
Circulant Graph Constructor

Gomango999:
This pull request implements the circulant graph constructor, as requested in #1851.


GroteGnoom:
Great stuff, there's a lot of little things you've done right which we now don't have to worry about anymore :) 

Don't forget to add the function to doc/generators.xxml, then the documentation will be included.

tests/unit/igraph_circulant.c
```diff
@@ -0,0 +1,28 @@
...
+#include <igraph.h>
+
+#include "test_utilities.inc"
+
+int main() {
+
+    // TODO: Add a test
+
+    return 0;
+}
```
GroteGnoom, line 28: You can set your editor to always put a newline at the end of a file, that prevents this issue ( no newline at end of file). It's a really minor thing, but if you for example want to view a lot of files in one go by using `cat` then this matters.

ntamas: Or, just install [`pre-commit`](https://pre-commit.com/) and run `pre-commit install` in your clone of the repo; this will install our preferred set of pre-commit hooks that get executed before every commit. The pre-commit hooks will take care of trailing newlines for you.

GroteGnoom: As you see, I'm also still learning things :D

Gomango999: Ah I was wondering what the failed test was. Thanks!

Gomango999:
Just added some basic tests, planning to add some more when I get time. I was wondering whether it is possible to test for invalid inputs which call `IGRAPH_ERRORF`? For example, testing that my program correctly errors when a negative value for n is given.

ntamas:
> I was wondering whether it is possible to test for invalid inputs which call `IGRAPH_ERRORF`?

`tests/unit/test_utilities.inc` includes a handy helper macro called `CHECK_ERROR`; see, e.g., the `igraph_sbm_game.c` test case for usage hints.

CONTRIBUTING.md
```diff
@@ -155,7 +155,7 @@ Follow the following steps if you would like to make a new pull request:
    ```bash
    git pull [--rebase] upstream <dev-branch>
    ```
-   
+
```
ntamas, line 158: Please do not add changes to unrelated files in the PR.

src/centrality/hub_authority.c
```diff
@@ -142,7 +142,7 @@ static igraph_error_t igraph_i_kleinberg_weighted(igraph_real_t *to,
  * \brief Kleinerg's hub and authority scores.
  *
  * The hub and authority scores of the vertices are defined as the principal
- * eigenvectors of <code>A*A^T</code> and <code>A^T*A</code>, respectively,
```
ntamas, line 145: Please revert this file to the original state in the PR; it is great that the trailing whitespace was removed, but the file is unrelated to the PR and I'd rather keep things clean. I'll clean up the trailing whitespace in this file in a separate commit.

src/constructors/circulant.c
```diff
@@ -0,0 +1,97 @@
...
+#include "igraph_constructors.h"
+#include "igraph_interface.h"
+
+/**
+ * \function igraph_circulant
+ * \brief Creates a circulant graph.
+ *
+ * A circulant graph <code>G(n, l)</code> consists of \p n vertices \c v_0, ...,
+ * \c v_(n-1) such that for each \c l_i in the list of offsets \p l, \c v_j is
+ * connected to <code> v_(j + l_i mod n) </code> for all j.
```
ntamas, line 28: I think this should be `v_((j + l_i) mod n)` just to be clear that the modulo applies to the entire sum. I'm not sure what the "official" precedence of the modulo operator is.

src/constructors/circulant.c
```diff
@@ -0,0 +1,97 @@
...
+ *
+ * </para><para>
+ * The function works with both directed and undirected graphs. Multiple edges are
+ * merged, and self loops are ignored. All offsets are taken modulo \p n, so values greater than n as well as negative offsets are permitted.
+ *
+ * \param graph Pointer to an uninitialized graph object, the result will
+ * be stored here.
+ * \param n Integer, \p n is the number of vertices in the circulant graph. It must
+ * be at least 1.
+ * \param l Integer Vector, \p l is a list of the offsets within the circulant graph.
```
ntamas, line 38: `Vector` should be lowercase.

src/constructors/generalized_petersen.c
```diff
@@ -24,17 +24,17 @@
 /**
  * \function igraph_generalized_petersen
  * \brief Creates a Generalized Petersen graph.
- *
```
ntamas, line 27: As above; please revert this file to its original content and then I'll fix up the whitespace in a separate commit.

tests/unit/igraph_circulant.c
```diff
@@ -0,0 +1,176 @@
...
+    igraph_bool_t iso;
+    igraph_integer_t i;
+
+    /* Testing n = 1 case */
+    /* n = 1, l = [1], undirected */
+    igraph_vector_int_init(&l, 0);
+    igraph_vector_int_push_back(&l, 1);
+    IGRAPH_ASSERT(igraph_circulant(&graph, 1, &l, IGRAPH_UNDIRECTED) == IGRAPH_SUCCESS);
+    igraph_small(&graph_test, 1, IGRAPH_UNDIRECTED, -1);
+    IGRAPH_ASSERT(igraph_is_same_graph(&graph, &graph_test, &iso) == IGRAPH_SUCCESS);
```
ntamas, line 36: This technique is okay, but for the future, the alternative is simply to print the graph to `stdout` with `print_graph_canon()` and provide an expected output file in `igraph_circulant.out`. Our unit test framework will compare the expected output with the printed output and print a diff if they are different so it's easier to see which edges are missing or which edges have been added incorrectly.

ntamas (changes requested):
Great job, thanks a lot! There are only a few minor issues to be fixed; see the comments that I added. I'll fix up the whitespace in `develop` in a separate commit.

ntamas:
Don't worry about the CI failures; the one on Alpine Linux seems to be unrelated and the MINGW/ucrt-x86_64 test case sometimes fails randomly in the `glpk_error` test case (see #1850) and we haven't figured out yet why.

CONTRIBUTING.md
szhorvat: Wasn't this a result of the pre-commit hook? I admit I never used the pre-commit hooks.

ntamas: pre-commit hooks are run only on the files that were changed, unless you run them explicitly on all files with `pre-commit run --all`. Anyhow, I have removed all trailing whitespace now in all source files so an easy way out is to rebase this PR on the current `develop`.

ntamas:
You can now rebase the PR on `develop` to get the whitespace issue fixed.

CONTRIBUTING.md
Gomango999: Yep  my bad,  I had run pre-commit run --all-files, and it had autofixed all the files. I've reverted these and rebased as well to avoid merge conflicts.

src/constructors/circulant.c
Gomango999: Done.

src/constructors/circulant.c
Gomango999: Done.

src/constructors/circulant.c
```diff
@@ -0,0 +1,97 @@
...
+            offset += n;
+        }
+        if (!directed) {
+            if (offset >= (n + 1) / 2) {
+                offset = n - offset;
+            }
+        }
+
+        /* only use offset if non-zero and we haven't seen it before */
+        if (offset != 0 && !VECTOR(offset_seen)[offset]) {
```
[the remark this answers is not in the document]
Gomango999: Done.

tests/unit/igraph_circulant.c
Gomango999: Yep that would have been alot more natural. I'll keep it in mind for next time!

Gomango999:
Ok I've both reverted the commit as well as rebased, so as to avoid any merge conflicts. I've also added some extra tests for invalid values of n, and made the corrections to style.  Please let me know if there's anything else!

src/constructors/circulant.c
```diff
@@ -0,0 +1,99 @@
...
+ */
+
+igraph_error_t igraph_circulant(igraph_t *graph, igraph_integer_t n, const igraph_vector_int_t *l, igraph_bool_t directed) {
+
+    igraph_vector_int_t edges;
+    igraph_vector_bool_t offset_seen;
+    igraph_integer_t i, j;
+    igraph_integer_t limit;
+
+    if (n <= 0) {
```
GroteGnoom, line 55: We can also accept zero and generate an empty graph.

src/constructors/circulant.c
```diff
@@ -0,0 +1,99 @@
...
+ * \param directed Boolean, \p directed determines whether the graph should be directed.
+ * \return Error code.
+ *
+ * \sa \ref igraph_ring(), \ref igraph_generalized_petersen().
+ *
+ * Time complexity: O(|V||L|), the number of vertices in the graph times the number
+ * of offsets.
+ */
+
+igraph_error_t igraph_circulant(igraph_t *graph, igraph_integer_t n, const igraph_vector_int_t *l, igraph_bool_t directed) {
```
GroteGnoom, line 48: You can pass `l` directly, it's a small struct.

src/constructors/circulant.c
Gomango999: Ok, I've accounted for that

src/constructors/circulant.c
Gomango999: Why would copying the entire struct be better than passing just the pointer? I also noticed that other functions used pointers as well, e.g. [igraph_realize_degree_sequence](https://igraph.org/c/doc/igraph-Generators.html#igraph_realize_degree_sequence)

GroteGnoom: You're correct. Your way is consistent with the rest of igraph, I don't know why I didn't notice.

src/constructors/circulant.c
```diff
@@ -0,0 +1,102 @@
...
+   This program is distributed in the hope that it will be useful,
+   but WITHOUT ANY WARRANTY; without even the implied warranty of
+   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+   GNU General Public License for more details.
+
+   You should have received a copy of the GNU General Public License
+   along with this program.  If not, see <https://www.gnu.org/licenses/>.
+*/
+
+#include "igraph_constructors.h"
```
szhorvat, line 19: Please add a newline after this line. We like to keep the header in which the prototype of this function lives at the top, and visually separated from the following ones.

src/constructors/circulant.c
```diff
@@ -0,0 +1,102 @@
...
+ * \param directed Boolean, \p directed determines whether the graph should be directed.
+ * \return Error code.
+ *
+ * \sa \ref igraph_ring(), \ref igraph_generalized_petersen().
+ *
+ * Time complexity: O(|V||L|), the number of vertices in the graph times the number
+ * of offsets.
+ */
+
+igraph_error_t igraph_circulant(igraph_t *graph, igraph_integer_t n, const igraph_vector_int_t *l, igraph_bool_t directed) {
```
szhorvat, line 48: I would prefer choosing a different notation than `l` due to its similarity to `1` and `I`. Either a full word, such as `shifts`, or a different letter, such as `k` or `j`.  

Is it just me who's bothered by the `l`?

src/constructors/circulant.c
```diff
@@ -0,0 +1,102 @@
...
+    if (n == 0) {
+        return igraph_empty(graph, 0, directed);
+    }
+
+    IGRAPH_VECTOR_INT_INIT_FINALLY(&edges, 0);
+    IGRAPH_VECTOR_BOOL_INIT_FINALLY(&offset_seen, n);
+
+    VECTOR(offset_seen)[0] = 1; /* do not allow self loops */
+
+    for (i = 0; i < igraph_vector_int_size(l); i++) {
```
szhorvat, line 67: Can you please move `igraph_vector_int_size()` outside if the loop? Record the value, then use it. Avoid calling `_size()` functions repeatedly in loops because they are expensive.

For this function, performance is not an issue. However, I would prefer to avoid this pattern because people sometimes look at the existing code and imitate it. 

In C++, it's fine to loop until `i < container.size()` because `.size()` is typically an inline function, thus cheap. However, in this case, `_size` functions are defined a different compilation unit, so a complete (and potentially slow) function call cannot be avoided (unless compiling with LTO).

src/constructors/circulant.c
```diff
@@ -0,0 +1,102 @@
...
+    igraph_integer_t limit;
+
+    if (n < 0) {
+        IGRAPH_ERRORF("number of nodes = %" IGRAPH_PRId " must be non-negative.", IGRAPH_EINVAL, n);
+    }
+    if (n == 0) {
+        return igraph_empty(graph, 0, directed);
+    }
+
+    IGRAPH_VECTOR_INT_INIT_FINALLY(&edges, 0);
```
szhorvat, line 62: Usually, there will be `size(l) * n` edges. Sometimes, there will be fewer because of graph simplification. Here, we _could_ reserve `2 * size(l) * n` amount of space for the edges vector using `IGRAPH_CHECK(igraph_vector_int_reserve(...))`, so that all the `igraph_vector_int_push_back()` calls later will not keep re-allocating.

Ultimately, it's not a big deal here because this is a not a performance-critical function. I still wanted to point this out because in _some_ cases such an optimization can make a noticeable different, and I expect that you will continue contributing @Gomango999, so it's good to be aware of it.

tests/unit/igraph_circulant.c
```diff
@@ -0,0 +1,195 @@
...
+    igraph_t graph, graph_test;
+    igraph_vector_int_t l;
+    igraph_bool_t iso;
+    igraph_integer_t i;
+
+    /* Testing invalid values */
+
+    /* n = -3, l = [1], undirected */
+    igraph_vector_int_init(&l, 0);
+    igraph_vector_int_push_back(&l, 1);
```
szhorvat, line 34: Tip for when you want to create a small vector with specific elements:

Take a look at `igraph_vector_int_init_int_end()` and `igraph_vector_int_init_int()`. We can create the vector (1,2,3) as:

```
igraph_vector_int_init_int(&vec, 3 /* length */, 1,2,3 /* elements */);

/* or */

igraph_vector_int_init_int(&vec, -1 /* end marker */, 1,2,3, /* elements */ -1 /* the end */);
```

It's up to you what you use. This is just so you are aware of possibilities. There is no need to rewrite this test.

tests/unit/igraph_circulant.c
```diff
@@ -0,0 +1,195 @@
...
+    IGRAPH_ASSERT(igraph_vcount(&graph) == 0);
+    IGRAPH_ASSERT(igraph_ecount(&graph) == 0);
+    igraph_vector_int_destroy(&l);
+
+
+    /* Testing n = 1 case */
+    /* n = 1, l = [1], undirected */
+    igraph_vector_int_init(&l, 0);
+    igraph_vector_int_push_back(&l, 1);
+    IGRAPH_ASSERT(igraph_circulant(&graph, 1, &l, IGRAPH_UNDIRECTED) == IGRAPH_SUCCESS);
```
szhorvat, line 53: You have allocated a graph on line 43. Here, you re-allocate on top of it, without having destroyed it first. This is why some tests fail:

https://dev.azure.com/igraph-team/igraph/_build/results?buildId=1753&view=logs&j=4c3a8381-b9de-5463-904f-918810769be8&t=a906eab8-d2a3-5e54-d199-606dbbfb0805&l=205

It's good to look at the `linux_static_vendored` test because it has various sanitizers enabled, and it prints stack traces even for non-sanitizer errors.

tests/unit/igraph_circulant.c
```diff
@@ -0,0 +1,195 @@
...
+    igraph_destroy(&graph_test);
+    igraph_vector_int_destroy(&l);
+
+    /* Testing n = 2 case */
+    /* n = 2, l = [1], undirected */
+    igraph_vector_int_init(&l, 0);
+    igraph_vector_int_push_back(&l, 1);
+    IGRAPH_ASSERT(igraph_circulant(&graph, 2, &l, IGRAPH_UNDIRECTED) == IGRAPH_SUCCESS);
+    igraph_small(&graph_test, 2, IGRAPH_UNDIRECTED, 0, 1, -1);
+    IGRAPH_ASSERT(igraph_is_same_graph(&graph, &graph_test, &iso) == IGRAPH_SUCCESS);
```
szhorvat, line 67: Nitpick: In several places, you use the name `iso` with `igraph_is_same_graph`. Can you please change this to a different name, such as `same`? `igraph_is_same_graph` does not test isomorphism, but checks if the graphs have the same edges and vertices. While you understand that, I would like to point other beginners to your PR to show them how a contribution goes. I worry they may be confused by this.

src/constructors/circulant.c
```diff
@@ -0,0 +1,102 @@
...
+ *
+ * \param graph Pointer to an uninitialized graph object, the result will
+ * be stored here.
+ * \param n Integer, \p n is the number of vertices in the circulant graph. It must
+ * be at least 0.
+ * \param l Integer vector, \p l is a list of the offsets within the circulant graph.
+ * \param directed Boolean, \p directed determines whether the graph should be directed.
+ * \return Error code.
+ *
+ * \sa \ref igraph_ring(), \ref igraph_generalized_petersen().
```
szhorvat, line 42: Let's add `igraph_extended_chordal_ring`. I'll do myself that when I go over the docs one last time.

szhorvat (changes requested):
It looks good! There were only a few tiny nitpicks, and the failing test needs to be fixed (destroy the graph). Before it's merged, I'd like to edit the docs a little bit, but that will be at the very last step.

src/constructors/circulant.c
ntamas: No, that's a good point.

src/constructors/circulant.c
Gomango999: Done

src/constructors/circulant.c
Gomango999: Yeah `l` was slightly bothering me as well. I've changed it to `shifts`, as well as the code inside to refer to shifts as well to avoid confusion.

src/constructors/circulant.c
Gomango999: Got it. I've never had to actually optimise C code before, so I'm super happy to be learning all the little tips and tricks!

src/constructors/circulant.c
Gomango999: Done. I expect to contributing again, so definitely another thing for me to make a mental note of next time.

tests/unit/igraph_circulant.c
Gomango999: Definitely alot cleaner, so I've gone ahead and updated the tests anyway. I've also changed the tests to use `shifts` instead of `l` to be consistent.

tests/unit/igraph_circulant.c
Gomango999: Done

tests/unit/igraph_circulant.c
Gomango999: Ah, thanks for finding that for me. I had accidently assumed it was another linux_alpine issue, so didn't look deeper. The graph has been properly destroyed now.

src/constructors/circulant.c
Gomango999: I've completed all the suggested changes. You're free to make the final adjustments now

ntamas (approved):

ntamas:
Alpine Linux build problems fixed on `develop`; I guess there's no need to rebase if the tests pass in all other environments.

szhorvat:
Thanks @Gomango999, excellent work on this!

Gomango999:
And thanks for all the support and reviews from you and everyone else during the process! I have exams coming up soon so it might be a while, but I will definitely be returning :)

PR #1860, opened 2021-11-30
Implement umap

GroteGnoom:
I've started trying to implement UMAP. There's so much new stuff, that I'm writing code just to explore the problem and see what we need. 

**Prototype iteration summary:**
- Initially start from data in a vector space: perhaps skip graph building and start directly with weighted graph (e.g. knn) in a later iteration
- Start directly with a sparse graph with "distances"
- Implement rough functions
- Implement conversion distance->probability in original graph, `mu(a)`
- Implement smoothed Euclidean distance->probability in embedded space, `nu(a)`, via Gauss-Newton fitting of hyperparameters a and b (using line search to deal with strong off-diagonal elements of the Hessian)
- Implement binary search for `sigma`, a rough estimate of the decay of the neighbor distance beyond the closest neighbor
- Implement corner cases for the `sigma` calculation
- Implement cross-entropy (CE) function
- Implement derivatives of the CE for attractive and repulsive forces
- Double check everything
- Put it all together
- Write header file
- Write unit test code
- Write unit test output
- Test produces reasonable results, no strict checking yet ('t's a stochastic algo, so not sure how to do that?)
- Check function naming
- Write function docstring (first draft)
- Tests pass locally (Fabio's computer)
- Add a bunch of brackets that depend on `#define UMAP_DEBUG` for printing stuff... is this any good?
- Redo the math on the CE and derivatives to make sure we are doing it correctly
 - Add unit test for unweighted graph
 - Lint code and docs
 - 3D (theoretically ND) version
 - Cover trivial graphs
 - improve docs on `(a,b)` fit scaling behaviour
 - Stress test under a few corner and not-so-corner cases
- Ask other igraph folks for review/suggestions
- Move to `layout`

Todo:
- Decide what to do about initializatoin choices
- Merge PR :smirk:

Possible optimization:
- Use openMP

edit (Fabio): added brief idea list as requested by @szhorvat


iosonofabio:
Thank you!!! So excited to see this happening! Let me know if I can help!

src/misc/umap.c
```diff
@@ -0,0 +1,262 @@
...
+                VECTOR(*smallest)[max_index] = i;
+            }
+        }
+    }
+    igraph_vector_destroy(&smallest_values);
+    IGRAPH_FINALLY_CLEAN(1);
+    return (IGRAPH_SUCCESS);
+}
+
+/*This creates a k-nearest neighbors graph with their distances.*/
```
iosonofabio, line 58: My intention would be to work with an *existing* graph (usually knn) instead of making one from whatever vector space, so that we can trash this part. Indeed, the perception that distance computation (e.g. correlation, Euclidean), graph construction (via knn or whatever other method), and layouting are one thing is the source of many confusions in the field I think.

iosonofabio: For now, whatever you find most useful, but the function signature at the end of the day would be expected... ehr, at least *I* would expect it - somewhat naively perhaps - to be:

graph pointer -> layout pointer

instead of:

data pointer -> layout pointer

src/misc/umap.c
```diff
@@ -0,0 +1,262 @@
...
+    igraph_vector_destroy(&node_distances);
+    igraph_vector_destroy(&data_row_i);
+    igraph_vector_destroy(&data_row_j);
+    igraph_vector_int_destroy(&closest);
+    igraph_vector_int_destroy(&edges);
+    IGRAPH_FINALLY_CLEAN(5);
+    return (IGRAPH_SUCCESS);
+}
+
+/*open set size is just the size of the distance to the closest neighbor*/
```
iosonofabio, line 99: notice that the concept of *distance* here is a little misleading. Physicists find it useful, but in practice it's usually defined as something like

    d (v1, v2) := 1 - correlation(v1, v2)

which is actually funny because then it's this weird object between 0 and 2. In other words, the edge weight as a *similarity* score, something like max(0, correlation(v1, v2)) is a better representation of the situation compared to rephrasing the whole problem as a distance matrix.

Of course, the embedding space itself is a 2D Euclidean thing, so distances matter there. But that does not change the fact that in the original space, similarity/edge weight has more meaning than distance.

Not sure how useful all of this actually is for you now ;-)

src/misc/umap.c
```diff
@@ -0,0 +1,262 @@
...
+        igraph_matrix_sub(layout, &gradient);
+    }
+
+    IGRAPH_FINALLY_CLEAN(1);
+
+    return (IGRAPH_SUCCESS);
+}
+
+igraph_error_t igraph_umap(igraph_matrix_t *data, igraph_matrix_t *layout) {
+    igraph_t knn_graph;
```
iosonofabio, line 235: As above, ideally we start directly from a similarity graph (call it knn if you like the name). How that is constructed should not be our business I believe?

src/misc/umap.c
szhorvat: > My intention would be to work with an existing graph

I would like to ask to briefly document these things, i.e. what the intention is, in the top post. Hopefully it's not too much work—the briefest note will do. The idea is that those of is who can't follow comments daily can still catch up later, without having to read through the entire thread, and go through past ideas that already got discarded during the journey.

GroteGnoom:
Thanks for your comments @iosonofabio, I will look a bit more at them tomorrow :)

src/misc/umap.c
GroteGnoom: (I'm using this process: https://umap-learn.readthedocs.io/en/latest/how_umap_works.html)
I used a graph twice, first to get a KNN graph, and then I used that graph to create a new graph, which is then used in the layout process. Do you mean the function should accept the first graph, which should then be transformed using the fuzzy open sets, which should then generate the new graph?
Or should this function only do the minimization of the cross entropy?

iosonofabio: Good point, I'm referring to the former here: we do not need to build the knn graph from a vector space in igraph I think (but am open to other opinions). As for option 2, even though having a function that does purely operation 2 might be useful in the future, ATM it sounds too specialized.

Since we are here, one word on the article. Basically our starting point - the way the current implementation works as well -  is a weighted **truncated** knn graph, i.e. a similarity graph in which each vertex/cell is connected (with weight equal to some measure, i.e. correlation) to its k neighbors *unless you run out of neighbors above a certain similarity* (tipically 0.2). Then you have to transform that into probabilities and minimize the cross-entropy to yield the 2D layout.

So even though the article says that you don't want a fixed radius on the high-dimensional balls but can use knns to "adapt to the local Riemann geometry", in practice software implementations do have an *additional* maximal radius (in practice, minimal simliarity), because otherwise you *always* end up connecting vertices/cells that are really different by "warping too much", and your topology goes to pot by basically folding the back of the sheet back onto itself. I had some fun with that cutoff in the past and biologically you can see that cells with correlation >0.9 should basically be merged (a la Louvain/Leiden), cells with 0.75-0.9 are "birds of a feather" (similar but not redundant), and cells <0.75 are already quite different from one another and are only connected together mainly to keep some appearance of "global geometry" on the space. So in my mind the whole "Riemaniann warping" technique works a little differently from how the mathematician imagined and would be an interesting topic of research in its own right. But I'm digressing!

Happy to chat on whatsapp/zoom if you prefer, but it sounds like you grasp this better than me already, unsurprisingly.

GroteGnoom: The python umap implementation has an enormous amount of parameters you can tweak. One of the parameters to tweak for our umap could be the cutoff for the distance. One problem here is that I first want to have a basic umap working, before adding lots of parameters, but without all those parameters it just doesn't do what you expect. For now I'll just use 0.2 as a hardcoded cutoff.

umap just seems crazily general. Make fuzzy balls, use it to make a graph where close things are connected, and then shake your layout about until some function of the graph and layout is minimized.

GroteGnoom: If I don't have the original data, any point in the umap graph can only be connected to other points if they're already connected in the knn graph, right? Because I have no clue about other distances.

iosonofabio: I think so

iosonofabio: >  The 0.2 is done before our part, so that's ok.

Yes that's part of my argument 😊

> umap just seems crazily general. Make fuzzy balls, use it to make a graph where close things are connected, and then shake your layout about until some function of the graph and layout is minimized.

See why I want to have it in igraph? I think it's just the first instance of a family of similar layouts (e.g. denseMAP, which I think is quite useless)

GroteGnoom:
(I'm now reading the paper a bit more, which clears things up. )
They use spectral embedding to get an initial layout. Shall we use our spectral embedding or avoid it? 
I first encountered some problems with it here: https://igraph.discourse.group/t/the-unit-tests-thread/597/48
Umap could be a reason to look a bit more at spectral embedding, test it, use it, and fix it. But right now I believe nobody is an expert on this, so it will be hard to fix and maintain.

iosonofabio:
> (I'm now reading the paper a bit more, which clears things up. ) They use spectral embedding to get an initial layout. Shall we use our spectral embedding or avoid it? I first encountered some problems with it here: https://igraph.discourse.group/t/the-unit-tests-thread/597/48 Umap could be a reason to look a bit more at spectral embedding, test it, use it, and fix it. But right now I believe nobody is an expert on this, so it will be hard to fix and maintain.

I'd be in favour of not caring about spectral embedding for now - as far as this PR is concerned. Sure, minimization will take a little longer, but we're not out for speed quite yet anyway.

GroteGnoom:
On page 19 of the [paper](https://arxiv.org/abs/1802.03426), they do a binary search for the σ parameter, which determines how fast the fuzziness of the fuzzy ball falls off with distance beyond the nearest neighbor. To me it seems like you should calculate this algebraically, not do a binary search. I don't even know what to search over. Does anybody have any idea why they do a binary search?

iosonofabio:
> On page 19 of the [paper](https://arxiv.org/abs/1802.03426) (algorithm 3), they do a binary search for the σ parameter, which determines how fast the fuzziness of the fuzzy ball falls off with distance beyond the nearest neighbor. To me it seems like you should calculate this algebraically, not do a binary search. I don't even know what to search over. Does anybody have any idea why they do a binary search?

Let me take a look. It sounds like a hack to me

GroteGnoom:
At page 21, the paper does curve fitting. I think the fit depends on min-dist, and I guess some random points to fit the two curves at. I'm not sure I should implement some curve fitting myself for this situation. Has anybody got a little more experience with curve fitting? Does anybody understand at what points the curves should be fit for the formula at definition 11? From algorithm 5 it seems the fit should be only done once, and a and b are called hyper-parameters at page 16, so I think it shouldn't depend on the data, but maybe I'm missing something.

GroteGnoom:
Ah, I thought I could do a matrix inverse with LAPACK's dgetri, but apparently dgetri is not in vendor/lapack, so I probably shouldn't use it.

ntamas:
Not all LAPACK functions have been vendored into igraph, only those that we actually needed. You can find the old scripts that we used to grab the LAPACK sources and translate them with `f2c` in `tools/lapack/getlapack.sh`. You need to get LAPACK 3.5.0, extract that function(s) that you need and feed them to `f2c` to get an equivalent C source. Let me know if you need help with this.

Also, _don't_ try to run `tools/lapack/getlapack.sh` directly; it comes from the days when igraph was built with autotools and not CMake, so basically nothing that the script does would be useful with the current build system. You should look at the script only to get a rough idea about how a function from LAPACK should be integrated into the vendored source.

szhorvat:
You should not use an inverse to repeatedly solve linear systems with the same coefficients. For that, look at e.g. `dgetrf` / `dgetrs` which are already exposed in igraph.

GroteGnoom:
Trying to see if I could do [Gauss-Newton](https://en.wikipedia.org/wiki/Gauss%E2%80%93Newton_algorithm) for a bit of curve fitting. It's only a small part of umap, but it seems like it shouldn't be hard once you have the basic matrix operations available of which only the inverse is missing.

szhorvat:
While I am not familiar with the method, it seems to me that it essentially needs solving a linear equation. You do not need to compute and store the inverse to do that. Instead of thinking "x = M^-1 y", think "solve M x = y".

Matrix inversion can get unstable. Do not invert matrices unless you know why you need to.

szhorvat:
https://www.johndcook.com/blog/2010/01/19/dont-invert-that-matrix/

szhorvat:
You should look at what's available in LAPACK (again, I'm not very familiar with it), but I think you simply need `igraph_lapack_dgesv()`.

iosonofabio:
Alright, the matrix to be inverted is 2x2 :smirk: I'm sure we'll make it even with an explicit inversion... let me code it

szhorvat:
If it's a 2 by 2, you might as well write down the formula:

```
In[]:= Inverse[{{a, b}, {c, d}}]
Out[]= {{d/(-b c + a d), -(b/(-b c + a d))}, {-(c/(-b c + a d)),  a/(-b c + a d)}}
```

But if you do so, why don't you just write down the formula for the solution of a linear equation directly instead of using `dgemm`?

```
In[]:= LinearSolve[{{a, b}, {c, d}}, {x, y}]
Out[]= {(d x - b y)/(-b c + a d), (c x - a y)/(b c - a d)}
```

iosonofabio:
I *think* the matrix is 2x2, so I just wrote out the formula. I think it should be fine

GroteGnoom:
So you're using only two residuals? Else it shouldn't be 2x2.

iosonofabio:
Of course not, but it's the transpose if the Jacobian right?

GroteGnoom:
Ah you're right, with m residuals and n variables, you get an m x n matrix, and its transpose times itself is n x n.

iosonofabio:
I hope at least, the comments in the code betray my insecurity :scream:

iosonofabio:
Alright, I transitioned to using LAPACK's solver like @szhorvat suggested instead of computing inverse matrix by hand. Makes no big difference, but since we have it, and the theoretical explanations in the code comments is a little more straightforward now. Also added some memory management and improved the line search.

Next thing would be either:
1. finish up the stochastic gradient thing, or
2. just use the full gradient for now and write the first unit test, so we can check if it works!

@GroteGnoom Up to you, I'll start doing them at some point anyway, no preference

GroteGnoom:
The ab starts out with an SSR of 1.124956 and later converges to 2.158813. That's not supposed to happen, right? In practice the end SSR should be the lowest you've ever seen.

iosonofabio:
hmm... not sure, here's what happens on my machine:

```
+start fit_ab
+start line search, SSR before delta: 1.124956, current SSR:, 1.790281
+during line search, k = 0, old SSR:, 1.124956, new SSR (half a,b):, 1.863988
+end of line search and iteration, squared_sum_res: 1.863988 
+
+start line search, SSR before delta: 1.790281, current SSR:, 79.240187
+during line search, k = 0, old SSR:, 1.790281, new SSR (half a,b):, 2.157009
+during line search, k = 1, old SSR:, 1.790281, new SSR (half a,b):, 7.304046
+end of line search and iteration, squared_sum_res: 7.304046 
+
+start line search, SSR before delta: 0.429805, current SSR:, 3.578511
+during line search, k = 0, old SSR:, 0.429805, new SSR (half a,b):, 1.204413
+during line search, k = 1, old SSR:, 0.429805, new SSR (half a,b):, 0.698009
+during line search, k = 2, old SSR:, 0.429805, new SSR (half a,b):, 0.540573
+during line search, k = 3, old SSR:, 0.429805, new SSR (half a,b):, 0.480010
+during line search, k = 4, old SSR:, 0.429805, new SSR (half a,b):, 0.453687
+during line search, k = 5, old SSR:, 0.429805, new SSR (half a,b):, 0.441450
+during line search, k = 6, old SSR:, 0.429805, new SSR (half a,b):, 0.435554
+during line search, k = 7, old SSR:, 0.429805, new SSR (half a,b):, 0.432661
+during line search, k = 8, old SSR:, 0.429805, new SSR (half a,b):, 0.431229
+during line search, k = 9, old SSR:, 0.429805, new SSR (half a,b):, 0.430516
+end of line search and iteration, squared_sum_res: 0.430516 
+
+no-change absolute convergence
```

It does look messy but the end is definitely lower then the beginning?

Anyway, both there and in the gradient there might be sign issues, let's double check the algebra. FWIW, Gauss-Newton is not guaranteed to decrease the objective function at each single iteration, just to converge in the long run I think.

iosonofabio:
Okay, things are moving well. Here a plot of the graph in the unit test after umap

igraph_umap

I would say: not too shabby!

The entropy is still oscillating and sometimes jumps up, I'll have to check on that. Moreover, the sigma computation is still ignoring corner cases, so I'll check that too.

iosonofabio:
I improved the logic of sigma for corner cases (e.g. unweighted/undistanced input graph) and it works well I think:

igraph_umap_2

iosonofabio:
Now I have added corner cases for `sigma` when all distances are `rho`, and cluttered the files with `#ifdef UMAP_DEBUG`. If the latter is poor practice, feel free to just change the code into whatever solution you like :smile:,,,

iosonofabio:
I renamed the internal functions as `igraph_i_...` and changed the test output, which now should pass (locally, at least).

We need at least a unit test with an unweighted graph (`distances` should be `NULL`) and then a bit of stress testing in the next days. If all looks good we can merge next week?

iosonofabio:
Assuming this works well, we can decide if we want to look into openMP for parallelizing the steps that are run independently on each vertex:

- computing `mu(a)`

I'm not sure the stochastic gradient descent can be parallelized very well. We noticed that when you move too many vertices around before checking on the cross-entropy the system tends to become a little unstable. But I imagine this is where the bulk of the runtime is spent, so perhaps?

GroteGnoom:
🙂  BTW, I think will never pass the tests on all systems if we output the exact layout, so we'll have to have to change the tests later, like only check for some general properties that should always hold.

iosonofabio:
Agreed

iosonofabio:
I changed the unit test to be more tolerant and added a few more customization options. I think as soon as we get it to pass CI tests (it passes locally) we can ask for a first review from folks.

iosonofabio:
Okay, a few tests fail but I think it's good enough for a first round of review from whoever has a bit of time.

Current CI failures relate to two things:
- macOS: caching is broken I think. The compiler warns about two `float` variables that are not even `float` anymore, and fails to compile. We should tell it to ignore whatever cache it's using, anyone knows how?
- Windows/a few other archs: one vertex can sometimes be shot into space and struggle to come back. I think this is a genuine bug with the repulsive force in the layouting, and needs to be fixed.

Thank you!

szhorvat:
There is no problem with caching on macOS. The error says:

```
/Users/runner/work/1/s/tests/unit/igraph_layout_umap.c:73:11: error: unused variable 'a' [-Werror,-Wunused-variable]
    float a, b;
```

You need to remove unused variables.

iosonofabio:
I can do that, yes

iosonofabio:
> 

Those

> There is no problem with caching on macOS. The error says:
> 
> ```
> /Users/runner/work/1/s/tests/unit/igraph_layout_umap.c:73:11: error: unused variable 'a' [-Werror,-Wunused-variable]
>     float a, b;
> ```
> 
> You need to remove unused variables.

Those variables have been removed 10 commits ago afaict

szhorvat:
No, they are there. Check carefully.

iosonofabio:
> > 
> 
> Those
> 
> > There is no problem with caching on macOS. The error says:
> > ```
> > /Users/runner/work/1/s/tests/unit/igraph_layout_umap.c:73:11: error: unused variable 'a' [-Werror,-Wunused-variable]
> >     float a, b;
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > You need to remove unused variables.
> 
> Those variables have been removed 10 commits ago afaict

Oh I'm stupid... Sorry

tests/unit/igraph_layout_umap.c
```diff
@@ -0,0 +1,121 @@
...
+        igraph_matrix_print(layout);
+    }
+}
+
+
+int main() {
+    igraph_t graph, empty_graph;
+    igraph_vector_t distances;
+    igraph_matrix_t layout;
+    float a, b;
```
szhorvat:
```suggestion
```

szhorvat:
Actually, I would keep the debug stuff. I would only remove it from the public header and the test.

I'll take a quick look at some formal things, but this is not going to be even close to a real review for now.

src/misc/umap.c
```diff
@@ -0,0 +1,846 @@
...
+   along with this program.  If not, see <https://www.gnu.org/licenses/>.
+   */
+
+#include "igraph_matrix.h"
+#include "igraph_interface.h"
+#include "igraph_umap.h"
+#include "igraph_constructors.h"
+#include "igraph_layout.h"
+#include "igraph_random.h"
+#include "igraph_lapack.h"
```
szhorvat, line 26:
```suggestion
#include "igraph_umap.h"

#include "igraph_constructors.h"
#include "igraph_interface.h"
#include "igraph_lapack.h"
#include "igraph_layout.h"
#include "igraph_matrix.h"
#include "igraph_random.h"
```

iosonofabio:
Ok I'll leave the debug stuff only in the `umap.c`.

You mean you have no time for a deep review?

src/misc/umap.c
```diff
@@ -0,0 +1,846 @@
...
+ * in PCA).
+ *
+ * However, UMAP uses a graph with distances associated to the edges as a key
+ * intermediate representation of the high-dimensional space, so it is also useful as
+ * a general graph layouting algorithm, hence its inclusion in igraph.
+ *
+ * Importantly, the edge-associated distances are derived from a similarity metric
+ * between the high-dimensional vectors, often Pearson correlation:
+ *
+ * corr(v1, v2) = v1 x v2 / [ sqrt(v1 x v1) * sqrt(v2 x v2) ]
```
szhorvat, line 47: Can you put formulas in `<code>` blocks?

src/misc/umap.c
```diff
@@ -0,0 +1,846 @@
...
+ *       subjective and potentially controversial operation: thread with care and at
+ *       your own risk. Two high-dimensional vectors might look similar or extremely
+ *       different depending on the point of view/angle, and there are a lot of
+ *       viewpoints when the dimensionality ramps up.
+ *
+ * </para><para>
+ * References:
+ *
+ * </para><para>
+ * Leland McInnes, John Healy, and James Melville, arXiv:1802.03426v3
```
szhorvat, line 84: Add permalink (dx.doi.org) to peer reviewed version. If does not exist, add link to arXiv.

src/misc/umap.c
```diff
@@ -0,0 +1,846 @@
...
+ * \param epochs Number of iterations of the main stochastic gradient descent loop on the
+ *   cross-entropy. If negative or zero, 500 epochs are used if the graph is the graph is small
+ *   (less than 50k edges), 50 epochs are used for larger graphs.
+ * \param sampling_prob The fraction of vertices moved at each iteration of the stochastic gradient descent (epoch). At fixed number of epochs, a higher fraction makes the algorithm slower. Vice versa, a too low number will converge very slowly, possibly too slowly. If negative of zero, a value of 0.3 is assumed.
+ *
+ * \return Error code.
+ *
+ */
+igraph_error_t igraph_layout_umap(const igraph_t *graph, const igraph_vector_t *distances,
+    igraph_matrix_t *layout, const igraph_real_t min_dist, const igraph_integer_t epochs, const igraph_real_t sampling_prob);
```
szhorvat, line 107: There should be a single declaration in the header. Do not repeat it here. Move docs to before the definition.

src/misc/umap.c
```diff
@@ -0,0 +1,846 @@
...
+ *       viewpoints when the dimensionality ramps up.
+ *
+ * </para><para>
+ * References:
+ *
+ * </para><para>
+ * Leland McInnes, John Healy, and James Melville, arXiv:1802.03426v3
+ *
+ * \param graph Pointer to the similarity graph to find a layout for (i.e. to embed). Weights in
+ *   this graph object are currently ignored.
```
szhorvat, line 87: What do you mean by "weights are ignored"? Are you referring to edge attributes? If so, this comment should be removed. All igraph functions that use weights take it as a separate vector parameter instead of relying on the presence of an attribute with a hardcoded name. Generally, igraph functions should be usable without having the attribute handler enabled.

src/misc/umap.c
```diff
@@ -0,0 +1,846 @@
...
+ * \param distances Pointer to a vector of "distances" between vertices. Similarity graphs for
+ *   UMAP are often originally meant in terms of similarity weights (e.g. correlation between
+ *   high-dimensional vectors) and converted into distances by crude dist := 1 - corr. That is
+ *   fine here too. If this argument is an empty pointer (NULL), connected vertices are assumed
+ *   to be very similar.
```
szhorvat, line 92: I don't quite understand this description. Is this an edge property? If so, phrase it accordingly (e.g. "edge lengths" instead of "distances between vertices") to make it clear that `distances[i]` belongs to edge `i`.

I'm not sure what "connected vertices are assumed to be very similar" means.

src/misc/umap.c
```diff
@@ -0,0 +1,846 @@
...
+ * \param layout Pointer to the n x 2 matrix where the layout coordinates will be stored. Only 2D
+ *   embedding are currently supported, as they are by far more common than any higher dimensions.
```
szhorvat, line 94: Would it be difficult to generalize to 3D or N-D? Just curious.

szhorvat: Also, a general comment: I don't think "justifications" such as "as they are by far more common than any higher dimensions" or "hence its inclusion in igraph" belong in the documentation.

src/misc/umap.c
```diff
@@ -0,0 +1,846 @@
...
+ * \param min_dist A fudge parameter that decides how close two unconnected vertices can be in the
+ *   embedding before feeling a repulsive force. Typically, 0.01 is a good number. If this is
+ *   negative or zero, 0.01 is assumed.
```
szhorvat, line 97: When the value is invalid, throw an error. Do not just fall back to a default.

I understand where this is coming from: it's supposed to be a mechanism to allow selecting a default value. But it does not fit well into high-level interfaces this way. It's much too easy to pass in a wrong value, and _unintentionally_ get the default. 

Defaults are handled differently on the user side in high-level languages, e.g. in Mathematica you pass `Automatic`, in other languages you might omit the parameter, etc. Defaults should go in places like `functions.yaml`, not in the C code.

src/misc/umap.c
```diff
@@ -0,0 +1,846 @@
...
+ * \param epochs Number of iterations of the main stochastic gradient descent loop on the
+ *   cross-entropy. If negative or zero, 500 epochs are used if the graph is the graph is small
+ *   (less than 50k edges), 50 epochs are used for larger graphs.
```
szhorvat, line 100: Throw an error for invalid values, do not fall back to default.

src/misc/umap.c
```diff
@@ -0,0 +1,846 @@
...
+ * \param sampling_prob The fraction of vertices moved at each iteration of the stochastic gradient descent (epoch). At fixed number of epochs, a higher fraction makes the algorithm slower. Vice versa, a too low number will converge very slowly, possibly too slowly. If negative of zero, a value of 0.3 is assumed.
```
szhorvat, line 101: Wrap this line. Throw an error for invalid values, do not fall back to defaults.

src/misc/umap.c
```diff
@@ -0,0 +1,846 @@
...
+    igraph_real_t sum;
+    igraph_real_t tol = 0.01;
+    igraph_integer_t maxiter = 100;
+    igraph_integer_t no_of_neis = igraph_vector_int_size(eids);
+    igraph_integer_t eid;
+    igraph_real_t step = sigma;
+    int seen_max = 0;
+
+    /* Binary search */
+    for(int iter = 0; iter < maxiter; iter++) {
```
szhorvat, line 129: All variables used to index vectors should be `igraph_integer_t`. Is there any reason this is an `int`?  Whenever I see an `int` or a `long`, I wonder if this is a bug, i.e. a failure to change a type during the integer transition process. When there is a reason to keep `int`, add a comment and explain why.

Formatting:Use `for (`, not `for(`, please.

src/misc/umap.c
```diff
@@ -0,0 +1,846 @@
...
+
+        /* Check for convergence */
+        if (fabs(sum - target) < tol) {
+            break;
+        }
+    }
+
+    *sigma_p = sigma;
+
+    return (IGRAPH_SUCCESS);
```
szhorvat, line 168: For consistent formatting:

```suggestion
    return IGRAPH_SUCCESS;
```

`return` is not a function.

src/misc/umap.c
```diff
@@ -0,0 +1,846 @@
...
+
+    *a_p = a;
+    *b_p = b;
+
+    return (IGRAPH_SUCCESS);
+
+}
+
+/* cross-entropy */
+static igraph_error_t igraph_compute_cross_entropy(const igraph_t *graph,
```
szhorvat, line 482: If this function is used only for debugging, it should also be conditional on `UMAP_DEBUG`.

szhorvat (commented):
I added a few comments, mostly based on the documentation. A proper review will take quite a bit of time, and I'd need to read the UMAP paper first. Adding more comments in front of each of the subroutines and their purposes might help a bit.

tests/unit/igraph_layout_umap.c
```diff
@@ -0,0 +1,125 @@
...
+   You should have received a copy of the GNU General Public License
+   along with this program.  If not, see <https://www.gnu.org/licenses/>.
+*/
+
+#include <igraph.h>
+#include "igraph_umap.h"
+#include "test_utilities.inc"
+
+
+int check_graph_twoclusters(const igraph_matrix_t *layout) {
```
szhorvat, line 24:
```suggestion
void check_graph_twoclusters(const igraph_matrix_t *layout) {
```

szhorvat, line 24: It seems this function returns error codes but they are never used by the called?

szhorvat:
Do you know what's happening on Windows? Is the method numerically unstable?

src/misc/umap.c
GroteGnoom: I added links, but I kept the arXiv, because that's one people like me can access 🙂

iosonofabio:
You mean to the nature paper? Actually that paper is about applications, not the algorithm itself, so I'd leave only the arXiv info with a hyperlink

src/misc/umap.c
GroteGnoom: Oh I see I pointed to the wrong UMAP paper, I'll look into it a bit more.

src/misc/umap.c
GroteGnoom: That was probably my fault, I'm usually working on a project where we have pretty weird formatting requirements, and `return (something)` is one of them.

tests/unit/igraph_layout_umap.c
```diff
@@ -113,11 +113,14 @@ int main() {
...
 
     printf("Empty graph:\n");
-	igraph_small(&empty_graph, 0, IGRAPH_UNDIRECTED, -1);
+    igraph_small(&empty_graph, 0, IGRAPH_UNDIRECTED, -1);
     IGRAPH_ASSERT(igraph_layout_umap(&empty_graph, NULL, &layout, 0.01, 500, -1) == IGRAPH_SUCCESS);
     igraph_matrix_print(&layout);
-    igraph_destroy(&empty_graph);
 
+    printf("Check error for negative min_dist.\n");
+    CHECK_ERROR(igraph_layout_umap(&empty_graph, NULL, &layout, -0.01, 500, -1), IGRAPH_EINVAL);
```
szhorvat: Actually, we don't need to explicitly check for a return code of `IGRAPH_SUCCESS` in most cases. If there is an error, it _should_ be triggered using `IGRAPH_ERROR`, which means that `igraph_error()` will be called, which by default terminated the program with `abort()`, and causes the test to fail. `igraph_abort()` is set up to print a stack trace whenever address sanitizer is enabled, so we can tell where the failure happened.

Checking the error code is most useful when the error handler is changed to something else (e.g. `igraph_error_handler_ignore()`)

Of course, this check does not hurt in any way, and does provide a little bit of security (what if a wrong error code is returned in other ways than `IGRAPH_ERROR`, e.g. a missing `return` statement).

tests/unit/igraph_layout_umap.c
```diff
@@ -76,7 +74,7 @@ int main() {
 #endif
 
     igraph_rng_seed(igraph_rng_default(), 42);
-    igraph_small(&graph, 4, IGRAPH_UNDIRECTED, // 4?
+    igraph_small(&graph, 12, IGRAPH_UNDIRECTED,
```
szhorvat, line 77: Just noting that in `igraph_small()` you may choose to provide `0` as the vertex count and let the function figure out the correct value.

GroteGnoom: Yes, I had no clue, but that's also apparently what happened before. Learned something new 😃

iosonofabio:
> The algorithm is pretty numerically unstable in general now, the second test fails if I change the sampling probability of the first one.

Correct, that's my current struggle to get it working on windows. The good thing is that that probability only factors in the last function, so there might be a bug there... I'll take a look

src/misc/umap.c
```diff
@@ -543,7 +543,12 @@ static igraph_error_t igraph_apply_forces(const igraph_t *graph,  const igraph_v
             continue;
         }
 
-        IGRAPH_CHECK(igraph_edge(graph, eid, &from, &to));
+        /* half the time, swap the from/to, otherwise some vertices are never moved */
+        if (RNG_UNIF01() > 0.5) {
+            IGRAPH_CHECK(igraph_edge(graph, eid, &from, &to));
```
szhorvat, line 548: Just to be aware, `igraph_edge()` can potentially be noticeably slower than `IGRAPH_FROM` and `IGRAPH_TO`. I would prefer to avoid it in new code, and personally I'd rather remove this function eventually.

szhorvat:
Please do not use names like `igraph_attract()`. It makes it look like this is a public function. It should be possible to tell at a glance what a function is.

 - Public functions should be named `igraph_foo()`
 - Internal functions that are not `static` should be named `igraph_i_foo()`.
 - `static` functions can be named anything you like, as they will not conflict with others, but please don't make them look like public ones. `attract()` is fine. If I see that, I can tell at a glance that it is local to this compilation unit, therefore used only for UMAP. Some have preferred to use `igraph_i_` even for static functions, although this is not technically necessary and IMO it again makes it difficult to tell if the function is specific to UMAP. So don't name it `igraph_i_attract()`. If you insist on the `igraph_i_` prefix, then use `igraph_i_umap_attract`

src/misc/umap.c
szhorvat: Ah yes, now I remember that `igraph_edge()` is exposed in high-level interfaces, so it has to stay. But we have faster alternatives for internal use.

src/misc/umap.c
```diff
@@ -0,0 +1,864 @@
...
+    IGRAPH_VECTOR_INIT_FINALLY(&umap_weights, no_of_edges);
+
+    /* Make combined graph with smoothed probabilities */
+    IGRAPH_CHECK(igraph_i_umap_find_prob_graph(graph, distances, &umap_weights));
+
+    /* From now on everything lives in probability space, it does not matter whether
+     * the original graph was weighted/distanced or unweighted */
+
+    /* Skip spectral embedding for now, initialize at random */
+    igraph_layout_random(graph, layout);
```
szhorvat, line 847: Are the spectral embedding functions that igraph has not suitable here?

BTW those functions should probably have `layout` counterparts. This is a very useful graph layout in its own right.  If we do that first, that will perhaps make it easier to use it in UMAP?

src/misc/umap.c
```diff
@@ -0,0 +1,864 @@
...
+    }
+
+    if (epochs < 0) {
+        IGRAPH_ERRORF("Number of epochs should be non-negative, but found %"
+                IGRAPH_PRId ".",
+                IGRAPH_EINVAL, epochs);
+    }
+
+    if (sampling_prob <= 0) {
+        IGRAPH_ERRORF("Sampling probability should be positive, but found %f.",
```
szhorvat, line 830: Generally, use `%g` unless there is a specific reason for `%f`.

```suggestion
        IGRAPH_ERRORF("Sampling probability should be positive, but found %g.",
```

src/misc/umap.c
```diff
@@ -0,0 +1,864 @@
...
+     * */
+    igraph_vector_int_t neis;
+
+    if (avoid_neighbor_repulsion) {
+        IGRAPH_VECTOR_INT_INIT_FINALLY(&neis, 0);
+    }
+
+    igraph_integer_t n_random_vertices = sqrt(no_of_nodes);
+
+    /* iterate over a random subsample of edges */
```
szhorvat, line 557: We have a function for efficiently selecting a random subset, `igraph_random_sample()`.

src/misc/umap.c
```diff
@@ -0,0 +1,864 @@
...
+    for (igraph_integer_t eid = 0; eid < no_of_edges; eid++) {
+        if (RNG_UNIF01() > prob) {
+            continue;
+        }
+
+        /* half the time, swap the from/to, otherwise some vertices are never moved */
```
szhorvat, line 563: Why is only one endpoint of the edge considered? I see that the paper says the same,

> Thus, whenever an attractive force is applied to an edge, one of that edge’s vertices is repulsed by a sampling of other vertices

but I'm still confused.

iosonofabio: I'm confused as well, and this might be a source of numerical instabilities. I'm investigating...

src/misc/umap.c
iosonofabio: @GroteGnoom mentioned our spectral layout is currently buggy and none of us has knowledge + time to debug it, so for now we thought rather then starting with something wrong let's use a random layout. Is that true according to your understanding @szhorvat? If our spectral embedding works, I can add an argument to the public API UMAP function that lets the user choose the initialization. Otherwise, I propose we merge this PR as is and then have a separate PR to bugfix the spectral embedding and include that argument here as well

szhorvat: I didn't know it was buggy. I have never used it. @GroteGnoom, if you are aware of specific issues that are not documented yet, can you write a bug report?

src/misc/umap.c
iosonofabio: yep, thanks

src/misc/umap.c
iosonofabio: implemented, thank you

src/misc/umap.c
iosonofabio: yep, now using the macros, thanks

src/misc/umap.c
iosonofabio: Seems like the algo is now actually stable enough to converge on all platforms. Part of it was clipping the gradient to avoid giant swings.

As for this particular observation, I think I understand what's going on. At every iteration (epoch), only some edges are adjusted to save time. Indeed, the whole algorithm does not "converge" per se, it's more like we reduce the learning rate/temperature until it becomes zero and at that point the layout is frozen in place. Whether that can be improved is a good question.... I guess so? At the same time, UMAP has rotational invariance, so perhaps if we don't freeze it it might start spinning rigidly? In any case, even if an edge is not picked for shift in this epoch, it might be picked in the next one, and the temperature is almost identical so it's alright. Same goes for the vertices that compose the edge: even if one is not picked this epoch, it might be in the next epoch and the difference is almost nothing. Of course, we could always update the vertices of an edge at the same time, but then both of them would be repelled by the rest of the graph, which is computationally as expensive as just waiting out another epoch and doing the other vertex. In addition, by moving both vertices simultaneously we add one more constraint to the minimization, which might somewhat reduce the ability of the algo to explore the solution space efficiently.

In conclusion, I think it's arbitrary but so is the choice of the edge sampling probability, and increasing the latter is almost equivalent to changing the former (i.e. moving both vertices of the edge). So it's okay for now I'd say.

If we do proper profiling in the future, we might be able to cut corners for speed and this might be one of them.

iosonofabio:
Alright guys! 99 commits into this PR we have a passing CI, tests, corner cases, clean docs, and are almost ready to go! It would be sweet to have it by commit 100 to celebrate :smirk:

The only thing left is to decide whether we can trust the spectral layout as an initialization or, if we have a known bug in that algo, skip if for now and move to fix that bug in a separate PR.

I can see the light at the end of the tunnel!

iosonofabio:
Commit 103, now we have a 2D and a 3D version... not sure if the double docs and stuff are a great idea?

szhorvat:
Don't duplicate the docs. Just refer back to the 2D version. Document only the parameter list.

GroteGnoom:
I've copied the spectral embedding problem to the issue tracker: https://github.com/igraph/igraph/issues/1971

src/layout/fruchterman_reingold.c
```diff
@@ -435,7 +435,7 @@ igraph_error_t igraph_layout_fruchterman_reingold(const igraph_t *graph,
  *
  * This is the 3D version of the force based
  * Fruchterman-Reingold layout (see \ref
- * igraph_layout_fruchterman_reingold for the 2D version
+ * igraph_layout_fruchterman_reingold for the 2D version)
```
szhorvat, line 438:
```suggestion
 * igraph_layout_fruchterman_reingold() for the 2D version)
```

iosonofabio:
Ok, given #1971 I propose we merge this as is and as part of the PR against that issue we also include spectral embedding as a umap starting condition.

Or would you like to let the user specify a certain starting layout manually (could be spectral, FR, random, circle, whatever)?

In that case we should have a flag argument about it (e.g. with an enum for `initialize={RANDOM, SPECTRAL, CUSTOM}`, in the last case do not touch the vector, just check that the dimensions are appropriate and start shifting vertices according to the cross-entropy minimization)

src/layout/umap.c
```diff
@@ -973,60 +973,10 @@ igraph_error_t igraph_layout_umap(const igraph_t *graph, const igraph_vector_t *
...
- *       different depending on the point of view/angle, and there are a lot of
- *       viewpoints when the dimensionality ramps up.
- *
- * </para><para>
- * References:
- *
- * </para><para>
- * Leland McInnes, John Healy, and James Melville. https://arxiv.org/abs/1802.03426
+ * This is the 3D version of the UMAP algorithm (see \ref
+ * igraph_layout_umap for the 2D version).
```
szhorvat, line 979:
```suggestion
 * igraph_layout_umap() for the 2D version).
```

szhorvat:
Can we wait until someone can do a full review? I'll be out for the rest of this week. Otherwise, is it 100% ready as far as you're concerned @iosonofabio ?

Can you build the docs on your machine and check that they look good? I see that many paragraph markers are missing, so everything will flow together.

src/layout/umap.c
```diff
@@ -0,0 +1,1038 @@
...
+            eid = VECTOR(*eids)[j];
+            sum += exp(-(VECTOR(*distances)[eid] - rho) / sigma);
+        }
+
+#ifdef UMAP_DEBUG
+        printf("SIGMA function (i = %" IGRAPH_PRId ", no_of_neis = %" IGRAPH_PRId ")- sum: %g, "
+               "target: %g, rho: %g, sigma: %g\n", i, no_of_neis, sum, target, rho, sigma);
+#endif
+
+        /* TODO: this seems fine, but is probably a little off for some corner cases */
```
szhorvat, line 60: If someone other than you comes back to this in the future, they will not be able to tell what this comment is about.

src/layout/umap.c
```diff
@@ -0,0 +1,1038 @@
...
+            IGRAPH_ERROR("Distances cannot be negative.", IGRAPH_EINVAL);
+        }
+    }
+
+    return IGRAPH_SUCCESS;
+}
+
+
+/* This is the main function that works for any dimensionality of the embedding (currently hard-constrained to 2 or 3 ONLY in the initialization) */
+igraph_error_t igraph_i_layout_umap(const igraph_t *graph, const igraph_vector_t *distances,
```
szhorvat, line 811: Shouldn't this be `static`?

src/layout/umap.c
```diff
@@ -0,0 +1,1038 @@
...
+
+    if (igraph_vector_size(distances) != no_of_edges) {
+        IGRAPH_ERROR("Distances must be the same number as the edges in the graph.", IGRAPH_EINVAL);
+    }
+
+    for (igraph_integer_t eid = 0; eid != no_of_edges; eid++) {
+        if (VECTOR(*distances)[eid] < 0) {
+            IGRAPH_ERROR("Distances cannot be negative.", IGRAPH_EINVAL);
+        }
+    }
```
szhorvat, line 804: We normally use `vector_min` for such checks, but pay attention that it cannot be called on empty vectors. At the same time, check for NaN.  Here's e.g. how it's done by `closeness()`:

```c
    if (no_of_edges > 0) {
        igraph_real_t minweight = igraph_vector_min(weights);
        if (minweight <= 0) {
            IGRAPH_ERROR("Weight vector must be positive.", IGRAPH_EINVAL);
        } else if (igraph_is_nan(minweight)) {
            IGRAPH_ERROR("Weight vector must not contain NaN values.", IGRAPH_EINVAL);
        }
    }
```

src/layout/umap.c
```diff
@@ -0,0 +1,1038 @@
...
+    IGRAPH_CHECK(igraph_i_umap_check_distances(distances, no_of_edges));
+
+    /* Trivial graphs (0 or 1 nodes) beget trivial - but valid - layouts */
+    if (no_of_nodes <= 1) {
+        IGRAPH_CHECK(igraph_matrix_resize(layout, no_of_nodes, ndim));
+        if (no_of_nodes == 1) {
+            for (int d = 0; d != ndim; d++) {
+                MATRIX(*layout, 0, d) = 0;
+            }
+        }
```
szhorvat, line 854: Use the `matrix_null()` function. Do not use `int` unless there is a reason.

src/layout/umap.c
```diff
@@ -0,0 +1,1038 @@
...
+    RNG_BEGIN();
+    IGRAPH_VECTOR_INIT_FINALLY(&umap_weights, no_of_edges);
+
+    /* Make combined graph with smoothed probabilities */
+    IGRAPH_CHECK(igraph_i_umap_find_prob_graph(graph, distances, &umap_weights));
+
+    /* From now on everything lives in probability space, it does not matter whether
+     * the original graph was weighted/distanced or unweighted */
+
+    /* Skip spectral embedding for now, initialize at random */
```
szhorvat, line 867: Please clarify comment here in case someone else would be working on it in the future.

src/layout/umap.c
```diff
@@ -0,0 +1,1038 @@
...
+    if (ndim == 2) {
+        igraph_layout_random(graph, layout);
+    } else {
+        igraph_layout_random_3d(graph, layout);
+    }
+
+    /* Definition 11 */
```
szhorvat, line 874: There should be a comment somewhere making it clear that these refer to (a specific version of) the arXiv preprint.

src/layout/umap.c
```diff
@@ -0,0 +1,1038 @@
...
+
+    if (avoid_neighbor_repulsion) {
+        IGRAPH_VECTOR_INT_INIT_FINALLY(&neis, 0);
+    }
+    IGRAPH_VECTOR_INT_INIT_FINALLY(&negative_vertices, 0);
+
+    igraph_integer_t n_random_vertices = sqrt(no_of_nodes);
+
+    /* iterate over a random subsample of edges */
+    for (igraph_integer_t eid = 0; eid < no_of_edges; eid++) {
```
szhorvat, line 623: We have a function to select a random subset of integers (or edge IDs), see `igraph_random_sample()`. Can you use it here? That should be more efficient for small `prob`. What are typical values of `prob`?

src/layout/umap.c
```diff
@@ -0,0 +1,1038 @@
...
+        if (RNG_UNIF01() > 0.5) {
+            from = IGRAPH_FROM(graph, eid);
+            to = IGRAPH_TO(graph, eid);
+        } else {
+            to = IGRAPH_FROM(graph, eid);
+            from = IGRAPH_TO(graph, eid);
+        }
+
+        /* Current coordinates of both vertices */
+        for (int d = 0; d != ndim; d++) {
```
szhorvat, line 638: Do not use `int`.

src/layout/umap.c
```diff
@@ -0,0 +1,1038 @@
...
+            VECTOR(from_emb)[d] = MATRIX(*layout, from, d);
+            VECTOR(to_emb)[d] = MATRIX(*layout, to, d);
+            VECTOR(delta)[d] = MATRIX(*layout, from, d) - MATRIX(*layout, to, d);
+        }
+
+        /* Apply attractive force since they are neighbors */
+        IGRAPH_CHECK(igraph_i_umap_attract(&delta, VECTOR(*umap_weights)[eid], a, b, &forces));
+        for (int d = 0; d != ndim; d++) {
```
szhorvat, line 646: Do not use `int`.

src/layout/umap.c
```diff
@@ -0,0 +1,1038 @@
...
+            to = VECTOR(negative_vertices)[j];
+            /* obviously you cannot repel yourself */
+            if (to >= from) {
+                to++;
+            }
+            /* do not repel neighbors for small graphs, for big graphs this
+             * does not matter as long as the k in knn << number of vertices */
+            if (avoid_neighbor_repulsion) {
+                /* FIXME: This algo terribly inefficient, try to improve using
+                 * adjacency lists or something but just for testing anyway */
```
szhorvat, line 663: Should this be done before merging?

src/layout/umap.c
```diff
@@ -0,0 +1,1038 @@
...
+                igraph_bool_t skip = 0;
+                igraph_incident(graph, &neis, from, IGRAPH_ALL);
+                for (igraph_integer_t k = 0; k < igraph_vector_int_size(&neis); k++) {
```
szhorvat, line 666: Please do not use `_size` functions in `for` loops. Save the size in advance and use that.

src/layout/umap.c
```diff
@@ -0,0 +1,1038 @@
...
+                        break;
+                    }
+                }
+                if (skip == 1) {
+                    continue;
+                }
+            }
+
+            /* Get layout of random neighbor and gradient in embedding */
+            for (int d = 0; d != ndim; d++) {
```
szhorvat, line 682: Do not use `int`.

src/layout/umap.c
```diff
@@ -0,0 +1,1038 @@
...
+                VECTOR(to_emb)[d] = MATRIX(*layout, to, d);
+                VECTOR(delta)[d] = MATRIX(*layout, from, d) - MATRIX(*layout, to, d);
+            }
+
+            /* This repels the other vertex assuming it's a negative example
+             * that is no weight, no edge */
+            IGRAPH_CHECK(igraph_i_umap_repel(&delta, 0, a, b, &forces));
+            /* The repulsive force is already *away* from the other (non-neighbor) vertex */
+            for (int d = 0; d != ndim; d++) {
```
szhorvat, line 691: Do not use `int`. I probably didn't catch all uses, please replace them all, except where `int` is truly needed (e.g. LAPACK error code).

szhorvat (changes requested):
I added a few more technical comments, but this is still not a proper review.

Do you consider it done now, are you planning to keep working on it? Some of the comments suggest that it's not quite done yet. I would suggest merging only when you do consider it done. Don't worry about being absent for a while, @GroteGnoom  and the rest of us will be able to take care of it :-)

I'll add a few more comments later, mostly with the aim to understand this well enough to be able to continue.

iosonofabio:
I just want to see this merged and a Python interface for it. My students will need it sooner rather than later.  It's complete but we my might tweak it a bit based on user feedback I'd say

Thank you
Fabio

src/layout/umap.c
```diff
@@ -0,0 +1,1038 @@
...
+       igraph_matrix_t *layout, igraph_integer_t epochs, igraph_real_t sampling_prob) {
+
+    igraph_real_t learning_rate = 1;
+
+#ifdef UMAP_DEBUG
+    igraph_real_t cross_entropy, cross_entropy_old;
+#endif
+
+    /* Explicit avoidance of neighbor repulsion, only useful in small graphs
+     * which are never very sparse */
```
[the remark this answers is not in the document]
iosonofabio: negative sampling as implemented relies on an approximation that only works if the graph is sparse, which is never quite true for small graphs. Welcome to the land of people being half honest in their publicity stunts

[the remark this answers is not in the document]
iosonofabio: Sparse as in V << E << V^2, which is never a very good assumtion if say V ~ 5 since the brackets are so close to one another

szhorvat:
I have pieces of code which I need for my work as well, and I use them locally with an unreleased version of IGraph/M. I keep a separate branch of the C core for this. Can your students use such a solution? I am not comfortable releasing unbaked functionality.

There are multiple comments in the code which make the impression of work-in-progress notes. We should bring this to a quality so that someone else than you would be able to continue working on it in the future, if necessary. This means understanding what is or isn't complete, what needs improvements, and why. Some of the current comments make it seem like there are some problems, but don't always make it clear what those are or how to test for them.

This brings me to my main question:

Did you follow the algorithm in the arXiv paper to the letter, or did you make changes to it? Is the algorithm there described completely, or do you need to fill in some parts? Can you make it clear which parts of the code follow the paper, which are filling in missing details, which are changed, and which are your own invention?

src/layout/umap.c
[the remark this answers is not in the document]
szhorvat: I don't understand what is being done here, so my main question to you: is this correct as it is, or do we need to change something? I'm not saying we need changes, I'm just asking.

szhorvat:
One last comment: if this does get merged, let us mark it as `\experimental` in the docs, meaning no commitment to current behaviour or even the API, when it gets released with 0.10.

iosonofabio:
This is 99% as in the paper and is complete. That's why we should merge.

Of course, a paper is not divine truth and algorithms can always be improved. We have a little bit of that here (e.g. explicitly avoid repelling neighbors for small graphs, much better conceptual framing without the whole topology smokescreen) but nothing essential. A few further, small improvements might emerge quite naturally from users' requests, in which case we should of course consider adding them.

I'm fixing the few things you noticed, will push in a minute. @ntamas it would be great if we could take a review soon, do you think that's on your radar?

src/layout/umap.c
```diff
@@ -0,0 +1,1038 @@
...
+         * function because the vertices are not moved in the high-dimensional
+         * space, so there is no need for differentiating that function.
+         *
+         * The residual is of course:
+         *
+         *    Q(d) - P(d) = ( 1 + a*d^2b )^-1 - [ d < min_dist ? 1 : e^{-(d - min_dist)} ]
+         *
+         * This function also sets the auxiliary vector powb.
+         * */
+        VECTOR(*powb)[i] = powf(VECTOR(*x)[i], 2 * b);
```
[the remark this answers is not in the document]
iosonofabio: yep

src/layout/umap.c
iosonofabio: yep, this is done

src/layout/umap.c
iosonofabio: yep

src/layout/umap.c
iosonofabio: that function allocates a vector of integers, which is not needed in this version. Since prob can be high if the user wants, I do not want to allocate a huge vector for no reason, hence the current version

src/layout/umap.c
iosonofabio: no need to go through the whole vector if the first or second value is negative (a likely scenario, if the user just got the wrong vector in). I added the NaN check

src/layout/umap.c
iosonofabio: no need, it's fast enough for small graphs and large graphs should not be using this part anyway, the approximation used in the paper is just fine

iosonofabio:
Okay, I made all requested changes and improved a few comments that made it look "artisanal". There's not point specifying to the letter what followed *exactly* the preprint and what took a *slightly* different approach: if someone wants to understand this is commented way better than any other implementation out there already, including the original authors'.

I'm in favour of tagging as `\experimental` in 0.10 if that makes you feel good. We can always break API in 0.11 anyway, so it's not a big deal I think. Since I'm not sure where you want the "tag" to appear, please push to the branch directly or perhaps @GroteGnoom can do it for you.

ntamas:
> [@ntamas](https://github.com/ntamas) it would be great if we could take a review soon, do you think that's on your radar?

Yes it is, give me a day or two, tops. (I might even do it today after lunch).

iosonofabio:
:pray: :pray:

szhorvat:
> Of course, a paper is not divine truth and algorithms can always be improved

Sure, the question was just to help with the review, not a comment on correctness.

src/layout/umap.c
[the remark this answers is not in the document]
szhorvat: Do add a clarifying comment in the sources though, otherwise this question will keep coming up.

src/layout/umap.c
szhorvat: Again, it's best to add a comment about this.

src/layout/umap.c
```diff
@@ -0,0 +1,1038 @@
...
+
+    igraph_integer_t no_of_nodes = igraph_vcount(graph);
+    igraph_integer_t no_of_edges = igraph_ecount(graph);
+    igraph_integer_t no_of_neis, eid;
+    igraph_vector_int_t eids, weight_seen;
+    igraph_real_t rho, dist_max, dist, sigma, weight, weight_inv, sigma_target;
+
+    /* if the original graph is unweighted, probabilities are 1 throughout */
+    if (distances == NULL) {
+        no_of_neis = igraph_vector_size(umap_weights);
```
szhorvat, line 102: I believe this line is not needed (variable is never used).

iosonofabio:
Hey guys, could we merge this? It's been such a tour de force... Happy to have it tagged "experimental" if you want, but at least I could build a Python interface and start using it to discover further bugs

szhorvat:
It makes sense to merge and expose to Python because then it's easier to test. Can you add `\experimental` to the docs first and remove that bit of unused code (if it's truly unused) that I mentioned above?

src/layout/umap.c
iosonofabio: good catch, I deleted it now

iosonofabio:
Okay, I added an "experimental" line like in another layout function, deleted the unused variable, and merged `develop` into this branch. I hope this passes CI because I really have almost zero time to fix things these days. It would be just amazing if one of you @GroteGnoom @ntamas took a look at the CI issues and merged the PR...

iosonofabio:
Phew tests passed now, let's merge! Yaay!

src/layout/umap.c
```diff
@@ -1027,6 +1026,8 @@ igraph_error_t igraph_layout_umap(const igraph_t *graph,
  *
  * \return Error code.
  *
+ * Added in version 0.10. This function is experimental and the API is not stable.
```
szhorvat, line 1029: I mean literally add `\experimental`, which will expand to a big yellow warning box :-) I can also do this, unless someone else gets to it first.

ntamas:
Give me at most a day to review this and then I'll merge.

ntamas (approved):

ntamas:
@GroteGnoom squash-and-merge or merge commit, which one would you prefer? Do you want to keep all the commits in the commit history of `develop` or is it okay to just have a single squashed commit?

src/layout/umap.c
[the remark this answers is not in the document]
ntamas: I've added a clarifying comment there.

src/layout/umap.c
ntamas: I have added a comment with Fabio's explanation. I've also added a TODO note mentioning that we should at least investigate using a geometric distribution to generate the next edge ID from the current if `prob` is small, similarly to how we do it in the G(n, p) Erdős-Rényi random graph model.

GroteGnoom:
> @GroteGnoom squash-and-merge or merge commit, which one would you prefer? Do you want to keep all the commits in the commit history of `develop` or is it okay to just have a single squashed commit?
I think squashed is better, there's a lot of commits in there that don't make any sense by themselves.

src/layout/umap.c
```diff
@@ -0,0 +1,1047 @@
...
+
+    }
+
+    /* UMAP is sometimes used on unweighted graphs, that means distances are always zero */
+    IGRAPH_CHECK(igraph_i_umap_check_distances(distances, no_of_edges));
+
+    /* Trivial graphs (0 or 1 nodes) beget trivial - but valid - layouts */
+    if (no_of_nodes <= 1) {
+        IGRAPH_CHECK(igraph_matrix_resize(layout, no_of_nodes, ndim));
+        if (no_of_nodes == 1) {
```
szhorvat, line 859: @GroteGnoom @iosonofabio Why did you use this conditional here? Does `matrix_null` not work for zero-size matrices? If so, that would be a bug that needs to be fixed.

szhorvat: As far as I can tell, there's no problem with calling `matrix_null` on zero-size matrices. I'll remove the conditional so we wouldn't keep wondering about it next time we see it. Let me know if I missed something.

PR #1868, opened 2021-12-03
Add stimulus to CI

szhorvat:


.github/workflows/stimulus.yml
```diff
@@ -0,0 +1,24 @@
...
+      #- name: Setup Python
+      #  uses: actions/setup-python@v2
+      - name: Configure igraph
+        run: |
+          mkdir build && cd build
+          cmake ..
+      - name: Run verification
+        run: |
+          cd 
+          python3 -m .venv venv
```
ntamas, line 21: This should be `python3 -m venv .venv` (the `.venv` is the name of the venv, the `venv` itself is the name of the Python module that creates virtualenvs)

.github/workflows/stimulus.yml
```diff
@@ -0,0 +1,24 @@
...
+          .venv/bin/pip install 'https://github.com/igraph/stimulus.git@0.8.0#egg=stimulus'
+          .venv/bin/stimulus -f functions.yaml -t types.yaml -l ci:validate -o test.cpp
+          c++ -std=c++14 -o test test.cpp -I ../include -I ../build/include -ferror-limit=1
```
ntamas, line 24: For CI purposes we can probably raise the error limit; having only one error is not helpful if someone breaks multiple definitions because you will falsely anticipate that fixing this single error will fix the build.

szhorvat:
Sorry about the mess, it's done now @ntamas

ntamas:
Thanks, merged!

PR #1881, opened 2021-12-08
Check GLPK version in CMake

### LINKED ISSUE #1879: cmake does not check the glpk version
**Describe the bug**
I had an older version of glpk. configuration for igraph detected it and tried to use it, with lots of failures.

The test for glpk should actually, once detected, include glpk.h and verify the major/minor versions are okay, since unfortunately there is no pkgconfig file.

**To reproduce**
Install glpk < 5.0.

**Version information**
release tarball for 0.9.5

szhorvat:
- Rewrite `FindGLPK.cmake` to allow for version check
 - Simplify `FindGMP.cmake`. While it looks in fewer places, it is consistent with `FindGLPK.cmake`. We can maybe do this for other libraries too. `GMP_PREFIX` is not necessary because there is the standard `GMP_ROOT` (provided automatically by CMake for any package)
 - Provide support for checking package versions in `dependencies.cmake` and check that GLPK >= 4.57

Fixes #1879

### DIFF
--- a/doc/installation.xml
+++ b/doc/installation.xml
@@ -53,7 +53,7 @@
     </listitem>
     <listitem>
       <para>
-        <ulink url="https://www.gnu.org/software/glpk/">GLPK</ulink>
+        <ulink url="https://www.gnu.org/software/glpk/">GLPK</ulink> (version 4.57 or later)
       </para>
     </listitem>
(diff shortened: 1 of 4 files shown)


ntamas:
Thanks!

Issue #1873, opened 2021-12-06
Check if a graph is a forest


szhorvat:
**What is the feature or improvement you would like to see?**

A function to check if a graph is a forest. It can mimic the interface of `igraph_is_tree()`.

Unlike `igraph_is_tree(), it should return true for the null graph.

_This is an easy task, but it's slightly more involved than other issues labelled as "good first issue". Look at `igraph_is_tree()` for inspiration._

**Use cases for the feature**

This is a generally useful basic check for a common graph type. This function will have better performance than decomposing a graph into components and calling `igraph_is_tree()` on each.

rohitt28:
Hey I would like to work on this. I have compiled the development version of igraph and read about some basics from the tutorial.

Basic Implementation: -
I think a static function graph_is_forest_visitor() will be required which is similar to the one made for tree but this takes the visited vector. Only checking the ecount < vcount-1 might not work as there may be some components with only 1 vertex.
So, the Forest can have both directed and undirected graphs. 

- In case of undirected graphs, I will make a root vector and add each vertex to it and call the function for each unvisited root, check if there is a loop at any vertex.
- For directed ones, will there be only 1 type of trees in the forest (in tree or out tree)?
   1. If yes, then I can add each vertex with degree 0 to the root vector and follow the same approach as in undirected graphs. I will change the result to 0 if the current node is already visited or if total visited vertices not equal to vcount.
   2. If no, then we can’t get roots by finding vertices with degree 0. I am not sure about how to proceed in this case.

Is this approach fine?

szhorvat:
Welcome! I assigned you to this issue, and we are looking forward to the implementation.

Generally, it works best if you open a draft PR early, so that we can see what you are doing and give tips as you go.

Your approach sounds good. 

In the undirected case, we only need to check that there are no cycles. Just in case I misunderstood your proposal: we want to end up with a `roots` vector that contains precisely one vertex from each connected component. So we can take an arbitrary vertex, add it to `roots`, then when we are done with the traversal, take the next unvisited vertex, add that to `roots`, and so one.

> * For directed ones, will there be only 1 type of trees in the forest (in tree or out tree)?

Only one type, as specified in the function's `mode` argument.



> Is this approach fine?

Sounds great!

Along the way, we can look for small opportunities to immediately discover that the graph is not a forest and return. For example, if `ecount > vcount - 1`, then it is certainly not a forest, and there is no need to look further.

szhorvat:
[...]

Issue #168, opened 2013-05-04
Assortativity coefficient, as defined by Newman


gaborcsardi:
See http://arxiv.org/pdf/cond-mat/0209450v2
It should be easy.

---

Imported from Launchpad using lp2gh.
- date created: 2009-07-05T09:13:16Z
- owner: gabor.csardi
- assignee: gabor.csardi
- the launchpad url was https://bugs.launchpad.net/bugs/395711

PR #1888, opened 2021-12-13
Check if a graph is a forest

rohitt28:


rohitt28:
Initial draft to implement the igraph_is_graph function mentioned [here](https://github.com/igraph/igraph/issues/1873)

GroteGnoom:
Don't forget to add your function to `doc/structural.xxml`, so it's documentation will be included.

src/properties/forest.c
```diff
@@ -0,0 +1,98 @@
...
+        
+        //for undirected graph
+
+        break;
+
+    case IGRAPH_IN:
+    case IGRAPH_OUT: {
+        igraph_vector_t degree;
+
+        IGRAPH_CHECK(igraph_vector_init(&degree, 0));
```
GroteGnoom, line 82: Right now you don't destroy this one.

rohitt28: Thanks for pointing it out. I will change that in next commit.

GroteGnoom:
I think you've got a lot of the structure already correctly figured out 🙂
You've not added documentation yet, you can already include that as well.
I think you can continue. I'm interested to see your algorithm 🙂

tests/unit/igraph_is_forest.c
```diff
@@ -0,0 +1,8 @@
+#include <igraph.h>
```
GroteGnoom, line 1: You're missing the copyright here.

rohitt28:
Hey @GroteGnoom, I tried to run unit tests locally and realized that ```igraph_is_forest``` function isn't recognised in the files. I thought adding it to ```igraph_structural.h``` would be enough. Can you help me fix this?

rohitt28:
The error when I run the tests ```168/401 Test #168: test::igraph_is_forest.............................***Failed    0.70 sec
Test exited abnormally with error: Access violation```

And when I hover over the function it shows this instead of the parameters

src/properties/forest.c
```diff
@@ -0,0 +1,214 @@
...
+#include "igraph_adjlist.h"
+#include "igraph_constructors.h"
+#include "igraph_interface.h"
+#include "igraph_stack.h"
+
+/* count the number of vertices reachable from the root */
+static int igraph_i_is_forest_visitor(igraph_integer_t root,igraph_vector_t *visited, const igraph_adjlist_t *al, igraph_integer_t *visited_count,igraph_integer_t *res,  igraph_neimode_t mode){
+    
+    igraph_stack_int_t stack;
+    long i;
```
GroteGnoom, line 27: use igraph_integer_t instead of longs.

rohitt28:
This is the current version, it works when I add the function on top of it.

GroteGnoom:
For me it builds fine but segfaults. I'll check things out and add some comments.

tests/unit/igraph_is_forest.c
```diff
@@ -0,0 +1,95 @@
...
+
+#include <igraph.h>
+
+#include "test_utilities.inc"
+
+void check_output(const igraph_t *graph,igraph_bool_t *res,igraph_neimode_t mode){
+    igraph_bool_t result;
+    igraph_vector_t roots;
+    igraph_vector_init(&roots, 0);
+    IGRAPH_FINALLY(igraph_vector_destroy, &roots);
```
GroteGnoom, line 24: You don't have to put things on the finally stack in tests.

tests/unit/igraph_is_forest.c
```diff
@@ -0,0 +1,95 @@
...
+    igraph_vector_print(&roots);
+    igraph_vector_destroy(&roots);
+}
+
+int main() {
+    igraph_t graph;
+    igraph_bool_t res;
+    igraph_neimode_t mode;
+
+    /*Empty Graph*/
```
GroteGnoom, line 37: If you print something before each test you can see which output belongs to which test. Else the test results are hard to check.

src/properties/forest.c
```diff
@@ -0,0 +1,214 @@
...
+        *res= visited_count==vcount;
+    }
+    /*If the graph is not a forest the the root vector will be empty*/
+    if(!*res){
+        IGRAPH_CHECK(igraph_vector_resize(roots,0));
+    }
+
+    igraph_vector_destroy(&visited);
+    igraph_adjlist_destroy(&al);
+    IGRAPH_FINALLY_CLEAN(3);
```
GroteGnoom, line 211: You do a CLEAN of 3, but you only destroy 2 things.

src/properties/forest.c
```diff
@@ -0,0 +1,214 @@
...
+        IGRAPH_CHECK(igraph_degree(graph, &degree, igraph_vss_all(), mode == IGRAPH_IN ? IGRAPH_OUT : IGRAPH_IN, /* loops = */ 1));
+        
+        for (i = 0; i < vcount; ++i){
+            if (VECTOR(degree)[i] == 0) {
+                igraph_vector_push_back(roots,i);
+                IGRAPH_CHECK(igraph_i_is_forest_visitor(i,&visited, &al, &visited_count,res,mode));
+            }
+        }
+
+        igraph_vector_destroy(&degree);
```
GroteGnoom, line 195: You don't FINALLY_CLEAN this one.

GroteGnoom:
Your undirected graph with cycle seems to segfault.

GroteGnoom:
As for your build problems: Did you do a `cmake ..` and then a `cmake --build . --target check` ?

rohitt28:
> Your undirected graph with cycle seems to segfault.

Oh! I accidentally added 6 vertices instead of 5. 


> Did you do a `cmake ..` and then a `cmake --build . --target check`

Yes, I will try to clone the repo again in another folder, maybe that might work.

rohitt28:
I have made the required changes, could you check if I missed something. Also currently this function is in `forest.c` (new file) but it could be moved to the `trees.c` file. Let me know if moving it there would be better.

src/properties/forest.c
```diff
@@ -0,0 +1,215 @@
+/*
+   IGraph library.
+   Copyright (C) 2021  The igraph development team <igraph@igraph.org>
```
GroteGnoom, line 3: I think the copyright headers should have like 3 extra newlines in them, which should be the case in most other files.

src/properties/forest.c
```diff
@@ -0,0 +1,215 @@
...
+
+#include "igraph_structural.h"
+
+#include "igraph_adjlist.h"
+#include "igraph_constructors.h"
+#include "igraph_interface.h"
+#include "igraph_stack.h"
+
+/* count the number of vertices reachable from the root */
+static int igraph_i_is_forest_visitor(igraph_integer_t root,igraph_vector_t *visited, const igraph_adjlist_t *al, igraph_integer_t *visited_count,igraph_integer_t *res,  igraph_neimode_t mode){
```
GroteGnoom, line 24: We don't do ints anymore, so this int should be an igraph_integer_t

src/properties/forest.c
```diff
@@ -0,0 +1,215 @@
...
+
+    igraph_stack_int_t stack;
+    igraph_integer_t i;
```
GroteGnoom, line 27: Can you put a newline between your declarations and your code? That makes it a little easier to read.

src/properties/forest.c
```diff
@@ -0,0 +1,215 @@
...
+/**
+ * \ingroup structural
+ * \function igraph_is_forest
+ * \brief Decides whether the graph is a forest.
+ *
+ * An undirected graph is a forest if it has no cycles.
+ * </para><para>
+ *
+ * In the directed case, a possible additional requirement is that two
+ * different trees should not have common vertex. Also, edges in each
```
GroteGnoom, line 87: Isn't there something missing from this definition? It seems to allow for a vertex with two parents. Like 0->1, 0->2, 1->3, 2->3 should not be in forest, right? I see that it's also in `is_tree`, but I'm still confused.

src/properties/forest.c
```diff
@@ -0,0 +1,215 @@
...
+ *        \c IGRAPH_OUT, \c IGRAPH_IN, \c IGRAPH_ALL. This argument is
+ *        ignored for undirected graphs.
+ * \return Error code:
+ *        \c IGRAPH_EINVAL: invalid mode argument.
+ *
+ * Time complexity: At most O(|V|+|E|), the
+ * number of vertices plus the number of edges in the graph.
+ *
+ */
+int igraph_is_forest(const igraph_t *graph,igraph_bool_t *res, igraph_vector_t *roots , igraph_neimode_t mode) {
```
GroteGnoom, line 115: The return type should be igraph_error_t

src/properties/forest.c
rohitt28: Oh thats right. That condition is checked in the code, I will add it in the description too.

src/properties/forest.c
```diff
@@ -0,0 +1,215 @@
...
+ *        component can be the root. When \p mode is \c IGRAPH_OUT
+ *        or \c IGRAPH_IN, all the vertices with zero in- or out-degree,
+ *        respectively are considered as root nodes.
+ * \param mode For a directed graph this specifies whether to test for an
+ *        out-tree, an in-tree or ignore edge directions. The respective
+ *        possible values are:
+ *        \c IGRAPH_OUT, \c IGRAPH_IN, \c IGRAPH_ALL. This argument is
+ *        ignored for undirected graphs.
+ * \return Error code:
+ *        \c IGRAPH_EINVAL: invalid mode argument.
```
GroteGnoom, line 109: This is actually an IGRAPH_EINVMODE

src/properties/forest.c
```diff
@@ -0,0 +1,215 @@
...
+
+        igraph_vector_destroy(&degree);
+        IGRAPH_FINALLY_CLEAN(1);
+        break;
+    }
+    default:
+        IGRAPH_ERROR("Invalid mode", IGRAPH_EINVMODE);
+    }
+    if(*res){
+        *res= visited_count==vcount;
```
GroteGnoom, line 203: Can you give me an example of when then visited_count is not the vcount? I think that shouldn't happen for this algorithm.

rohitt28: If there is a single vertex in directed graph with a self loop then its degree wont be 0 so `igraph_forest_visitor` wont be called for that vertex and function would return that it is a forest.

GroteGnoom: Aha! Thanks!

GroteGnoom:
There's also another issue that needs to be resolved: the conflict with the include/igraph_structural.h
Did you maybe base your branch on master instead of develop?

rohitt28:
>  Did you maybe base your branch on master instead of develop?

Yes, thats why the functions still had int return type. I think that merge conflict can be resolved by just removing the int functions. Will this work?

GroteGnoom:
I think so, it seems like it's going to work out.

src/properties/forest.c
```diff
@@ -0,0 +1,220 @@
...
+    IGRAPH_CHECK(igraph_stack_int_init(&stack, 0));
+    IGRAPH_FINALLY(igraph_stack_int_destroy, &stack);
+    /* push the root into the stack */
+    IGRAPH_CHECK(igraph_stack_int_push(&stack, root));
+
+    while (! igraph_stack_int_empty(&stack)) {
+        igraph_integer_t u;
+        igraph_vector_int_t *neighbors;
+        igraph_integer_t ncount;
+        u = igraph_stack_int_pop(&stack);
```
GroteGnoom, line 41: also here some space between declarations and code please :)

GroteGnoom:
All right, that merge cleaned a lot of things up, thanks :) I think I will stop reviewing soon, and maybe continue tomorrow. We should try to pass the tests, but right now your `res` is of different types in the two functions, so it will not compile.

src/properties/forest.c
```diff
@@ -0,0 +1,221 @@
...
+    }
+    /* Ignore mode for undirected graphs. */
+    if (! igraph_is_directed(graph)) {
+        mode = IGRAPH_ALL;
+    }
+    IGRAPH_CHECK(igraph_adjlist_init(graph, &al, mode, IGRAPH_LOOPS, IGRAPH_MULTIPLE));
+    IGRAPH_FINALLY(igraph_adjlist_destroy, &al);
+
+    *res = 1; /* assume success */
+    IGRAPH_CHECK(igraph_vector_init(&visited, 0));
```
GroteGnoom, line 153: You can init this with vcount immediately, so you don't have to resize it later. All the elements are set to 0 on init, so you also can remove the loop.

src/properties/forest.c
```diff
@@ -0,0 +1,221 @@
...
+ * Time complexity: At most O(|V|+|E|), the
+ * number of vertices plus the number of edges in the graph.
+ *
+ */
+igraph_error_t igraph_is_forest(const igraph_t *graph,igraph_bool_t *res, igraph_vector_t *roots , igraph_neimode_t mode) {
+    igraph_adjlist_t al;
+    igraph_vector_t visited;
+    igraph_integer_t visited_count=0;
+    igraph_integer_t vcount, ecount;
+
```
GroteGnoom, line 125: You assume the roots vector starts with  a size of 0 later in the algorithm, but that doesn't always hold, so you need to resize it yourself.

src/properties/forest.c
```diff
@@ -0,0 +1,216 @@
...
+    if (! igraph_is_directed(graph)) {
+        mode = IGRAPH_ALL;
+    }
+    IGRAPH_CHECK(igraph_adjlist_init(graph, &al, mode, IGRAPH_LOOPS, IGRAPH_MULTIPLE));
+    IGRAPH_FINALLY(igraph_adjlist_destroy, &al);
+
+    *res = 1; /* assume success */
+    IGRAPH_CHECK(igraph_vector_init(&visited,vcount));
+    IGRAPH_FINALLY(igraph_vector_destroy, &visited);
+    IGRAPH_CHECK(igraph_vector_resize(roots,0));
```
GroteGnoom, line 155: This resize should be moved up higher, because at line 142 you already do         igraph_vector_push_back(roots,0); for `vcount == 1`

rohitt28: Done

GroteGnoom:
I think looks pretty great, and you done very good work. It was a pleasure working with you on this, thanks 🙂 I will leave it up to someone else do the final checks. @ntamas, can you take a look?

rohitt28:
Thanks a lot for your helpful reviews. I will keep these suggestions in mind while contributing again.

include/igraph_structural.h
```diff
@@ -54,10 +54,13 @@ IGRAPH_EXPORT igraph_error_t igraph_is_loop(const igraph_t *graph, igraph_vector
                                  igraph_es_t es);
 IGRAPH_EXPORT igraph_error_t igraph_is_multiple(const igraph_t *graph, igraph_vector_bool_t *res,
                                      igraph_es_t es);
+
 IGRAPH_EXPORT igraph_error_t igraph_is_mutual(const igraph_t *graph, igraph_vector_bool_t *res, igraph_es_t es);
 IGRAPH_EXPORT igraph_error_t igraph_is_simple(const igraph_t *graph, igraph_bool_t *res);
 IGRAPH_EXPORT igraph_error_t igraph_is_tree(const igraph_t *graph, igraph_bool_t *res, igraph_integer_t *root, igraph_neimode_t mode);
+IGRAPH_EXPORT igraph_error_t igraph_is_forest(const igraph_t *graph, igraph_bool_t *res, igraph_vector_t *roots, igraph_neimode_t mode);
 IGRAPH_EXPORT igraph_error_t igraph_maxdegree(const igraph_t *graph, igraph_integer_t *res,
+
```
ntamas, line 63: This empty line is unnecessary here.

include/igraph_structural.h
```diff
@@ -54,10 +54,13 @@ IGRAPH_EXPORT igraph_error_t igraph_is_loop(const igraph_t *graph, igraph_vector
                                  igraph_es_t es);
 IGRAPH_EXPORT igraph_error_t igraph_is_multiple(const igraph_t *graph, igraph_vector_bool_t *res,
                                      igraph_es_t es);
+
```
ntamas, line 57: This empty line is unnecessary here.

doc/structural.xxml
```diff
@@ -150,6 +150,10 @@
 <!-- doxrox-include igraph_to_prufer -->
 </section>
 
+<section id="forest"><title>Forest</title>
```
ntamas, line 153: I'd rather put the `igraph_is_forest()` function in the Trees section for the time being. I don't know how many forest-related functions we are going to have, but forests and trees are closely related anyway so we can keep them in the same section. Just rename the section to "Trees and forests"

src/properties/forest.c
```diff
@@ -0,0 +1,216 @@
...
+
+#include "igraph_structural.h"
+
+#include "igraph_adjlist.h"
+#include "igraph_constructors.h"
+#include "igraph_interface.h"
+#include "igraph_stack.h"
+
+/* count the number of vertices reachable from the root */
+static  igraph_error_t igraph_i_is_forest_visitor(igraph_integer_t root,igraph_vector_t *visited, const igraph_adjlist_t *al, igraph_integer_t *visited_count, igraph_bool_t *res, igraph_neimode_t mode){
```
ntamas, line 27: Please wrap long lines at around 80 chars or so to make it easier to read on narrow screens or for people who use two code panels side-by-side.

src/properties/forest.c
```diff
@@ -0,0 +1,216 @@
...
+ *        \c IGRAPH_OUT, \c IGRAPH_IN, \c IGRAPH_ALL. This argument is
+ *        ignored for undirected graphs.
+ * \return Error code:
+ *        \c IGRAPH_EINVMODE: invalid mode argument.
+ *
+ * Time complexity: At most O(|V|+|E|), the
+ * number of vertices plus the number of edges in the graph.
+ *
+ */
+igraph_error_t igraph_is_forest(const igraph_t *graph,igraph_bool_t *res, igraph_vector_t *roots , igraph_neimode_t mode) {
```
ntamas, line 120: This is my only real concern with the current state of this PR: the `roots` argument should be of type `igraph_vector_int_t`, _not_ `igraph_vector_t`. `igraph_vector_t` contains doubles, not integers.

Also, nitpicking: please include a space after every comma, and no space before commas.

src/properties/forest.c
```diff
@@ -0,0 +1,216 @@
...
+    }
+    // A forest can have maximum vcount-1 edges.
+    if (ecount > vcount - 1) {
+        *res = 0;
+        return IGRAPH_SUCCESS;
+    }
+    /*A single-vertex graph is a forest, provided it has no edges (checked in the previous if (..)) */
+    if (vcount == 1) {
+        *res = 1;
+        igraph_vector_push_back(roots,0);
```
ntamas, line 143: Please add an `IGRAPH_CHECK()` here. It is very unlikely that the user won't have free memory for pushing one more item to the roots vector, but it may still happen and it's best if you get into the habit of wrapping all igraph function calls in `IGRAPH_CHECK()`.

src/properties/forest.c
```diff
@@ -0,0 +1,216 @@
...
+     *
+     * If all the vertices are visited exactly once, then the graph is a forest.
+     */
+
+    switch (mode) {
+    case IGRAPH_ALL:{
+
+        for(i =0; i<vcount;++i){
+            if (!VECTOR(visited)[i]) {
+                igraph_vector_push_back(roots,i);
```
ntamas, line 175: `IGRAPH_CHECK()` is missing here.

src/properties/forest.c
```diff
@@ -0,0 +1,216 @@
...
+        }
+        break;
+    }
+    case IGRAPH_IN:
+    case IGRAPH_OUT: {
+        igraph_vector_int_t degree;
+
+        IGRAPH_CHECK(igraph_vector_int_init(&degree, 0));
+        IGRAPH_FINALLY(igraph_vector_int_destroy, &degree);
+        IGRAPH_CHECK(igraph_degree(graph, &degree, igraph_vss_all(), mode == IGRAPH_IN ? IGRAPH_OUT : IGRAPH_IN, /* loops = */ 1));
```
ntamas, line 187: @szhorvat has recently introduced a macro named `IGRAPH_REVERSE_MODE(...)`, which reverses the `mode` argument like you do it here. Let's start using it. This is not in the docs yet so that's why I'm telling now :)

src/properties/forest.c
```diff
@@ -0,0 +1,216 @@
...
+     * Directed Graph:- For each tree, the root is the node with no
+     * incoming/outgoing connections, depending on 'mode'. We add each vertex
+     * with zero degree to the roots vector and mark all other vertices that are
+     * reachable from it as visited.
+     *
+     * If all the vertices are visited exactly once, then the graph is a forest.
+     */
+
+    switch (mode) {
+    case IGRAPH_ALL:{
```
ntamas, line 171: In the rest of the codebase, `case` branches are indented by one indentation unit (4 spaces) relative to the `switch` statement. Please update the coding style accordingly.

src/properties/forest.c
```diff
@@ -0,0 +1,216 @@
...
+    if (vcount == 1) {
+        *res = 1;
+        igraph_vector_push_back(roots,0);
+        return IGRAPH_SUCCESS;
+    }
+    /* Ignore mode for undirected graphs. */
+    if (! igraph_is_directed(graph)) {
+        mode = IGRAPH_ALL;
+    }
+    IGRAPH_CHECK(igraph_adjlist_init(graph, &al, mode, IGRAPH_LOOPS, IGRAPH_MULTIPLE));
```
ntamas, line 150: I would not use an adjacency list for this algorithm, I'd simply use `igraph_neighbors()` whenever I need the neighbors of a vertex. The reason is that basically `igraph_adjlist_init()` is pre-querying `igraph_neighbors()` for all the vertices, and this is wasteful if you are likely to visit every vertex at most once. We typically only use `igraph_adjlist_init()` where we know that `igraph_neighbors()` queries will be repeated over and over multiple times for many vertices, otherwise there is no performance gain in using `igraph_adjlist_init()` but there is an additional cost you pay in storage.

ntamas (changes requested):
@rohitt28 Thank you for your work on this PR, great job! I've added a few comments; the two major concerns that I raised are:

* the `roots` argument should be an `igraph_vector_int_t`, not an `igraph_vector_t` because the latter stores doubles
* there is no need to use an adjacency list because you will visit each vertex at most once. `igraph_adjlist_init()` is doing an `igraph_neighbors()` query for each vertex in the graph in advance, which pays off performance-wise only if you would visit the same vertex multiple times. This is not the case here.

@GroteGnoom Thanks for the initial round of reviews! I'll take over from here until this PR is merged.

ntamas:
Also, please take a look at the code coverage report [here](https://app.codecov.io/gh/igraph/igraph/compare/1888/diff) and try to exercise all newly added code paths in the unit tests. It seems like the case of the single-vertex graph is not covered. There is a branch in the forest visitor that the unit tests do not cover either.

szhorvat:
I'd like to have the chance to review this before it is merged. Please give me a couple of days.

doc/structural.xxml
```diff
@@ -141,13 +141,14 @@
 <!-- doxrox-include igraph_similarity_inverse_log_weighted -->
 </section>
 
-<section id="trees"><title>Trees</title>
+<section id="trees"><title>Trees and Forest</title>
```
szhorvat, line 144: "forests" should be plural.

"forests" should not be capitalized: let's keep the style consistent with the rest of this document.

src/properties/trees.c
```diff
@@ -349,3 +349,197 @@ igraph_error_t igraph_is_tree(const igraph_t *graph, igraph_bool_t *res, igraph_
...
+ */
+igraph_error_t igraph_is_forest(const igraph_t *graph, igraph_bool_t *res, 
+                            igraph_vector_int_t *roots, igraph_neimode_t mode) {
+    igraph_vector_bool_t visited;
+    igraph_integer_t visited_count=0;
+    igraph_integer_t vcount, ecount;
+
+    vcount = igraph_vcount(graph);
+    ecount = igraph_ecount(graph);
+    IGRAPH_CHECK(igraph_vector_int_resize(roots, 0));
```
szhorvat, line 459: Let us use `igraph_vector_int_clear()` here and drop `IGRAPH_CHECK` (because `igraph_vector_int_clear()` never fails and thus does not return an error code).

ntamas, line 459: `igraph_is_forest()` should allow the user to pass in a null pointer for `roots` if the user is not interested in the roots and only wants a yes/no answer in `res`. it should also allow `res` to be null if the user is only interested in the roots. Basically you would need to guard accesses to `*res` and `*roots` with `if (...)` statements.

szhorvat (changes requested):
I left a few comments, but most of these are nitpicks that are easily fixed, and they are not what I'm worried about.

My major concern is that I am not convinced that the algorithm is correct, i.e. that cycles will always be detected correctly. Specifically, in the undirected case, only _unvisited_ neighbours of a vertex are pushed onto the stack. Visited ones are skipped. This is done to avoid traversing the graph backwards towards a vertex's parent in the DFS tree. However, the check you do is not specifically for the parent, but for any visited vertex.

Could it happen that a vertex will not be pushed onto the stack due to this check, preventing the detection of a real cycle?  Can you prove that this can never happen?  It may be true, but it is far from obvious to me.

I don't have time to dig into deeper today, so I'll stop here for now.

I will also note that counting visited vertices and checking that all of them were reached seems pointless here. For `is_tree` we did it to ensure that the graph was connected: unlike forests, all trees are connected. For forests, this is not necessary.

szhorvat:
Actually I think this is correct. The question can be put as: if there is a cycle in the graph, is it certain that at least one vertex will be "discovered" more than once (i.e. pushed onto the stack more than once) before it is fully processed.

The remaining part is the counting of visited vertices: why did you keep this check?

rohitt28:
> counting of visited vertices: why did you keep this check?

Do you mean this one?
``if(*res)
{
          *res= visited_count==vcount;
}``

If yes then, in case of directed graph there could be 1 separate component with only 1 node that has self loop.
That node will have degree 1 so the  ``igraph_forest_visitor`` function will not be called for it. Then if rest of graph is a forest then we will get wrong output.

rohitt28:
>  if there is a cycle in the graph, is it certain that at least one vertex will be "discovered" more than once (i.e. pushed onto the stack more than once) before it is fully processed.

You are probably right about that. I tried some test cases in which I thought current algo will fail but it seems to pass.
For eg: `igraph_small(&graph, 6, 0, 0, 1, 0, 2, 2, 1, -1);` Anyways, I will just use another vector to keep track of parents.

szhorvat:
> If yes then, in case of directed graph there could be 1 separate component with only 1 node that has self loop.
> That node will have degree 1 so the `igraph_forest_visitor` function will not be called for it. Then if rest of graph is a forest then we will get wrong output.

Yes, you are correct about this.

Please give me a couple more days to convince myself about the correctness of the algorithm, or otherwise find a counterexample, and I'll get back to you. One reason why it is tricky is that if we skip pushing visited vertices on the stack then unvisited vertices will be pushed more than once only in the undirected case, not the directed case. That's of course not a problem because we only check for the visited status in the undirected case. I'm just explaining why I found this confusing: even though the code for the directed and undirected cases is almost identical, the way they work is quite different.

szhorvat:
@rohitt28 Thanks again for implementing this! Everything's fine with the algorithm. Can you fix the small issues I commented on, and double-check the formatting? Then I'll add a few comments and a few small documentation edits and we can merge this.

szhorvat:
Summary of changes I made:

There were a couple of places where `roots` was accessed without checking that it is not NULL.

We now re-use the stack and neighbours vector across calls to the visitor function.

Added another test.

The rest are cosmetic changes.

szhorvat:
I think this is good to merge now.

Thanks so much for your work @rohitt28!

ntamas (approved):

ntamas:
Thanks a lot @rohitt28 !

PR #1892, opened 2021-12-15
Reimplementation of igraph_ring()

### LINKED ISSUE #1870: What should igraph_ring() return for n=1 and 2?
When asking close "closed rings", `igraph_ring()` returns:

 - for `n=1`, the singleton graph with no edges
 - for `n=2`, two vertices connected by a single edge

image

Should it instead return:

 - for `n=1`, a singleton graph with a self-loop
 - for `n=2`, two vertices connected by two edges

image

This is what I would expect for the "1-cycle" and the "2-cycle".

The reason why it behaves this way is that it's just a wrapper for `igraph_lattice()`. I would rewrite the function fro scratch and make sure that modifications to `igraph_lattice()` would not affect it.

ayushagr2002:
I have written initial implementation for the case of closed circular graphs when calling `igraph_ring()` when n=1 or 2. Please review and further guidance on it.

Fixes #1870.


GroteGnoom:
You pushed three binary files, can you remove them?

examples/simple/igraph_ring.c
```diff
@@ -42,5 +42,14 @@ int main() {
     /* Destroy the graph. */
     igraph_destroy(&graph);
 
+    igraph_ring(&graph,1,IGRAPH_UNDIRECTED,0,1);
```
GroteGnoom, line 45: You've changed the example, so the output should change as well, you can fix this in the igraph_ring.out file.

src/constructors/regular.c
```diff
@@ -311,12 +311,45 @@ int igraph_lattice(igraph_t *graph, const igraph_vector_t *dimvector,
...
-
     if (n < 0) {
         IGRAPH_ERRORF("The number of vertices must be non-negative, got %" IGRAPH_PRId ".", n, IGRAPH_EINVAL);
     }
+    if(n == 1 && circular)
+    {
+        igraph_vector_t edges;
+        
+        //initialise edge vector and insert edge 0-0 for self loop
+        igraph_vector_init(&edges,2);
```
GroteGnoom, line 322: you can use igraph_vector_init_int to clean things up.

GroteGnoom, line 322: You can use igraph_vector_init_int to clean things up.

GroteGnoom: When you initialize a vector it needs to be added to the 'finally stack' with
IGRAPH_FINALLY(igraph_vector_destroy, &edges);

src/constructors/regular.c
```diff
@@ -311,12 +311,45 @@ int igraph_lattice(igraph_t *graph, const igraph_vector_t *dimvector,
...
+        igraph_vector_set(&edges,0,0);
+        igraph_vector_set(&edges,1,0);
+
+        //create the graph
+        IGRAPH_CHECK(igraph_create(graph,&edges,0,directed));
+        igraph_vector_destroy(&edges);
+        //IGRAPH_FINALLY_CLEAN(1);
```
GroteGnoom, line 329: This should be uncommented.

GroteGnoom:
Thanks for your draft :) I've added some comments.

ayushagr2002:
Thank you for the suggestions. I will make changes accordingly and make commits again.

examples/simple/igraph_ring.c
ayushagr2002: Can you explain what should i change here?

src/constructors/regular.c
ayushagr2002: Done

src/constructors/regular.c
```diff
@@ -311,12 +311,47 @@ int igraph_lattice(igraph_t *graph, const igraph_vector_t *dimvector,
...
+
+        return IGRAPH_SUCCESS;
+    }
+
+    if(n == 2 && circular && directed)
+    {
+        igraph_vector_t edges;
+
+        //initialise edge vector and insert edge 0-1 and 1-0
+        igraph_vector_init_int(&edges,4,0,1,1,0);
```
GroteGnoom, line 340: I'm sorry I didn't think of this earlier, but you don't have to first create edges and then a graph, if you're not building up the edges algorithmically. You can call `igraph_small` directly, which makes everything even simpler.

examples/simple/igraph_ring.c
GroteGnoom: Look at the output of this example, check if it's correct, and if so, put it in  examples/simple/igraph_ring.out so the tests pass.

GroteGnoom:
General nitpick: please put spaces after your commas.

src/constructors/regular.c
```diff
@@ -311,12 +311,25 @@ int igraph_lattice(igraph_t *graph, const igraph_vector_t *dimvector,
...
+    }
 
+    if(n == 2 && circular && directed)
+    {
+        //create the graph
+        IGRAPH_CHECK(igraph_small(graph,2,directed,0,1,1,0,-1));
+        return IGRAPH_SUCCESS;
+    }
+
+    igraph_vector_t v = IGRAPH_VECTOR_NULL;
```
GroteGnoom, line 332: This declaration can stay at the top of the function. It also doesn't need to be initialized, so you can change it to `igraph_vector_t v;`

src/constructors/regular.c
```diff
@@ -311,12 +311,25 @@ int igraph_lattice(igraph_t *graph, const igraph_vector_t *dimvector,
...
 
-    igraph_vector_t v = IGRAPH_VECTOR_NULL;
-
     if (n < 0) {
         IGRAPH_ERRORF("The number of vertices must be non-negative, got %" IGRAPH_PRId ".", n, IGRAPH_EINVAL);
     }
+    
+    if(n == 1 && circular)
+    {
+        //create the graph
```
GroteGnoom, line 320: These comments aren't needed.

src/constructors/regular.c
ayushagr2002: That was already there when i cloned the repository. I can remove if you want.

src/constructors/regular.c
ayushagr2002: Ok, I will remove them.

GroteGnoom:
The binary file `examples/simple/igraph_ring` is still there and can be removed.

GroteGnoom:
If you can still put those spaces after your commas, then you can press 'ready for review' and I'll hand it over to someone else for a final checkup. Thanks for your work 🙂

GroteGnoom:
(The undirected n=2 cases still needs to be checked)

ayushagr2002:
> 

Can you tell what should the graph look like for this case?

ayushagr2002:
For the case of n=2, the edge from 0-1 is the same as 1-0, so does it make sense for a closed loop in an undirected graph with two vertices? Please give your views.

GroteGnoom:
It matters if there is one edge between 0 and 1, or if there are two. With one edge you can't make a cycle, and with two you can.

ayushagr2002:
```
2-cycle graph:
0 1
0 1
```
Is this the expected output for undirected n=2?

szhorvat:
> Is this the expected output for undirected n=2?

Yes.  It is a 2-cycle:

image

szhorvat:
This is one possible solution, but as I said [in this comment](https://github.com/igraph/igraph/issues/1870#issuecomment-993612539), we would like to make `igraph_ring()` completely independent of `igraph_lattice()`. Can you write an implementation that does not call `igraph_lattice()`?

szhorvat:
There are a number of problems here that need to be fixed before we can go on with the review.

First, this PR appears to be based on a very old version of the `develop` branch, much older than what was current when you opened the PR, and from well before igraph was refactored to use `igraph_integer_t` consistently. I'm not sure how this happened, but there are many conflicts with the current `develop`. Make sure you base the work on the latest `develop` branch and that there are no merge conflicts. In fact, it might be easiest for you to create a new PR.

Second, you seem to have reformatted the entire `regular.c` file in a style that is different from what the rest of igraph uses. Please do not modify any code unrelated to `igraph_ring`. For the code that you do need to  touch, use the same style that you see in the rest of igraph.

Finally, do not commit irrelevant files (such as the `examples/simple/igraph_ring` executable).

szhorvat:
@ayushagr2002 Are you still working on this?

ayushagr2002:
Hi, I apologise for the inactivity due to some health issues. I will start working on it from now.

szhorvat:
No problem, hope you are well now. There is no rush, I just wanted to make sure that you are still interested in completing this PR.

szhorvat:
I hope you are doing better. Let us know if you are still working on this.

szhorvat:
@ayushagr2002 Are you still planning to finish this PR? If you don't have time, that's alright, but we need to know how to move forward.

szhorvat:
Closing due to no response.

Issue #1817, opened 2021-10-16
Identify widest paths


mcmahanp:
*What is the feature or improvement you would like to see?*

It would be great to implement a [widest paths](https://en.wikipedia.org/wiki/Widest_path_problem) algorithm in igraph. The 'width' of a path in a weighted graph is the minimum weight of the edges in that path, and the widest path between two points is the path with the greatest width between them. 

*Use cases for the feature*

The widest path has applications in various flow-type networks (e.g. traffic, networking, metabolic networks, …), and in the [Schulze method](https://en.wikipedia.org/wiki/Schulze_method) of identifying a winner in ranked-choice voting. It's [currently  possible brute force this](https://stackoverflow.com/questions/57522871/how-to-calculate-a-maximum-bottleneck-path-with-igraph) in igraph by enumerating all simple paths and calculating the minimax over the edge weights, but this is very inefficient (especially for networks with any sort of density).

*References*

Based on a cursory read, it looks like the best algorithm for this depend on whether the graph is directed or undirected and on the overall density of the network. However there are general purpose algorithms as well.

The [Wikipedia page on widest paths](https://en.wikipedia.org/wiki/Widest_path_problem) cites a number of algorithms, and the page on the [Schulze method](https://en.wikipedia.org/wiki/Schulze_method#Implementation) lists a simple algorithm.

iosonofabio:
If this isn't too hard my new intern could take this

mcmahanp:
> Can you please elaborate a bit more on _actual_ use cases, so we can make a better judgement about what the most common needs are, and what to implement?

Great point. In my case, I'm only interested in the Schulze method specifically. For this specific use case the networks will tend to be small, dense, and directed.

But I know that the single-source, single-target solution is also useful for other applications. E.g. When trying to find the highest-bandwidth path through a computer network you would want to specify the source and target in an *undirected* network. Also, for ecological and metabolic networks (like the one asked about in [this stackoverflow question](https://stackoverflow.com/questions/57522871/how-to-calculate-a-maximum-bottleneck-path-with-igraph) it is common to need the widest path from a single source to a single target in a *directed* network. So an ideal igraph function would choose its algorith based on the type of network.

Incidentally, the modified Floyd-Warshall algorithm laid out in the [Shulze method wikipedia page](https://en.wikipedia.org/wiki/Schulze_method#Implementation) is very simple to implement for the all-pairs version -- though I suspect there are more efficient algorithms as well.

[...]

Issue #1880, opened 2021-12-08
Consistent representation of trees through "parent vectors"


szhorvat:
There are several functions in igraph which effectively return a tree through a "parent vector", sometimes called "predecessors" or "father". We should make sure that functions that use this format are consistent with each other. We should also use a consistent and neutral naming (such as "parent").

Here is an incomplete list of relevant functions:

 - `igraph_bfs`, `igraph_bfs_simple`, `igraph_dfs`. @ntamas proposed to overhaul the interface of these functions.
 - `igraph_get_shortest_paths`, `igraph_get_shortest_paths_dijkstra`, `igraph_get_shortest_paths_bellman_ford` return the shortest path tree in a similar format. (fixed in #2070 and 9fddebf)
 - `igraph_dominator_tree` returns a tree like this as well
 - There is a proposal in #1840 for a function to create a tree from this format.
 - Any other functions that are not yet mentioned? Please edit the issue and complete this list.

There are small differences in the representation:

 - DFS/BFS functions used NaN to represent the lack of a parent, but this was switched to -1 after the integer transition
 - Shortest path functions use -1 for unvisited vertices, but they use the same index as the vertex itself for the root vertex. According to the current proposal in #1840 such a vector would be rejected by the tree creation function. edit: shortest path functions now confirm to this proposal.

We need to make everything consistent, and since this is API-breaking, ideally do this before 0.10.

**Other related issues to consider:**

 - Do we want other functions to optionally return data in the same format? How about spanning tree functions—is it useful at all for them?
 - Do we want to introduce a format which can encode not just a spanning tree, but a complete edge traversal? This relates to modifying DFS/BFS which currently just record nodes, but not the edges. For some applications, the edges are necessary (e.g. cycle basis). It also relates to weighted shortest path functions, which also return the predecessor edge ID of each vertex—this is necessary to identify the weight of the incoming edge. But it is not sufficient for a complete representation of an edge traversal: for that we additionally need the predecessor vertex of each edge (thus two vectors: predecessor edge of each vertex and predecessor vertex of each edge). Finally, it also related to `igraph_unfold_tree()`.

szhorvat:
Let me make a concrete suggestion about what restrictions parent vectors must have and what their interpretations should be.

 - Values are either negative or a valid vertex ID.
 - `parent[i]` is the vertex ID of the parent/predecessor of vertex `i`.
[...]

PR #1893, opened 2021-12-15
Widest Path Implementations

Gomango999:
This pull request implements various means of calculating widest paths, as requested in #1817.

The function headers are essentially identical to the `shortest_path_dijkstra` variants, since the problem of finding the widest path from a source node to every other node can be implemented as a modified Dijsktra. The only difference is that the `igraph_widest_paths` function (which only computes the widths and not the actual paths between all pairs of nodes) is split into Dijkstra and Floyd-Warshall implementations. The Floyd-Warshall algorithm will allow more efficient computation of widest paths between all pairs of nodes on dense graphs, as required by the [Schulze method](https://en.wikipedia.org/wiki/Schulze_method#Implementation). Meanwhile, the Dijkstra implementation is more efficient for sparse graphs.

Note that there are several more implementations mentioned on [Wikipedia](https://en.wikipedia.org/wiki/Widest_path_problem), some of which are  faster for certain types of graphs then the ones listed above. However, I believe the Dijkstra (O(M log M)) and Floyd-Warshall (O(N^3)) implementations together should be suitable enough for the use cases described in the original issue.


include/igraph_paths.h
```diff
@@ -197,6 +197,43 @@ IGRAPH_EXPORT igraph_error_t igraph_spanner(const igraph_t *graph,
...
+                                             const igraph_vector_t *weights,
+                                             igraph_neimode_t mode);
+IGRAPH_EXPORT igraph_error_t igraph_get_all_shortest_paths(const igraph_t *graph,
+                                                   igraph_vector_ptr_t *vertices,
+                                                   igraph_vector_ptr_t *edges,
+                                                   igraph_vector_int_t *nrgeo,
+                                                   igraph_integer_t from, igraph_vs_t to,
+                                                   const igraph_vector_t *weights,
+                                                   igraph_neimode_t mode);
+IGRAPH_EXPORT igraph_error_t igraph_widest_paths_floyd_warshalls(const igraph_t *graph,
```
ntamas, line 223: Typo here: it should be `floyd_warshall`

src/paths/widest_paths.c
```diff
@@ -0,0 +1,105 @@
...
+#include "igraph_dqueue.h"
+#include "igraph_interface.h"
+#include "igraph_memory.h"
+#include "igraph_stack.h"
+#include "igraph_qsort.h"
+
+#include "core/indheap.h"
+#include "core/interruption.h"
+
+#include <string.h>   /* memset */
```
ntamas, line 37: Not sure if `string.h` will be needed in this file; also it is generally a good idea to cut down on the number of includes first and then progressively re-add includes as you move ahead with the implementation.

src/paths/widest_paths.c
```diff
@@ -0,0 +1,105 @@
...
+                                       igraph_integer_t from,
+                                       igraph_vs_t to,
+                                       const igraph_vector_t *weights,
+                                       igraph_neimode_t mode,
+                                       igraph_vector_int_t *predecessors,
+                                       igraph_vector_int_t *inbound_edges) {
+
+    // TODO: implement a modified dijkstra to calculate widest paths from a source to every other node
+
+    return IGRAPH_SUCCESS;
```
ntamas, line 52: Please return `IGRAPH_UNIMPLEMENTED` until we actually have an implementation.

tests/unit/igraph_widest_paths.c
```diff
@@ -0,0 +1,28 @@
...
+
+#include <igraph.h>
+
+#include "test_utilities.inc"
+
+int main() {
+
+    // TODO: Add tests
+
+    return 0;
```
ntamas, line 27: Make sure to fail the test by returning 1, or to skip the test by returning 77 (magic exit code) so we don't accidentally merge this PR without writing tests.

tests/CMakeLists.txt
```diff
@@ -248,6 +248,7 @@ add_examples(
   igraph_simplify
   igraph_topological_sorting
   igraph_transitivity
+  igraph_widest_paths
```
ntamas, line 251: You have added a unit test, not an example so please do not add it to the `add_examples(...)` block; those are meant for exercising code in `examples/simple`.

ntamas:
Looks fine to me so far, apart from the comments I added above. The headers look okay, we can adjust later as you go ahead with the PR if needed.

include/igraph_paths.h
Gomango999: Done

src/paths/widest_paths.c
Gomango999: Was mainly copying from the `dijkstra.c` file, but definitely good advice. I've removed some of the headers, and I'll re-add as needed.

src/paths/widest_paths.c
Gomango999: Done

tests/unit/igraph_widest_paths.c
Gomango999: Done, used exit code 77

tests/CMakeLists.txt
Gomango999: Oops my bad! It's now in the correct spot (add_legacy_tests).

Gomango999:
Hello! Just had a chance to read over the existing dijkstra code in detail (specifically, the implementation of `igraph_shortest_paths_dijkstra()` in `paths/dijkstra.c`), and I have a question about the 2 way heap. What exactly do `igraph_2wheap_has_elem()` and `igraph_2wheap_has_active()` do? They seem to return something related to `VECTOR(h->index2)[idx]`, which judging by `igraph_2wheap_get` is an index in the data vector (albeit offset by 2). What does the index in the data vector have to do with 'having an element' or whether an element 'is active'?

ntamas:
Okay, so, the `2wheap` isn't really well-documented (and it's not part of the public API either). I'm sorry about the lack of docs, but this is a good opportunity to improve the documentation :)

So, `igraph_2wheap_t` contains a `size` property and a `data` vector; these constitute a "normal" heap of doubles so there's not much to see here. Then we have `index`, which is an index (an integer) associated to each element in the heap. Let's call this "item" instead of index because it will become confusing later :)

So, when you push values into the heap, you push item-value pairs with `igraph_2wheap_push_with_index()`. The _value_ that you push into the heap goes in `data`, while the _item_ corresponding to the value (think vertex or edge indices) goes in `index`. When you swap two elements in the heap while maintaining the heap invariant, you swap them both in `data` (which swaps the values) and also swap them in `index` (which swaps the items); they are always moved as if they were a single pair. The maximum element is always at `data[0]`, and the corresponding item is in `index[0]`.

Now, the problem with the data structure so far is that looking up the _value_ corresponding to an _item_ is an O(n) operation because you need to scan the entire `index` array to find the item you are looking for. E.g., if you use the max-heap to store path lengths leading to vertices, and you want to know the length of the path leading to vertex 42, you need to scan the entire `index` array, find where `42` is in it, and then look at the corresponding cell in `data`. That's why we have `index2` -- looking up `index2[42]` would tell you where item 42 is in `index` and `data` in O(1) time.

The dirty little secret of `index2` is that it is _not_ true that `index[index2[42]] == 42` and `data[index2[42]] == the-value-corresponding-to-item-42` because there are two special values: 0 and 1:

* When `index2[x] == 0`, it means that item 42 is _not_ in the heap at all. We need to do it like this because `igraph_vector_long_init(&h->index2, ...)` initializes the `index2` array to all-zeros, and we don't want to rewrite the values again to another special value (say, -1). So, zero in `index2` means "item does not exist". `igraph_2wheap_has_elem(heap, item)` will return false if `heap->index2[item]` is zero and true otherwise.
* The other special value in `index2` is 1, which means "item is deactivated". Deactivated items are just like removed items; `igraph_2wheap_deactivate_max()` basically pops the max item from the heap just like `igraph_2wheap_delete_max()`, the only difference is that you can distinguish between genuine removal and deactivation with `igraph_2wheap_has_elem()` and `igraph_2wheap_has_active()`. If the item is in the heap, both return `true`. If the item was deactivated but not deleted, `igraph_2wheap_has_elem()` returns `true` and `igraph_2wheap_has_active()` returns `false`. If the item was deleted, both return `false`.
* And, of course you may have guessed it by now: if the item is truly in the heap and has not been deactivated, `heap->index2[item] - 2` tells you the index where you should look for its value in `heap->data[...]`.

I hope this helps.

Gomango999:
That explanation was awesome, thanks a bunch! I hadn't realised that removing and deactivating were two separate things, but it makes _so_ much more sense now :). If I get time I'd like to contribute a PR paraphrasing your explanation into the relevant file, if that's okay?

ntamas:
Yes, please, go ahead and feel free to improve the comments about `igraph_2wheap_t`. There's no need to make it super pretty as it is not part of the public API so it won't appear anywhere on the homepage. Just make it detailed enough that the next time we need it I won't have to read the source code again to figure out how things work :)

Gomango999:
Okay, all the code is done. All that remains is to write up some more comphrensive tests, as well as further documentation.

I was originally planning to implement `igraph_get_all_widest_paths`, but I couldn't figure out a way to do it efficiently. It's not like `igraph_get_all_shortest_paths` because in that function, each path only traverses nodes in an order such that the distance from the source strictly increases which greatly narrows down the number of paths you can take. For `igraph_get_all_widest_paths`, the width is only non-increasing as you go further away from the source. This means that for certain cases (for example one where all the edge weights are the same), then finding all widest paths is equivalent to finding every path from one node to every other in an unweighted graph. This can be done with a DFS, but it's extremely inefficient, and made me reconsider whether it would be worth implementing a function that has such an slow runtime. Of course I'd love to hear your thoughts on this @ntamas .

szhorvat:
I won't be able to go through it for a while, just wanted to make sure that you've seen this wiki page:

https://github.com/igraph/igraph/wiki/Checklist-for-new-%28and-old%29-functions

(I think you can skip the functions.yaml part for now, but do consider the other points, if you haven't yet.)

Gomango999:
Yep, already looked through them for the most part., though I'll check the input vaidity in more detail when I write better tests.

Gomango999:
Ok I've finished the tests and documentation, as well as gone through everything in the "implementing functions checklist", this PR is ready for review. In particular, I'm curious about my usage of `IGRAPH_ALLOW_INTERRUPTION()`. I couldn't find documentation on it, so I took my best guess for it.

szhorvat:
`IGRAPH_ALLOW_INTERRUPTION()` is like `IGRAPH_ERROR` (or like `IGRAPH_CHECK`). It checks if the user requested interruption, and if yes, it effectively raises an "error". Of course, it's not an _error_, but it's handled exactly the same way. It just has its own error code.

Thus, you need to pay attention to the same things as with any error check (make sure that if an error is found, all resources are properly freed). 

Additionally, it's good to pay attention to performance, and not call it in very tight loops. How long the interrupt check takes depends on what's in the interrupt handler: it's different for each high-level interface. But generally it should be very fast.

Gomango999:
Merge conflicts are resolved, and `IGRAPH_ALLOW_INTERRUPTION()` is called just a little less frequently now.

src/paths/widest_paths.c
```diff
@@ -0,0 +1,745 @@
...
+    igraph_real_t my_neginfinity = IGRAPH_NEGINFINITY;
+    igraph_bool_t all_to;
+    igraph_vector_int_t indexv;
+
+    if (!weights) {
+        IGRAPH_ERROR("Weight vector is required.", IGRAPH_EINVAL);
+    }
+
+    if (igraph_vector_size(weights) != no_of_edges) {
+        IGRAPH_ERRORF("Weight vector length (%ld) does not match number "
```
szhorvat, line 640: This is the cause of the test failure. Instead of `"...%ld..."`, use `"...%"IGRAPH_PRId"..."`

Gomango999: Ah thanks for that. I was able to compile successfully locally, and forgot to check the Github tests after pushing. 

Out of curiosity, is there a way to run the tests locally before pushing publically to the PR? Sometimes I get worried that people will start reviewing before I have a chance to fix whatever bugs come up in the test.

szhorvat:
> Out of curiosity, is there a way to run the tests locally before pushing publically to the PR?

Of course! Build the `check` target. If you use `ninja` (recommended), then `ninja check`. If you use `make`, then `make check`.

Please do ask any such question freely, and early. It will safe you a lot of time and trouble. I always assumed that anyone who contributes a PR would run tests locally.

ntamas:
Also, if you want to run a few tests only, you can save some time by running `ninja build_tests && ctest -R name_of_test`. `name_of_test` has to be a word or something that is part of the test name; `ctest` will run all tests that matches the regex given there. It also has an option to print the stdout in case of failure.

szhorvat:
We should write up a development guide detailing all of this.

Gomango999:
@szhorvat Just checking the status of this PR. I've fixed the cause of the testing error and this commit is ready for review. Apologies if this was already on your radar!

szhorvat:
It is on my radar, and thanks for the great contribution! Personally, I was very happy to see this work. 

Unfortunately, I won't be able to do a proper review for a few weeks due to being overwhelmed with job applications. In order to test this thoroughly, I'd need to update the Mathematica interface of igraph to use the `develop` branch, which will take a while.

There is some urgency to merging this PR, as well as PR #1763, because @ntamas is working on changes that might affect this (replacing "pointer vectors" with "vector lists"). If someone else could take a thorough look at this, and test is from _some_ high-level interface, that would be really great. Not sure if @vtraag or @GroteGnoom can afford the time.

GroteGnoom:
Why test it from some high-level interface specifically?

szhorvat:
It's not strictly necessary, it's just how I like to test things. It makes it much easier for me to verify that the result is correct in some special cases. Sometimes I implement a very simple (and usually very inefficient) version of the same operation in a high-level language and compare results.

If people are confident about correctness without that sort of verification, then please go ahread.

GroteGnoom:
If @vtraag doesn't have time I can do it, but it will probably take me a couple of workdays, including understanding the theory, which means it will take weeks real time.

GroteGnoom:
Can you use some helper functions to clean up your tests? There seems to be quite a bit of copied code. For example, two tests which only differ in weights should be two function calls with different weights, not two code blocks.

Also, can you remove unnecessary comments, like ```/* edges */ &edges,```. I think that would also make it easier to read.

ntamas:
Eventually this PR will have to be updated so that the function signatures use the newly introduced `igraph_vector_int_list_t` type instead of an `igraph_vector_ptr_t` as the former is type-safe while the latter is not.

I'd suggest to go ahead with the review of the PR as is. If everything seems okay, I'll rewrite the functions to use `igraph_vector_int_list_t` before merging.

Gomango999:
The widest paths tests should be a lot more readable now that I've moved most of the repetitive stuff into functions. I've run tests locally and it passes.

Looking at the details of why my tests are failing, I'm assuming it's because it is testing my branch after it's been merged into develop, which already has `igraph_vector_ptr_t` removed?

ntamas:
`igraph_vector_ptr_t` was not removed but `igraph_paths.h` is not including `igraph_vector_ptr.h` any more in the `develop` branch.  You can re-add the include and it should compile just fine.

Gomango999:
There seems to be a line in`src/cliques/glet.c` that stops the code from compiling. Specifically line 373 contains the line
```c

igraph_vector_int_list_pop_back(&mycliques, cl);
```
which I believe should be 
```c
igraph_vector_int_list_remove_back(&mycliques, cl);
```
Exchanging it allows it to compile locally. Should I open a PR for this issue?

Gomango999:
All good then, thanks!

Gomango999:
Seems to be more compilation issues with vector lists, this time in `src/core/typed_list.pmt:543`. I think I'm going to leave this for now, until develop is a bit more stable.

ntamas:
`develop` should compile again now (except the docs, which are being worked on).

src/paths/widest_paths.c
```diff
@@ -0,0 +1,747 @@
...
+    width to the current node and the width of the edge.
+
+    We denote the widest path from a node to itself as infinity, and the widest
+    path from a node to a node it cannot reach as negative infinity.
+    */
+
+    igraph_integer_t no_of_nodes = igraph_vcount(graph);
+    igraph_integer_t no_of_edges = igraph_ecount(graph);
+    igraph_real_t my_posinfinity = IGRAPH_POSINFINITY;
+    igraph_real_t my_neginfinity = IGRAPH_NEGINFINITY;
```
GroteGnoom, line 143: why do you rename these?

GroteGnoom:
Since this is going to take a while to evaluate, I'll just make a checklist for confirmed tests, and then they don't have to be checked in one go:

- [ ] 1. null weight vector -- maybe it should accept them?
- [ ] 2. Number of weights don't match number of edges
- [ ] 3. NaN values in weights
- [ ] 4. Empty graph
- [ ] 5. Testing 1 Node Graph
- [ ] 6. Testing Unreachable Nodes
- [ ] 7. Testing Self Loops
- [ ] 8. Testing Multiple Edges
- [ ] 9. Testing Directed Graphs
- [ ] 10. Testing Mode
- [ ] 11. Testing Multiple Widest Paths
- [ ] 12. Testing 5 Node Simple Graph
- [ ] 13. Testing 7 Node Wikipedia Graph
- [ ] 14. Testing Negative Weights
- [ ] 15. Testing Disconnected Graphs

src/paths/widest_paths.c
szhorvat: I guess this pattern was lifted from the shortest path functions. I'm also not sure why it was done this way there. My guess is that in the past, `IGRAPH_POSINFINITY` may have been defined through a function, i.e. evaluation was expensive:

https://github.com/igraph/igraph/blob/0.8.5/include/igraph_types.h#L68

@gaborcsardi Do you remember why it was done this way?

tests/unit/igraph_widest_paths.c
```diff
@@ -0,0 +1,587 @@
...
+}
+
+/* confirm that all widest paths algorithms fail */
+void check_invalid_input(igraph_t *g, igraph_vector_t *w, igraph_matrix_t *res1,
+                        igraph_vs_t *from, igraph_vs_t *to,
+                        igraph_vector_ptr_t *vertices, igraph_vector_ptr_t *edges,
+                        igraph_vector_int_t *predecessors, igraph_vector_int_t *inbound_edges,
+                        igraph_vector_int_t *vertices2, igraph_vector_int_t *edges2) {
+    CHECK_ERROR(igraph_widest_paths_dijkstra(/* graph */ g, res1,
+                            *from, *to, /* weights */ NULL,
```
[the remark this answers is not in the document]
Gomango999: Yep you're 100% right. Looking again, I found that, as well as even more silly mistakes that were left in the code, probably because I rushed the last couple commits (my fault, sorry!). It'll be a couple days before I can really go over the test outputs properly and make sure I didn't do anything else silly.

src/paths/widest_paths.c
Gomango999: A good chunk of the code is copied from the equivalent function in `src/paths/dijkstra.c`. Specifically:

- `igraph_widest_paths_dijkstra` is `igraph_shortest_paths_dijkstra` 
- `igraph_get_widest_paths` is `igraph_get_shortest_paths_dijkstra`
- `igraph_get_widest_path` is `igraph_get_shortest_path_dijkstra`

I was originally planning on writing my own faster functions to replace dijkstra, but realised that the difference in run time in dijkstra was already good enough for most use cases and there was no need for me to reinvent the wheel.

Gomango999:
> It'll be a couple days before I can really go over the test outputs properly and make sure I didn't do anything else silly.

Saw that the checklist hadn't been started yet, so took the chance to fix up a few minor things with the text. Namely, making matrices the correct sizes, and making sure to add some asserts to better show the exact thing I'm testing.

tests/unit/igraph_widest_paths.c
```diff
@@ -0,0 +1,578 @@
...
+    igraph_vs_1(&from, 0);
+    igraph_vs_seq(&to, 0, n-1);
+
+    init_vertices_and_edges(n, &vertices, &edges, &predecessors, &inbound_edges, &vertices2, &edges2);
+
+    run_widest_paths(&g, &w, &res1, &res2, &from, &to, IGRAPH_OUT);
+    run_get_widest_paths(&g, &w, /* source */ 0, /* destination */ 3, &to, IGRAPH_OUT,
+                            &vertices, &edges, &predecessors, &inbound_edges, &vertices2, &edges2);
+
+    IGRAPH_ASSERT(MATRIX(res1, 0, 3) == IGRAPH_NEGINFINITY);
```
GroteGnoom, line 309: There's a few of these assertions which are already checked by comparing the output to the .out file (like the result `[ Inf 2.000000 2.000000 -Inf ]`). I think these can be removed, they don't need a double check.

tests/unit/igraph_widest_paths.c
```diff
@@ -0,0 +1,578 @@
...
+    igraph_vs_seq(&to, 0, n-1);
+
+    init_vertices_and_edges(n, &vertices, &edges, &predecessors, &inbound_edges, &vertices2, &edges2);
+
+    run_widest_paths(&g, &w, &res1, &res2, &from, &to, IGRAPH_OUT);
+    run_get_widest_paths(&g, &w, /* source */ 0, /* destination */ 2, &to, IGRAPH_OUT,
+                            &vertices, &edges, &predecessors, &inbound_edges, &vertices2, &edges2);
+
+    IGRAPH_ASSERT(MATRIX(res1, 0, 2) == -3);
+    check_and_print_matrices(&res1, &res2);
```
GroteGnoom, line 545: Why don't you do a `print_results` here?

src/paths/widest_paths.c
```diff
@@ -0,0 +1,747 @@
...
+#include "igraph_interface.h"
+#include "igraph_memory.h"
+#include "igraph_qsort.h"
+
+#include "core/indheap.h"
+#include "core/interruption.h"
+
+/**
+ * \function igraph_get_widest_paths
+ * \brief Widest paths from a single vertex
```
GroteGnoom, line 37: There's a `.` missing :)
```suggestion
 * \brief Widest paths from a single vertex.
```

src/paths/widest_paths.c
```diff
@@ -0,0 +1,747 @@
...
+ *        if they are both null pointers.
+ * \param from The id of the vertex from/to which the widest paths are
+ *        calculated.
+ * \param to Vertex sequence with the IDs of the vertices to/from which the
+ *        widest paths will be calculated. A vertex might be given multiple
+ *        times.
+ * \param weights The edge weights. Edge weights can be negative. If this
+ *        is a null pointer or if any edge weight is NaN, then an error
+ *        is returned.
+ * \param mode The type of widest paths to be use for the
```
GroteGnoom, line 69:
```suggestion
 * \param mode The type of widest paths to be used for the
```

src/paths/widest_paths.c
```diff
@@ -0,0 +1,747 @@
...
+ *        \cli IGRAPH_EINVVID
+ *           \p from is invalid vertex ID, or the length of \p to is
+ *           not the same as the length of \p vertices (if not NULL) or
+ *           the length of \p edges (if not NULL)
+ *        \cli IGRAPH_EINVMODE
+ *           invalid mode argument.
+ *        \endclist
+ *
+ * Time complexity: O(|E|log|E|+|V|), where |V| is the number of
+ * vertices and |E| is the number of edges
```
GroteGnoom, line 110: Vertices and edges of the graph? Or vertices in the `to`?

src/paths/widest_paths.c
GroteGnoom: It makes sense that it used to be expensive because it was a function call. Let's now use IGRAPH_POSINFINITY etc. directly. They can also be changed in `src/paths/dijkstra.c`.

src/paths/widest_paths.c
```diff
@@ -0,0 +1,747 @@
...
+
+    if (igraph_vector_is_any_nan(weights)) {
+        IGRAPH_ERROR("Weight vector must not contain NaN values.", IGRAPH_EINVAL);
+    }
+
+    IGRAPH_CHECK(igraph_vit_create(graph, to, &vit));
+    IGRAPH_FINALLY(igraph_vit_destroy, &vit);
+
+    if (vertices && IGRAPH_VIT_SIZE(vit) != igraph_vector_ptr_size(vertices)) {
+        IGRAPH_ERROR("Size of `vertices' and `to' should match.", IGRAPH_EINVAL);
```
GroteGnoom, line 170: Why do you give an error here? Shouldn't you resize `vertices`? That's what the docs suggest, and what for example `igraph_get_shortest_paths_dijkstra` also does.

Same goes for the `edges`.

src/paths/widest_paths.c
```diff
@@ -0,0 +1,747 @@
...
+            igraph_integer_t size, act, edge;
+            igraph_vector_int_t *vvec = 0, *evec = 0;
+            if (vertices) {
+                vvec = VECTOR(*vertices)[i];
+                igraph_vector_int_clear(vvec);
+            }
+            if (edges) {
+                evec = VECTOR(*edges)[i];
+                igraph_vector_int_clear(evec);
+            };
```
GroteGnoom, line 294:
```suggestion
            }
```

GroteGnoom:
Just like in the current PR, I've also copied huge chunks of code from `src/paths/dijkstra.c`, but I this makes it much harder to check and maintain the code. I think we should try to use more helper functions.

tests/unit/igraph_widest_paths.c
Gomango999: My reasoning was to better highlight the specific purpose of the test for the reviewer, rather than just dump every single output the program makes. 

In hindsight, I agree that the tests are simple enough that the purpose of the test should be obvious. This has been fixed.

tests/unit/igraph_widest_paths.c
Gomango999: Great question :) This has been fixed.

src/paths/widest_paths.c
Gomango999: Done :)

src/paths/widest_paths.c
```diff
@@ -0,0 +1,747 @@
...
+ * using a modified Dijkstra's algorithm.  If there is more than one path with
+ * the largest width between two vertices, this function gives only one of them.
+ * \param graph The graph object.
+ * \param vertices The result, the IDs of the vertices along the paths.
+ *        This is a pointer vector, each element points to a vector
+ *        object. These should be initialized before passing them to
+ *        the function, which will properly clear and/or resize them
+ *        and fill the IDs of the vertices along the paths from/to
+ *        the vertices. Supply a null pointer here if you don't need
+ *        these vectors. Normally, either this argument, or the \c
```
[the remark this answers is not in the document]
Gomango999: I did find that comment to be a bit overly verbose too. I've removed it for both comments.

src/paths/widest_paths.c
Gomango999: Done.

src/paths/widest_paths.c
Gomango999: |V| is the number of nodes in the graph, since even if there was only one node in `to`, the Dijkstra could explore every other node in the graph first if that single node is the furthest from the source. 

I'll update the comment to clarify this.

src/paths/widest_paths.c
Gomango999: Genuinely not sure how that happened, since I would have had to manually change it to be something different! Either way, this is now fixed to resize `vertices` and `edges`, as noted by the documentation.

src/paths/widest_paths.c
Gomango999: Done.

GroteGnoom:
We've talked about it a little bit, and it seems like it's not such an easy problem here to prevent code duplication without affecting performance. For now it seems the best course of action is to just keep going with this review, and later take a good look at all the copied Dijkstra code.

GroteGnoom:
We need to think about what to do with #1880 and this PR. It's probably nicer to let everything conform to #1880, but that's not the case for shortest path functions right now, and maybe we want to merge this PR before changing everything.

src/paths/widest_paths.c
```diff
@@ -0,0 +1,743 @@
...
+ * \return Error code:
+ *        \clist
+ *        \cli IGRAPH_ENOMEM
+ *           not enough memory for temporary data.
+ *        \cli IGRAPH_EINVVID
+ *           \p from is invalid vertex ID, or the length of \p to is
+ *           not the same as the length of \p vertices (if not NULL) or
+ *           the length of \p edges (if not NULL)
+ *        \cli IGRAPH_EINVMODE
+ *           invalid mode argument.
```
GroteGnoom, line 102: IGRAPH_EINVAL can also occur as an error.

ntamas:
I think it's easier to merge this first and then take care of #1880 than to do the same thing in parallel in both the `develop` branch and here.

If you feel like this PR is ready apart from the handling of the parent vectors, let's merge this and deal with the parent vector representation later in #1880.

GroteGnoom:
@Gomango999 Sorry for the long and weird review process, thanks for all the great work :)

Gomango999:
Nope, it's been great looking at how much thought you all put into every PR. Thanks so much everyone for all the feedback and help!!!

iosonofabio:
Good job @Gomango999 !

vtraag:
This PR closes issue #1817.

PR #1896, opened 2021-12-20
Removing igraph_get_eids_multi() and adding igraph_get_all_eids_between()

ntamas:
This PR attempts to remove `igraph_get_eids_multi()` from the core API and replace it with `igraph_get_all_eids_between()`. It also attempts to remove the `path` argument of `igraph_get_eids()` in order to simplify the core API.

The primary reason for removing `igraph_get_eids_multi()` is that it does not make it possible to retrieve the edge IDs of all the edges between a pair of vertices without knowing in advance how many such edges there are between the pair. As such, I don't believe it has any use, which seems to be confirmed by the fact that the edge selector relying on this core API has been left incomplete for years now (for instance, the function that would have returned the number of edges selected by the selector was unimplemented).

The reason for removing the `path` argument of `igraph_get_eids()` is that it can be replicated by `igraph_get_eids()` and its `pairs` argument. `igraph_get_eids()` in its old version was essentially trying to do two different things (retrieving edge IDs by either pairs of vertices or a path of vertex IDs); now it does only one thing. A new function named `igraph_expand_path_to_pairs()` is provided to obtain a "pairs vector" from a "path vector".

This PR is a work in progress; I'm primarily soliciting feedback at this stage and exercising Codecov to figure out which parts need unit testing.


ntamas:
`igraph_get_all_eids_between()` does not work yet, FYI. Stay tuned.

PR #1906, opened 2022-01-01
fix: correct error handling when reading Pajek file

szhorvat:
@ntamas I need some input here because I am completely new to Bison. If this approach is good, we can use it for all the parsers.

With this fix, the locally run fuzzer did not find crashes, but it does hit the memory limit occasionally. This happens because Pajek files are explicit about the number of vertices, and it is very easy to just use a number which is too large.

### DIFF
--- a/src/io/pajek-header.h
+++ b/src/io/pajek-header.h
@@ -30,6 +30,7 @@ typedef struct {
     void *scanner;
     int eof;
     char errmsg[300];
+    igraph_error_t igraph_errcode;
     igraph_vector_int_t *vector;
     igraph_bool_t directed;
     igraph_integer_t vcount, vcount2;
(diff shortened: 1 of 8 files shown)


szhorvat:
It looks like there are still some memory leaks on invalid inputs, but I don't have the time anymore today.

https://dev.azure.com/igraph-team/igraph/_build/results?buildId=2041&view=logs&jobId=4c3a8381-b9de-5463-904f-918810769be8&j=4c3a8381-b9de-5463-904f-918810769be8&t=a906eab8-d2a3-5e54-d199-606dbbfb0805

ntamas:
This approach seems good to me in general, I can't think of a better solution either. Feel free to merge.

ntamas (approved):

PR #1917, opened 2022-01-05
Fix system plfit cmake detection

antonio-rojas:
The headers are installed under a pflit subdir by default, so the current cmake module can't find them.


ntamas (approved):

ntamas:
Thanks for the fix!

PR #1921, opened 2022-01-06
benchmarks: plfit folder: specify

jgmbenoit:
This patch specify the folder of the `plfit` headers
 for the benchmark `igraph_power_law_fit.c`.

### DIFF
--- a/tests/benchmarks/igraph_power_law_fit.c
+++ b/tests/benchmarks/igraph_power_law_fit.c
@@ -1,5 +1,5 @@
 #include <igraph.h>
-#include <plfit_sampling.h>
+#include <plfit/plfit_sampling.h>
 
 #include "bench.h"
 


szhorvat:
I don't think if this is necessary. Can you please check if things work correctly after the patch in #1917?

szhorvat:
I now verified that an external PLFIT is correctly used after #1917. The correct fix was in that PR, and this change breaks using the vendored PLFIT. I'm closing this now, please let us know if #1917 isn't sufficient.

jgmbenoit:
The fix is needed when you build benchmarks outside the source and with an external `plfit`.
A more proper way might be to add a `Requires` field in the `igraph.pc` (see `pkg-config(1)`  section METADATA FILE SYNTAX).

szhorvat:
Do you mean that you are not building the benchmarks using igraph's build system and the `build_benchmarks` target? How are you building them then?

jgmbenoit:
I put the benchmarks along the other tests as examples. This allows to run bencmarks durint the CI tests.

jgmbenoit:
I am aware of the possible mismatches at building time.
I guess that I am attached to how GSL manages its headers. GSL gathers its headers in a `gsl` folder  and  calls them with a `gsl/` prefix so that we do not have to specify the include path in ``traditional'' installations. igraph and plfit gather both their headers in a folder but they call their headers with no prefix so that we have to specify the include path even in ``traditional`` installations.

szhorvat:
Could you please show what exactly is broken, through a code example? Then we can find a solution for it without breaking other stuff.

For reference, this is how I tested it:

 - Install PLFIT with `-DCMAKE_INSTALL_PREFIX=/my/prefix`. Then the headers go into `/my/prefix/include/plfit`.
 - Configure igraph with `-DCMAKE_PREFIX_PATH=/my/prefix`.
 - Build the benchmarks using `cmake --build . --target build_benchmarks`

szhorvat:
> I put the benchmarks along the other tests as examples. This allows to run benchmarks during the CI tests.

I'm sorry, I didn't see this message when I responded.

You can run the benchmarks by building the `benchmark` target, i.e. `make benchmark` if you use `make`. The benchmarks, at this point, are more or less just a rough development tool. They are not streamlined like the tests and they don't even work on all platforms (Windows). If you want to run them, it's best to use `make` instead of `ninja` because of the immediate output.

Does this help?

The benchmarks need a lot more work before they can be made part of some automated CI run, such as extracting and summarizing the timings. It would be nice to do this, there just wasn't enough time yet.

Issue #1922, opened 2022-01-06
Test that graph readers and writers are consistent


szhorvat:
It would be good to have a set of tests that verify that graph writers and readers are consistent:

 - Create a graph (with attributes and vertex names).
 - Write it to a file.
 - Read it back from the file.
 - If the format does not naturally preserve the vertex ordering, then reorder vertices to have the vertex names in alphabetical order (`igraph_permute_vertices()`).
 - Compare that the graph read back from the file, as well as its attributes, are the same as the graph originally written out. See `igraph_is_same_graph()`, and when checking edge attributes, canonicalize edge ordering.

szhorvat:
I suggest adding these tests on `develop` only. I made several robustness fixes on `develop` recently, and now it's difficult to port fixes between `master` and `develop`.

NCOL is a good one to start with because it is simple and has no tests so far. On `develop`, I changed the NCOL writer to preserve edge ordering. This can be tested too.

With formats that do not preserve the edge ordering, I think we can proceed like this in verifying that the edge attributes match: (1) match up vertex ordering between the graphs as discussed above (2) iterate through edges as `IGRAPH_EDGEORDER_FROM`.  I _think_ this would make the edge ordering predictable assuming that there are no multi-edges. If there are multi-edges, there's no way to match up edges between the two graphs other than through attributes.

szhorvat:
This was partially done in fuzz targets that I added recently.

Issue #1930, opened 2022-01-15
`permute_vertices` is confusing: it uses the inverse permutation


szhorvat:
**tl;dr**

 - `permute_vertices` uses the inverse of the permutation it is given to permute vertices.
 - The documentation of `permute_vertices` is incorrect, in all interfaces.
 - The way `permute_vertices` works is confusing, inconvenient, and inconsistent with other parts of the library.

----

### The problem

Regarding `igraph_permute_vertices()`, in [the Python docs](https://igraph.org/python/api/latest/igraph._igraph.GraphBase.html#permute_vertices) it says:

> Permutes the vertices of the graph according to the given permutation and returns the new graph.

In R,

> permutation | A numeric vector giving the permutation to apply.

**But that's not what it does. It actually applies the inverse permutation.**

In Mathematica, I use a separate implementation instead of `igraph_permute_vertices()`, so I won't discuss that.

**Note:** In this issue, I will use 1-based indexing.

Consider this graph. I marked both names and indices on the vertices:

<img width="360" alt="image" src="https://user-images.githubusercontent.com/1212871/149617630-1740ef93-5fc6-42dd-b08d-6cf64ad44a69.png">

If we were to apply the permutation `(2, 3, 1)`, we would get:

<img width="360" alt="image" src="https://user-images.githubusercontent.com/1212871/149617677-aa13709e-c48f-4fb5-b242-e5e56ef2e3c3.png">

In a high-level interface, this could be expressed as either of these:

 - Permutation to apply to vertices: Apply the permutation `2, 3, 1`. These integers can also be interpreted as the vertex indices in the original graph.
 - New name order: I want the new vertex order to be `b, c, a`.

The original order is `a,b,c`. Applying the permutation `2,3,1` gives the new order `b,c,a`. The inverse permutation of `2,3,1` is `3,1,2`. Applying this to the new order returns the original order.

However, what `permute_vertices()` does is _not_ apply a permutation, as I showed here. Instead, if applies the inverse. This is not what the R and Python docs say. What about the C docs? They are somewhat more accurate, but also a bit confused. They say:

> permutation The permutation to apply.

This is very misleading. But then it goes on:

> Vertex 0 is mapped to the first element of the vector, vertex 1 to the second, etc. 

So, if the input is `3,1,2` (note: 1-based), then the "mapping" is as

```
name  old_index  new_index
a        1         3
b        2         1
c        3         2
```

[...]

PR #1944, opened 2022-01-22
add NCOL read/write consistency test

GroteGnoom:
Fixes part of #1922


szhorvat:
Could we compare the graph that was read back with the original using `igraph_is_same_graph()`?  In order to do so, we will need to order the vertices identically. We can ensure identical vertex order using the following canonicalization process:

 1. Find the permutation that puts the vertex names in lexicographic order
 2. [Invert that permutation ...](https://github.com/igraph/igraph/issues/1930)  `igraph_vector_int_order` will effectively do the inversion.
 3. Use `igraph_permute_vertices()` to reorder the graph vertices

After having done this both for the original graph, as well as the graph that was read back, we can compare them using `igraph_is_same_graph()`.

GroteGnoom:
Could you explain why? Because it's more readable than a graphml? Or could there be information in there that we're not seeing now?

szhorvat:
I asked for this type of test after I modified the NCOL writer to write edges in a different order and introduced a bug, associating weights with the wrong edges (the bug was never pushed). I am looking for a test which can detect this type of bug.

This modification, even if bug-free, would change the GraphML output here. Then the output would need to be verified _manually_, diminishing the usefulness of the test.

I was looking for a test that verifies that the graph does not change after writing it to a file and reading it back again. The test should fail if and only if the graph that is read back is not identical to the original graph. It should not fail if, for example, the edge or vertex ordering changes.

szhorvat:
> The test should fail if and only if the graph that is read back is not identical to the original graph. It should not fail if, for example, the edge or vertex ordering changes.

This is really what I'm looking for. "Identical" here means having the same edges and vertices, with correct attribute assignments. Any solution that achieves this is fine for me.

GroteGnoom:
I've replaced the graphml write with the is_same_graph, but then checking for attributes separately doesn't seem easy. To compare edge attributes, you need to know which edge in graph1 corresponds to which edge in graph2. is_same_graph uses the internal representation of graphs for that.

Also doesn't is_same_graph still fail if the vertex ordering changes? It uses an ordered list of edges, but they're ordered by the second vertex.

szhorvat:
Could you rebase on current develop and make sure that the test still passes? This doesn't quite do everything I wanted, but I think we can improve it later. It would be useful to merge it as it is.

szhorvat:
I changed it to `abort()` instead of `exit(1)`. The reason is that if the program exits normally using `exit()`, without freeing all memory, LSan will complain about memory leaks. If it exits with `abort()`, it will be considered an abnormal exit, and there won't be distracting messages about memory leaks.

Issue #1932, opened 2022-01-16
Convenience function to check if a graph is acyclic


szhorvat:
**What is the feature or improvement you would like to see?**

A function to check if a graph is acyclic.

This is simple to implement: check if the graph is directed and a DAG (`igraph_is_dag()`) or if it is undirected and a forest (`igraph_is_forest()`).

_This issue is targeted at beginners. See [the wiki](https://github.com/igraph/igraph/wiki) on how to get started._

**Use cases for the feature**

Checking if a graph is acyclic is a common operation, so it is useful to have a single convenience function for it.

borsgeorgica:
could I work on this ?

GroteGnoom:
Sure!

szhorvat:
I would also like to bring to your attention these two issues: #1931 and #1585. They may be easier to get started with, and they also give you more freedom in doing things your own way. Unlike library code, examples do not need to be perfect.

If you still prefer to work on this issue though, you are welcome to do so.

szhorvat:
@borsgeorgica Let us know if you decided to go ahead with this issue.

Useful resources to look at before getting started:

 - Installation and Tutorial sections in the docs: https://igraph.org/c/html/latest/
 - Several beginner focused sections of the wiki: https://github.com/igraph/igraph/wiki

borsgeorgica:
@szhorvat thanks for linking the resources! sorry, I didn't get a chance to start on the issue yet but will do soon. I also looked at the other issues you suggested, I will give this a go first and maybe try the other ones if this is too challenging to start with.

borsgeorgica:
@szhorvat you mentioned `igraph_is_forest()`, is that something that's already implemented in `igraph` ? I looked in the manual and code but couldn't find it. I could see only `igraph_is_dag()`.

GroteGnoom:
It was added here: https://github.com/igraph/igraph/pull/1888 to the develop branch.

Issue #1882, opened 2021-12-08
random_walk() should take weights


szhorvat:
`igraph_random_edge_walk()` takes weights, but `igraph_random_walk()` does not. `igraph_random_walk()` should also be extended to take a `weights` argument. 

The underlying weighted implementation of `random_walk()` can mirror `random_edge_walk()` but it should be separate from the unweighted one. This is because the unweighted one uses an adjlist, while the weighted one needs an inclist. At this time, inclist-based traversal is slower (see benchmarks in #1823).

Notes:

 - We have a benchmark for random walks, keep an eye on it when implementing this.
 - Discussed here before: https://github.com/igraph/igraph/pull/1046#issuecomment-355291543

----

_While this is marked as a "good first issue", it is a bit more involved than other such issues. If you decide to work on this, mirror the implementation in `igraph_random_edge_walk()`, which already supports weights._

mp0963:
Hello, I am new to igraph and I would love to try working on this. Is this still open?

meetpatel0963:
Hello, I am new to igraph and I would love to try working on this. Is this still open?

szhorvat:
Yes, it is open, but note the last comment in italics. If you feel confident, go ahead and open a PR.  There's some info on getting started here: https://github.com/igraph/igraph/wiki/Quickstart-for-new-contributors

szhorvat:
`IGRAPH_OTHER` is slower than `IGRAPH_TO`/`IGRAPH_FROM`, so it's better to use the correct version.  Note that `IGRAPH_OTHER` contains an unpredictable conditional, which can potentially affect performance (see https://stackoverflow.com/q/11227809/695132 for an example).

Now, performance is a complicated topic, and it's unclear how great the practical performance difference will be. But I prefer that you keep the switch here, both for performance and clarity.

meetpatel0963:
oh, got it. I'll do the same in `igraph_random_walk()`. Also, Thank you for sharing this example. It was an amazing piece of info 🙌.

szhorvat:
See #2013.

PR #1945, opened 2022-01-22
Add convenience function to check if a graph is acyclic

borsgeorgica:
Issue https://github.com/igraph/igraph/issues/1932

### DIFF
--- a/src/properties/trees.c
+++ b/src/properties/trees.c
@@ -619,25 +619,9 @@ igraph_error_t igraph_is_forest(const igraph_t *graph, igraph_bool_t *res,
  * vertices and edges in the original input graph.
  */
 igraph_error_t igraph_is_acyclic(const igraph_t *graph, igraph_bool_t *res) {
-    igraph_bool_t dag, forest;
-
     if (igraph_is_directed(graph)) {
-        // check if it's dag
-        IGRAPH_CHECK(igraph_is_dag(graph, &dag));
-        if (dag) {
-            // it's acyclic
-            *res = 1;
-            return IGRAPH_SUCCESS;
-        }
+        return igraph_is_dag(graph, res);
     } else {
-        // check if it's forest - need to figure out roots
-        IGRAPH_CHECK(igraph_is_forest(graph, &forest, NULL, IGRAPH_ALL));
-        if (forest) {
-            *res = 1;
-            return IGRAPH_SUCCESS;
-        }
+        return igraph_is_forest(graph, res, NULL, IGRAPH_ALL);
     }
-
-    *res = 0;
-    return IGRAPH_SUCCESS;
 }
--- a/src/properties/trees.c
+++ b/src/properties/trees.c
@@ -22,6 +22,7 @@
 */
 
 #include "igraph_structural.h"
+#include "igraph_topology.h"
 
 #include "igraph_adjlist.h"
 #include "igraph_constructors.h"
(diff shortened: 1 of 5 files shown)


src/properties/trees.c
```diff
@@ -601,3 +602,42 @@ igraph_error_t igraph_is_forest(const igraph_t *graph, igraph_bool_t *res,
...
+        // check if it's forest - need to figure out roots
+        IGRAPH_CHECK(igraph_is_forest(graph, &forest, NULL, IGRAPH_ALL));
+        if (forest) {
+            *res = 1;
+            return IGRAPH_SUCCESS;
+        }
+    }
+
+    *res = 0;
+    return IGRAPH_SUCCESS;
```
szhorvat, line 642: These two lines are not reachable, so they should be removed.

szhorvat:
There is actually a more concise implementation, that makes it very clear that execution can never reach the end of the function:

```
if (igraph_is_directed(graph)) {
    return igraph_is_dag(graph, res);
} else {
    return igraph_is_forest(graph, res, NULL, IGRAPH_ALL);
}
```

szhorvat:
So far it looks good! Could you please add a test, which covers at least the following cases?

 - null graph
 - singleton graph
 - both directed an undirected cases
 - be sure that graphs which contain self-loops are not considered acyclic, in either the directed or undirected case
 - try to think of special cases where things might go wrong, and tests them; self-loops were the example I could think of

borsgeorgica:
> There is actually a more concise implementation, that makes it very clear that execution can never reach the end of the function:
> 
> ```
> if (igraph_is_directed(graph)) {
>     return igraph_is_dag(graph, res);
> } else {
>     return igraph_is_forest(graph, res, NULL, IGRAPH_ALL);
> }
> ```

this is much nicer! thanks

szhorvat:
When you are done, mark this PR as "ready for review". I looked through it and everything looks fine to me.

Others may want to comment on the location where this function is included in the docs.

szhorvat (approved):

ntamas:
The PR seems okay to me as well; I'll merge it and we can still move the section around in the docs if we find a better place for it.

For instance, there's an entire chapter about "graph cycles" (https://igraph.org/c/html/latest/igraph-Cycles.html), and it barely has any content. Wouldn't it be better to put this function there, along with maybe the entire "Topological sorting, directed acyclic graphs" section and `igraph_girth`?

ntamas:
Thanks for your contribution @borsgeorgica !

szhorvat:
@borsgeorgica Indeed, thanks for the PR!

One of the goals of issue #1932 was as a gentle introduction to igraph internals. Now you have some familiarity with error codes, documentation, and perhaps you also saw how memory management works. If you are interested in continuing to work with igraph, now you are ready for more interesting contributions. You can look through the existing issues, make a proposal for a feature, or simply tell us what you are looking for and we may be able to suggest tasks for which no GitHub issue has been written up yet.

szhorvat:
@ntamas Can you add this to `functions.yaml` before we forget?

borsgeorgica:
Thank you for giving me the opportunity to work on the issue and approving!

I would love to work on other issues as well, it would be nice to learn more about the library, get a better understanding of some of the graph algorithms and get more used to writing C. I was also looking at [this issue](https://github.com/igraph/igraph/issues/1882), would this be a good issue for me to do next? or shall I look at others? Also, I would be more than happy to work on any other problems that don’t have an issue open on Github yet.

GroteGnoom:
#1882 seems fine 🙂 There shouldn't be any surprising difficulties, and you already know how to add a function, so go ahead!

PR #1946, opened 2022-01-27
Add examples for assortativity functions

### LINKED ISSUE #1931: Add an example for igraph_assortativity
Start from the `develop` branch and add to examples/simple:
- igraph_assortativity.c, which calls the function igraph_assortativity (in /src/misc/mixing.c) with some graph and prints out the result,
- igraph_assortativity.out which contains the correct printed result.

Then, add the example to tests/CMakeLists.txt and to the docs of the function in /src/misc/mixing.c 

For an example of a simple example program: see examples/simple/igraph_small.c

The same should done for igraph_assortativity_nominal and igraph_assortativity_degree

- [ ] igraph_assortativity
- [ ] igraph_assortativity_nominal
- [x] igraph_assortativity_degree (https://github.com/igraph/igraph/pull/1946)

This issue is meant specifically for newcomers, and we'll gladly guide you through the process if you need any help :slightly_smiling_face:

Here's some more info about assortativity: http://networksciencebook.com/chapter/7#introduction7

----

Note: This is a part of #1585.

valdaarhun:
Add examples showing usage of assortativity related functions to `examples/simple`.

Should close #1931 once completed.


examples/simple/igraph_assortativity.c
```diff
@@ -0,0 +1,32 @@
...
+                 6, 4, 6, 7,
+                 7, 8, 7, 9,
+                 8, 9,
+                 9, 0,
+                 -1);
+
+    igraph_vector_init(&types, 0);
+    igraph_degree(&g, &types, igraph_vss_all(), IGRAPH_ALL, /* No self-loops */ 0);
+
+    igraph_assortativity(&g, &types, 0, &assortativity, /* ignore edge directions */ 0);
```
GroteGnoom, line 25: NULL instead of 0 is usually a bit clearer for pointers.

examples/simple/igraph_assortativity.c
```diff
@@ -0,0 +1,32 @@
+/* -*- mode: C -*-  */
```
szhorvat, line 1: Please remove this. Remember that the example is included in the documentation, and is aimed at all readers, including those who don't use emacs.

examples/simple/igraph_assortativity.c
```diff
@@ -0,0 +1,32 @@
...
+                 6, 4, 6, 7,
+                 7, 8, 7, 9,
+                 8, 9,
+                 9, 0,
+                 -1);
+
+    igraph_vector_init(&types, 0);
+    igraph_degree(&g, &types, igraph_vss_all(), IGRAPH_ALL, /* No self-loops */ 0);
+
+    igraph_assortativity(&g, &types, NULL, &assortativity, /* ignore edge directions */ 0);
```
szhorvat, line 25: Please use `IGRAPH_UNDIRECTED` and `IGRAPH_DIRECTED` instead of `0` and `1` to demonstrate this manner of improving code readability.

examples/simple/igraph_assortativity.c
```diff
@@ -0,0 +1,32 @@
...
+                 3, 6,
+                 4, 5,
+                 6, 4, 6, 7,
+                 7, 8, 7, 9,
+                 8, 9,
+                 9, 0,
+                 -1);
+
+    igraph_vector_init(&types, 0);
+    igraph_degree(&g, &types, igraph_vss_all(), IGRAPH_ALL, /* No self-loops */ 0);
```
szhorvat, line 23: It seems to me that in your example, excluding self-loops does not serve any purpose. If you check the "time complexity" section of the `igraph_degree` documentation, you will see that removing self-loops has a performance cost.  It's best for examples to illustrate good practices (even when performance doesn't matter _for this specific example_) so I suggest using `1`.

(You can also use `IGRAPH_LOOPS` instead of `1`, which is self-explanatory.)

examples/simple/igraph_assortativity.c
valdaarhun: I have changed the appropriate 0s to NULLs.

examples/simple/igraph_assortativity.c
valdaarhun: Sorry, I didn't think of that. I'll remove it.

szhorvat:
This is a good start that demonstrates the function's usage for a programmer.

Now let's think if we can make this example more educational. Can you come up with an example that is conceptually interesting, that perhaps is educational for a student of network science (and doesn't just illustrate an API usage)? This is your chance to be creative.

Have you worked with the concept of assortativity before?

szhorvat:
This is a useful resource: http://networksciencebook.com/chapter/7

examples/simple/igraph_assortativity.c
valdaarhun: I have changed the numbers to their respective macros.

examples/simple/igraph_assortativity.c
valdaarhun: I have changed it to `IGRAPH_LOOPS`.

szhorvat: Feel free to click "Resolve conversation" whenever you address one of these comments.

valdaarhun:
> Have you worked with the concept of assortativity before?

No. But I am interested in learning more about it.

szhorvat:
First take a look a the chapter I linked, and if you need tips on what to show in the example, just let us know.

valdaarhun:
> First take a look a the chapter I linked, and if you need tips on what to show in the example, just let us know.

Sure, I'll let you know if I have issues or get an idea. Thank you.

examples/simple/igraph_assortativity.c
```diff
@@ -0,0 +1,30 @@
+#include <igraph.h>
+#include <stdio.h>
+
+int main(){
+    igraph_t g;
+    igraph_vector_t types;
+    igraph_real_t assortativity;
+
+/* Create directed graph */
```
szhorvat, line 9:
```suggestion
    /* Create directed graph */
```

valdaarhun:
> First take a look a the chapter I linked, and if you need tips on what to show in the example, just let us know.

Hi. I read through the chapter. I have an idea of an example that probably is aimed at a beginner from a programmer's and network science student's perspective. I was thinking of demonstrating how graphs with identical degree sequences can have completely different assortativities.

szhorvat:
> I was thinking of demonstrating how graphs with identical degree sequences can have completely different assortativities.

Excellent idea!

Do you need tips on how to efficiently produce two graphs with the same degrees but different assortativity?

examples/simple/igraph_assortativity.c
```diff
@@ -0,0 +1,30 @@
...
+                 3, 6,
+                 4, 5,
+                 6, 4, 6, 7,
+                 7, 8, 7, 9,
+                 8, 9,
+                 9, 0,
+                 -1);
+
+    igraph_vector_init(&types, 0);
+    igraph_degree(&g, &types, igraph_vss_all(), IGRAPH_ALL, /* Consider self-loops */ IGRAPH_LOOPS);
```
szhorvat, line 21: tiny nitpick, just for consistency

```suggestion
    igraph_degree(&g, &types, igraph_vss_all(), IGRAPH_ALL, /* consider self-loops */ IGRAPH_LOOPS);
```

If you like, you can click the "merge" button here, then `git pull` on your side

valdaarhun:
> Do you need tips on how to efficiently produce two graphs with the same degrees but different assortativity?

Would it be possible to efficiently produce the graphs using `igraph_realize_degree_sequence`?

szhorvat:
A more efficient, and also fairly standard way is to use `igraph_rewire()`. This is what I suggest. The number of rewiring steps can be a few times the edge count.

valdaarhun:
> A more efficient, and also fairly standard way is to use `igraph_rewire()`.

Alright. I'll take a look at the usage of this function. Thanks.

valdaarhun:
I have changed assortativity.c so that it demonstrates how graphs with identical degree sequences can have different assortativities.

szhorvat:
As for the parameters of `igraph_barabasi_game()`, use:

 - undirected
 - Experiment with values of `m=1,2,3,...` to get a nice result with clearly changing assortativity
 - `power=1` for simplicity
 - `A` can be zero for undirected; this is simplest
 - `outpref` is irrelevant for undirected
 - `algo` should be `IGRAPH_BARABASI_PSUMTREE` for simple graphs and good performance

valdaarhun:
> The 5 result all look very different, but why? This is because the data (the graph) is tiny, so we can't get meaningful statistics. If the graph were large, all 5 results would be very similar.

Sorry, I didn't understand this properly. Why does this happen? Is it only related to the number of nodes/edges or is it also related to how the rewiring is done?

szhorvat:
Take 5 random numbers between 0 and 1, and compute the mean. The mean will fluctuate wildly, values of 0.2 or 0.7 are not uncommon. Now take 1000 random numbers between 0 and 1, and you'll find that the mean is always very close to 0.5. It's the same effect. The graph is too small, so there is high fluctuation.

valdaarhun:
> Take 5 random numbers between 0 and 1, and compute the mean. The mean will fluctuate wildly, values of 0.2 or 0.7 are not uncommon. Now take 1000 random numbers between 0 and 1, and you'll find that the mean is always very close to 0.5. It's the same effect. The graph is too small, so there is high fluctuation.

Got it. This makes sense now.

valdaarhun:
>     * Create a large graph whose assortativity differs from the typical value for graphs having the same degrees. I suggest you use `igraph_barabasi_game()` in undirected mode. This function implements the preferential attachment model, connecting low-degree vertices to high-degree ones. Thus, it has large negative assortativity. Print this out.
> 
>     * Rewire a single time, and print the assortativity again. It will be significantly different from the original value. If the graph is not tiny, then repeated rewirings will not change the assortativity by much. Rewiring gives the "typical" assortativity of graphs with these specific degrees. This experiment shows that the preferential attachment graph is not typical (is has much more negative assortativity).

I have understood what this means. Before working in the `igraph_assortativity.c` file, I thought I would play around with the function in a test program and experimented with the various options (number of nodes, power, m, A).

The assortativity before rewiring comes out to be in the (-0.05, -0.02) range while after rewiring, their assortativities tend to be around -0.005. Are these values appropriate or should the inital assortativity be closer to -1?

valdaarhun:
Here's an example.

```
#include <igraph.h>
#include <stdio.h>

int main(){
    igraph_t g;
    igraph_integer_t nodes = 100000, A = 0, power = 1, m = 1;
    
    igraph_barabasi_game(&g, nodes, power, m, NULL, IGRAPH_UNDIRECTED, A, IGRAPH_UNDIRECTED, IGRAPH_BARABASI_PSUMTREE, NULL);
    
    igraph_integer_t edge_count = igraph_ecount(&g);
    printf("Number of edges = %d\nnodes = %d\nA = %d\npower = %d\nm = %d\n\n", edge_count, nodes, A, power, m); 
    
    igraph_vector_t degree;
    igraph_vector_init(&degree, 0); 
    igraph_degree(&g, &degree, igraph_vss_all(), IGRAPH_ALL, IGRAPH_LOOPS);

    igraph_real_t assortativity;
    igraph_assortativity(&g, &degree, NULL, &assortativity, IGRAPH_UNDIRECTED);
    printf("Assortativity of undirected graph before rewiring = %g\n\n", assortativity);

    for(int i = 0 ; i < 5 ; i++){
        igraph_rewire(&g, 10 * edge_count, IGRAPH_REWIRING_SIMPLE);
        igraph_assortativity(&g, &degree, NULL, &assortativity, IGRAPH_UNDIRECTED);
        printf("Assortativity of undirected graph = %g\n", assortativity);
    }   

    igraph_vector_destroy(&degree);
    printf("\n");

    igraph_destroy(&g);
}
```

Output:

```
Number of edges = 99999
nodes = 100000
A = 0
power = 1
m = 1

Assortativity of undirected graph before rewiring = -0.0204148

Assortativity of undirected graph = -0.00814841
Assortativity of undirected graph = -0.00855524
Assortativity of undirected graph = -0.00742923
Assortativity of undirected graph = -0.00699636
Assortativity of undirected graph = -0.00691494
```
```

**Comment by ntamas** [comment]
@valdaarhun Actually, I think your result is correct and the expectation that BA networks should have a large negative assortativity (in terms of absolute value) is not correct; see, e.g., [this paper](https://appliednetsci.springeropen.com/articles/10.1007/s41109-019-0152-1):

> Concerning the assortativity properties of BA networks, in previous work some estimates of the Newman coefficient r were found ([Newman 2002](https://appliednetsci.springeropen.com/articles/10.1007/s41109-019-0152-1#ref-CR14)). According to these estimates, for large N (number of nodes), r vanishes as −ln(2N/N). It was therefore generally believed that BA networks are almost uncorrelated, and numerical simulations appeared to confirm this. However, more recent asymptotic estimates ([Fotouhi and Rabbat 2018](https://appliednetsci.springeropen.com/articles/10.1007/s41109-019-0152-1#ref-CR10); [Bertotti and Modanese 2019](https://appliednetsci.springeropen.com/articles/10.1007/s41109-019-0152-1#ref-CR5)) yield a different result: r vanishes only as −ln(2N / sqrt(N)) for large N. It should be recalled that for real networks with the same scale-free exponent (γ=3), the r coefficient is always small in absolute value, so even this small total disassortativity is significant.

@szhorvat Given the above, how shall we proceed with the example?

**Comment by szhorvat** [comment]
If I understand this right, the paper claims that the assortativity of BA networks approaches zero as the vertex count increases. 

This, however, does not exclude the possibility that they are clearly more disassortative than a typical network with the same degrees.

Let's generate many BA networks with `m=1`, rewire each, and compare the assortativity histograms between the original and rewired networks. This is what we get:

100 vertices:

image

1000 vertices:

image

10000 vertices:

image

For 10000-vertex graphs, rewiring is almost guaranteed to make the graph less disassortative. I think this can still make a nice demonstration: generate a BA network, rewire it, print both assortativities. Repeat 5 times to show that rewiring consistently reduces disassortativity.

----

If we wanted strongly assortative/disassortative networks, we could use `igraph_realize_degree_sequence` with the "largest first" and "smallest first" methods. However, I would prefer to keep this for an example program associated with `igraph_realize_degree_sequence` (which you could easily do after this one @valdaarhun).

**Comment by valdaarhun** [comment]
> which you could easily do after this one @valdaarhun

I would love to work on that after this.

**Comment by szhorvat** [comment]
I just realize that this example is not run in the CI tests (and in fact there is an error in it which the compiler should have warned you about). Please add it to `test/CMakeLists.txt` within the `add_examples` just below `# structural_properties.at`.

**Comment by szhorvat** [comment]
> > which you could easily do after this one @valdaarhun
> 
> I would love to work on that after this.

Yes, that will be great.

**Comment by valdaarhun** [comment]
> and in fact there is an error in it which the compiler should have warned you about

I believe you are talking about the usage of `igraph_vector_t` instead of `igraph_vector_int_t` in `igraph_degree()`. That's the warning that I got after running the test.

I checked the docs for the 'develop' branch and noticed that `igraph_degree()` uses `igraph_vector_int_t` while `igraph_assortativity()` uses `igraph_vector_t`. So I instead used `igraph_strength()` (the weights for all the edges being equal to 1).

I am not sure if this method is appropriate for resolving the warnings.

**Comment by szhorvat** [comment]
I have not forgotten about this PR. I wanted to clean up the assortativity functionality (and its documentation) before proceeding with examples. This is happening in #1953, and will soon be completed.

**Comment by valdaarhun** [comment]
Sure, no problem. Let me know when I can continue.

File: examples/simple/igraph_assortativity.c [hunk 7]
```diff
@@ -0,0 +1,35 @@
...
+    for (int i = 0; i < 5; i++) {
+        igraph_real_t assortativity;
+
+        /* Generate undirected graph with 100000 nodes */
+        igraph_barabasi_game(&g, nodes, power, m, NULL, IGRAPH_UNDIRECTED, A, IGRAPH_UNDIRECTED, IGRAPH_BARABASI_PSUMTREE, /* start from */ NULL);
+
+        /* Get degree sequence */
+        igraph_strength(&g, &degree, igraph_vss_all(), IGRAPH_ALL, /* consider self-loops */ IGRAPH_LOOPS, /* no weights */ NULL);
+       
+        igraph_assortativity(&g, &degree, NULL, &assortativity, /* ignore edge directions */ IGRAPH_UNDIRECTED);
```
szhorvat, line 22: Let us change this to use `igraph_assortativity_degree()` and make this into an example for that function instead. Then we can get rid of `igraph_strength()` completely.

I am concerned that if people see this example, in its current form, they will believe that this is the usual way to compute degree assortativity (the most common use case). In fact, they should be using `igraph_assortativity_degree()` for that task, as it's more convenient.

szhorvat (changes requested):
The example looks good. Only a few small changes need to be made, then we can move on to the next PR.

 - Please merge the `develop` branch into your branch. I made a few changes there, including adding a new argument to assortativity functions.
 - Let's change to `igraph_assortativity_degree()`.
 - When you are done with the above two points, include the example in the documentation of `igraph_assortativity_degree()` by adding an `\example examples/simple/yourfilename.c` line at the end of the documentation block.

valdaarhun:
The PR is ready for review.

szhorvat:
Would you like to add an example for `igraph_realize_degree_sequence()` now?

I would suggest the following:

 - Create a random graph with a similar number of vertices and edges
 - Take its degrees
 - Create two graphs from this degree sequence, one with the "largest first" and one with the "smallest first" methods
 - Print the degree assortativity of the three graphs
 - Add comments in the example, pointing out that one method generates high-assortativity, the other low-assortativity graphs

szhorvat:
It would be nice to add an example for `igraph_assortativity_nominal()` as well:

 - Use one of the graph generators that creates networks based on "vertex types". For some, like the "stochastic block model", you effectively input the types. For others, like "establishment game" or "preference game", the types are generated. Choose one you like an make sure you have the "types" (= categories) in a vector.
 - Show that the nominal assortativity is high based on these types.

Feel free to open PRs directly for any of these. Now that you see how adding examples works, everything should go much faster.

valdaarhun:
I enjoyed working on this issue. I would like to work on the other two issues as well.

szhorvat:
Thanks for the work you did so far! Feel free to open PRs for these two issues as well.

PR #1948, opened 2022-01-29
Typedef igraph_error_t to igraph_error_type_t

### LINKED ISSUE #1867: igraph_error_t should be an enum
`igraph_error_t` should be an enum, now it's a typedef for an `int`.

Be sure to test with C++.

GroteGnoom:
To fix #1867

### DIFF
--- a/src/cliques/cliquer/cliquer.h
+++ b/src/cliques/cliquer/cliquer.h
@@ -28,7 +28,7 @@ struct _clique_options {
 				 clique_options *);
 	FILE *output;
 
-	boolean (*user_function)(set_t,graph_t *,clique_options *);
+	igraph_error_t (*user_function)(set_t,graph_t *,clique_options *);
 	void *user_data;
 	set_t *clique_list;
 	int clique_list_length;
--- a/tests/unit/single_target_shortest_path.c
+++ b/tests/unit/single_target_shortest_path.c
@@ -26,11 +26,10 @@
 #include "test_utilities.inc"
 
 void igraph_warnings_ignore(const char *reason, const char *file,
-                            int line, int igraph_errno) {
+                            int line) {
     IGRAPH_UNUSED(reason);
     IGRAPH_UNUSED(file);
     IGRAPH_UNUSED(line);
-    IGRAPH_UNUSED(igraph_errno);
     /* Do nothing */
 }
 
--- a/tests/unit/single_target_shortest_path.c
+++ b/tests/unit/single_target_shortest_path.c
@@ -25,14 +25,6 @@
 
 #include "test_utilities.inc"
 
-void igraph_warnings_ignore(const char *reason, const char *file,
-                            int line) {
-    IGRAPH_UNUSED(reason);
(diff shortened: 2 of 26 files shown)


src/cliques/cliquer/cliquer.h
```diff
@@ -28,7 +28,7 @@ struct _clique_options {
 				 clique_options *);
 	FILE *output;
 
-	boolean (*user_function)(set_t,graph_t *,clique_options *);
```
GroteGnoom, line 31: This one I'm not completely sure of. This changes the interface, but we were putting errors in a boolean.

GroteGnoom:
https://github.com/igraph/igraph/runs/4996866883?check_suite_focus=true

@ntamas It seems like every return value is assumed to be an int in stimulus. Can that be fixed? I can't find where that happens. Is it because of C implicit declarations returning an int?

tests/unit/single_target_shortest_path.c
```diff
@@ -26,11 +26,10 @@
 #include "test_utilities.inc"
 
 void igraph_warnings_ignore(const char *reason, const char *file,
```
szhorvat, line 28: We already have `igraph_warning_handler_ignore()`. Can such tests simply use it instead of providing their own?

ntamas:
@GroteGnoom 

> It seems like every return value is assumed to be an int in stimulus. Can that be fixed? 

The default _abstract_ return type in Stimulus is `ERROR`. `ERROR` is mapped to `int` in `types.yaml`, so you can change it there.

src/cliques/cliquer/cliquer.h
ntamas: I think Cliquer's `user_function` is a private API so we can do whatever we want with it.

ntamas:
Looks good to me -- feel free to merge after addressing the minor outstanding comments.

src/cliques/cliquer/cliquer.h
szhorvat: As far as igraph is concerned it's private, yes. Diverging from Cliquer is a bit uncomfortable, but the Cliquer author did say that he is not planning new releases.

szhorvat:
Is there a compiler warning which is triggered when assigning a non-enum value to an enum variable? If yes, maybe it's worth activating to see if any issues are revealed.

szhorvat:
> Is there a compiler warning which is triggered when assigning a non-enum value to an enum variable? If yes, maybe it's worth activating to see if any issues are revealed.

The only warning option I found is `-Wc++-compat` for GCC (not Clang). However, this triggers so many unrelated warnings (which I can't seem to disable separately) that it is not useful.

Let's not bother with this.

tests/unit/single_target_shortest_path.c
GroteGnoom: Yes. I also looked for other cases, but those were printing handlers, and those made sense because igraph's own handlers print line numbers, which is not nice in tests.

ntamas:
Thanks a lot!

PR #1951, opened 2022-01-30
Add igraph_blas_dgemm

GroteGnoom:


src/linalg/blas.c
```diff
@@ -82,6 +82,85 @@ igraph_error_t igraph_blas_dgemv(igraph_bool_t transpose, igraph_real_t alpha,
...
+    k = (int) ncol_oa;
+    n = (int) ncol_ob;
+    lda = (int) igraph_matrix_nrow(a);
+    ldb = (int) igraph_matrix_nrow(b);
+    ldc = (int) igraph_matrix_nrow(c);
+
+
+#ifdef HAVE_GFORTRAN
+    igraphdgemm_(&trans_a, &trans_b, &m, &n, &k, &alpha, VECTOR(a->data),
+                 &lda, VECTOR(b->data), &ldb, &beta, VECTOR(c->data), &ldc, /*trans_len*/ 1);
```
ntamas, line 155: Here you need `..., &ldc, /*trans_a_len*/ 1, /*trans_b_len*/ 1` at the end. `gfortran` has this weird calling convention when bridging C and Fortran such that you need to pass the _lengths_ of string arguments at the _end_ of the argument list. Since you have two string arguments here, you need to pass the lengths twice as well.

ntamas: This is needed for the R interface only at the moment.

src/linalg/blas.c
```diff
@@ -82,6 +82,85 @@ igraph_error_t igraph_blas_dgemv(igraph_bool_t transpose, igraph_real_t alpha,
...
+    if (ncol_oa != nrow_ob) {
+        IGRAPH_ERROR("Number of columns of Matrix A not equal to "
+                "number of rows of matrix B (after possible transpositions of A and B).", IGRAPH_EINVAL);
+    }
+    if (beta && ncol_oa != igraph_matrix_ncol(c)) {
+        IGRAPH_ERROR("Number of columns of Matrix A not equal to "
+                "number of columns of matrix C (after possible transposition of A).", IGRAPH_EINVAL);
+    }
+    if (beta && nrow_oa != igraph_matrix_nrow(c)) {
+        IGRAPH_ERROR("Number of rows of Matrix A not equal to "
```
ntamas, line 132: Can you add a test case in the unit test that covers this branch? It seems like this one was missed.

ntamas (changes requested):

CHANGELOG.md
```diff
@@ -544,6 +544,7 @@
  - `igraph_is_forest()` to check whether a graph is a forest (#1888, thanks to @rohitt28)
  - `igraph_is_acyclic()` to check whether a graph is acyclic (#1945, thanks to @borsgeorgica)
  - `igraph_es_all_between()` to create an edge selector that selects all edges between a pair of vertices.
+ - `igraph_blas_dgemm()` to multiply to matrices.
```
szhorvat, line 547:
```suggestion
 - `igraph_blas_dgemm()` to multiply two matrices.
```

src/linalg/blas.c
```diff
@@ -82,6 +82,85 @@ igraph_error_t igraph_blas_dgemv(igraph_bool_t transpose, igraph_real_t alpha,
...
+ * size k x m.
+ *
+ * \return \c IGRAPH_EOVERFLOW if the matrix is too large for BLAS,
+ *         \c IGRAPH_EINVAL if the matrices have incompatible sizes,
+ *         \c IGRAPH_SUCCESS otherwise.
+ *
+ * \example examples/simple/blas_dgemm.c
+ */
+igraph_error_t igraph_blas_dgemm(igraph_bool_t transpose_a, igraph_bool_t transpose_b,
+        igraph_real_t alpha, const igraph_matrix_t* a, const igraph_matrix_t* b,
```
szhorvat, line 113: Could we please write pointers as `type *var` instead of `type* var`, as in the rest of the codebase?

I always find `type* var` a bit misleading because `type* a, b` in fact declares only `a` as a pointer, not `b`. So the `*` attached to the var name, not the type name.

ntamas: I was wondering whether we should simply start using `clang-format`, set it up the way we like it, and then set our editors to call it upon saving. I started doing this in other C projects (in fact, I also added a pre-commit hook that pipes all changed source files through `clang-format`) and it works great and there is no need to take extra care when formatting code.

src/linalg/blas.c
```diff
@@ -82,6 +82,85 @@ igraph_error_t igraph_blas_dgemv(igraph_bool_t transpose, igraph_real_t alpha,
...
+    char trans_b = transpose_b ? 'T' : 'N';
+    int m, n, k, lda, ldb, ldc;
+    igraph_integer_t nrow_oa = transpose_a ? igraph_matrix_ncol(a) : igraph_matrix_nrow(a);
+    igraph_integer_t ncol_oa = transpose_a ? igraph_matrix_nrow(a) : igraph_matrix_ncol(a);
+    igraph_integer_t nrow_ob = transpose_b ? igraph_matrix_ncol(b) : igraph_matrix_nrow(b);
+    igraph_integer_t ncol_ob = transpose_b ? igraph_matrix_nrow(b) : igraph_matrix_ncol(b);
+
+    if (ncol_oa != nrow_ob) {
+        IGRAPH_ERROR("Number of columns of Matrix A not equal to "
+                "number of rows of matrix B (after possible transpositions of A and B).", IGRAPH_EINVAL);
```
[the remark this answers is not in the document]
szhorvat: This comment applies both here and elsewhere below. "Matrix A" is capitalized where it should not be.

If you still prefer naming then A, B, C, maybe you can spell out the formula as 

> Matrices A and B have incompatible dimensions for computing A*B + C.

src/linalg/blas.c
```diff
@@ -82,6 +82,85 @@ igraph_error_t igraph_blas_dgemv(igraph_bool_t transpose, igraph_real_t alpha,
...
+ * can be transposed.
+ *
+ * \param transpose_a whether to transpose the matrix \p a
+ * \param transpose_b whether to transpose the matrix \p b
+ * \param alpha       the constant \c alpha
+ * \param a           the matrix \c a
+ * \param b           the matrix \c b
+ * \param beta        the constant \c beta
+ * \param c           the matrix \c c. The result will also be stored here.
+ *                    If beta is zero, c wil be resized to fit the result.
```
szhorvat, line 101:
```suggestion
 *                    If beta is zero, c will be resized to fit the result.
```

src/linalg/blas.c
```diff
@@ -82,6 +82,85 @@ igraph_error_t igraph_blas_dgemv(igraph_bool_t transpose, igraph_real_t alpha,
...
+ *
+ * Time complexity: O(nmk) where matrix a is of size n x k, and matrix b is of
+ * size k x m.
```
szhorvat, line 104:
```suggestion
 * Time complexity: O(n m k) where matrix a is of size n × k, and matrix b is of
 * size k × m.
```

src/linalg/blas.c
```diff
@@ -82,6 +82,85 @@ igraph_error_t igraph_blas_dgemv(igraph_bool_t transpose, igraph_real_t alpha,
     return IGRAPH_SUCCESS;
 }
 
+/**
+ * \function igraph_blas_dgemm
+ * \brief Matrix-matrix multiplication using BLAS.
+ *
+ * This function is a somewhat more user-friendly interface to
+ * the \c dgemm function in BLAS. \c dgemm performs the operation
+ * y = alpha*a*b + beta*c, where a, b and c are matrices, of which a and b
```
szhorvat, line 91: Should we say `c := alpha*a*b + beta*c`?

ntamas (approved):

ntamas:
@szhorvat My concerns have been addressed so I leave it up to you to merge it if you are happy with it.

PR #1786, opened 2021-05-03
Cycle basis calculations

szhorvat:
Fixes #1252

PR #1957, opened 2022-02-10
Cycle basis calculations

szhorvat:
Continuing from #1786 

This PR implements some cycle basis calculations.

 - Fundamental cycle basis
 - Unweighed minimum cycle basis

It has a special feature where it can limit the length of cycles found, which drastically reduces computation time. This will be a unique feature of igraph.

Within the minimum cycle basis computation, I represent cycles as sorted vectors of edge IDs. This makes certain operations, such as "adding" cycles, efficient. However, in the end we obtain the cycles with an edge ID ordering that does not match the cycle structure. Therefore, in a final step we order the cycles. This operation was much easier with a map data structure, so I did it in C++. I could do it in pure C without writing a map from scratch, but it would be much more complicated, and I don't have the time now. There is an option that controls whether this reordering should be done. Setting it to `false` will speed the function up, though I did not yet time to benchmark by how much.

This PR also adds a `igraph_vector_list_remove_consecutive_duplicates` function.


szhorvat:
Please DO NOT MERGE, just give feedback. I will merge myself later. At first, I expected technical comments, as understanding the algorithm would take time. I discussed the algorithm with @vtraag last year, but I am not sure if he remembers.

include/igraph.h
```diff
@@ -99,4 +99,6 @@
 #include "igraph_eulerian.h"
 #include "igraph_graphicality.h"
 
+#include "igraph_cycles.h"
```
ntamas, line 102: There isn't really any structure in `igraph.h` so I'd just add this header without an empty line between them and the others. The way I see it, there are three groups of headers in this file: very-basic-lowlevel-stuff, templated types and all the rest. This one belongs in the "all the rest" section.

include/igraph_cycles.h
```diff
@@ -0,0 +1,27 @@
...
+IGRAPH_EXPORT igraph_error_t igraph_fundamental_cycles(
+        const igraph_t *graph,
+        igraph_integer_t start_vid,
+        igraph_integer_t cutoff,
+        igraph_vector_int_list_t *result);
+
+IGRAPH_EXPORT igraph_error_t igraph_minimum_cycle_basis(
+        const igraph_t *graph,
+        igraph_integer_t cutoff,
+        igraph_bool_t complete,
```
ntamas, line 21: I was wondering whether we should accumulate these two booleans into a "flags" parameter where individual bits correspond to these booleans. The advantage would be that we could simply design the flags in a way that 0 means the "default" settings, i.e. the ones that a user is likely to use the most frequently.

include/igraph_cycles.h
```diff
@@ -0,0 +1,27 @@
...
+#include "igraph_types.h"
+#include "igraph_vector_list.h"
+
+__BEGIN_DECLS
+
+IGRAPH_EXPORT igraph_error_t igraph_fundamental_cycles(
+        const igraph_t *graph,
+        igraph_integer_t start_vid,
+        igraph_integer_t cutoff,
+        igraph_vector_int_list_t *result);
```
ntamas, line 16: We haven't agreed on the policy for the placement of the `result` argument. I vaguely recall that in most functions the ordering is: the graph comes first, then the output arguments, then all the input arguments. It is my subjective perception only, though. My point is only that we should probably try to stick to whatever is predominant now in the library.

src/core/typed_list.pmt
```diff
@@ -951,6 +951,45 @@ igraph_error_t FUNCTION(sort_ind)(
     return IGRAPH_SUCCESS;
 }
 
+/**
+ * \ingroup vector_list
+ * \function igraph_vector_list_remove_consecutive_duplicates
+ * \brief Removes consecutive duplicates from a vector list.
+ *
+ * Takes an unsorted list \p v as input and computes an array of
+ * indices \p inds such that v[ inds[i] ], with i increasing from 0, is
```
ntamas, line 960: It looks like this is not true any more. The current implementation seems to simply compare consecutive items and removes duplicates in-place. The documentation needs to be updated. We can mention that this actually removes all duplicates if the list is pre-sorted first.

src/misc/cycle_bases.c
```diff
@@ -0,0 +1,507 @@
+/*
+   IGraph library.
+   Copyright (C) 2021  The igraph development team <igraph@igraph.org>
```
ntamas, line 3: This should now be 2021-2022

src/misc/cycle_bases.c
```diff
@@ -0,0 +1,507 @@
...
+
+#include "core/interruption.h"
+#include "misc/order_cycle.h"
+
+/**** Fundamental cycles *****/
+
+/* Computes fundamental cycles for the connected component containing 'start_vid',
+ * and appends them to 'result'.
+ *
+ * 'visited' must be a vector of length vertex_count.
```
ntamas, line 36: Is `vertex_count` the number of vertices in the _whole_ graph or in the component containing `start_vid` only?

src/misc/cycle_bases.c
```diff
@@ -0,0 +1,507 @@
...
+
+    if (start_vid < 0 || start_vid >= no_of_nodes) {
+        IGRAPH_ERROR("Invalid starting vertex id.", IGRAPH_EINVAL);
+    }
+
+    IGRAPH_VECTOR_INT_INIT_FINALLY(&pred_edge, no_of_nodes);
+    IGRAPH_VECTOR_INT_INIT_FINALLY(&u_back, 0);
+    IGRAPH_VECTOR_INT_INIT_FINALLY(&v_back, 0);
+
+    igraph_dqueue_int_init(&q, 0);
```
ntamas, line 67: Shouldn't we wrap this in an `IGRAPH_CHECK()`?

src/misc/cycle_bases.c
```diff
@@ -0,0 +1,507 @@
...
+    IGRAPH_FINALLY(igraph_dqueue_int_destroy, &q);
+
+    igraph_dqueue_int_push(&q, start_vid); /* vertex id */
```
ntamas, line 70: `IGRAPH_CHECK()` needed here and in the next line.

src/misc/cycle_bases.c
```diff
@@ -0,0 +1,507 @@
...
+
+                IGRAPH_CHECK(igraph_vector_int_list_push_back(result, &cycle));
+                IGRAPH_FINALLY_CLEAN(1); /* pass ownership of 'cycle' to 'result' */
+            } else {
+                /* encountering u for the first time, queue it for processing */
+
+                /* Only queue vertices with distance at most 'cutoff' from the root. */
+                /* Negative 'cutoff' indicates no cutoff. */
+                if (cutoff < 0 || vdist < cutoff) {
+                    igraph_dqueue_int_push(&q, u);
```
ntamas, line 153: `IGRAPH_CHECK()` needed here and in the next line.

src/misc/cycle_bases.c
```diff
@@ -0,0 +1,507 @@
...
+ *
+ * </para><para>
+ * Edge directions are ignored. Multi-edges and self-loops are supported.
+ *
+ * \param graph The graph object.
+ * \param start_vid If negative, a complete fundamental cycle basis is returned.
+ *   If a vertex ID, the fundamental cycles associated with the BFS tree rooted
+ *   in that vertex will be returned, only for the weakly connected component
+ *   containing that vertex.
+ * \param cutoff If negative, a complete cycle basis is returned. Otherwise, only
```
ntamas, line 189: Could we define this parameter in a way that it represents the maximum length of the cycles we are looking for? This would be more intuitive from the end user's point of view.

ntamas: Especially in higher-level interfaces where the autocompletion of the IDE might show you that you have a "cutoff" parameter and then you simply assume "oh, so it's the maximum length of the cycles we are looking for".

src/misc/cycle_bases.c
```diff
@@ -0,0 +1,507 @@
...
+        IGRAPH_FINALLY(igraph_inclist_destroy, &inclist);
+
+         /* TODO: estimate space to reserve. 'rank' is a lower bound only. */
+        IGRAPH_VECTOR_INT_LIST_INIT_FINALLY(&candidates, 0);
+        IGRAPH_CHECK(igraph_vector_int_list_reserve(&candidates, rank));
+
+        mark = 0;
+        for (i=0; i < no_of_nodes; ++i) {
+            igraph_integer_t degree = VECTOR(degrees)[i];
+            igraph_bool_t vis = VECTOR(visited)[i] % 3;
```
ntamas, line 422: Explain where the `% 3` comes from. I understand, but it's better to be explicit, especially because we are talking about "vertices of degree 3 or greater" one line below, and this 3 has nothing to do with that 3.

Also note that with this trick, we are limiting ourselves to processing graphs where `no_of_nodes <= IGRAPH_INTEGER_MAX / 3`, otherwise the markers start to wrap around in `visited`. It's not a problem for me, just let us be aware of it.

include/igraph.h
szhorvat: Of course, I actually meant fixed this before asking for a review.

include/igraph_cycles.h
szhorvat: I have no preference. As I recall, @vtraag suggested the output arg to come last, that's why it's last here. Let me know where you want it and I can easily refactor it.

src/core/typed_list.pmt
szhorvat: Thanks, copy-paste mistake!

include/igraph.h
szhorvat: How about merging `igraph_eulerian.h` and `igraph_cycles.h`? Both deal with "cycles" in some sense.

src/misc/cycle_bases.c
szhorvat: Note to self: implement overflow checks.

src/core/typed_list.pmt
szhorvat: Allowing `NULL` is actually not easily possible because the same framework is used for lists of `igraph_t` as well. That case is different enough that I wouldn't want to allow a default interpretation just yet.

src/misc/cycle_bases.c
szhorvat: I did think about this of course, and it would look more natural to use a cutoff on cycle length. However, I'm not sure that's a good idea right now. That is part of the reason why I want to make the API `\experimental`.

This is feature is completely novel, as far as I am aware. No other graph library implements it, nor have I seen it described. It gives a massive speedup, effectively descreasing the exponent by one in the complexity of the algorithms, when the graph satsifies certain conditions. I think it's better to be explicit about how it works, i.e. that it is the BFS depth. Also, we can only limit the cycle lengths to odd numbers.

include/igraph_cycles.h
ntamas: Okay then, mark it as `\experimental` and then we are free to change it later. I just don't like booleans in argument lists in general and even less if there is more than one of them :)

include/igraph_cycles.h
ntamas: I'll make a quick survey in the functions of the library to see where `res` and `result` are typically placed and then get back to you.

src/misc/cycle_bases.c
ntamas: Then let's call it `bfs_depth_limit` for the time being and then it's explicit that it's not the cycle length that is being limited here.

include/igraph.h
ntamas: `igraph_eulerian.h` also deals with Eulerian _paths_, doesn't it? (But we also have `igraph_paths.h` :)).
Let's keep things as is, just move `igraph_cycles.h` together with the rest of the group above.

include/igraph_cycles.h
ntamas: See the result on our Slack channel

szhorvat:
Where feedback might be useful is an efficient algorithm to put edges in cycle order.

Currently, the `minimum_cycle_basis()` function produces cycles as sorted lists of edge IDs. This is actually sufficient for a surprisingly large number of applications, but in many cases users will expect the edge IDs to be ordered along the cycle instead. To do this, I used `std::map` from C++, hence the separate C++ source file. I'm not very happy with this, but it was the fastest way to handle this issue.

How the algorithm works is that it produces BFS-based fundamental cycles starting from all vertices (of degree d >= 3). This produces a list of candidate cycles of length `(E - V + 1)*V`, where E and V are edge and vertex counts. I.e., it produces a pretty huge list. These cycles are originally in the correct order, but they are sorted by edge ID for further processing. What we would do is to duplicate the original list in memory, and keep handles to the unsorted versions of cycles from the sorted one. However, this would complicate the code quite a bit (couldn't use `vector_list` anymore), and would more than double the already significant memory requirements. So I don't like it. This is why I opted to restore the cycle order on the final result instead.

szhorvat:
Yet another thing where I need feedback is a refactoring for `IGRAPH_HANDLE_EXCEPTIONS(...)`. Now the argument is a large block of code, but this causes problems.  Notice the separate `typedef std::map<igraph_integer_t, eid_pair_t> inclist_t;` I have in that file. Why do I need that? Because writing `map<igraph_integer_t, eid_pair_t>` within `IGRAPH_HANDLE_EXCEPTIONS(...)` confuses the C preprocessor with that comma, and just won't work.

We probably want a begin/end style solution for `IGRAPH_HANDLE_EXCEPTIONS`, but I want to see concrete suggestions please.

ntamas:
Added a begin/end style solution as the first iteration. I think it's good enough. Another alternative would have been to move the function body to a separate function and just wrap a call to _that_ function within `IGRAPH_HANDLE_EXCEPTIONS()`.

src/core/exceptions.h
```diff
@@ -8,12 +8,23 @@
...
+#define IGRAPH_HANDLE_EXCEPTIONS_BEGIN \
+    try {
+#define IGRAPH_HANDLE_EXCEPTIONS_END \
+    } \
     catch (const std::bad_alloc &e) { IGRAPH_ERROR(e.what(), IGRAPH_ENOMEM); /* LCOV_EXCL_LINE */ } \
     catch (const std::exception &e) { IGRAPH_ERROR(e.what(), IGRAPH_FAILURE); } \
     catch (...) { IGRAPH_ERROR("Unknown exception caught.", IGRAPH_FAILURE); }
 
+#define IGRAPH_HANDLE_EXCEPTIONS(code) \
+    IGRAPH_HANDLE_EXCEPTIONS_BEGIN; \
```
szhorvat, line 26: This `;` is a bit strange here as it expands to `try {;`.

ntamas:
> Where feedback might be useful is an efficient algorithm to put edges in cycle order.

Do you mean that the input of the algorithm is a graph and a list of edge IDs in arbitrary order that are guaranteed to form a cycle, and the result should be a list where the edge IDs are in the order they are traversed along the cycle? I don't see any particular problem with the current implementation.

src/core/exceptions.h
ntamas: feel free to remove the `;`, it has no effect. I just wanted to test that one is allowed to use `IGRAPH_HANDLE_EXCEPTIONS_BEGIN` or `_END` with or without a semicolon at the end without breaking thigs.

szhorvat:
Let me clarify.

Here's an example of a graph with edges labelled by ID:

image

3, 4, 8, 5 form a cycle, in this order. However, the computation returns the edge ID in sorted order, 3, 4, 5, 8. I use a separate function, `igraph_i_order_cycle()`, to put them back in cycle order, i.e. 3, 4, 8, 5.  This function is written in C++, not C, because it uses `std::map`.

I found this a bit ugly.

include/igraph_cycles.h
szhorvat: Done.

src/misc/cycle_bases.c
szhorvat: Renamed to `bfs_cutoff` (keep it short).

ntamas (approved):

PR #1962, opened 2022-02-12
Expose float comparison with epsilon functions

### LINKED ISSUE #1956: Expose igraph_almost_equals and igraph_cmp_epsilon
`igraph_almost_equals` and `igraph_cmp_epsilon` should be exposed as public functions. The return value of `igraph_almost_equals` should be an `igraph_bool_t`. They should also be documented.

When you do this, remember to remove `#include "core/math.h"` from relevant tests.

GroteGnoom:
Fixes #1956

### DIFF
--- a/src/math/utils.c
+++ b/src/math/utils.c
@@ -267,8 +267,8 @@ int igraph_is_neginf(double x) {
  *         error is defined as <code>abs(a-b) / (abs(a) + abs(b))</code>. The
  *         two numbers are considered equal if this is less than \c eps.
  *
- * \return Nonzero if the two floats are nearly equal to each other within
- *         the given level of tolerance, zero otherwise.
+ * \return True if the two floats are nearly equal to each other within
+ *         the given level of tolerance, false otherwise.
  */
 igraph_bool_t igraph_almost_equals(double a, double b, double eps) {
     return igraph_cmp_epsilon(a, b, eps) == 0 ? 1 : 0;
--- a/src/math/utils.c
+++ b/src/math/utils.c
@@ -22,6 +22,7 @@
 */
 
 #include "igraph_types.h"
+#include "igraph_nongraph.h"
 
 #include "core/math.h"
 
@@ -269,7 +270,7 @@ int igraph_is_neginf(double x) {
  * \return nonzero if the two floats are nearly equal to each other within
  *         the given level of tolerance, zero otherwise
  */
-int igraph_almost_equals(double a, double b, double eps) {
+igraph_bool_t igraph_almost_equals(double a, double b, double eps) {
     return igraph_cmp_epsilon(a, b, eps) == 0 ? 1 : 0;
 }
 
--- a/src/math/utils.c
+++ b/src/math/utils.c
@@ -256,19 +256,19 @@ int igraph_is_neginf(double x) {
 
 /**
  * \function igraph_almost_equals
- * Compare two double-precision floats with a tolerance
+ * \brief Compare two double-precision floats with a tolerance.
  *
  * Determines whether two double-precision floats are "almost equal"
  * to each other with a given level of tolerance on the relative error.
  *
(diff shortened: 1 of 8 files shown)


szhorvat:
Looks good to me. Some small comments:

 - I am ambivalent about exposing `IGRAPH_SHORTEST_PATH_EPSILON`. What do others think?
 - Can you add the `\brief`s to the docs, and add a fullstop at the end?
 - There's a `\c` in one of the docs followed by a long quoted expression.  I think this won't render properly, and we'll need `<code>` instead. Can you check this please?

ntamas:
> There's a \c in one of the docs followed by a long quoted expression. I think this won't render properly,

You are correct, it won't. It works in Doxygen but not in our dumb parser; the doc syntax that we use was introduced when we migrated from Doxygen to Docbook and we wanted to save us the hassle of re-writing the doc markup, but it does not handle everything that Doxygen does.

> I am ambivalent about exposing IGRAPH_SHORTEST_PATH_EPSILON. What do others think? On the one hand, I'd say that this is an internal implementation detail, which shouldn't be exposed. On the other hand, it does make sense to document the threshold we use

I think that the usage of _some_ epsilon is a necessity due to floating-point comparisons, and the user should be aware that there is such a threshold. At that point, it does not hurt to document and expose what we actually use.

GroteGnoom:
> Do you mind if I edit the docs a bit and change "floats" to "real numbers"?

I do wonder why 🙂 They are way more float than real numbers. I do see more people do it, and of course we have igraph_real_t, but it does seem wrong to me. Reals are much different than floats.

szhorvat:
- Well "float" is not even a proper word. It is a colloquialism used by some programmers. igraph's main audience is scientists, not programmers.
 - There is a chance for confusion with the `float` type of C.
 - The concept here applies to real numbers in the mathematical sense, not merely to their floating-point representation.
 - igraph uses the terminology "real", not "float". And it's not alone in this, so do several programming languages such as Fortran or Mathematica.

src/math/utils.c
```diff
@@ -255,21 +256,21 @@ int igraph_is_neginf(double x) {
...
+ * \param  a  The first float.
+ * \param  b  The second float.
+ * \param  eps  The level of tolerance on the relative error. The relative
+ *         error is defined as <code>abs(a-b) / (abs(a) + abs(b))</code>. The
  *         two numbers are considered equal if this is less than \c eps.
  *
- * \return nonzero if the two floats are nearly equal to each other within
- *         the given level of tolerance, zero otherwise
+ * \return Nonzero if the two floats are nearly equal to each other within
+ *         the given level of tolerance, zero otherwise.
```
szhorvat, line 271: Let's say true and false here. The return type is not a Boolean (it used to be an integer).

GroteGnoom:
> * Well "float" is not even a proper word. It is a colloquialism used by some programmers. igraph's main audience is scientists, not programmers.
Maybe I'm missing some context here. Do you think there are a lot of people who use the C-part of igraph who don't know what floats are? Also, I thought especially mathematicians would balk at saying they're reals. They're a limited set of rational numbers with some extra things like NaN on top.

>     * There is a chance for confusion with the `float` type of C.
Ok, but that's why it says double-precision a few times, and the types are right there in the function. Couldn't people also be confused by thinking they're actually infinite precision?

>     * The concept here applies to real numbers in the mathematical sense, not merely to their floating-point representation.
But it's not a conceptual mathematical function and it deals with floating point numbers. I thought these functions were mainly used because of issues that would come up with floats, not reals.

>     * igraph uses the terminology "real", not "float". And it's not alone in this, so do several programming languages such as Fortran or Mathematica.
I know, and I don't fully know why. I thought for igraph it was just an easy way to not say float or double but leave it open which one you mean without finding another word that means floating point number. C uses 'float', why use Mathematica as an example instead of C?

Issue #1960, opened 2022-02-11
Swap method is inconsistent between vector and vector_list


szhorvat:
For `vector`, the method that swaps two elements is called `swap_elements()`. The method that swaps two vectors is called `swap()`, but is is badly implemented (O(n) complexity instead of O(1) and requires vectors to have the same lengths).

For `vector_list`, the method that swaps two elements is called `swap()`. There is no method to swap two vector lists.

Suggestion:

 - The naming should be made consistent between `vector` and `vector_list`
 - The current `vector_swap()` function should either be removed (it's never used, performs badly, and is unnecessarily restrictive), or it should be made O(1) and should be provided both for `vector` and `vector_list`.

ntamas:
Fixed in f9152e9

PR #1964, opened 2022-02-14
Reduced time complexity of igraph_vector_swap from O(n) to O(1).

araujo88:
This pull request references #1960 and reduces time complexity of igraph_vector_swap() from O(n) to O(1).

### DIFF
--- a/src/core/matrix.pmt
+++ b/src/core/matrix.pmt
@@ -910,6 +910,7 @@ igraph_error_t FUNCTION(igraph_matrix, cbind)(TYPE(igraph_matrix) *to,
  * The contents of the two matrices will be swapped.
  * \param m1 The first matrix.
  * \param m2 The second matrix.
+ * \return Error code.
  *
  * Time complexity: O(1).
  */
--- a/src/core/matrix.pmt
+++ b/src/core/matrix.pmt
@@ -907,19 +907,14 @@ igraph_error_t FUNCTION(igraph_matrix, cbind)(TYPE(igraph_matrix) *to,
  * \function igraph_matrix_swap
  * Swap two matrices.
  *
- * The contents of the two matrices will be swapped. They must have the
- * same dimensions.
+ * The contents of the two matrices will be swapped.
  * \param m1 The first matrix.
  * \param m2 The second matrix.
- * \return Error code.
  *
- * Time complexity: O(mn), the number of elements in the matrices.
+ * Time complexity: O(1).
  */
 
 igraph_error_t FUNCTION(igraph_matrix, swap)(TYPE(igraph_matrix) *m1, TYPE(igraph_matrix) *m2) {
-    if (m1->nrow != m2->nrow || m1->ncol != m2->ncol) {
-        IGRAPH_ERROR("Cannot swap non-conformant matrices", IGRAPH_EINVAL);
-    }
     return FUNCTION(igraph_vector, swap)(&m1->data, &m2->data);
 }
 
--- a/src/core/vector.pmt
+++ b/src/core/vector.pmt
@@ -2446,19 +2446,19 @@ igraph_error_t FUNCTION(igraph_vector, update)(TYPE(igraph_vector) *to,
 
 igraph_error_t FUNCTION(igraph_vector, swap)(TYPE(igraph_vector) *v1, TYPE(igraph_vector) *v2) {
 
-    TYPE(igraph_vector) *tmp;
+    TYPE(igraph_vector) tmp;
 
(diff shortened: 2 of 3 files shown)


src/core/vector.pmt
```diff
@@ -2438,30 +2438,28 @@ igraph_error_t FUNCTION(igraph_vector, update)(TYPE(igraph_vector) *to,
...
 
 igraph_error_t FUNCTION(igraph_vector, swap)(TYPE(igraph_vector) *v1, TYPE(igraph_vector) *v2) {
 
-    igraph_integer_t i, n1 = FUNCTION(igraph_vector, size)(v1);
-    igraph_integer_t n2 = FUNCTION(igraph_vector, size)(v2);
-    if (n1 != n2) {
-        IGRAPH_ERROR("Vectors must have the same number of elements for swapping.",
-                     IGRAPH_EINVAL);
-    }
+    TYPE(igraph_vector) *tmp;
```
GroteGnoom, line 2449: If you're going to make a temporary vector variable it can't be a pointer, because it points to nowhere, and then `tmp->stor_begin = v1->stor_begin;` will try to place v1->str_begin in some arbitrary place in memory.

src/core/vector.pmt
```diff
@@ -2446,19 +2446,19 @@ igraph_error_t FUNCTION(igraph_vector, update)(TYPE(igraph_vector) *to,
 
 igraph_error_t FUNCTION(igraph_vector, swap)(TYPE(igraph_vector) *v1, TYPE(igraph_vector) *v2) {
 
-    TYPE(igraph_vector) *tmp;
+    TYPE(igraph_vector) tmp;
 
-    tmp->stor_begin = v1->stor_begin;
-    tmp->stor_end = v1->stor_end;
-    tmp->end = v1->end;
+    tmp.stor_begin = (*v1).stor_begin;
```
GroteGnoom, line 2451: You can just leave the `(*v1).` to be `v1->`, that's the usual way of writing it 🙂

src/core/vector.pmt
```diff
@@ -2438,30 +2438,28 @@ igraph_error_t FUNCTION(igraph_vector, update)(TYPE(igraph_vector) *to,
  * \function igraph_vector_swap
  * \brief Swap elements of two vectors.
  *
- * The two vectors must have the same length, otherwise an error
- * happens.
  * \param v1 The first vector.
  * \param v2 The second vector.
- * \return Error code.
```
szhorvat, line 2445: Please don't remove this line.

src/core/vector.pmt
```diff
@@ -2438,30 +2438,28 @@ igraph_error_t FUNCTION(igraph_vector, update)(TYPE(igraph_vector) *to,
...
  *
- * Time complexity: O(n), the length of the vectors.
+ * Time complexity: O(1), the length of the vectors.
```
szhorvat, line 2444:
```suggestion
 * Time complexity: O(1).
```

szhorvat:
I just realized that to make this complete `matrix_swap` also needs to be updated (fix the complexity in the documentation, relax limitation of having to have the same size).

vtraag:
Hi @araujo88, I cleaned up some merge conflicts with the `develop` branch. There are still a few open code review comments that need to be addressed.

As @szhorvat  writes:

> Actually, couldn't we just write the swap in terms of `tmp` directly instead of their components?

This would simplify the implementation. It essentially comes down to this:

```
igraph_vector_t tmp;
tmp = (*v1);
(*v1) = (*v2);
(*v2) = tmp;
```

Since `igraph_vector_t` is just a `struct`, by assigning another struct to it (i.e. the struct pointed to by `v1`, so `*v1`), will assign all members of the `struct` (i.e. `stor_begin`, `stor_end` and `end` in this case). This way, we can be oblivious to the details of what are the exact members in the `igraph_vector_t` struct, and we wouldn't need to update this function in case that changes.

src/core/matrix.pmt
```diff
@@ -907,19 +907,14 @@ igraph_error_t FUNCTION(igraph_matrix, cbind)(TYPE(igraph_matrix) *to,
  * \function igraph_matrix_swap
  * Swap two matrices.
  *
- * The contents of the two matrices will be swapped. They must have the
- * same dimensions.
+ * The contents of the two matrices will be swapped.
  * \param m1 The first matrix.
  * \param m2 The second matrix.
- * \return Error code.
```
vtraag, line 914: Similarly here, please don't remove this line.

vtraag:
Thanks for the contribution @araujo88 ! If you are interested in some other issues, contributions are always welcome!

PR #1966, opened 2022-02-14
Ensure dominance for all labels in label propagation (fixes #1963)

### LINKED ISSUE #1963: community/label_propagation.c stops even if there is a vertex violating the goal condition
According to the original paper [1] describing the label propagation algorithm (the paper is also mentioned in label_propagation.c comment), the iterative process of label propagation is performed until _every node in the network has a label to which the maximum number of its neighbors belongs_, but the implementation [label_propagation.c ](https://github.com/igraph/igraph/blob/master/src/community/label_propagation.c) sometimes stops even if there is a vertex which has a different label than the maximum number of its neighbors.

If I understand it correctly, the problem is on lines 294-301 in the code, especially in the condition `if (VECTOR(label_counters)[(long)VECTOR(*membership)[v1]] != max_count)`. There is an example of wrong behavior:

Let's have a triangle graph with vertices A, B, and C. A and C have "red" labels, B has a different "green" label.

The currently processed vertex is A (`v1` on line 240). The number of green and red neighbors is the same. The condition `if (VECTOR(label_counters)[(long)VECTOR(*membership)[v1]] != max_count)` is false, because the number of neighbors with the same label as A (red) is equal to the maximum number of its neighbors with one label (both is 1), so `running` stays 0. The new label is chosen randomly because the number of green and red neighbors is the same. Let's say that green was chosen.

Now, the currently processed vertex is B. The number of green and red neighbors is the same, so it works the same as for A. `running` stays 0, the new label is chosen randomly, let's say that red was chosen.

Now, the currently processed vertex is C. The number of green and red neighbors is the same again. `running` stays 0 again and the new label is chosen randomly. In the end, there are two vertices with the same labels and one vertex with another label. The problem is, that the iteration through all vertices (line 239) is done and `running==0`, so the function finishes, but there is a vertex that does not have a label to which the maximum number of its neighbors belongs.

In a pseudo-code of the algorithm in the original paper, the control if every node in the network has a label to which the maximum number of its neighbors belongs is performed in step (v), after the relabeling in step (iv). Here, these two steps are merged together.

**To reproduce**

```python
import igraph as ig

# 300 was enough for me to see the problem several times, but since the algorithm is randomized, so it might be necessary to run it more times
for i in range(300):     
    G = ig.Graph()
    G.add_vertices(3)
    G.add_edges([(0,1),(1,2),(2,0)])
    
    partition = G.community_label_propagation()
    ig.plot(partition, "labelprop"+str(i)+".png") 
```

**Version information**
I am using igraph 0.9.9 running inside Python 3.8.10 (default, Nov 26 2021, 20:14:08), but I think that this is a problem in the algorithm in C code, not in Python interface, that  is the reason why I am not opening an issue in igraph/python-igraph.

The line numbers I am referring to are from commit fb5b8bed7133d53ff1ce4e5033206cf4d4f2a80b.

[1]  Raghavan, U.N. and Albert, R. and Kumara, S.: Near linear time algorithm to detect community structures in large-scale networks. Phys Rev E 76, 036106. (2007). <https://doi.org/10.1103/PhysRevE.76.036106>

Katterrina:
I do not want to make the code much slower or memory consuming. I tried to implement the following idea: To check if the end condition holds, we need to count neighbors with diverse labels for each vertex (or weights of edges to the neighbors in a weighted case). We need to do the same for choosing a new label. I tried to count this only once, store the chosen new labels and apply it only if there is a next iteration to do. 

It does not work, but I still hope I can figure out how to fix the problem. It has never finished for the unit tests right now. It could be because the updating process in my solution is synchronous, so there is a problem with oscillation -- it is described in the original paper that it could be a problem in a synchronous setting. (Or there might be some bug in my code, I do not have much experience in C programming.)


vtraag:
Thanks for taking a stab at this @Katterrina !

I think it would be better to stick to the asynchronous implementation of LPA, instead of switching to a synchronous implementation. What do you think @lovre?

You now seem to be taking the approach that you do one iteration over all nodes where you just check for dominance, and note move any label, correct? That generally sounds like a good direction to address the issue you identified in #1963.

When you are finished with your implementation, mark your PR as ready for review, and I'll take a closer look. Don't hesitate to ask any questions in the meantime!

Katterrina:
> You now seem to be taking the approach that you do one iteration over all nodes where you just check for dominance, and not move any label, correct?

Yes. I am not sure if it is the fastest solution, because it goes through all vertices twice, but I hope it works. I do not think the synchronous implementation was a good idea, but I did not realized the problems before I started programming.

I also modified the tests for LPA, but I am not sure with tests/unit/community_label_propagation2.out. The output for directed graph with unlabelled nodes not reachable from any labelled ones is a bit different now. I don't understand how it should work for directed graphs, so I can't check if it is correct.

tests/unit/community_label_propagation.c
```diff
@@ -32,7 +32,26 @@ int main() {
...
+    igraph_integer_t i, j;
+
+    /* Simple triangle graph, the output should be always one community */
+    igraph_small(&g, 0, IGRAPH_UNDIRECTED, 0,  1,  0,  2,  1,  2, -1);
+
+    for (j = 0; j < 100; j++) {
+        /* label propagation is a stochastic method */
+        igraph_rng_seed(igraph_rng_default(), j);
+
+        igraph_vector_int_init(&membership, 0);
```
vtraag, line 44: You now initialize the vector `membership` repeatedly in the for loop. A vector needs to be initialised only once. Could you please move this to outside the for loop (i.e. to L39)? Note that in L79 it now is also initialised still, that initialisation should be removed.

src/community/label_propagation.c
```diff
@@ -259,7 +260,9 @@ igraph_error_t igraph_community_label_propagation(const igraph_t *graph,
         igraph_vector_int_t *ineis;
         igraph_bool_t was_zero;
 
-        running = 0;
+        if (control_iteration) {
+            running = 0;
+        }
 
         /* Shuffle the node ordering vector */
         IGRAPH_CHECK(igraph_vector_int_shuffle(&node_order));
```
vtraag, line 268: You only need to shuffle the nodes when not doing a control iteration (the shuffling is immaterial for checking for dominance).

vtraag (commented):
Looks good to me overall! And good idea to reuse the exact same code, but then just for checking dominance. Could you perhaps add a few sentence in the implementation to explain the use of `control_iteration` and how you use it? That would make future maintenance easier.

There are few other minor items that should be corrected still, but that shouldn't be too difficult. See the specific points raised in the code review.

> I also modified the tests for LPA, but I am not sure with `tests/unit/community_label_propagation2.out`. The output for directed graph with unlabelled nodes not reachable from any labelled ones is a bit different now. I don't understand how it should work for directed graphs, so I can't check if it is correct.

Don't worry about this. This is just because you now call the RNG more often, meaning that the state of the RNG that is used to label the unreachable nodes is different at that point. However,  the shuffling of the nodes is unnecessary when you do the checking, so that is one point that you could improve still (see specific point in the code review).

Katterrina:
Thank you very much for your comments, I corrected it.

examples/simple/igraph_community_label_propagation.c
```diff
@@ -25,6 +25,7 @@ int main() {
     igraph_t graph;
     igraph_vector_int_t membership;
     igraph_real_t modularity;
+    igraph_integer_t i;
```
szhorvat, line 28: This will fix the CI failure:

```suggestion
```

src/community/label_propagation.c
```diff
@@ -337,6 +343,8 @@ igraph_error_t igraph_community_label_propagation(const igraph_t *graph,
             }
         }
         RNG_END();
+
+        control_iteration = 1 - control_iteration;
```
szhorvat, line 347: Let us not rely on Boolean values begin exactly 0 and 1.

```suggestion
        control_iteration = ! control_iteration;
```

src/community/label_propagation.c
```diff
@@ -319,15 +322,18 @@ igraph_error_t igraph_community_label_propagation(const igraph_t *graph,
...
-                /* Select randomly from the dominant labels */
-                k = RNG_INTEGER(0, igraph_vector_int_size(&dominant_labels) - 1);
-                k = VECTOR(dominant_labels)[k];
-                /* Check if the _current_ label of the node is also dominant */
-                if (VECTOR(label_counters)[(long)VECTOR(*membership)[v1]] != max_count) {
-                    /* Nope, we need at least one more iteration */
-                    running = 1;
+                if (control_iteration) {
+                    /* Check if the _current_ label of the node is also dominant */
+                    if (VECTOR(label_counters)[(long)VECTOR(*membership)[v1]] != max_count) {
```
szhorvat, line 327: This cast to `long` should not have been there in the original code (it was a bug). Let's remove it now.

```suggestion
                    if (VECTOR(label_counters)[VECTOR(*membership)[v1]] != max_count) {
```

szhorvat:
@vtraag Don't forget to update the changelog after merging.

If you have time, it would also be nice to:

 - Backport to `master` (I expect another 0.9 release). This was we can ship the fix in the next python-igraph 0.9.x release.
 - Update the unit test to use the `IGRAPH_ASSERT(a==b)` style instead of the `if (a != b) { return 1; }` style.  @Katterrina you could do this point if you like.

src/community/label_propagation.c
```diff
@@ -251,6 +251,7 @@ igraph_error_t igraph_community_label_propagation(const igraph_t *graph,
         IGRAPH_FINALLY(igraph_vector_int_destroy, &node_order);
     }
 
+    control_iteration = 1;
```
ntamas, line 254: Please add a comment here explaining what `control_iteration` means when it's 1 and what it means when it's zero. (I understand it by reading the code now, but I'm not sure I'd still remember it if I look at the same piece of code two months later).

ntamas:
@szhorvat @vtraag Thanks for taking care of this PR while I am away; feel free to merge if you feel that it's solid enough. I went through the commits now and I only had one minor comment.

src/community/label_propagation.c
```diff
@@ -259,10 +264,12 @@ igraph_error_t igraph_community_label_propagation(const igraph_t *graph,
         igraph_vector_int_t *ineis;
         igraph_bool_t was_zero;
 
-        running = 0;
-
-        /* Shuffle the node ordering vector */
-        IGRAPH_CHECK(igraph_vector_int_shuffle(&node_order));
+        if (control_iteration) {
+            running = 0;
```
vtraag, line 268: Could you also please put a comment here to explain what is done here?

src/community/label_propagation.c
```diff
@@ -337,6 +347,8 @@ igraph_error_t igraph_community_label_propagation(const igraph_t *graph,
             }
         }
         RNG_END();
+
+        control_iteration = !control_iteration;
```
vtraag, line 351: Could you also please put a comment here to say we are alternating between control iterations and label updating iterations?

vtraag (commented):
Great, thanks, it looks good to me! If you can just put a few more small comments in to explain what you're doing, I'll get it merged.

Katterrina:
Thank you for all the comments and suggestions!

vtraag (approved):

szhorvat:
I don't think this ever made it into the changelog ... if you still remember the details, can you add (with attribution) it for the section with the appropriate version @vtraag ?

vtraag:
Yes, you're right, corrected now.

szhorvat:
Thanks!

PR #1969, opened 2022-02-19
Make strvector more like a vector

### LINKED ISSUE #1965: igraph_strvector_t should have a reserve method
`igraph_strvector_t` should have a reserve method, similar to `vector_t`. `strvector_add()` (see #1959) should double the size of storage, just like `vector_push()` does, for better performance. There is currently a timeout issue on OSS-Fuzz which may or may not be related to this one.

At the same time, we should consider refactoring `strvector_t` to use a similar `stor_begin`, `stor_end`, `end` structure like `vector`. Currently, it only has a data pointer and a length, which does not allow for reserving.

GroteGnoom:
Trying to fix #1965

The first commit just changes the struct without doing anything useful. But the ability to have a storage size different from the size can later be used to have less allocations.

Todo: 
- check thoroughly (fix codecov warnings so you know changes are at least covered by tests (no, all codecov warnings are about allocation failures))
- add reserve function and let `stor_end` be different from `end`


szhorvat:
I'll put this here now, as later I'm going to forget it. It will also serve as a review checklist.

 - [ ] `reserve` method; all reallocation should go through it, like in `vector_t`
 - [ ] `push_back` method, `add` should be deprecated
 - [ ] protection against overflow, especially when doubling storage in push_back / add
 - [ ] functions that fail should leave the data structure in a consistent state; especially push_bash / add
 - [ ] `get` functions should return result directly

ntamas:
Tried to resolve conflicts with `develop` by taking your changes, except these two:

* `igraph_strvector_move_interval()` was removed
* `igraph_strvector_add()` was renamed to `_push_back()`; `_add()` was kept as a deprecated alias.

szhorvat:
Can you please add this to the top?

```
#include "internal/hacks.h" /* strdup */
```

I don't want to change this file and create more conflicts.

`strdump` is not in C99 and we have a replacement for it, but the relevant header is not included in all files that use this function.

src/core/strvector.c
```diff
@@ -21,13 +21,15 @@
 
 */
 
+#include <string.h>         /* memcpy & co. */
+#include <stdlib.h>
```
szhorvat, line 25: Normally we order headers from most specific to least specific, meaning that standard headers come last. This helps avoid introducing a bug where an igraph header only works if a standard header is included before it. This is the usual recommendation for C and C++.

szhorvat:
This is just a reminder that you can use `strdup` to duplicate strings instead of having to manually `strlen`, `IGRAPH_CALLOC`, `strcpy`. Remember to check the return value of `strdup`, though.

szhorvat:
Many of the standard library (or POSIX) string handling functions come in two variants: one assumes null-terminated strings, the other takes the string length explicitly (but still handles early null terminators). Compare `strcpy` vs `strncpy` or `strchr` vs `memchr`.

Some of the parser code could be simplified and also made more efficient if we had the same thing for adding new entries to `strvector`s. Now we have `add/push_back` which assume a null-terminator. It would be nice to add a version which takes the string length instead. 

The `flex` interface uses pointer + length pairs instead of null-terminated strings because it effectively works with one long stream of characters (file contents) of which it needs to take subsets. In order to use null-terminated strings, it would have to make copies of parts of this stream. Instead, it just produces pointers to parts of the stream, along with character counts.

GroteGnoom:
Some of the code considers that strings might be NULL pointers, but that shouldn't be possible when using the strvector interface, and if it is possible that's a bug that should be fixed. So I'll remove the code that checks for NULL pointer strings.

szhorvat:
I believe that some of the attribute handler code uses NULL pointers in strvector. I am not very familiar with the attribute handler as I don't use it in the Mathematica interface.

I don't like the idea of allowing NULL pointers though.

szhorvat:
One more thought:

There are some cases where `NULL`s should be tolerated though, notably in `igraph_strvector_destroy()`. This is so that partially initialized strvectors can be safely freed. However, this should work with no extra effort, as `free()` is safe to call with a `NULL` argument.

GroteGnoom:
> This is just a reminder that you can use strdup to duplicate strings instead of having to manually strlen, IGRAPH_CALLOC, strcpy. Remember to check the return value of strdup, though. Some of the existing code can be simplified with this.

That doesn't do the same as a REALLOC though. A strdup will always try to allocate memory, but with a realloc you might place the string at the same location.

szhorvat:
Yes, you're right. Don't replace realloc. But it can be used for `push_back`?

src/internal/hacks.c
```diff
@@ -43,6 +43,25 @@ char* igraph_i_strdup(const char *s) {
     return result;
 }
 
+/**
+ * Drop-in replacement for strndup.
+ * Used only in compilers that do not have strndup or _strndup
+ */
+char *igraph_i_strndup(const char *s1, size_t n) {
+    size_t i;
+    /* We need to check if the string is shorter than n characters.
```
szhorvat, line 52: Tip: I think `memchr` might be useful here.

szhorvat: But also: don't spend more time than necessary. Let's make things work before we make them pretty. A TODO note about memchr() might be good for later.

GroteGnoom:
igraph_strvector_permdelete is a public function which says it's `for internal use`, but it's not actually used internally. What should we do with it?

ntamas:
Remove it if it is not used anywhere.

src/core/strvector.c
```diff
@@ -370,60 +346,55 @@ void igraph_strvector_clear(igraph_strvector_t *sv) {
...
-        if (error) {
-            /* There was an error, free everything we've allocated so far */
-            for (j = 0; j < i; j++) {
-                if (v->data[v->len + i] != 0) {
-                    IGRAPH_FREE(v->data[v->len + i]);
+            sv->stor_begin[oldsize + i] = IGRAPH_CALLOC(1, char);
+            if (sv->stor_begin[oldsize + i] == NULL) {
+                /* LCOV_EXCL_START */
+                for (j = 0; j < i; j++) {
+                    IGRAPH_FREE(sv->stor_begin[oldsize + i]);
```
szhorvat, line 368: Did you mean the following?

```suggestion
                    IGRAPH_FREE(sv->stor_begin[oldsize + j]);
```

szhorvat (changes requested):
I had a quick peek and only found one real issue (see `i` vs `j`)

It would be nice to fix up the error message (consistent use of "strvector" or "string vector") as well as the docstrings, but these can come after merging this PR.

I am curious to see how this fares on OSS-fuzz.

I may have missed things, so it'd be good if @ntamas could take a look as well.

src/core/strvector.c
```diff
@@ -182,25 +177,22 @@ igraph_error_t igraph_strvector_set(igraph_strvector_t *sv, igraph_integer_t idx
  */
 igraph_error_t igraph_strvector_set2(igraph_strvector_t *sv, igraph_integer_t idx,
                           const char *value, size_t len) {
-    if (idx < 0 || idx >= sv->len) {
+    char *tmp;
+
+    if (idx < 0 || idx >= sv->stor_end - sv->stor_begin) {
         IGRAPH_ERROR("String vector index out of bounds.", IGRAPH_EINVAL);
     }
```
szhorvat, line 184: IMO this should be a fatal error (i.e. an `IGRAPH_ASSERT`) because we want this to fail hard in tests, and especially on OSS-fuzz. This is always a sign of a bug in the C code, so it should not be hidden when the error handler is set to "ignore".

szhorvat:
One more thing: since we now have `reserve` and `capacity`, we should probably have a `resize_min` as well.

szhorvat:
Can you explain in the docs of `_resize()` and `_init()` that new elements are `""`? I found code that does not assume that this is the case. It'll be easier to update these if the docs are very clear.

src/core/strvector.c
```diff
@@ -312,18 +294,18 @@ igraph_error_t igraph_strvector_append(igraph_strvector_t *to,
     igraph_bool_t error = 0;
     IGRAPH_CHECK(igraph_strvector_resize(to, len1 + len2));
     for (i = 0; i < len2; i++) {
-        if (from->data[i][0] != '\0') {
-            IGRAPH_FREE(to->data[len1 + i]);
-            to->data[len1 + i] = strdup(from->data[i]);
-            if (!to->data[len1 + i]) {
+        if (from->stor_begin[i][0] != '\0') {
+            IGRAPH_FREE(to->stor_begin[len1 + i]);
```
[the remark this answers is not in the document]
GroteGnoom: I think the code is wrong and it should be fixed 🙂 I will use a reserve instead of resize.

After that fix, there's another thing wrong here. This only copies strings if they're not empty. That seems like a strange exception, which I will remove.

szhorvat:
Regarding commit [remove superflulous allocations in append](https://github.com/igraph/igraph/pull/1969/commits/46618e36fabe5e8b752cb996368d15fb65a12d5f), this function is no longer correct. We need to clean up the vector in case of errors: free so-far allocated pieces and ensure that the original size is retained. Otherwise, the size has to be adjusted.

src/core/strvector.c
```diff
@@ -182,25 +177,22 @@ igraph_error_t igraph_strvector_set(igraph_strvector_t *sv, igraph_integer_t idx
  */
 igraph_error_t igraph_strvector_set2(igraph_strvector_t *sv, igraph_integer_t idx,
```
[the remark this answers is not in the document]
ntamas: I think `set_len` would be better. If we already had `set2` before, that should become a deprecated alias to `set_len`

ntamas:
I like `set_len` and `push_back_len` better because `push_back2` is pretty unclear about what the `2` means. I'll review this PR now and try to get the last bits sorted out.

szhorvat:
Make sure that if an error occurs, the `strvector` stays consistent, and does not change its size or contents. This means that some of the changes will need to be undone. You _might_ need to use `ENTER`/`EXIT` if there's a chance that `IGRAPH_ERROR` was called when the failure occurred.

src/core/strvector.c
```diff
@@ -211,19 +211,29 @@ igraph_error_t igraph_strvector_set_len(igraph_strvector_t *sv, igraph_integer_t
...
+    igraph_integer_t size = igraph_strvector_size(sv);
     igraph_integer_t i;
 
     IGRAPH_ASSERT(sv != NULL);
     IGRAPH_ASSERT(sv->stor_begin != NULL);
 
-    for (i = from; i < to; i++) {
-        IGRAPH_FREE(sv->stor_begin[i]);
+    if (from < 0) {
+        from = 0;
```
szhorvat, line 221: I don't like this because it hides mistakes. This should be an `IGRAPH_ASSERT()`. Mistakes should be brought to the surface so that they can be fixed.

src/core/strvector.c
ntamas: I've removed the check; we don't do this for vectors either, so if the goal is to make strvector more similar to the vector type, then we just leave it to the AddressSanitizer to do bounds checks.

src/core/strvector.c
ntamas: How about the truncation of `to` to the length of the vector? Should that also become an `IGRAPH_ASSERT()`? Note that the current implementation allows clearing the vector from a given index onwards to the end by supplying a very large number to `to`, which would not be possible if we check `to` rigorously.

Also, how about the case of `from > to`? Technically, that's also a mistake, so should that also become an `IGRAPH_ASSERT()`? Most programming languages with slicing operators happily allow the case of `from > to`, they just simply interpret it as an empty slice.

szhorvat: If you think that there is a good use case for one of these, then do not prevent it with an assert.

Personally think that supplying a large number to `to` is sloppy: one should supply the size instead: `strvector_remove_section(strvec, from, strvector_size(strvec))`.

As for `from > to`, my gut feeling is that it's fine to interpret this as empty, but I can't quite explain why, in a rational manner. Probably it's my association to C++ ranges: a typical implementation is often similar to 

```
while (from < to) {
    do_stuff(from);
    from++;
}
```

This naturally leads to an empty range when `from >= to`. But if `from < 0` or `to > size` then we have an out-of-bounds access.

Another point is that, subjectively, I think that there's a higher risk of out-of-bounds accesses than an accidental `from > to`, so catching out-of-bounds cases is more useful than catching `from > to`.

szhorvat: Thinking more about this,

 - Allowing from < 0 and to > size is mathematically consistent. We can have any range within (-infty, +infty) and the overlap of this range with the vector will be removed.
 - But what matters here is not mathematical consistency, but usefulness. Is there a realistic use case for from < 0 or to > size? I can't think of one right now, even for convenience. It seems to me that if such values are passed in, the most likely reason is that the programmer made a mistake (there's a bug). Thus, the most useful behaviour is to make them aware of the mistake, just in case the mistake would have other consequences as well.

szhorvat:
Should `igraph_strvector_get` return a `char *` or `const char *`? Are we supposed to directly modify the strings in the vector? I don't think igraph does this at the moment. It's just a thought I had, no hard preference from my part.

src/core/strvector.c
ntamas: I am still of the opinion that downright crashing the app when we could do better is downright rude; supplying negative indices to `to` and `from` are mistakes but they aren't mistakes that we could not solve on our own inside `remove_section()` (unlike things like null pointers in `stor_begin` that we clearly cannot solve). Since the alternative that I've implemented is mathematically consistent, I'd rather leave it that way. (I am partly inspired here by higher-level languages like Python or JavaScript that do not make a fuss about out-of-bounds indices when slicing an array). So, unless you have strong feelings against forcing `from` and `to` into a sensible range, I'll leave it this way.

ntamas:
> Should `igraph_strvector_get` return a `char *` or `const char *`?

I'll try `const char*` and see where it leads. It has an avalanche effect throughout the codebase, but so far it seems manageable.

szhorvat:
Can you resolve the conflict with `develop` so that CI runs can resume?

src/core/strvector.c
szhorvat: Code written in higher-level language is bug prone exactly because of this.

My point is that if the programmer made a mistake, a crash from `remove_section()` is very likely not the only consequence. Hiding the mistake does a disservice to the programmer: they will notice the problem later and will have to spend more time debugging it.  A "crash" is not evil. It is simply the way errors show up in C. What's helpful to do is to make it a crash that returns useful information (i.e. an `assert`) instead of a plain segfault.

This is the tradeoff between choosing a strongly typed low-level language and a "sloppy" high-level one. In the former, we need to do a bit more work because every aspect of the program must be specified precisely, but in return we get guarantees that we can rely on. The latter lets us get away with a quick and dirty job. That's useful if I just want to get some data analysis done with no fuss, but quick and dirty work will get you dirty results. Eventually something will go wrong. There is _some_ input for which the program will fail.

Since we are using a strict language, and we are working on the innermost, core part of the library, we should go with the philosophy of strict languages: provide as strong guarantees as possible, and surface errors as fast as possible. The high-risk approach of Python is not suitable for C, especially not for a library that must perform reliably under all conditions.

szhorvat:
I looked through it, added a few minor touches and `resize_min()`. I wanted `resize_min()` because `push_back()` will now increase the capacity exponentially.

I still strongly dislike the idea of hiding out-of-bounds errors. I laid out my arguments above. However, I also want to move forward with this PR ASAP. Therefore I ask you to just make your decision, and merge without waiting for further input from me.

Next steps are:

 - Wait two days to see how it fares on OSS-fuzz
 - Eliminate all uses of deprecated functions (if any)
 - Refactor some of the graph reader code to make use of the new capabilities (`reserve`/`capacity`/`resize_min`)

src/core/strvector.c
szhorvat: Also, this behaviour extends automatically to `strvector_remove()`. Ignoring out-of-range errors there is even worse.

ntamas: >  A "crash" is not evil. It is simply the way errors show up in C. 

In certain contexts it _is_ evil. I wouldn't want igraph to run on any of my drones at this moment due to the tendency of diverting to fatal errors even when it would be perfectly fine to let the user handle the error gracefully. For instance, I could consider using igraph on a drone as a path planner on the visibility graph of a fixed set of hard-coded obstacles, but if I cannot trust igraph not to throw fatal errors, I would not use it. If igraph simply returned an error code, I could just make it a policy not to ignore the return values of igraph functions (there are static code checkers to ensure that), and handle all errors while keeping the basic flight stabilization routines of the firmware operational at all times. It's not like scientific computing where the worst thing that can happen is that you get thrown back to the shell and you need to restart the calculation that you were working on (although this can also be pretty annoying if you submit a few thousand jobs to a cluster and they all crash after a few hours of number crunching because of some internal error in igraph that we decided to make fatal even if we could have handled it).

So, basically, I'd like to leave fatal errors for situations when something is so utterly broken beyond all repair that we cannot do anything sensible about it. Otherwise we could simply take all cases when we return `IGRAPH_EINVAL` into fatal errors, couldn't we?

Anyhow, I think this discussion has far-fetching implications for the entire coding style of the library and it's probably not a PR where we should discuss this in detail. If you wish, we can put this on the agenda of the next dev meeting where all the team members can voice their opinions.

szhorvat: If the function received an out-of-bounds value when it was not supposed to, then how can you trust anything your program does at this point? Something went wrong, it's just hidden. Your drone is going to fly down when it's supposed to go up, and you won't even know why.

igraph will only throw fatal errors when there is a bug. If you don't want it to throw fatal errors, you should find the bugs and fix them. Hiding the failures is an anti-pattern. It leads to bugs lurking for longer before they are found and fixed.

BTW fatal errors _can_ be handled in those special situations where it's necessary to be able to recover even from a disaster. But at least we know that there was a disaster.

szhorvat: > Otherwise we could simply take all cases when we return `IGRAPH_EINVAL` into fatal errors, couldn't we?

No, as I said last time this is part of the standard argument checking, which high-level interfaces need to do anyway. Invalid input to those functions which are exposed in high-level interfaces is something we expect, and handle explicitly. In other words, if the someone passes in a negative probability from R, that's a bug in the user's R code, not a bug in the C core. That's why it's not a fatal error.

Issue #1970, opened 2022-02-20
Assertion failure in igraph_add_edges()


szhorvat:
**Describe the bug**

There may be an assertion failure in `igraph_add_edges()` when `igraph_i_attribute_add_edges()` fails. Specifically, it happens when using an error handler that returns and an error occurs on this line:

https://github.com/igraph/igraph/blob/master/src/graph/type_indexededgelist.c#L313

This was found by OSS-fuzz, and is triggered when reading a GraphML file that has one attribute specified twice with different types.

It's unclear to me if this is ultimately a bug in the GraphML reader, the attribute handler, or `igraph_add_edges()` itself. It seem to me that it is the latter.

**To reproduce**

Use this C code to read the included GraphML file.

C code:

```c
#include <igraph.h>

int main(int argc, char *argv[]) {
    if (argc != 2)
        return 1;

    igraph_set_attribute_table(&igraph_cattribute_table);

    igraph_set_error_handler(igraph_error_handler_printignore);

    FILE *file;
    file = fopen(argv[1], "r");
    if (! file) {
        fprintf(stderr, "Failed to open file.");
        return 2;
    }

    igraph_t graph;
    igraph_error_t err = igraph_read_graph_graphml(&graph, file, 0);
    
    fclose(file);

    if (err != IGRAPH_SUCCESS) {
        fprintf(stderr, "Failed read.");
        return 3;
    }

    igraph_write_graph_graphml(&graph, stdout, 1);

    igraph_destroy(&graph);

    return 0;
}
```

GraphML file, as generated by OSS-fuzz:

```xml
<graphml><key id="" for="e" attr.type="d"/><key id="" for="e" attr.type=""></key><graph></graph></graphml>
```

Here's a "less invalid" GraphML input, which also has duplicate attributes of different types:

```xml
<graphml><key id="one" for="edge" attr.type="double"/><key id="one" for="edge" attr.type="string"></key><graph></graph></graphml>
```

If using an error handler that aborts, the output is (with address sanitizer for stack trace):

```
Error at src/graph/cattributes.c:1757 : Attribute types do not match - Invalid value.

Stack trace:
    #0 0x11038c467 in __sanitizer_print_stack_trace (libclang_rt.asan_osx_dynamic.dylib:x86_64h+0x6b467)
    #1 0x10ff4161a in igraph_abort error.c:69
    #2 0x10ff405b9 in igraph_error_handler_abort error.c:199
[...]

PR #1972, opened 2022-02-21
Multi-level error handling

szhorvat:
This is an attempt at fixing #1970 

The idea is that the finally stack is now structured into separate levels:

```
8   ...           LEVEL 2 (CURRENT LEVEL)
7   ...
------------
6   ...
5   ...           LEVEL 1
4   ...
------------
3   ...
2   ...           LEVEL 0
1   destr2()
0   destr1()
```

When `IGRAPH_FINALLY_FREE()` is called by the error handler*, then destructors are executed only from the topmost level. This ensures that we have a chance to still work with data structures that are present at lower levels in the stack.

This is useful not only for #1970, but also for cases when using a library that operates with callback functions. This applies to almost every graph reader (Bison based parsers do effectively this) as well as some other functions.  We might need to set up a data structure for the library to work with, then call the library, which then calls back to a function provided by us. If a failure occurs in this function, it should be result in destroying the data that the library works with. That should be deferred until the library has fully returned. This can be ensured by "entering" a new finally stack level before calling the library.

Entering/exiting a level is done with `IGRAPH_FINALLY_ENTER()/EXIT()`. See the changes to `igraph_add_edges()` in this PR to see the typical programming patterns this leads to. We may want to wrap it up into a simpler-to-use macro, but I'm not sure it's actually worth it.

----

**TODO**

 - [ ] Agree on and finalize design
 - [ ] Use it in `igraph_trie_get()` (which sets special error handlers)
 - [ ] Use it in GraphML reader (which sets special error handlers for a similar purpose)
 - [ ] Use it in Bison-based graph readers, where the current error handling relies on Bison (not) allocating memory in a certain way

----

Footnotes:

\* If we go with a solution like this, we might as well make calling `IGRAPH_FINALLY_FREE()` mandatory and do it in `igraph_error()` instead of leaving it to the error handler.

### DIFF
--- a/include/igraph_error.h
+++ b/include/igraph_error.h
@@ -580,7 +580,7 @@ IGRAPH_EXPORT const char* igraph_strerror(const igraph_error_t igraph_errno);
  * information. We don't use the exception handling code though.  */
 
 struct igraph_i_protectedPtr {
-    int all;
+    int level;
     void *ptr;
     void (*func)(void*);
 };
@@ -615,6 +615,9 @@ IGRAPH_EXPORT void IGRAPH_FINALLY_CLEAN(int num);
  */
 
 IGRAPH_EXPORT void IGRAPH_FINALLY_FREE(void);
+
+IGRAPH_EXPORT void IGRAPH_FINALLY_ENTER(void);
+IGRAPH_EXPORT void IGRAPH_FINALLY_EXIT(void);
 
 /**
  * \function IGRAPH_FINALLY_STACK_SIZE
--- a/src/core/error.c
+++ b/src/core/error.c
@@ -229,47 +229,71 @@ igraph_error_handler_t *igraph_set_error_handler(igraph_error_handler_t *new_han
 /***** "Finally" stack *****/
 
 IGRAPH_THREAD_LOCAL struct igraph_i_protectedPtr igraph_i_finally_stack[100];
+IGRAPH_THREAD_LOCAL int igraph_i_finally_stack_size = 0;
+IGRAPH_THREAD_LOCAL int igraph_i_finally_stack_level = 0;
 
 /*
  * Adds another element to the free list
  */
 
 void IGRAPH_FINALLY_REAL(void (*func)(void*), void* ptr) {
-    int no = igraph_i_finally_stack[0].all;
+    int no = igraph_i_finally_stack_size;
     if (no < 0) {
+        /* Reset finally stack in case fatal error handler does a longjmp instead of terminating the process: */
+        igraph_i_finally_stack_size = 0;
+        igraph_i_finally_stack_level = 0;
         IGRAPH_FATALF("Corrupt finally stack: it contains %d elements.", no);
     }
-    if (no >= 100) {
+    if (no >= (int) (sizeof(igraph_i_finally_stack) / sizeof(igraph_i_finally_stack[0]))) {
+        /* Reset finally stack in case fatal error handler does a longjmp instead of terminating the process: */
+        igraph_i_finally_stack_size = 0;
+        igraph_i_finally_stack_level = 0;
         IGRAPH_FATALF("Finally stack too large: it contains %d elements.", no);
     }
     igraph_i_finally_stack[no].ptr = ptr;
     igraph_i_finally_stack[no].func = func;
-    igraph_i_finally_stack[0].all ++;
-    /* printf("--> Finally stack contains now %d elements\n", igraph_i_finally_stack[0].all); */
+    igraph_i_finally_stack[no].level = igraph_i_finally_stack_level;
+    igraph_i_finally_stack_size++;
 }
 
 void IGRAPH_FINALLY_CLEAN(int minus) {
-    igraph_i_finally_stack[0].all -= minus;
-    if (igraph_i_finally_stack[0].all < 0) {
-        int left = igraph_i_finally_stack[0].all + minus;
-        /* Set to zero in case fatal error handler does a longjmp instead of terminating the process: */
-        igraph_i_finally_stack[0].all = 0;
+    igraph_i_finally_stack_size -= minus;
+    if (igraph_i_finally_stack_size < 0) {
+        int left = igraph_i_finally_stack_size + minus;
+        /* Reset finally stack in case fatal error handler does a longjmp instead of terminating the process: */
+        igraph_i_finally_stack_size = 0;
+        igraph_i_finally_stack_level = 0;
         IGRAPH_FATALF("Corrupt finally stack: trying to pop %d element(s) when only %d left.", minus, left);
     }
-    /* printf("<-- Finally stack contains now %d elements\n", igraph_i_finally_stack[0].all); */
 }
 
 void IGRAPH_FINALLY_FREE(void) {
-    int p;
-    /*   printf("[X] Finally stack will be cleaned (contained %d elements)\n", igraph_i_finally_stack[0].all);  */
-    for (p = igraph_i_finally_stack[0].all - 1; p >= 0; p--) {
+    for (; igraph_i_finally_stack_size > 0; igraph_i_finally_stack_size--) {
+        int p = igraph_i_finally_stack_size - 1;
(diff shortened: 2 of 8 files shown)


vtraag:
The implementation looks good to me @szhorvat, I don't have any immediate remarks about that. The strange role of the `all` entry in `igraph_i_protectedPtr` is probably good to revise anyway.

The logic in dealing with this construction seems less obvious directly. That is, it seems quite contorted how to use this construction in practice, as it is implemented now in `type_indexededgelist.c`. The `CHECK_ERR` you implemented in `type_indexededgelist.c` now essentially performs a type of `catch` operation, catching the error and performing the restoration of the graph internals, before rethrowing the exception again through `IGRAPH_ERROR`. Could we perhaps make it work so that this becomes more clear? I don't immediately have a better idea, I'll have to think about. 

You mention that the error handling in `igraph` is based on some of the ideas in https://ldeniau.web.cern.ch/ldeniau/html/oopc/oopc.html#Exceptions, but this is much more involved it seems (although a `finally` statement seems to be missing, ironically). Have you looked at this in more detail, and what are your thoughts?

> * If we go with a solution like this, we might as well make calling `IGRAPH_FINALLY_FREE()` mandatory and do it in `igraph_error()` instead of leaving it to the error handler.

This is probably a good idea anyway, wouldn't it?

szhorvat:
> You mention that the error handling in `igraph` is based on some of the ideas in https://ldeniau.web.cern.ch/ldeniau/html/oopc/oopc.html#Exceptions, but this is much more involved it seems (although a `finally` statement seems to be missing, ironically). Have you looked at this in more detail, and what are your thoughts?

I have not looked in details, as I realized that that system is quite different. I think it uses a separate stack for each `try` location, and uses longjmp for control flow. It's a full-blown exception mechanism. We _may_ want to try to do something like that, or even borrow the whole system, but that's a huge change. This PR attempts a minimal fix.

> The logic in dealing with this construction seems less obvious directly. That is, it seems quite contorted how to use this construction in practice, as it is implemented now in `type_indexededgelist.c`. The `CHECK_ERR` you implemented in `type_indexededgelist.c` now essentially performs a type of `catch` operation, catching the error and performing the restoration of the graph internals, before rethrowing the exception again through `IGRAPH_ERROR`. Could we perhaps make it work so that this becomes more clear? I don't immediately have a better idea, I'll have to think about.

It would be nice to wrap this up into a more structured construct that has a "try" and a "catch/cleanup" section. The problem is that the cleanup code tends to change as we progress through the block. For example, see the difference between these two cleanup blocks:

https://github.com/igraph/igraph/blob/develop/src/core/trie.c#L309

https://github.com/igraph/igraph/blob/develop/src/core/trie.c#L314

Thus sometimes we can't use a single piece of cleanup code throughout.

szhorvat:
Let's see if Tamás has any ideas when he's back from holidays. For now, I'm refraining from directly pinging him.

szhorvat:
BTW we already have an OSS-fuzz issue which I believe is related to the `trie` code I mentioned: https://oss-fuzz.com/testcase-detail/5699537782571008  It can be fixed by this, and the symptom would be eliminated by #1969, I believe.

vtraag:
> It would be nice to wrap this up into a more structured construct that has a "try" and a "catch/cleanup" section. The problem is that the cleanup code tends to change as we progress through the block.

Yes, this would be nicer, although I'm not sure how this could best be implemented without creating too many problems. Of course if there is different cleanup code, then we can just have different try/catch sections. This should in itself not really be problematic, should it?

vtraag:
Just thinking out loud here: Instead of having a macro that returns directly, could we perhaps set it up as follows:

```C
IGRAPH_FINALLY_ENTER()
igraph_error_t err = IGRAPH_SUCCESS;
do
{
    CHECK_ERR(x);
}
while(0)
if (err != IGRAPH_SUCCESS)
{
    // Handle specific problems
    ...

   IGRAPH_FINALLY_EXIT();
   IGRAPH_ERROR("Propagate error up", err);
}
IGRAPH_FINALLY_EXIT()
```

where then

```
#define CHECK_ERR(expr) \
      err = (expr); \
      if (err != IGRAPH_SUCCESS) { \
           break; \
      } 
```

We could then create macros that facilitate this workflow, e.g. create a `CATCH` macro as

```C
#define CATCH(x)
if (err != IGRAPH_SUCCESS) \
{ \
    x \
 \
   IGRAPH_FINALLY_EXIT(); \
   IGRAPH_ERROR("Propagate error up", err); \
}
```

Perhaps we could similarly make a `TRY` macro that facilitates this, so that we could somehow end up writing

```C
IGRAPH_TRY_BEGIN();
  x
IGRAPH_TRY_END();
CATCH(
  y
)
```

ntamas:
> Let's see if Tamás has any ideas when he's back from holidays. For now, I'm refraining from directly pinging him.

Now that I've taken a look at it, it might be worth looking at how Python's `Py_BEGIN_ALLOW_THREADS` and `Py_END_ALLOW_THREADS` macro works. These macros are meant to be used in pairs, just like `IGRAPH_FINALLY_ENTER` and `IGRAPH_FINALLY_EXIT` (at least I cannot imagine why we would want to break these up in separate functions). Python's `Py_BEGIN_ALLOW_THREADS` macro opens a new block implicitly, and `Py_END_ALLOW_THREADS` closes the block, so the user is absolutely forced to use them in pairs (otherwise the code wouldn't compile). This is a neat trick that we might want to borrow. Note that this implies that handling and re-throwing an error code has to be moved out of the `IGRAPH_FINALLY_ENTER` .. `IGRAPH_FINALLY_EXIT` block; I consider this as an advantage because then you won't have multiple `_EXIT` calls for a single `_ENTER`.

I'm not really fond of trying to conjure up a `try..catch` syntax-alike in C with macros, though. Let's go ahead with this proposal first and see what sort of patterns emerge; after that we can still re-visit the question of adding more macros if that can potentially make the code clearer.

ntamas:
I'd probably also move this code snippet:

```
igraph_i_finally_stack_size = 0;
igraph_i_finally_stack_level = 0;
```

to a separate `igraph_i_reset_finally_stack()` function as it is now being repeated many times.

szhorvat:
These are not macros in the current implementation, just functions. The plan was to first use these raw functions, then see what programming patterns we end up with, then finally try to wrap them up into more structured-looking macros.

The problem is that I do not see how one can avoid `EXIT`ing in the middle when an error occurs. Exiting in the middle in some way is necessary for error handling. How the exit works is the question. It can be a direct call to `EXIT` followed by a `return` or some jump instruction to the part of the code that does the cleanup (a `goto`, a `break` in ` do`, some `if`/`else` or whatever—but these are hard to make flexible). If we don't want to exit in the middle at all, then we need to wrap every potentially failing function call into an ENTER/EXIT pair. Doing that conflicts with not repeating the cleanup code many times.

So the challenge is: how do we minimize the number of times the cleanup code has to be stated?  How do we deal with the size of the cleanup code "growing" as the algorithm proceeds through the block? 

Luckily, the need for this ENTER/EXIT construct is very rare, so it IMO the solution does not have to be perfect. If it's a bit verbose that's okay.  In fact, personally I'd be even happy with just using the raw calls.  In the end, I just want a solution to fix #1970 and related issues, but I'm not fussy about how it is done.

ntamas:
Okay, I can let go of the requirement that `ENTER` and `EXIT` are always used in pairs, in which case you can use `EXIT` wherever you want. Which basically takes us back to the present state of the PR, right?

szhorvat:
In the end I did not create convenient wrapper macros. I would very much like to see this done, but I can't come up with anything I am happy with, and I don't want to spend more time on it. I also didn't touch the GraphML code since it seems to be working fine as it is, though in the future it might be refactored a bit to use this functionality.

If you'd like to, feel free to polish this, otherwise please just merge.

ntamas (approved):

ntamas:
Thanks a lot!

szhorvat:
So once again, if someone can make this prettier that would be really great ...

PR #1974, opened 2022-02-23
Fix initial modularity for walktrap

### LINKED ISSUE #1927: `modularity()` results differ when called on communities vs with graph + membership
This is similar to an [old question on SO](https://stackoverflow.com/questions/20313608/the-correct-use-and-interpretation-of-modularity) but not due to weights.

The two different ways given in `?modularity`  give different results yet based on the answer to the SO question I don't think this should be the case.

The two different ways to compute modularity given in the help are:
```r
    wtc <- cluster_walktrap(g)
    modularity(wtc)
    # and
    modularity(g, membership(wtc))
```

For this example these give different results

```r
g2 <- make_graph(c(1,2,2,3,3,1), directed = FALSE)

set.seed(444)
wtc2 <- cluster_walktrap(g2)
modularity(wtc2)
## 0
wtc2
##IGRAPH clustering walktrap, groups: 3, mod: 0
##+ groups:
## $`1`
##  [1] 1
## 
##  $`2`
## [1] 2
##  
##  $`3`
##  [1] 3
modularity(g2, membership(wtc2))
## -0.3333
```

This is with `igraph_1.2.5`.

## Note

This shows up on much more complicated graphs as well -- this is the simplest case where I could reproduce

vtraag:
Walktrap maintains a list of modularity along each step of its merging of various communities. The initial modularity was always incorrectly set to 0, instead of the modularity of the singleton partition. Walktrap returns the actual membership that contains the maximum modularity.

The actual partition that was identified in walktrap in #1927 is the partition that puts the whole graph into a single cluster, which has a modularity of 0. However, because the initial modularity before any merge was incorrectly set to 0, it incorrectly returned the singleton partition instead. This PR fixes #1927.

### DIFF
--- a/examples/simple/walktrap.out
+++ b/examples/simple/walktrap.out
@@ -1,5 +1,5 @@
 Merges:
- 6 +  8 -> 10 (modularity 0.00)
+ 6 +  8 -> 10 (modularity -0.10)
  1 +  2 -> 11 (modularity -0.07)
  7 + 10 -> 12 (modularity -0.04)
  3 + 11 -> 13 (modularity 0.02)
--- a/src/community/walktrap/walktrap_communities.cpp
+++ b/src/community/walktrap/walktrap_communities.cpp
@@ -482,6 +482,13 @@ Communities::Communities(Graph* graph, int random_walks_length,
         /*     } */
     }
 
+    double Q = 0.;
+    for (int i = 0; i < nb_communities; i++) {
+      if (communities[i].sub_community_of == 0) {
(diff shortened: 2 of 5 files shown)


szhorvat (approved):

szhorvat:
My only comment is that it would be nice to backport this to 0.9 as well so it can go into the 1.3 R release... but if you think it's not important, just skip it.

vtraag:
Yeah, I agree, I will backport to 0.9.

PR #1996, opened 2022-03-12
Add igraph_sample_dirichlet tests

GroteGnoom:
Part of #1592


szhorvat:
If this helps, one of the invariants is that the total of the result vector is in [0, 1].

szhorvat:
This function returns the result a bit differently than I expected.

 - The result vectors must be the same length as the `alpha` vector
 - The sum of each vector must be 1.
 - The vector elements are in (0,1)
 - `n` vectors are returned as the columns of a matrix

szhorvat (approved):
Any reason you didn't use `IGRAPH_ASSERT` instead of `printf` + `exit(1)`?

It looks good to me, feel free to merge.

PR #2010, opened 2022-03-21
strvector_merge()

szhorvat:
Let's agree on a design through this strvector example, then I will add an analogous function for vector_list. EDIT: Actually, vector_list doesn't even have `append`, so I will leave adding these for a separate PR.

### DIFF
--- a/src/core/strvector.c
+++ b/src/core/strvector.c
@@ -27,6 +27,7 @@
 #include "igraph_error.h"
 
 #include "internal/hacks.h" /* strdup */
+#include "math/safe_intop.h"
 
 #include <string.h>         /* memcpy & co. */
 #include <stdlib.h>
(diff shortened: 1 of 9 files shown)


szhorvat:
There is an extra commit to rename `strvector_copy()` to `strvector_init_copy()` for consistency with `vector_init_copy()` and to make it clear that it is a constructor.

src/core/strvector.c
```diff
@@ -333,6 +355,44 @@ igraph_error_t igraph_strvector_append(igraph_strvector_t *to,
     return IGRAPH_SUCCESS;
 }
 
+/**
+ * \ingroup strvector
+ * \function igraph_strvector_merge
+ * \brief Merges a string vector into another, clearing the first one.
```
ntamas, line 361: The word "first" is potentially confusing here because the order of the _arguments_ is "to" and then "from" so the vector being cleared is the _second_ argument, but the _first_ vector in the sentence. It would be good to rephrase this somehow to avoid confusion. Something like: merges a _source_ vector into a _destination_ vector and clears the _source_ afterwards.

ntamas (changes requested):
Just one minor comment, otherwise it looks good I think.

src/core/strvector.c
szhorvat: How about this? _"Merges a string vector into another without reallocating the contents"_  This makes it clear what the major difference is from `append`. Or "... without reallocating individual strings".

Or maybe: "Moves the contents of a string vector into another" or "... to the end of another." The word "move" implies that there is no copying.

ntamas: good! Feel free to merge if this is fixed.

ntamas: I mean, I vote for the "moves the contents" variant.

PR #2012, opened 2022-03-22
Fix up igraph_diversity()

szhorvat:
fix: fix up igraph_diversity()

 - Simplify code and avoid duplication
 - Check that weights are non-negative and non-NaN
 - Now return 0 for degree-1 vertices

----

First, a recap on the definition of "diversity" (not a particularly popular measure).

Take a vertex and all its incident edges. Divide the edge weights by the vertex strengths to obtain "probabilities" (values that sum to the vertex). The compute the Shannon entropy of these values, `S = - \sum_i p_i log p_i`. Finally, normalize it by the maximum value, which is `log k` for `k` values.

----

This is partly triggered by the tidygraph test suite failure with `diversity()`. The reason for the failure is that now `igraph_diversity()` accepts only undirected graphs. This decision was correct in my opinion. Arguably, the definition used by the function only makes sense for undirected graphs. For directed ones, one might choose to take only out-edges or similar. However, I don't want igraph to take the lead on generalizing very niche graph measures, and the cited paper only deals with the undirected case (as I recall).

The test case used by tidygraph is a k-ary tree, effectively: `diversity(make_tree(10,2), weights=1:9)`. This triggers an error as this is a directed graph. But let's make it undirected and see what we get:

```
> diversity(make_tree(10,2,mode='undirected'), weights=1:9)
 [1] 0.9182958 0.8868595 0.9214632 0.9342064 0.8904916       NaN      -Inf       NaN
 [9]       NaN       NaN
```

Notice that the tree leaves (degree-1 vertices) get a NaN value, except for one, which gets -Inf. This is inconsistent, they should all be the same. This is likely the result of numerical roundoff.

This PR makes it so that degree-1 vertices always get a value of 0. This is the Shannon entropy when there is only one probability value (necessarily p=1). The normalized version would be 0/0 = NaN, but I think it makes sense to keep zeros as zero, even after using some sort of normalization.

Additionally, it now returns an explicit `IGRAPH_NAN` for isolated vertices instead of letting the NaN value arise as the result of calculations.

The function now checks that the weights are non-negative, which is necessary to take their logarithm. I opted to allow zero weights, and use 0*log(0) = 0, which is standard for entropy calculations.

Finally, the code has been refactored to avoid duplication.

Since some of these decisions may be controversial, here's a PR.  It should go in `master` and be part of R/igraph 1.3.0.

### DIFF
--- a/src/properties/basic_properties.c
+++ b/src/properties/basic_properties.c
@@ -175,11 +175,24 @@ int igraph_diversity(const igraph_t *graph, const igraph_vector_t *weights,
         long int v = IGRAPH_VIT_GET(vit);
 
         IGRAPH_CHECK(igraph_incident(graph, &incident, v, /*mode=*/ IGRAPH_ALL));
-        k = igraph_vector_size(&incident);
+        k = igraph_vector_size(&incident); /* degree */
+
+        /*
+         * Non-normalized diversity is defined as
+         * d = -sum_i w_i/s log (w_i/s)
+         * where s = sum_i w_i. In order to avoid two passes through the w vector,
+         * we use the equivalent formulation of
+         * d = log s - (sum_i w_i log w_i) / s
+         * However, this formulation may not give an exact 0.0 for some w when k=1,
+         * due to roundoff errors (examples: w=3 or w=7). For this reason, we
+         * special-case the computation for k=1 even for the unnormalized diversity
+         * insted of just setting the normalization factor to 1 for this case.
+         */
         if (k == 0) {
             d = IGRAPH_NAN;
         } else if (k == 1) {
-            d = 0.0;
+            if (VECTOR(*weights)[0] > 0) d = 0.0; /* s > 0 */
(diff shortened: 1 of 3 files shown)


vtraag (approved):
Looks good to me!

szhorvat:
As recommended by @vtraag in chat, I eliminated the separate branch for the `igraph_vss_all()` case. There should be no performance difference when compiled with LTO. This version reserves the appropriate amount of memory for `res.`

ntamas (approved):

ntamas:
Thanks a lot!

ntamas:
Porting this to the `develop` branch now, FYI.

szhorvat:
Actually, I think the check for NaN weights was not strictly necessary. We need to do that for betweenness because NaN will put the algorithm on the wrong track and risks a crash. But here, if a weight is NaN, it just causes some of the results to be NaN, which does make sense (especially with NA in R).

PR #2013, opened 2022-03-23
Merge random_edge_walk() into random_walk()

meetpatel0963:
This pull request merges a function `igraph_random_edge_walk` into `igraph_random_walk` to support both the vertices and edges in a single function - part of #1882.

- It uses `igraph_i_random_walk_adjlist` for unweighted graphs when only vertices are required.
- It uses `igraph_i_random_walk_inclist` when edges are required and/or vertices are required (with weighted graphs).


meetpatel0963:
Hi. I have added missing test cases for `Misses` indicated by Codecov Report.

szhorvat:
FYI — if you see the "422 - test::glpk_error (SEGFAULT)" test failure with MinGW, it's not because of the changes you made. That's an unrelated intermittend issue, #1850. I'll re-run that test now, which should get rid of the red x.

vtraag:
Hi @meetpatel0963, thanks for contributing this! 

I just want to double check with @szhorvat about the idea of the implementation. We currently have `igraph_random_edge_walk` and `igraph_random_walk`, and the only difference is whether they return the path in terms of edges or in terms of vertices. In [`igraph_get_shortest_path`](https://igraph.org/c/doc/igraph-Structural.html#igraph_get_shortest_path) we integrate that into a single function, which can return both the vertices and the edges in a single function. To me, this makes more sense. For that reason, I would therefore propose to just create a single new function `igraph_random_walk` for 0.10 (breaking API) with the signature:

```c
igraph_error_t igraph_random_edge_walk(const igraph_t *graph,
                            igraph_vector_int_t *vertices,
                            igraph_vector_int_t *edges,
                            igraph_integer_t start,
                            igraph_neimode_t mode,
                            igraph_integer_t steps,
                            igraph_random_walk_stuck_t stuck)
```

where `vertices` contains the vertices in the random walk and `edges` contains the edges in the random walk, similar to how we approach this for `igraph_get_shortest_path`. This avoids a lot of code duplication between the two. The current code of `igraph_random_edge_walk` could well serve as the basis then, where it essentially is just a matter of pushing each `start` vertex into a vector at each step.

szhorvat:
Having such an API makes sense, but I don't want to remove the current vertex walk implementation. The `adjlist`-based version is clearly faster than the `inclist`-based one, so we should keep this special case, and use it when the user only requests vertices and no weights are present (very common use case).

Thus, I would propose the same API that you suggest above (though I would prefer to call it `igraph_random_walk()` and to just deprecate `igraph_random_edge_walk()`). This would be backed by two _internal_ functions: one using adjlist (used when only vertices are requested and there are no weights) and one using inclist (used in all other cases). We already have these two, it's just a question of a minor restructuring.

szhorvat:
Something different that I'd like your opinion on @vtraag :

Once this work is done, it would be interesting to add a new `RESTART` value to `igraph_random_walk_stuck_t`. When the random walk gets stuck, it would restart from a random vertex. This is not particularly useful for real work, but it would be educational, and useful for demonstrating PageRank. While this can be implemented relatively easily using the current capabilities of this function, why not support it directly? Instructors/teachers might appreciate it, and I think we should support them well with igraph.

vtraag:
> Thus, I would propose the same API that you suggest above (though I would prefer to call it `igraph_random_walk()` and to just deprecate `igraph_random_edge_walk()`).

Yes, sorry, I meant to call it `igraph_random_walk` as well.

> The `adjlist`-based version is clearly faster than the `inclist`-based one, so we should keep this special case, and use it when the user only requests vertices and no weights are present (very common use case).

Yes, right, sorry, I didn't realise that the `igraph_random_edge_walk` always uses an `inclist`.

So, all in all, we then come to the proposal that we two internal functions

```C
igraph_error_t igraph_i_random_walk_adjlist(const igraph_t *graph,
                            igraph_vector_int_t *vertices,
                            igraph_integer_t start,
                            igraph_neimode_t mode,
                            igraph_integer_t steps,
                            igraph_random_walk_stuck_t stuck)
```

and

```C
igraph_error_t igraph_i_random_walk_inclist(const igraph_t *graph,
                            igraph_vector_int_t *vertices,
                            igraph_vector_int_t *edges,
                            igraph_integer_t start,
                            igraph_neimode_t mode,
                            igraph_integer_t steps,
                            igraph_random_walk_stuck_t stuck)
```

with a single

```C
igraph_error_t igraph_random_walk(const igraph_t *graph,
                            igraph_vector_int_t *vertices,
                            igraph_vector_int_t *edges,
                            igraph_integer_t start,
                            igraph_neimode_t mode,
                            igraph_integer_t steps,
                            igraph_random_walk_stuck_t stuck)
```

which essentially just calls

```C
...
if (edges != NULL || weights != null)
    igraph_i_random_walk_inclist( ... );
else
    igraph_i_random_walk_adjlist( ... );

...
```

Are we on the same page then?

vtraag:
> Once this work is done, it would be interesting to add a new `RESTART` value to `igraph_random_walk_stuck_t`. When the random walk gets stuck, it would restart from a random vertex. This is not particularly useful for real work, but it would be educational, and useful for demonstrating PageRank. While this can be implemented relatively easily using the current capabilities of this function, why not support it directly? Instructors/teachers might appreciate it, and I think we should support them well with igraph.

Yes, interesting idea! Perhaps better if that is done in a separate issue & PR?

szhorvat:
> Are we on the same page then?

Yes!

----

(Also, you might have gotten an email about my last comment, which I just deleted, as it was based on a misunderstanding. Please ignore it.)

vtraag:
Great!

@meetpatel0963 , are you comfortable updating the code in the direction described above? If something is unclear about it, do ask. If you are not comfortable making these changes, then just let us know, and we will handle it.

mp0963:
Hi @vtraag.
```
are you comfortable updating the code in the direction described above?
```
Yes sure, I got the point. It makes sense. I will add the requested changes.

meetpatel0963:
Hi @vtraag.
> are you comfortable updating the code in the direction described above?

Yes sure, I got the point. It makes sense. I will add the requested changes.

mp0963:
Hello. In case when both vertex IDs and edge Ids are needed, 
`size of the vector of vertex IDs will be = (steps + 1)` and 
`size of the vector of edge IDs will be = steps`, right?

In this case, we'll consider the steps as the number of edges traversed on the walk and include the vertex IDs of both start and end vertex of the walk in vertices. Is that correct or something else is expected?

meetpatel0963:
Hello. In case when both vertex IDs and edge Ids are needed,
`size of the vector of vertex IDs will be = (steps + 1)` and
`size of the vector of edge IDs will be = steps`, right?

In this case, we'll consider the steps as the number of edges traversed on the walk and include the vertex IDs of both start and end vertex of the walk in vertices. Is that correct or something else is expected?

vtraag:
> Hello. In case when both vertex IDs and edge IDs are needed, `size of the vector of vertex IDs will be = (steps + 1)` and `size of the vector of edge IDs will be = steps`, right?
> 
> In this case, we'll consider the steps as the number of edges traversed on the walk and include the vertex IDs of both start and end vertex of the walk in vertices. Is that correct or something else is expected?

Yes, let us do it like this, I think this makes most sense. The current two implementations are now not exactly consistent actually, because `igraph_random_walk` traverse `steps - 1` edges, while `igraph_random_edge_walk` traverse `steps` edges. It's good that this is now consistent.

> 
> Also, according to this:
> 
> ```c
> ...
> if (edges != NULL || weights != null)
>     igraph_i_random_walk_inclist( ... );
> else
>     igraph_i_random_walk_adjlist( ... );
> 
> ...
> ```
> 
> We are assuming that vertices are always needed, right? Or it may happen that we only need edges and not vertices in which case a separate check is required in `igraph_i_random_walk_inclist`.

No, the `vertex` argument is not necessarily needed. If we do a weighted random walk we need to access the edge weight, for which we need an `inclist`, not an `adjlist`. But, that doesn't mean that the `vertex` argument is needed.

szhorvat:
Hit "ready for review" again when you're done with everything (it seems it's done?)

vtraag:
Sorry @meetpatel0963, I haven't yet found the time to do a proper review, I should be able to do so later this week.

meetpatel0963:
Hi @vtraag, Cool. Not a problem.

tests/unit/igraph_random_walk.c
```diff
@@ -19,67 +19,216 @@
...
+    
+    /* Initialize g_line graph weights */
+    ec = igraph_ecount(&g_line);
+    igraph_vector_resize(&g_line_weights, ec);
+    for (i = 0; i < ec; ++i) {
+        VECTOR(g_line_weights)[i] = igraph_rng_get_unif01(igraph_rng_default());
+    }
+
+
+    /* Length of the vertices vector -> (steps + 1) -> includes starting and ending vertices of the walk,
```
vtraag, line 84: I would remove these comments, they are not really clarifying anything.

tests/unit/igraph_random_walk.c
```diff
@@ -19,67 +19,216 @@
...
 
+    /* This directed graph has loop edges.
+       It also has multi-edges when considered as undirected. */
+    igraph_de_bruijn(&graph, 3, 2);
+
+    /* Initialize de bruijn graph weights */
+    ec = igraph_ecount(&graph);
+    igraph_vector_resize(&graph_weights, ec);
+    for (i = 0; i < ec; ++i) {
+        VECTOR(graph_weights)[i] = igraph_rng_get_unif01(igraph_rng_default());
```
vtraag, line 73: Please seed the `rng` before creating the graph.

tests/unit/igraph_random_walk.c
```diff
@@ -19,67 +19,216 @@
...
 
     printf("Line graph backward:\n");
-    random_walk(&g_line, 4, IGRAPH_IN, 4, IGRAPH_RANDOM_WALK_STUCK_RETURN);
+    random_vertex_walk(&g_line, NULL, 4, IGRAPH_IN, 4, IGRAPH_RANDOM_WALK_STUCK_RETURN);
 
     printf("Loop at vertex 0:\n");
-    random_walk(&g_loop, 0, IGRAPH_OUT, 4, IGRAPH_RANDOM_WALK_STUCK_RETURN);
+    random_vertex_walk(&g_loop, NULL, 0, IGRAPH_OUT, 4, IGRAPH_RANDOM_WALK_STUCK_RETURN);
+    
+    printf("Checking an actual random walk with seed 42:\n");
```
vtraag, line 110: Please seed the `rng` here.

vtraag (changes requested):
Great, nice work! This looks very good, and you have tested it very comprehensively. There are a couple of minor comments I have. There is one larger comment and that is whether we shouldn't deprecate `igraph_random_edge_walk` first before removing it completely. I have no strong opinion on this, so I'll defer to @ntamas. Depending on what he thinks, we should first deprecate it (instructions for doing so are in the review).

ntamas:
I vote for deprecating `igraph_random_edge_walk()` properly first, and then removing it later. The deprecated version can be re-written to call the new `random_walk()` function with appropriate arguments if that is possible with the current refactoring.

vtraag:
> I vote for deprecating `igraph_random_edge_walk()` properly first, and then removing it later

OK, let's do that then.

> The deprecated version can be re-written to call the new `random_walk()` function with appropriate arguments if that is possible with the current refactoring.

Yes, that should be possible.

@meetpatel0963, the comments in the review should provide guidance for doing this. But if some things are unclear, please do let us know.

FYI, the deprecated `igraph_random_edge_walk` does not need to be separately tested.

tests/unit/igraph_random_walk.c
meetpatel0963: I have set the rng seed at line 58. Do we need to set rng seed every time we create a random weights vector? Or is it just to place the line that is setting the rng seed close to where it is used? I'm not sure about this.

include/igraph_paths.h
```diff
@@ -202,18 +202,15 @@ IGRAPH_EXPORT igraph_error_t igraph_get_all_simple_paths(const igraph_t *graph,
...
+IGRAPH_EXPORT igraph_error_t igraph_random_walk(const igraph_t *graph,
+                                     const igraph_vector_t *weights,
+                                     igraph_vector_int_t *vertices,
+                                     igraph_vector_int_t *edges,
+                                     igraph_integer_t start,
+                                     igraph_neimode_t mode,
                                      igraph_integer_t steps,
                                      igraph_random_walk_stuck_t stuck);
 
-IGRAPH_EXPORT igraph_error_t igraph_random_edge_walk(const igraph_t *graph,
```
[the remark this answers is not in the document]
meetpatel0963: Thank you very much for these instructions. I've added the changes accordingly.

src/paths/random_walk.c
```diff
@@ -132,85 +95,37 @@ static void vec_destr(igraph_vector_t *vec) {
...
- * \param weights A vector of non-negative edge weights. It is assumed
- *   that at least one strictly positive weight is found among the
- *   outgoing edges of each vertex. Additionally, no edge weight may
- *   be NaN. If either case does not hold, an error is returned. If it
- *   is a NULL pointer, all edges are considered to have equal weight.
- * \param edgewalk An initialized vector; the indices of traversed
+ * This function performs a random walk with a given length on a graph,
+ * from the given start vertex.
+ * It's used for igraph_random_walk when vertex IDs and edge IDs of the
+ * visited vertices and traversed edges respectively, both are needed.
```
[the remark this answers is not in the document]
meetpatel0963: Yes, sure. I have added it.

src/paths/random_walk.c
```diff
@@ -32,86 +32,49 @@
...
-                       igraph_random_walk_stuck_t stuck) {
-
-    /* TODO:
-       - multiple walks potentially from multiple start vertices
-       - weights
-    */
-
-    igraph_lazy_adjlist_t adj;
-    igraph_integer_t vc = igraph_vcount(graph);
+igraph_error_t igraph_i_random_walk_adjlist(const igraph_t *graph,
```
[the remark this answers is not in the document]
meetpatel0963: Yes, I have declared the both the internal functions static.

src/paths/random_walk.c
```diff
@@ -132,85 +95,37 @@ static void vec_destr(igraph_vector_t *vec) {
...
- * \return Error code.
- *
  */
-igraph_error_t igraph_random_edge_walk(const igraph_t *graph,
-                            const igraph_vector_t *weights,
-                            igraph_vector_int_t *edgewalk,
-                            igraph_integer_t start, igraph_neimode_t mode,
-                            igraph_integer_t steps,
-                            igraph_random_walk_stuck_t stuck) {
+igraph_error_t igraph_i_random_walk_inclist(const igraph_t *graph,
```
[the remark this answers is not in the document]
meetpatel0963: Yes, I have declared the both the internal functions static. Thanks for the advice.

src/paths/random_walk.c
```diff
@@ -301,3 +234,104 @@ igraph_error_t igraph_random_edge_walk(const igraph_t *graph,
...
+ * \param vertices An allocated vector, the result is stored here as
+ *   a list of vertex IDs. It will be resized as needed.
+ *   It includes vertex ids of starting and ending vertices.
+ * \param edges An initialized vector, the indices of traversed
+ *   edges are stored here. It will be resized as needed.
+ * \param start The start vertex for the walk.
+ * \param steps The number of steps to take. If the random walk gets
+ *   stuck, then the \p stuck argument specifies what happens.
+ *   Steps is the number of edges to traverse on the walk.
+ *   Length of the vertices vector: stepts + 1,
```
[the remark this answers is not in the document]
meetpatel0963: Sure. I removed the lengths from here and added \p in front of arguments.

src/paths/random_walk.c
```diff
@@ -301,3 +234,104 @@ igraph_error_t igraph_random_edge_walk(const igraph_t *graph,
...
+ * \param graph The input graph, it can be directed or undirected.
+ *   Multiple edges are respected, so are loop edges.
+ * \param weights A vector of non-negative edge weights. It is assumed
+ *   that at least one strictly positive weight is found among the
+ *   outgoing edges of each vertex. Additionally, no edge weight may
+ *   be NaN. If either case does not hold, an error is returned. If it
+ *   is a NULL pointer, all edges are considered to have equal weight.
+ * \param vertices An allocated vector, the result is stored here as
+ *   a list of vertex IDs. It will be resized as needed.
+ *   It includes vertex ids of starting and ending vertices.
```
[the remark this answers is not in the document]
meetpatel0963: Thanks for the advice. I've added length of vertices and edges vector under its own description as a pram.

src/paths/random_walk.c
```diff
@@ -301,3 +234,104 @@ igraph_error_t igraph_random_edge_walk(const igraph_t *graph,
...
+ * \param edges An initialized vector, the indices of traversed
+ *   edges are stored here. It will be resized as needed.
```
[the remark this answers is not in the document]
meetpatel0963: Thanks for the advice. I've added length of vertices and edges vector under its own description as a pram.

src/paths/random_walk.c
```diff
@@ -301,3 +234,104 @@ igraph_error_t igraph_random_edge_walk(const igraph_t *graph,
...
+ *   Length of the vertices vector: stepts + 1,
+ *   Length of the edges vector   : steps
+ * \param mode How to walk along the edges in directed graphs.
+ *   \c IGRAPH_OUT means following edge directions, \c IGRAPH_IN means
+ *   going opposite the edge directions, \c IGRAPH_ALL means ignoring
+ *   edge directions. This argument is ignored for undirected graphs.
+ * \param stuck What to do if the random walk gets stuck.
+ *   \c IGRAPH_RANDOM_WALK_STUCK_RETURN means that the function returns
+ *   with a shorter walk; \c IGRAPH_RANDOM_WALK_STUCK_ERROR means
+ *   that an error is reported. In both cases \p walk is truncated
```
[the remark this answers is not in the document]
meetpatel0963: Sure. It slipped out of my mind. I've corrected it.

src/paths/random_walk.c
```diff
@@ -301,3 +234,104 @@ igraph_error_t igraph_random_edge_walk(const igraph_t *graph,
...
+            else if (igraph_is_nan(min)) {
+                IGRAPH_ERROR("Weights must not contain NaN values.", IGRAPH_EINVAL);
+            }
+        }
+    }
+
+    if (edges || weights) {
+        return igraph_i_random_walk_inclist(graph, weights, vertices, edges,
+                                            start, mode, steps, stuck);
+    }
```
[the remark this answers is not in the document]
meetpatel0963: Yes, correct. I've added else block here.

tests/benchmarks/igraph_random_walk.c
```diff
@@ -24,27 +25,35 @@ int main() {
     }
 
     BENCH(" 1 Random edge walk,   directed,   unweighted, small graph  ",
```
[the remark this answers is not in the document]
meetpatel0963: Thanks for the advice. I didn't notice that. Corrected it.

tests/unit/igraph_random_walk.c
meetpatel0963: Sure, removed these comments.

tests/unit/igraph_random_walk.c
```diff
@@ -19,67 +19,216 @@
...
+    igraph_rng_seed(igraph_rng_default(), 42);
+    
     igraph_small(&g_1, 1, IGRAPH_UNDIRECTED, -1);
     igraph_small(&g_line, 5, IGRAPH_DIRECTED, 0,1, 1,2, 2,3, 3,4, -1);
     igraph_full(&g_full, 20, IGRAPH_UNDIRECTED, 0);
     igraph_small(&g_loop, 1, IGRAPH_DIRECTED, 0,0, -1);
 
+    /* This directed graph has loop edges.
+       It also has multi-edges when considered as undirected. */
+    igraph_de_bruijn(&graph, 3, 2);
```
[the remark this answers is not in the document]
meetpatel0963: Yes, that's a good idea. I've updated the graph name to g_de_bruijn and graph_weights to g_de_bruijn_weights.

meetpatel0963:
Hi @vtraag, Thank you very much for your comments and suggestions. I've added the requested changes.

src/paths/random_walk.c
```diff
@@ -132,85 +95,43 @@ static void vec_destr(igraph_vector_t *vec) {
...
- * Performs a random walk with a given length on a graph, from the given
- * start vertex. Edge directions are (potentially) considered, depending on
- * the \p mode argument.
- *
- * \param graph The input graph, it can be directed or undirected.
- *   Multiple edges are respected, so are loop edges.
+ * This function performs a random walk with a given length on a graph,
+ * from the given start vertex.
+ * It's used for igraph_random_walk: 
+ *  - when edge IDs of the traversed edges are needed (weighted or unweighted graphs)
```
vtraag, line 101: I would simplify the description here:

```suggestion
 * when weights are used or when edge IDs of the traversed edges are requested.
```

tests/unit/igraph_random_walk.c
vtraag: In principle the seed can be set only once in the entire test. However, it was mentioned explicitly in some of the `printf` statements that it was testing with seed 42, that's why I indicated multiple times to seed things. The RNG is now definitely seeded plenty :)

src/paths/random_walk.c
```diff
@@ -301,3 +240,152 @@ igraph_error_t igraph_random_edge_walk(const igraph_t *graph,
...
+ *   edge directions. This argument is ignored for undirected graphs.
+ * \param stuck What to do if the random walk gets stuck.
+ *   \c IGRAPH_RANDOM_WALK_STUCK_RETURN means that the function returns
+ *   with a shorter walk; \c IGRAPH_RANDOM_WALK_STUCK_ERROR means
+ *   that an error is reported. In both cases, \p edgewalk is truncated
+ *   to contain the actual interrupted walk.
+ *
+ * \return Error code.
+ *
+ * \deprecated-by igraph_strvector_set_len 0.10.0
```
vtraag, line 380:
```suggestion
 * \deprecated-by igraph_random_walk 0.10.0
```

src/paths/random_walk.c
```diff
@@ -32,86 +32,49 @@
...
-                      IGRAPH_PRId ".", IGRAPH_EINVAL, steps);
-    }
-
+    igraph_lazy_adjlist_t adj;
+    
     IGRAPH_CHECK(igraph_lazy_adjlist_init(graph, &adj, mode, IGRAPH_LOOPS, IGRAPH_MULTIPLE));
     IGRAPH_FINALLY(igraph_lazy_adjlist_destroy, &adj);
 
-    IGRAPH_CHECK(igraph_vector_int_resize(walk, steps));
+    IGRAPH_CHECK(igraph_vector_int_resize(vertices, steps));
```
vtraag, line 55:
```suggestion
    IGRAPH_CHECK(igraph_vector_int_resize(vertices, steps + 1));
```

Right?

src/paths/random_walk.c
```diff
@@ -301,3 +240,152 @@ igraph_error_t igraph_random_edge_walk(const igraph_t *graph,
...
+        IGRAPH_ERROR("Invalid mode parameter.", IGRAPH_EINVMODE);
+    }
+    
+    if (start < 0 || start >= vc) {
+        IGRAPH_ERRORF("Starting vertex must be between 0 and the "
+                      "number of vertices in the graph (%" IGRAPH_PRId
+                      "), got %" IGRAPH_PRId ".", IGRAPH_EINVAL,
+                      vc, start);
+    }
+    if (steps < 0) {
```
vtraag, line 317: I'm thinking briefly about the edge case for which `steps == 0`. In that case, `vertices[0]` will not be valid, but this is now not caught. Probably the easiest is just to add an `if (steps == 0)` on l. 336 just before calling the relevant subroutines to return without doing anything more.

src/paths/random_walk.c
```diff
@@ -301,3 +240,152 @@ igraph_error_t igraph_random_edge_walk(const igraph_t *graph,
...
+igraph_error_t igraph_random_walk(const igraph_t *graph,
+                       const igraph_vector_t *weights,
+                       igraph_vector_int_t *vertices,
+                       igraph_vector_int_t *edges,
+                       igraph_integer_t start,
+                       igraph_neimode_t mode,
+                       igraph_integer_t steps,
+                       igraph_random_walk_stuck_t stuck) {
+
+    /* TODO:
```
vtraag, line 300: I would remove this `TODO` comment. I don't think we'll ever implement this in C directly anyway, and people can just call higher-level function repeatedly to get multiple walks from multiple start vertices.

vtraag (changes requested):
Great, thank you so much! I just have two minor textual comments left. There is also one point about an edge case (`steps == 0`) and I think one vector should be resized slightly longer. After that, I believe we are all good to go!

src/paths/random_walk.c
meetpatel0963: Yes, that's actually better. 
```
 * It's used for igraph_random_walk: 
 *    when weights are used or when edge IDs of the traversed edges
 *    and/or vertex IDs of the visited vertices are requested.
```
Does this look fine?

src/paths/random_walk.c
meetpatel0963: Oh, Sorry for this. I was referring to some other PRs to see how the functions are deprecated and forgot to change this. I've corrected it.

src/paths/random_walk.c
meetpatel0963: Yes, you're right! Also, I need to change the for loop on l. 60 from
```C
for (i = 1; i < steps; i++)
```
to 
```C
for (i = 1; i <= steps; i++)
```

vtraag: Ah, yes, indeed, good catch!

src/paths/random_walk.c
vtraag: No problem, that happens...

src/paths/random_walk.c
meetpatel0963: Sure. I've added it. Also, added
```
* If \p steps is 0, the function returns without doing anything more.
```
under the description of steps params in doc.

src/paths/random_walk.c
meetpatel0963: Sure. I removed it.

src/paths/random_walk.c
vtraag: Thanks! Not sure if that's needed though, it's also implicit that if 0 steps are being taken, that you will simply get vectors of length 0 and 1 respectively right, so in a sense there's nothing special about `steps == 0`, we just need to make sure it works correctly.

meetpatel0963:
Thank you very much for your help and suggestions! I've added the requested changes.

vtraag (commented):
Sorry for the long delay, I got sick last week so it took me a bit longer to get back to this. The special case for `steps == 0` that I proposed was actually incorrect, sorry. If we can correct this now, then after these last changes, I believe we are all good to go.

CHANGELOG.md
```diff
@@ -429,7 +429,9 @@ Some of the highlights are:
    types of the nodes in the generated graph.
 
  - `igraph_random_walk()` now uses an `igraph_vector_int_t` for its
-   `walk` parameter.
+   `walk` parameter. Also, the function now takes both vertices and edges as
```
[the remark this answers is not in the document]
meetpatel0963: Not a problem. Sure, I'll update it. So now when steps=0 we'll return a single vertex and the edges vector will be empty, right?

ntamas:
Hmmm, what happened with this PR? Suddenly there are hundreds of files that appear as modified...

vtraag:
Yeah, good question, hadn't seen it previously. Probably something went wrong merging in the `develop` branch?

szhorvat:
I think it's fine now.

vtraag:
Thanks @szhorvat! GitHub garbled up something then, which is corrected by switching the base branch?

vtraag (approved):

vtraag:
Thanks @meetpatel0963! This was a wonderful contribution, we are very pleased with the high-quality code you provided. If you want to contribute something else in the future, that would be more than welcome.

meetpatel0963:
Thank you for all your help and suggestions @vtraag! Being new to open source, I believe it was a very good start for me. Sure, I'll love to work on other tasks soon.

ntamas:
Post-PR musings: I started wondering whether we used the term _edge walk_ correctly in the past. In our implementation, we always have an active vertex and a random incident edge is selected; the next active vertex will be the _other_ endpoint of that vertex. This means that it is possible to jump back and forth between the endpoints of that same edge so the "current" edge in the edge walk stays the same. However, in my interpretation of a proper edge walk, there is an active _edge_ instead, and the next active _edge_ is selected from the set of edges incident on at least one endpoint of the current edge, except the current edge, so it is not possible to encounter the same edge in two consecutive steps of the random walk. (In other words, the edge walk is a vertex walk on the line graph of the input graph).

This means (if my interpretation is correct) that at some point we need to get rid of `igraph_random_edge_walk()` in the C core, or transform it into a _proper_ edge walk with the properties mentioned above instead of piggybacking on `igraph_random_walk()`.

Issue #2015, opened 2022-03-26
Clean up attribute types


szhorvat:
Currently, we have the following attribute types:

```
typedef enum { IGRAPH_ATTRIBUTE_DEFAULT = 0,
               IGRAPH_ATTRIBUTE_NUMERIC = 1,
               IGRAPH_ATTRIBUTE_BOOLEAN = 5,
               IGRAPH_ATTRIBUTE_STRING = 2,
               IGRAPH_ATTRIBUTE_R_OBJECT = 3,
               IGRAPH_ATTRIBUTE_PY_OBJECT = 4
             } igraph_attribute_type_t;
```

There are custom types for R and Python. Are these truly needed? C code should never encounter a situation where _both_ R and Python objects are present.

I suggest one of the following refactorings:

### Option 1

```
typedef enum { IGRAPH_ATTRIBUTE_DEFAULT = 0,
               IGRAPH_ATTRIBUTE_NUMERIC = 1,
               IGRAPH_ATTRIBUTE_BOOLEAN = 2,
               IGRAPH_ATTRIBUTE_STRING = 3,
               IGRAPH_ATTRIBUTE_CUSTOM = 128
             } igraph_attribute_type_t;
```

We renumber existing enum values from 0 to 3 (as ABI backwards compatibility for 0.10 is long out the window anyway).

We introduce `IGRAPH_ATTRIBUTE_CUSTOM = 128`, which each high-level interface can treat as they wish.

### Option 2

```
typedef enum { IGRAPH_ATTRIBUTE_DEFAULT = 0,
               IGRAPH_ATTRIBUTE_NUMERIC = 1,
               IGRAPH_ATTRIBUTE_BOOLEAN = 2,
               IGRAPH_ATTRIBUTE_STRING = 3
             } igraph_attribute_type_t;
```

We renumber existing enum values from 0 to 3 (as ABI backwards compatibility for 0.10 is long out the window anyway).

We document that values 0-127 are reserved for future use, but 128 onward can safely be used by high-level interfaces. We do not introduce a specific named value, but in C enum values are proper integers anyway, so it's safe for a high-level interface create their own

```
#define IGRAPH_ATTRIBUTE_PY_LIST 128
#define IGRAPH_ATTRIBUTE_PY_DICT 129
#define IGRAPH_ATTRIBUTE_PY_OBJECT 130
...
```

or whatever. This way high-level interfaces can conveniently handle multiple types without having to bundle them up under a single one.

[...]

PR #2016, opened 2022-03-26
GML reader improvements

### LINKED ISSUE #174: Support graph attributes in GML reader/writer.
Should not be very hard, at least for non-composite attributes.

---

- date created: 2009-08-19T21:27:51Z
- owner: gabor.csardi
- the launchpad url was https://bugs.launchpad.net/bugs/416137

szhorvat:
### Done

 - Replaces `igraph_gml_string_t` with `char *` for simplicity. `igraph_gml_string_t` would be useful for non-null terminated strings with a specified length. This is indeed the way Bison/Flex handle strings. However, we need to copy the strings into the GML tree anyway, and we were already converting them to null-terminated.
 - Records line numbers into the GML tree, so they can be used when reporting some errors.
 - A warning is given for each composite attribute that is encountered and ignored, mentioning the attribute name. Note that no warning is given when only a _single value_ is ignored as opposed to all values, e.g. when one value of a certain attribute is composite but the others have simple types.
 - The default numerical attribute value is now NaN instead of 0. This is consistent with how the C attribute handler works in general, and helps the user sort out missing values from zero values.
 - Nodes with no IDs are no longer rejected. These are valid as isolated nodes according to the GML spec. These nodes get an `"id"` attribute value of NaN, in accordance with the above point.
 - Empty lists, `[ ]`, are now allowed, in accordance with the GML spec. This allows specifying isolated nodes as `node [ ]`.
 - The GML _writer_ will now check that the supplied ids are valid integers. If some are not, all are discarded. This fix was necessary because reading some GML files will now create id values of NaN.
 - Duplicate node 'id' fields, or edge 'source' / 'target' fields now cause an error, duplicate `directed` causes only a warning.
 - Graph attributes are now supported in the reader (the writer already supported them). Fixes #174.
 - GML writer will only show each warning at most once for each edge/vertex attribute. Previously, warnings were shown as many times as the edge/vertex count.
 - The behaviour of the `creator` argument is now different, see https://github.com/igraph/igraph/pull/2016#issuecomment-1079925812
 - More tests.

### Decisions

 - Duplicate `id` fields in `node` and duplicate `source`/`target` in `edge` are an error. Example: `node [ id 1 id 2 ]`.
 - Duplicate `directed` field in `graph` is only a warning for now, but it could be made an error as well.
 - When duplicate attributes are present, all but the last is ignored. Example: `node [ x 1 x 2 ]`. This is how this function behaved in the past, and implementing duplicate detection would be some trouble for little gain. According to the GML spec, this should be treated as an array named `x` with two entries, but we don't support this, and can't support this since igraph only has scalar types.

### Notes

 - `_` characters in key names are supported in the reader, but not the writer. The writer skips them. The GML spec does not indicate support for these characters, so I think this is fine. We want to be tolerant when reading malformed files, but try not to write them.


szhorvat:
@ntamas What do we do with the `creator` argument in `write_graph_gml()`? Currently, it does not control what is written in the `Creator` field, but (contrary to the documentation) it only sets what to add to the standard `igraph x.y.z` instead of the date.

I suggest the following:

 - If `creator` is `NULL`, do the same thing as now.
 - If `""`, omit the `Creator` field
 - Otherwise, write the value of `creator` verbatim, without the `igraph x.y.z`

szhorvat:
I do some ugliness here by storing `-1` in an `igraph_attribute_type_t`. This is not technically a valid value, but it happens to work as C represents enums with integers (but it does seem to use unsigned ints on my system!)

https://github.com/igraph/igraph/blob/be67455271fb7fd044acae709841680768e7ccab/src/io/gml.c#L304
https://github.com/igraph/igraph/blob/be67455271fb7fd044acae709841680768e7ccab/src/io/gml.c#L345
https://github.com/igraph/igraph/blob/be67455271fb7fd044acae709841680768e7ccab/src/io/gml.c#L355

The third of these lines linked above is prone to triggering compilers warnings (although it's not a warning that's enabled by default).

What do you think about using a legal, named values instead of -1, specifically `IGRAPH_ATTRIBUTE_DEFAULT` which is not used at the moment, @ntamas @vtraag ? See also #2015.

szhorvat:
I'm ready for a basic review before continuing with adding support for graph attributes.

src/io/gml.c
```diff
@@ -319,65 +339,89 @@ igraph_error_t igraph_read_graph_gml(igraph_t *graph, FILE *instream) {
...
+                        IGRAPH_ERRORF("Duplicate node id in GML file, line %" IGRAPH_PRId ".", IGRAPH_PARSEERROR,
+                                      igraph_gml_tree_line(node, j));
+                    }
                     hasid = 1;
                 }
             }
             if (!hasid) {
-                IGRAPH_ERROR("Node without 'id' while parsing GML file.", IGRAPH_PARSEERROR);
+                /* Isolated nodes are allowed not to have an id.
+                 * We generate an "n"-prefixed string id to be used in the trie. */
```
ntamas, line 387: Let's try to do something less predictable; I can imagine a graph where the first node is named `n1` and the next node is unnamed so it would also get `n1` as its name.

Ideas:

* Use `_generated_id_%d` or something similar that clearly indicates that this is a generated ID and that is less likely to be used in normal circumstances.
* Stick with `n%d` but then make sure to pick the first unused ID. If the generated ID is already used, increase the number until we obtain a name that is not used.

ntamas:
Re the `creator` field -- I'm not familiar with the GML spec, is this something that is mandated by the GML file format? If so, my preference would be to allow the user to override the creator field completely, so your proposal seems good (`NULL` should do what we do now, empty string should omit it if it is allowed by the GML spec, otherwise treat it as `NULL`, and any other value should override the creator field completely). After all, it could be the case that igraph is embedded in a larger app that wants to write its _own_ name and version number in GML files instead of igraph's.

ntamas:
> What do you think about using a legal, named values instead of -1, specifically IGRAPH_ATTRIBUTE_DEFAULT which is not used at the moment, @ntamas @vtraag ? See also https://github.com/igraph/igraph/issues/2015.

Yes, that's better than using -1 for sure.

src/io/gml.c
szhorvat: These generated IDs are never exposed to the user, and can never conflict. These are just used in the trie that translates GML ids to igraph vertex IDs.

Normally, GML ids can be any integer, while igraph vertex IDs must be consecutive integers. They way the GML reader works is that it converts the GML id to a string, and inserts it into a trie. Then the index of that string in the trie will be the igraph vertex ID. By adding the `n` prefix, we can ensure that these "generated string ids" never conflict with a explicit GML id. By keeping the prefix short (a single letter), we allow the trie to be more efficient for both storage and lookup.

szhorvat:
> I'm not familiar with the GML spec

It's a rather loose spec. You'll find a link to the Internet Archive copy of it in the WIkiepdia page: 

https://en.wikipedia.org/wiki/Graph_Modelling_Language
https://web.archive.org/web/20190303094704/http://www.fim.uni-passau.de:80/fileadmin/files/lehrstuhl/brandenburg/projekte/gml/gml-technical-report.pdf



> is this something that is mandated by the GML file format?

No, it is not mandatory.

src/io/gml.c
ntamas: Ah, okay, so there is no way for the user to create a node with ID equal to `n1`? That's okay then.

szhorvat: This is explained in the comments here:

https://github.com/igraph/igraph/blob/3dd467fcc074db9e8d5f54ef8566e9f5c66a8955/src/io/gml.c#L298

szhorvat:
@ntamas This is almost done. There's one little thing I need to decide about.  When writing a Boolean attribute, the GML writer shows a warning about converting it to numeric. This has already been so in the previous version. This warning is shown for every single edge and every single vertex that is written. Do we want to add a guard to prevent it from being printed more than once for each attribute? I have added other, new warnings for the GML writer, and prevented them from being printed more than once through `igraph_vector_bool_t ignore_warning_shown`, see e.g. [here](https://github.com/igraph/igraph/blob/b7dc73e787d16ab378265d09d0e024f33820f5b9/src/io/gml.c#L918). However, now I realize that all three of our high-level interfaces have some sort of protection against excessive warning message, so maybe it's not worth the effort.

szhorvat:
The rest of the code is ready for review, comments welcome.

ntamas:
Yes, I'd add a boolean guard so the warning is not printed more than once.

szhorvat:
Done, and also corrected the size of the `warning_shown` vector (must be the attribute count, not the edge/vertex count). Ready for review.

When merging, please do NOT squash. There's too much here to be all squashed.

ntamas:
This looks good to me now! We can merge if the CI runs pass.

szhorvat:
Alright, I'll merge. If there are any bugs, OSS-fuzz should find them.

PR #2017, opened 2022-04-02
Add complex matrix unit tests, docs, fix bugs

GroteGnoom:
ncol and nrow weren't correctly set after creation of a complex matrix.

### DIFF
--- a/src/core/matrix.c
+++ b/src/core/matrix.c
@@ -95,6 +95,20 @@ igraph_error_t igraph_matrix_complex_fprint(const igraph_matrix_complex_t *m,
     return IGRAPH_SUCCESS;
 }
 
+/**
+ * \ingroup matrix
+ * \function igraph_matrix_complex_real
+ * \brief Gives the real part of a complex matrix.
+ *
+ * \param v Pointer to a complex matrix.
+ * \param real Pointer to an initialized matrix. The result will be stored here.
+ * \return Error code.
+ *
+ * Time complexity: O(n),
+ * n is the
+ * number of elements in the matrix.
+ */
+
 igraph_error_t igraph_matrix_complex_real(const igraph_matrix_complex_t *v,
                                igraph_matrix_t *real) {
     igraph_integer_t nrow = igraph_matrix_complex_nrow(v);
@@ -104,6 +118,20 @@ igraph_error_t igraph_matrix_complex_real(const igraph_matrix_complex_t *v,
     return IGRAPH_SUCCESS;
 }
 
+/**
+ * \ingroup matrix
+ * \function igraph_matrix_complex_imag
+ * \brief Gives the imaginary part of a complex matrix.
+ *
+ * \param v Pointer to a complex matrix.
+ * \param imag Pointer to an initialized matrix. The result will be stored here.
+ * \return Error code.
+ *
+ * Time complexity: O(n),
+ * n is the
+ * number of elements in the matrix.
+ */
+
 igraph_error_t igraph_matrix_complex_imag(const igraph_matrix_complex_t *v,
                                igraph_matrix_t *imag) {
     igraph_integer_t nrow = igraph_matrix_complex_nrow(v);
@@ -112,6 +140,21 @@ igraph_error_t igraph_matrix_complex_imag(const igraph_matrix_complex_t *v,
     IGRAPH_CHECK(igraph_vector_complex_imag(&v->data, &imag->data));
     return IGRAPH_SUCCESS;
 }
+
+/**
+ * \ingroup matrix
+ * \function igraph_matrix_complex_realimag
+ * \brief Gives the real and imaginary parts of a complex matrix.
+ *
+ * \param v Pointer to a complex matrix.
+ * \param real Pointer to an initialized matrix. The real part will be stored here.
+ * \param imag Pointer to an initialized matrix. The imaginary part will be stored here.
+ * \return Error code.
+ *
+ * Time complexity: O(n),
+ * n is the
+ * number of elements in the matrix.
+ */
 
 igraph_error_t igraph_matrix_complex_realimag(const igraph_matrix_complex_t *v,
                                    igraph_matrix_t *real,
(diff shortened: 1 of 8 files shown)


szhorvat:
Can you please try:

```
#define _USE_MATH_DEFINES
#include <math.h>
```

instead of using an internal igraph header? There's value in keeping internal stuff out of tests when possible.

https://docs.microsoft.com/en-us/cpp/c-runtime-library/math-constants?view=msvc-170

I'm not sure if this will work _after_ `#include <igraph.h>`, but I hope it does. `igraph.h` will indirectly include `math.h` itself. It would be useful to find out.

szhorvat:
It's a bit disturbing to me that this function was broken and no one even noticed. Do we really need this functionality? Theoretically, it should be useful for some of the spectral stuff, but apparently it's not currently used. igraph should not aim to provide matrices for general use. There are much better libraries for that.

src/core/matrix.c
```diff
@@ -95,6 +95,20 @@ igraph_error_t igraph_matrix_complex_fprint(const igraph_matrix_complex_t *m,
...
+ * \param v Pointer to a complex matrix.
+ * \param real Pointer to an initialized matrix. The result will be stored here.
+ * \return Error code.
+ *
+ * Time complexity: O(n),
+ * n is the
+ * number of elements in the matrix.
+ */
+
 igraph_error_t igraph_matrix_complex_real(const igraph_matrix_complex_t *v,
```
szhorvat, line 112: Let's call the matrix `m` instead of `v`, both in the argument list and in the docs. Is it easy for you to do this with your IDE? If not, I can help. Most IDEs have a "rename" feature.

src/core/matrix.c
```diff
@@ -125,19 +168,88 @@ igraph_error_t igraph_matrix_complex_realimag(const igraph_matrix_complex_t *v,
...
 }
 
+/**
+ * \ingroup matrix
+ * \function igraph_matrix_complex_create_polar
+ * \brief Creates a complex matrix from a magnitude and an angle.
+ *
+ * \param v Pointer to an uninitialized complex matrix.
+ * \param r Pointer to the magnitude of the complex matrix.
+ * \param theta Pointer to the angle of the complex matrix.
```
szhorvat, line 221:
```suggestion
 * \param r Pointer to a real matrix containing magnitudes.
 * \param theta Pointer to a real matrix containing arguments (phase angles).
```

src/core/matrix.c
```diff
@@ -125,19 +168,88 @@ igraph_error_t igraph_matrix_complex_realimag(const igraph_matrix_complex_t *v,
...
+        VECTOR(v->data)[i] = igraph_complex(VECTOR(real->data)[i], VECTOR(imag->data)[i]);
+    }
+
     return IGRAPH_SUCCESS;
 }
 
+/**
+ * \ingroup matrix
+ * \function igraph_matrix_complex_create_polar
+ * \brief Creates a complex matrix from a magnitude and an angle.
```
szhorvat, line 217:
```suggestion
 * \brief Creates a complex matrix from a magnitude and phase angle.
```

src/core/matrix.c
```diff
@@ -125,19 +168,88 @@ igraph_error_t igraph_matrix_complex_realimag(const igraph_matrix_complex_t *v,
...
+
 igraph_error_t igraph_matrix_complex_create_polar(igraph_matrix_complex_t *v,
                                        const igraph_matrix_t *r,
                                        const igraph_matrix_t *theta) {
-    IGRAPH_CHECK(igraph_vector_complex_create_polar(&v->data, &r->data,
-                 &theta->data));
+    igraph_integer_t nrowr = igraph_matrix_nrow(r);
+    igraph_integer_t ncolr = igraph_matrix_ncol(r);
+    igraph_integer_t nrowi = igraph_matrix_nrow(theta);
+    igraph_integer_t ncoli = igraph_matrix_ncol(theta);
```
szhorvat, line 235: Nitpick: Can we rename these from `nrowi` to `nrowt` or similar? `r` and `i` look like real / imag.

tests/unit/matrix_complex.c
```diff
@@ -0,0 +1,93 @@
...
+    {
+        igraph_real_t e3[] = {1, 2, 3};
+        igraph_real_t e4[] = {5, 6, 7, 8};
+        printf("Check if unequal number of imaginary and real columns is handled correctly.\n");
+        igraph_matrix_view(&real, e3, 2, 1);
+        igraph_matrix_view(&imag, e4, 2, 2);
+        CHECK_ERROR(igraph_matrix_complex_create(&c, &real, &imag), IGRAPH_EINVAL);
+        CHECK_ERROR(igraph_matrix_complex_create_polar(&c, &real, &imag), IGRAPH_EINVAL);
+    }
+
```
szhorvat, line 91: Let's add another `VERIFY_FINALLY_STACK()` at the end.

src/core/matrix.c
```diff
@@ -125,19 +168,88 @@ igraph_error_t igraph_matrix_complex_realimag(const igraph_matrix_complex_t *v,
...
+    if (nrowr != nrowi) {
+        IGRAPH_ERRORF("Number of rows in real matrix (%" IGRAPH_PRId
+                ") not equal to number of rows in imaginary matrix (%"
+                IGRAPH_PRId ").", IGRAPH_EINVAL, nrowr, nrowi);
+    }
+    if (ncolr != ncoli) {
+        IGRAPH_ERRORF("Number of columns in real matrix (%" IGRAPH_PRId
+                ") not equal to number of columns in imaginary matrix (%"
+                IGRAPH_PRId ").", IGRAPH_EINVAL, ncolr, ncoli);
+    }
```
szhorvat, line 203: Nitpick: Can we simplify these two into a single, shorter message?

```
"Dimensions of real (%d by %d) and imaginary (%d by %d) matrices must match."
```

(Of course with `IGRAPH_PRId`.)

src/core/matrix.c
```diff
@@ -125,19 +168,88 @@ igraph_error_t igraph_matrix_complex_realimag(const igraph_matrix_complex_t *v,
...
+
+    igraph_matrix_complex_init(v, nrowr, ncolr);
```
szhorvat, line 205: Needs `IGRAPH_CHECK`?

src/core/matrix.c
```diff
@@ -104,6 +118,20 @@ igraph_error_t igraph_matrix_complex_real(const igraph_matrix_complex_t *v,
...
+ * \param v Pointer to a complex matrix.
+ * \param imag Pointer to an initialized matrix. The result will be stored here.
+ * \return Error code.
+ *
+ * Time complexity: O(n),
+ * n is the
+ * number of elements in the matrix.
+ */
+
 igraph_error_t igraph_matrix_complex_imag(const igraph_matrix_complex_t *v,
```
szhorvat, line 135: Let's call the matrix m instead of v, both in the argument list and in the docs.

src/core/matrix.c
```diff
@@ -125,19 +168,88 @@ igraph_error_t igraph_matrix_complex_realimag(const igraph_matrix_complex_t *v,
...
+ * \param real Pointer to the real part of the complex matrix.
+ * \param imag Pointer to the imaginary part of the complex matrix.
+ * \return Error code.
+ *
+ * Time complexity: O(n),
+ * n is the
+ * number of elements in the matrix.
+ */
+
 igraph_error_t igraph_matrix_complex_create(igraph_matrix_complex_t *v,
```
szhorvat, line 186: Let's call the matrix m instead of v, both in the argument list and in the docs. This notation is probably the result of a copy-paste form vector functions.

src/core/matrix.c
```diff
@@ -125,19 +168,88 @@ igraph_error_t igraph_matrix_complex_realimag(const igraph_matrix_complex_t *v,
...
+    if (nrowr != nrowi) {
+        IGRAPH_ERRORF("Number of rows in magnitude matrix (%" IGRAPH_PRId
+                ") not equal to number of rows in angle matrix (%"
+                IGRAPH_PRId ").", IGRAPH_EINVAL, nrowr, nrowi);
+    }
+    if (ncolr != ncoli) {
+        IGRAPH_ERRORF("Number of columns in magnitude matrix (%" IGRAPH_PRId
+                ") not equal to number of columns in angle matrix (%"
+                IGRAPH_PRId ").", IGRAPH_EINVAL, ncolr, ncoli);
+    }
```
szhorvat, line 246: Same nitpick as above about making the message shorter.

src/core/matrix.c
```diff
@@ -125,19 +168,88 @@ igraph_error_t igraph_matrix_complex_realimag(const igraph_matrix_complex_t *v,
...
+
+    igraph_matrix_complex_init(v, nrowr, ncolr);
```
szhorvat, line 248: Missing `IGRAPH_CHECK`.

szhorvat (changes requested):
I added a few comments.

GroteGnoom:
> It's a bit disturbing to me that this function was broken and no one even noticed. Do we really need this functionality? Theoretically, it should be useful for some of the spectral stuff, but apparently it's not currently used. igraph should not aim to provide matrices for general use. There are much better libraries for that.

It seems like they're only used in igraph_eigen_matrix, which itself seems to be only used in tests. So maybe we should remove complex matrices altogether? 

If we don't remove them, then the question remains if these functions are required. They seem pretty minimal to me, and the _real and _imag functions seem useful once something provides you with a complex matrix. We don't seem to have a real use for creating a complex matrix yourself, but it seems strange to only be able to split into real and imaginary parts, and not combine them.

We should probably conclude this first, before I go on to the other remarks :)

szhorvat:
I vote for keeping these functions for now, but I do not have a very strong opinion.

vtraag:
Unless we provide a useful implementation of complex matrices, I would vote to remove it if we don't need it ourselves.

GroteGnoom:
> Unless we provide a useful implementation of complex matrices, I would vote to remove it if we don't need it ourselves.

I'm not sure if I understand you correctly, but we do have a functional implementation right now, because the usual vector functions are also generated for complex matrices. I think only the functions for which you need an ordering are excluded.

vtraag:
Ah ok, from your earlier comments I got the impression that we didn't support complex matrices as well as regular matrices, but only provided a small part of the functionality. In that case, I'd say we should keep complex matrices in general.

ntamas:
+1 for keeping the functionality that we have now, but I also agree with @szhorvat that it should not be our goal to provide a generic matrix library.

OTOH I wonder whether the now-removed spectral coarse graining code made use of the complex matrix functionality.

src/core/matrix.c
GroteGnoom: vim handled it quite well once I understood how to do it :)

szhorvat (approved):

szhorvat:
Go ahead and merge when you're happy @GroteGnoom

PR #2023, opened 2022-04-03
Remove vector_permdelete

GroteGnoom:
It said it's for internal use only, but is not used internally.

It was once used here, for example: https://github.com/igraph/igraph/commit/0988d600b2e70454b8920c4394249230bbc06377#diff-a9e323228f51816f9d45d22cbcede355646cbd280c8132719b3c0ea3f01c21e4R633

### DIFF
--- a/include/igraph_vector_pmt.h
+++ b/include/igraph_vector_pmt.h
@@ -281,8 +281,6 @@ IGRAPH_EXPORT igraph_error_t FUNCTION(igraph_vector, move_interval)(
 IGRAPH_EXPORT IGRAPH_DEPRECATED igraph_error_t FUNCTION(igraph_vector, move_interval2)(
         TYPE(igraph_vector) *v, igraph_integer_t begin, igraph_integer_t end,
         igraph_integer_t to);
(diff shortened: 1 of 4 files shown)


ntamas (approved):

ntamas:
I've approved the changes in general, but I'd still like to put this up for a vote: shall we remove the `_permdelete()` function or not?

On one hand, it is not used for anything _inside_ igraph.
On the other hand, people might have started using it for stuff _outside_ igraph. This function provides functionality that is not meaningless outside igraph's context; it is a generic vector operation that basically generalizes shuffling the vector with an index vector in a way that also allows the deletion of some elements. I can imagine that this was actually used by people outside igraph, and if this is the case, we shouldn't remove it.

vtraag:
I would vote for removal.

szhorvat:
I agree with Tamás that this is potentially useful, and I won't vote for removal. But I also won't stand in it's way. I'll abstain this time.

vtraag:
So then we let @GroteGnoom break the tie? :)

PR #2026, opened 2022-04-04
remove array3

GroteGnoom:
It's not used anywhere, only in tests.

I opened a quick PR to show that everything in the C core still runs perfectly without these arrays. If we want a subtler approach or there a good reason to keep array3, it's fine to close this :)


ntamas:
Removal is fine by me but please mention it in the changelog.

szhorvat:
This is different from the "hashtable" which was not really a hashtable. 3D arrays are a more generally useful data structure, with much less flexibility in how they may be implemented. I could easily see a need for these in these future. If there is one, the implementation will look pretty much identical to what we have now.

I vote to keep it.

GroteGnoom:
> This is different from the "hashtable" which was not really a hashtable. 3D arrays are a more generally useful data structure, with much less flexibility in how they may be implemented. I could easily see a need for these in these future. If a need does arise, the implementation will look pretty much identical to what we have now.

If it isn't needed in all of igraph now, I can't really imagine what kinds of special cases we'd need it for in the future.

It's probably easier to add it back in if the need ever arises than to maintain it just in case.

szhorvat:
OK, now I'm going to point out that these are public functions, and summon @ntamas who will ask to deprecate it before removal :-)

----

There is one specific application that I have in mind for these 3D arrays. For quite a while now, I wanted to add functions to igraph which convert skeleton images to usable graphs. This included 3D images, so we need 3D arrays. This is an extremely common use case for biological image analysis, as well as for working with spatial networks in other contexts.

For my personal work, I usually use Mathematica's own functions for this operation, but they are unsatisfactory in many situations. So for a while now I've been planning to write my own.

GroteGnoom:
Ok, sounds cool :) Since you have a specific usage in mind, I'll just close this, and at least we can later find the argumentation for keeping it :) If I don't see skeletons in a year, I'll start complaining again ;)

PR #2038, opened 2022-04-16
add community_eb_get_merges tests

GroteGnoom:
I currently don't understand how the bridges result fits with the docs.

Why does this line
https://github.com/igraph/igraph/blob/develop/src/community/edge_betweenness.c#L287
`VECTOR(*bridges)[midx] = i + 1;`
add one to the index? It seems to me like it should just be `i`, not `i + 1`


ntamas:
That's weird for me as well, I also think it should be `i` and not `i + 1`.

szhorvat:
I double-checked that these are indeed one-based indices. I do not add one to them in the Mathematica interface, and the result I get this way is correct. 

Note that these are NOT edge IDs. They are step numbers, i.e. they can be used to index into the `result` vector.

In `functions.yaml` on the `develop` branch, this output argument is marked as `OPTIONAL OUT VERTEXSET bridges`, which is wrong in several ways: (1) these are not zero-based indices, so they need no +1 for R (2) they are definitely not vertex IDs, but also not edge IDs. Can you have a look at this @ntamas, also on `master` and in the R interface v1.3?

If this is changed, it will need to be marked clearly as a breaking change.

szhorvat:
On `master` it is `OUT VECTOR bridges`, so that's fine. But should `result` not be an `EDGESETLIST` or whatever a list of edge IDs is called on `master`, @ntamas ?

ntamas:
You are correct, the definition file should be fixed. Note that fixing the definition file is not a breaking change because the function is hand-written on the R side, and there we do not treat the `bridges` component of the result as a list of edge IDs.

szhorvat:
I mean that if we change from 1-based to 0-based indexing for 0.10, that's a breaking change that should be clearly noted.

I think it makes sense to do this change.

Is there a stimulus type that denotes an "index vector" (not merely an integer vector, but an index vector), @ntamas ?

GroteGnoom:
This is now one result of the tests:

```
Check graph with three vertices, two edges, remove one:
Bridges:0 -4702111234474983746
Modularity:-0.333333 -0.222222 -1.83255e-06
Membership:0 0 1
Merges:
1 0
0 -4702111234474983746
```
It's because the because the sizes of modularity, res and bridges are calculated up front using `no_of_nodes - no_comps`, which seems wrong. The size should at least depend on the number of edges that are removed. Maybe `no_of_nodes - no_comps` is an upper bound and everything should be resized to `midx` afterwards?

CHANGELOG.md
```diff
@@ -617,6 +617,8 @@ Some of the highlights are:
  - `igraph_adjacency()` no longer accepts a negative number of edges in its adjacency matrix.
    When negative entries are found, an error is generated.
 
+ - The `igraph_eb_get_merges()` bridges parameter now starts the edge indices at 0, not 1.
```
szhorvat, line 620: These are not edge indices. They are indices into the `result` vector, i.e. they represent the number of the step where the removal occurred.

GroteGnoom: You're right, I wrote the changelog before reading your comments, and I forgot to fix it.

ntamas:
Yes, using `no_of_nodes - no_comps` seems wrong; it is this way because initially this function was simply a helper function for edge betweenness, where we assumed that all the edges were removed, and that obviously the number of merges was `no_of_nodes - no_comps` because this many steps were needed to decompose everything into individual vertices.

This should be sorted out properly. I wouldn't care too much about efficiency here as this is a niche function that is unlikely to be used in performance-critical scenarios, so just go with whichever option feels simplest.

szhorvat:
> Indeed, but I'd consider that as a bugfix -- especially because the R interface currently adds +1 to the index vector so in R the indexing is effectively 2-based now, which absolutely doesn't make sense.

Just mention it under "breaking changes" in the changelog, so people are aware. I relied on this behaviour in the Mathematica interface and will need to remember to update it. I'll be reading the changelog for that.

GroteGnoom:
I think that removing 1 edge from a triangle should cause 0 merges, but the result is now 1. That's because igraph_i_community_eb_get_merges2 starts out with a membership vector where every vertex is it's own cluster. Any edge you choose is then counted as a merge. It seems plain incorrect to me, like it's checking what happens when _adding_ edges instead of removing them.

GroteGnoom:
All the edges in the graph need to be removed. That's in the docs, but I interpreted it wrong. I now added an error. This `solves` all the issues for only removing a few edges.

GroteGnoom:
The only error is `437 - test::glpk_error (SEGFAULT)`

ntamas:
Thanks!

szhorvat:
I think it would make sense to rename this function. It is not really related to edge betweenness in any way. Even the documentation states this. The naming since to be a leftover from a time when it was only a helper function for the Girvan-Newman community detector.

szhorvat:
But _if_ we remove it, it makes sense to think about whether it could be extended to the case when not all edges are removed, but only some of them. I don't have the mental energy for that now. Ping @vtraag since you are more fluent in clustering problems.

src/community/edge_betweenness.c
```diff
@@ -226,9 +230,29 @@ igraph_error_t igraph_community_eb_get_merges(const igraph_t *graph,
...
     igraph_integer_t no_comps;
+    igraph_integer_t no_removed_edges = igraph_vector_int_size(edges);
+    igraph_integer_t max_merges;
+    igraph_integer_t min;
+    igraph_integer_t max;
+
+    if (no_removed_edges > 0) {
+        igraph_vector_int_minmax(edges, &min, &max);
+        if (min < 0) {
+            IGRAPH_ERRORF("Edge ids should not be negative, found %" IGRAPH_PRId ".", IGRAPH_EINVAL, min);
```
szhorvat, line 245: Let us not refer to negative edge IDs, as it does not make sense in high-level interfaces that use 1-based indexing. Can we say something like "out-of-bounds (too small)"? Any other suggestions?

src/community/edge_betweenness.c
```diff
@@ -226,9 +230,29 @@ igraph_error_t igraph_community_eb_get_merges(const igraph_t *graph,
...
+        }
+        if (max > no_of_edges) {
+            IGRAPH_ERRORF("Edge id %" IGRAPH_PRId " is not in the graph.", IGRAPH_EINVAL, max);
```
szhorvat, line 248: Same comment as above. I would prefer to avoid printing edge IDs because they make no sense in either R or Mathematica.

szhorvat:
I'm wondering what happens when an edge ID appears twice in the input. Will it report an error cleanly?

szhorvat:
It looks like it returns a nonsense result, the same as when the number of removed edges is not the same as the edge count (i.e. as if that check didn't exist)?

It might be useful to guard against this: allocate a boolean vector of the same size as the edge count and tick off all edges as they are handled. Is this too big of a performance / memory use hit?

ntamas:
There are many more ways in which this function can be messed with; for instance, supplying invalid edge IDs are very likely to yield erroneous results without any error codes. Personally, I've always been thinking about this function as one that is meant for internal use only; in that sense it's kinda okay to have it as is. If we do anticipate that users will use this function on its own by supplying their own edge ID vectors, then of course we need more strict error checks, but in my opinion it is easier to make this function private instead as I don't think this would be used often by end users.

Note that the function is not even exposed in any of the higher-level interfaces so there mustn't have been overwhelming demand for it in the past.

szhorvat:
> supplying invalid edge IDs

That's checked for now, after @GroteGnoom's patch.

> Note that the function is not even exposed in any of the higher-level interfaces

True, but I think there's value in this function, and I was just planning to expose it in IGraph/M.

----

As far as I can tell, there's only one input check needed: make sure that each edge ID appears precisely once. Am I missing anything? If not, I'm willing to add this check.

ntamas:
I think that's all then, but note that there are actually two implementations (there's an internal static function) so make sure that the error checks affect both.

ntamas:
Plus, we might need a better name. This function does not simply "get the merges", it processes an edge removal sequence and calculates Lots-Of-Stuff<sup>TM</sup> based on it.

PR #2039, opened 2022-04-17
UMAP: Custom initial layout

iosonofabio:
Let the user choose a custom initial layout for UMAP, especially useful for perturbative layouting.

Mentioned in igraph/python-igraph#529.

TODO:
- [x] Test case
- [x] Check custom initial layout dimensions

### DIFF
--- a/src/layout/umap.c
+++ b/src/layout/umap.c
@@ -869,6 +869,21 @@ static igraph_error_t igraph_i_layout_umap(
         return IGRAPH_SUCCESS;
     }
 
+    if (skip_initialization) {
+        if((igraph_matrix_nrow(layout) != no_of_nodes) || (igraph_matrix_ncol(layout) != ndim)) {
+          IGRAPH_ERRORF("Initial layout should be %d x %d, got %d x %d.",
+                  IGRAPH_EINVAL, no_of_nodes, ndim,
+                  igraph_matrix_nrow(layout), igraph_matrix_ncol(layout));
+        }
+    } else {
+        /* Skip spectral embedding for now (see #1971), initialize at random */
+        if (ndim == 2) {
+            igraph_layout_random(graph, layout);
+        } else {
+            igraph_layout_random_3d(graph, layout);
+        }
+    }
+
     RNG_BEGIN();
     IGRAPH_VECTOR_INIT_FINALLY(&umap_weights, no_of_edges);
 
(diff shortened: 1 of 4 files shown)


iosonofabio:
Would you prefer I code a dimensional check for custom initial layouts? Something like if it's not the right number of rows/columns the function should fail early?

iosonofabio:
Clearly it was worth it, since it found a bug in the *only* test I had written :scream:

szhorvat:
This parameter is named `use_seed` in other layout functions. Can we make it consistent with them, in naming and behaviour?

ntamas:
I just wanted to say exactly what @szhorvat said, otherwise it seems fine to me

szhorvat:
One more comment:

When giving error messages about vertex coordinates, I would prefer to avoid referring to "matrices" directly. This is because the manner in which high-level interfaces expose coordinates may not be a matrix, and if it _is_ a matrix, it may be the transpose of what the C interface uses.

Can you re-write it in terms of something like "number of points" and "dimension" or "embedding dimension", or whatever representation-agnostic phrasing you prefer?

Finally, with actual matrices let's not use `3 x 4` but `3 by 4`, as `x` looks weird in some environments.

iosonofabio:
Thanks guys. I changed the name to `use_seed`, reordered the API function args to be consistent with other functions around, and changes the error message to not mention "matrces" anymore, but just "points in X dimensions".

I also added a test to catch a corner case for trivial graphs with a seed layout, which was segfaulting.

If it passes CI I'll merge.

src/layout/umap.c
```diff
@@ -860,11 +865,32 @@ static igraph_error_t igraph_i_layout_umap(
...
     IGRAPH_CHECK(igraph_i_umap_check_distances(distances, no_of_edges));
 
-    /* Trivial graphs (0 or 1 nodes) beget trivial - but valid - layouts */
-    if (no_of_nodes <= 1) {
-        IGRAPH_CHECK(igraph_matrix_resize(layout, no_of_nodes, ndim));
-        igraph_matrix_null(layout);
-        return IGRAPH_SUCCESS;
+    if (use_seed) {
+        if((igraph_matrix_nrow(res) != no_of_nodes) || (igraph_matrix_ncol(res) != ndim)) {
+          IGRAPH_ERRORF("Seed layout should have %d points in %d dimensions, got %d points in %d dimensions.",
```
szhorvat, line 870: `%d` is for `int` only, but we have `igraph_integer_t`, which maps to different types on different platforms. As a rule of thumb, you should never see `%d` in igraph code. If you do, it's likely a bug.

```suggestion
          IGRAPH_ERRORF("Seed layout should have %" IGRAPH_PRId " points in %" IGRAPH_PRId " dimensions, got %" IGRAPH_PRId " points in %" IGRAPH_PRId " dimensions.",
```

This will fix the test failure.

PR #2040, opened 2022-04-17
add st_vertex_connectivity tests

GroteGnoom:
I think two connected vertices with a IGRAPH_VCONN_NEI_IGNORE should return 0, not 1. Except for the direct path, there should be nothing to remove. Right?

In general, two connected vertices with a IGRAPH_VCONN_NEI_IGNORE seems to return one too many vertices, but maybe we want to change the docs, not the result.

### DIFF
--- a/src/flow/flow.c
+++ b/src/flow/flow.c
@@ -1763,10 +1763,12 @@ static igraph_error_t igraph_i_st_vertex_connectivity_check_errors(const igraph_
                                                     igraph_integer_t source,
                                                     igraph_integer_t target,
                                                     igraph_vconn_nei_t neighbors,
-                                                    igraph_bool_t *done) {
+                                                    igraph_bool_t *done,
+                                                    igraph_integer_t *no_conn) {
     igraph_integer_t no_of_nodes = igraph_vcount(graph);
     igraph_bool_t conn;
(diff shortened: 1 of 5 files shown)


ntamas:
I think that ` IGRAPH_VCONN_IGNORE` should mean "just pretend that the edge is not there, but I don't want to bother removing it". So, if there are two vertices that are _only_ connected by a single edge and there exists no other path between them, the result should be zero.

src/flow/flow.c
```diff
@@ -1758,51 +1758,88 @@ igraph_error_t igraph_mincut_value(const igraph_t *graph, igraph_real_t *res,
...
+        IGRAPH_CHECK(igraph_are_connected(graph, source, target, &conn));
+        if (conn) {
             *res = no_of_nodes;
             return IGRAPH_SUCCESS;
         }
         break;
     case IGRAPH_VCONN_NEI_IGNORE:
-        break;
+        {
+            igraph_vector_int_t neis;
```
GroteGnoom, line 1804: I will replace this with igraph_count_multiple

ntamas:
LGTM. Is this ready to merge now?

GroteGnoom:
Sure!

PR #2043, opened 2022-04-18
Walktrap fixes

### LINKED ISSUE #2041: Walktrap accepts negative weights
The walktrap community detection code accepts negative weights and NaN values. According to my recollection of how this algorithm works, such values should be disallowed. The weights are effectively used as non-normalized probabilities. Opinions, @vtraag ?

### LINKED ISSUE #2042: Walktrap community detection fails on some disconnected weighted graphs
**Describe the bug**

`igraph_community_walktrap()` fails on some disconnected weighted graphs.

This was reported here: https://igraph.discourse.group/t/number-of-steps-is-greater-than-number-of-rows-in-merges-matrix/1184?u=szhorvat

**To reproduce**

In Mathematica,

```
pg = IGShorthand["1,2-4,2-5,3-5,4-5", 
  EdgeWeight -> {0.2, 0.2, 0.1, 0.4}, VertexLabels -> Automatic, 
  EdgeLabels -> "EdgeWeight"]
```

image

```
In[285]:= IGCommunitiesWalktrap[pg]

During evaluation of In[285]:= IGraphM::error: src/community/community_misc.c:112 - Number of steps is greater than number of rows in merges matrix: found 4 steps, 3 rows.

During evaluation of In[285]:= IGraphM::error: igraph returned with error: Invalid value.

Out[285]= $Failed
```

I found this example with fuzzing, then manual tweaking of the weights for a simpler output. Original fuzzing code:

```
Do[g = RandomGraph[BernoulliGraphDistribution[5, 0.3], 
   EdgeWeight -> {_ :> RandomReal[{-.1, 0.5}]}];
 Check[IGCommunitiesWalktrap[g], Abort[]],
 Infinity]
g
IGEdgeProp[EdgeWeight][g]
```

(However, negative weights are not required.)

I cannot find examples with all-integer weights, or with no weights. I tried the first 1000 graphs from the graph atlas, and none failed.

**Version information**

Current `master` branch (i.e. soon after 0.9.8).

szhorvat:
Fixes #2041, fixes #2042

List of fixes, for pasting into the changelog (incomplete):

 - `igraph_community_walktrap()` would return an invalid `modularity` vector when the `merges` matrix was not requested.
 - `igraph_community_walktrap()` would return a `modularity` vector that was too long for disconnected graphs. This would cause a failure in some weighted graphs when the `membership` vector was requested.
 - `igraph_community_walktrap()` now uses double precision floating point operations internally instead of single precision.
 - `igraph_community_walktrap()` now checks the weight vector: only non-negative weights are accepted, and all vertices must have non-zero strength.

Notes:

 - I decided not to disallow zero weights, even though a vertex with zero strength would cause problems. If all vertices have a positive strength, everything should work fine.
 - The example program's output is unstable: the input graph is two cliques connected by a single edges. Thus a variety of cluster merge orders are equally valid. I made a small change (added an edge) so that the tests would pass, but keep in mind that this is fragile by nature.


szhorvat:
I am rounding modularity values that are very close to zero to exact zero _in the test_, otherwise we get slightly different values across platforms (notably between x87 and SSE math).

However, this always affects the last value in the modularity vector, which I think should be zero anyway. This corresponds to a partitioning into connected components, so the modularity is zero. Should we just coerce it to an exact zero directly in the walktrap code, @vtraag ?

szhorvat:
BTW the reason why I changes `float`s to `double`s is so that we can have consistent precision across the library. I would expect a value on the order of 10^-7 to be truly different from zero, while something on the order of 10^-15 is typically the result of roundoff errors. However, if we use `float`, we have 10^-7 due to roundoff errors.

This change increases memory usage and might decrease performance somewhat (I didn't measure). I'm not too concerned about this. Let me know if you want me to revert it.

szhorvat:
@ntamas I'd like to get this in before R/igraph 1.3.1.

szhorvat:
> However, this always affects the last value in the modularity vector, which I think should be zero anyway. This corresponds to a partitioning into connected components, so the modularity is zero. Should we just coerce it to an exact zero directly in the walktrap code, @vtraag ?

Other hierarchical methods suffer from the same problem: the last modularity value should be zero, but it is often only approximately zero. We don't need to deal with this before merging this PR. The priority is to include this in R/igraph 1.3.1.

szhorvat:
> However, this always affects the last value in the modularity vector, which I think should be zero anyway. This corresponds to a partitioning into connected components, so the modularity is zero. Should we just coerce it to an exact zero directly in the walktrap code, @vtraag ?

This is not true. The last value is not always zero in disconnected graphs. Ignore my comment please.

ntamas (approved):

ntamas:
Do you still need a review from Vincent or can I merge this?

vtraag:
I can take a look now, shouldn't take too long.

szhorvat:
@vtraag Ah, I just saw your comment. Since you're here, I would prefer that you take a look before merging.

vtraag:
> > However, this always affects the last value in the modularity vector, which I think should be zero anyway. This corresponds to a partitioning into connected components, so the modularity is zero. Should we just coerce it to an exact zero directly in the walktrap code, @vtraag ?
> 
> This is not true. The last value is not always zero in disconnected graphs. Ignore my comment please.

No, this is indeed not the case, as you say. However, modularity is guaranteed to be 0 for a partition consisting of a single community only. This is something that we could do specifically? (We could also do this in the `igraph_modularity` function.)

> * I decided not to disallow zero weights, even though a vertex with zero strength would cause problems. If all vertices have a positive strength, everything should work fine.

In principle, it would be fine to allow zero weights. It's not directly clear to me though if something will crash currently if it has zero strength? Or will calculations fail to be correct at some point, without it being clear that calculations are incorrect?

> The only theoretical issue here is: are there any uses/interpretations for negative weights that I missed? Is there any reason not to forbid negative weights?

I would indeed disallow negative weights, as they don't make sense in the context of modularity.

szhorvat:
> It's not directly clear to me though if something will crash currently if it has zero strength?

I was going to say that in the worst case you get some Inf/NaN, but nothing should blow up.

However, I just did some fuzzing by setting random edge weights, then setting the strength of one vertex to 0, and this is what I found:

 - By far the most common case where some modularity values are not finite is when we have a star graph, with the star centre having strength zero.
 - However, _very rarely_, I get a crash. So something can indeed blow up, but I don't know what yet.

szhorvat:
If we're short on time because of R/igraph 1.3.1, I would merge as-is and investigate this later. After all, this problem was present in the previous version as well so there's no regression.

szhorvat:
My fuzzing code, for reference:

```
n = 7; (* vertex count *)
v = 2; (* strength of v will be set to 0 *)
\

Do[g = IGTryUntil[ConnectedGraphQ]@
   RandomGraph[BernoulliGraphDistribution[n, 0.5], 
    EdgeWeight -> {_ :> RandomReal[]}, VertexLabels -> Automatic];
 AnnotationValue[{g, IncidenceList[g, v]}, EdgeWeight] = 0;
 cl = IGCommunitiesWalktrap[g];
 mod = cl["Modularity"];
 If[Not@VectorQ[mod, NumericQ], Break[]],
 {i, Infinity}
 ]
g
InputForm[cl]
```

ntamas:
Let's just merge as is and then we can return to this later.

szhorvat:
> I would indeed disallow negative weights, as they don't make sense in the context of modularity.

I didn't even think of modularity, just the interpretation of weights as (non-normalized) probabilities in walktrap.

The `spinglass` community detection does support negative weights and does compute some sort of modularity. How is this modularity defined then? It should at least be documented.

szhorvat:
The crash is likely related to comparisons of the `delta_sigma` sigma value (since comparison with NaN is always false, `! x < y` does not imply `x >= y`, so assumptions may be broken).

szhorvat:
@ntamas Wait a bit with the merge please, I will fix the zero-strength issue in a few min.

PR #2046, opened 2022-04-20
Add example demonstrating igraph_assortativity_nominal to example/simple

valdaarhun:
The example demonstrates that graphs with the same vertex types can have different assortativities.

Part of https://github.com/igraph/igraph/issues/1931.

### DIFF
--- a/examples/simple/igraph_assortativity_nominal.c
+++ b/examples/simple/igraph_assortativity_nominal.c
@@ -8,7 +8,7 @@ int main() {
     igraph_matrix_init(&pref_matrix, types, types);
 
     igraph_rng_seed(igraph_rng_default(), 42);
-    printf("Randomly generated graph with %ld nodes and %ld vertex types\n\n", (int)nodes, (int)types);
+    printf("Randomly generated graph with %" IGRAPH_PRId " nodes and %" IGRAPH_PRId " vertex types\n\n", nodes, types);
 
     /* Generate preference matrix giving connection probabilities for different vertex types */
     for (igraph_integer_t i = 0; i < types; i++) {
--- a/examples/simple/igraph_assortativity_nominal.c
+++ b/examples/simple/igraph_assortativity_nominal.c
@@ -8,7 +8,7 @@ int main() {
     igraph_matrix_init(&pref_matrix, types, types);
 
     igraph_rng_seed(igraph_rng_default(), 42);
-    printf("Randomly generated graph with %lld nodes and %lld vertex types\n\n", nodes, types);
+    printf("Randomly generated graph with %ld nodes and %ld vertex types\n\n", (int)nodes, (int)types);
 
     /* Generate preference matrix giving connection probabilities for different vertex types */
     for (igraph_integer_t i = 0; i < types; i++) {
@@ -19,11 +19,11 @@ int main() {
     
     igraph_t g;
     igraph_vector_int_t node_type_vec;
+    igraph_vector_int_init(&node_type_vec, nodes);
 
     // igraph_matrix_print(&pref_matrix);
     for (int i = 0; i < 5; i++) {
         igraph_real_t assortativity;
-        igraph_vector_int_init(&node_type_vec, nodes);
 
         /* Generate undirected graph with 1000 nodes and 50 vertex types */
         igraph_preference_game(&g, nodes, types, /* type_dist= */ NULL, /* fixed_sizes= */ 1, &pref_matrix, &node_type_vec, IGRAPH_UNDIRECTED, IGRAPH_LOOPS);
@@ -38,7 +38,7 @@ int main() {
         printf("Assortativity after rewiring = %g\n\n", assortativity);
 
         igraph_destroy(&g);    
-        igraph_vector_int_destroy(&node_type_vec);
     }
+    igraph_vector_int_destroy(&node_type_vec);
     igraph_matrix_destroy(&pref_matrix);
 }
--- a/examples/simple/igraph_assortativity_nominal.c
+++ b/examples/simple/igraph_assortativity_nominal.c
@@ -2,25 +2,26 @@
 #include <stdio.h>
 
 int main() {
-    igraph_integer_t nodes = 250, types = 5;
+    igraph_integer_t nodes = 120, types = 4;
 
     igraph_matrix_t pref_matrix;
     igraph_matrix_init(&pref_matrix, types, types);
 
     igraph_rng_seed(igraph_rng_default(), 42);
-    printf("Randomly generated graph with 250 nodes and 5 vertex types\n\n");
+    printf("Randomly generated graph with %ld nodes and %ld vertex types\n\n", nodes, types);
 
     /* Generate preference matrix giving connection probabilities for different vertex types */
     for (igraph_integer_t i = 0; i < types; i++) {
-        for (igraph_integer_t j = i; j < types; j++) {
-            MATRIX(pref_matrix, i, j) = RNG_UNIF01();
-            MATRIX(pref_matrix, j, i) = MATRIX(pref_matrix, i, j);
+        for (igraph_integer_t j = 0; j < types; j++) {
+            MATRIX(pref_matrix, i, j) = (i == j ? 0.1: 0.01);
         }
     }
 
+    // igraph_matrix_print(&pref_matrix);
+    igraph_vector_int_t node_type_vec;
+
     for (int i = 0; i < 5; i++) {
         igraph_t g;
-        igraph_vector_int_t node_type_vec;
         igraph_real_t assortativity;
         igraph_vector_int_init(&node_type_vec, nodes);
 
@@ -37,8 +38,8 @@ int main() {
         printf("Assortativity after rewiring = %g\n\n", assortativity);
 
         igraph_destroy(&g);
-        igraph_vector_int_destroy(&node_type_vec);
     }
-
+    
+    igraph_vector_int_destroy(&node_type_vec);
     igraph_matrix_destroy(&pref_matrix);
 }
--- a/examples/simple/igraph_assortativity_nominal.c
+++ b/examples/simple/igraph_assortativity_nominal.c
@@ -25,7 +25,7 @@ int main() {
     	igraph_vector_int_init(&node_type_vec, nodes);
 
         /* Generate undirected graph with 1000 nodes and 50 vertex types */
-        igraph_preference_game(&g, nodes, types, NULL, 1, &pref_matrix, &node_type_vec, IGRAPH_UNDIRECTED, IGRAPH_LOOPS);
+        igraph_preference_game(&g, nodes, types, /* type_dist= */ NULL, /* fixed_sizes= */ 1, &pref_matrix, &node_type_vec, IGRAPH_UNDIRECTED, IGRAPH_LOOPS);
 
         igraph_assortativity_nominal(&g, &node_type_vec, &assortativity, IGRAPH_UNDIRECTED, 1);
         printf("Assortativity before rewiring = %g\n", assortativity);
--- a/examples/simple/igraph_assortativity_nominal.c
+++ b/examples/simple/igraph_assortativity_nominal.c
@@ -2,9 +2,7 @@
 #include <stdio.h>
 
 int main() {
-    igraph_t g;
     igraph_integer_t nodes = 1000, types = 50;
-    igraph_vector_int_t node_type_vec;
 
     igraph_matrix_t pref_matrix;
     igraph_matrix_init(&pref_matrix, types, types);
@@ -21,15 +19,17 @@ int main() {
     }
 
     for (int i = 0; i < 5; i++) {
+        igraph_t g;
+        igraph_vector_int_t node_type_vec;
         igraph_real_t assortativity;
-    	igraph_vector_int_init(&node_type_vec, nodes);
+        igraph_vector_int_init(&node_type_vec, nodes);
 
         /* Generate undirected graph with 1000 nodes and 50 vertex types */
         igraph_preference_game(&g, nodes, types, /* type_dist= */ NULL, /* fixed_sizes= */ 1, &pref_matrix, &node_type_vec, IGRAPH_UNDIRECTED, IGRAPH_LOOPS);
 
         igraph_assortativity_nominal(&g, &node_type_vec, &assortativity, IGRAPH_UNDIRECTED, 1);
         printf("Assortativity before rewiring = %g\n", assortativity);
-        
+
         /* Rewire graph */
         igraph_rewire(&g, 10 * igraph_ecount(&g), IGRAPH_REWIRING_SIMPLE);
 
--- a/examples/simple/igraph_assortativity_nominal.c
+++ b/examples/simple/igraph_assortativity_nominal.c
@@ -25,7 +25,7 @@ int main() {
         igraph_t g;
 
         /* Generate undirected graph with 1000 nodes and 50 vertex types */
-        igraph_preference_game(&g, nodes, types, /* type_dist= */ NULL, /* fixed_sizes= */ 1, &pref_matrix, &node_type_vec, IGRAPH_UNDIRECTED, IGRAPH_LOOPS);
+        igraph_preference_game(&g, nodes, types, /* type_dist= */ NULL, /* fixed_sizes= */ 1, &pref_matrix, &node_type_vec, IGRAPH_UNDIRECTED, IGRAPH_NO_LOOPS);
 
         igraph_assortativity_nominal(&g, &node_type_vec, &assortativity, IGRAPH_UNDIRECTED, 1);
         printf("Assortativity before rewiring = %g\n", assortativity);
--- a/examples/simple/igraph_assortativity_nominal.c
+++ b/examples/simple/igraph_assortativity_nominal.c
@@ -17,13 +17,13 @@ int main() {
         }
     }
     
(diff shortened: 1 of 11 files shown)


examples/simple/igraph_assortativity_nominal.c
```diff
@@ -0,0 +1,44 @@
...
+            MATRIX(pref_matrix, j, i) = MATRIX(pref_matrix, i, j);
+        }
+    }
+
+    for (int i = 0; i < 5; i++) {
+        igraph_real_t assortativity;
+    	igraph_vector_int_init(&node_type_vec, nodes);
+
+        /* Generate undirected graph with 1000 nodes and 50 vertex types */
+        igraph_preference_game(&g, nodes, types, NULL, 1, &pref_matrix, &node_type_vec, IGRAPH_UNDIRECTED, IGRAPH_LOOPS);
```
szhorvat, line 28:
```suggestion
        igraph_preference_game(&g, nodes, types, /* type_dist= */ NULL, /* fixed_sizes= */ 1, &pref_matrix, &node_type_vec, IGRAPH_UNDIRECTED, IGRAPH_LOOPS);
```

examples/simple/igraph_assortativity_nominal.c
```diff
@@ -0,0 +1,44 @@
...
+    for(igraph_integer_t i = 0; i < types; i++){
+        for(igraph_integer_t j = i; j < types; j++){
+            MATRIX(pref_matrix, i, j) = igraph_rng_get_unif(igraph_rng_default(), 0, 1);
+            MATRIX(pref_matrix, j, i) = MATRIX(pref_matrix, i, j);
+        }
+    }
+
+    for (int i = 0; i < 5; i++) {
+        igraph_real_t assortativity;
+    	igraph_vector_int_init(&node_type_vec, nodes);
```
szhorvat, line 25: This line has a tab character. Please use only spaces.

src/misc/mixing.c
```diff
@@ -94,7 +94,9 @@
  * \sa \ref igraph_assortativity() for computing the assortativity
  * based on continuous vertex values instead of discrete categories.
  * \ref igraph_modularity() to compute generalized modularity.
- */
+ *
+ * \example examples/simple/igraph_assortativity_nominal.c
+*/
```
szhorvat, line 99:
```suggestion
 */
```

examples/simple/igraph_assortativity_nominal.c
```diff
@@ -0,0 +1,44 @@
...
+    igraph_matrix_t pref_matrix;
+    igraph_matrix_init(&pref_matrix, types, types);
+
+    igraph_rng_seed(igraph_rng_default(), 42);
+    printf("Randomly generated graph with 1000 nodes and 50 vertex types\n\n");
+
+    /* Generate preference matrix giving connection probabilities for different vertex types */
+    for(igraph_integer_t i = 0; i < types; i++){
+        for(igraph_integer_t j = i; j < types; j++){
+            MATRIX(pref_matrix, i, j) = igraph_rng_get_unif(igraph_rng_default(), 0, 1);
```
szhorvat, line 18:
```suggestion
            MATRIX(pref_matrix, i, j) = RNG_UNIF01();
```

examples/simple/igraph_assortativity_nominal.c
```diff
@@ -0,0 +1,44 @@
+#include <igraph.h>
+#include <stdio.h>
+
+int main(){
+    igraph_t g;
+    igraph_integer_t nodes = 1000, types = 50;
+    igraph_vector_int_t node_type_vec;
```
szhorvat, line 7: Since `g` and `node_type_vec` are used only in the loop, can you move their declarations there?

examples/simple/igraph_assortativity_nominal.c
```diff
@@ -0,0 +1,44 @@
...
+
+    igraph_matrix_t pref_matrix;
+    igraph_matrix_init(&pref_matrix, types, types);
+
+    igraph_rng_seed(igraph_rng_default(), 42);
+    printf("Randomly generated graph with 1000 nodes and 50 vertex types\n\n");
+
+    /* Generate preference matrix giving connection probabilities for different vertex types */
+    for(igraph_integer_t i = 0; i < types; i++){
+        for(igraph_integer_t j = i; j < types; j++){
```
szhorvat, line 17:
```suggestion
    for (igraph_integer_t i = 0; i < types; i++) {
        for (igraph_integer_t j = i; j < types; j++) {
```

examples/simple/igraph_assortativity_nominal.c
```diff
@@ -0,0 +1,44 @@
+#include <igraph.h>
+#include <stdio.h>
+
+int main(){
```
szhorvat, line 4:
```suggestion
int main() {
```

szhorvat (changes requested):
Thanks for the contribution! The general approach is good, but this example is too slow. It runs for almost an entire minute on the CI, in debug mode. Can you get it down below 0.1 seconds?

https://dev.azure.com/igraph-team/igraph/_build/results?buildId=3052&view=logs&j=4c3a8381-b9de-5463-904f-918810769be8&t=a906eab8-d2a3-5e54-d199-606dbbfb0805&l=907

The reason is that this graph has a very large number of edges due to using high connection probabilities. Try using a smaller and much sparser graph.

szhorvat:
Some more explanation:

 - Why is it so slow? Because tests are run with minimal optimization and a number of debugging options turned on (including address sanitizer and `IGRAPH_VERIFY_FINALLY_STACK`)
 - Why do we want to make it fast even in debug mode? Because we frequently run the test suite while working on igraph, thus having to wait for it has a real productivity cost. Examples are run as part of the test suite.

examples/simple/igraph_assortativity_nominal.c
valdaarhun: Sorry. I'll be more careful.

valdaarhun:
> Some more explanation:
> 
>     * Why is it so slow? Because tests are run with minimal optimization and a number of debugging options turned on (including address sanitizer and `IGRAPH_VERIFY_FINALLY_STACK`). That slow them down by 1-2 orders of magnitude.
> 
>     * Why do we want to make it fast even in debug mode? Because we frequently run the test suite while working on igraph, thus having to wait for it has a real productivity cost. Examples are run as part of the test suite.

Makes sense. I'll work on it ASAP.

szhorvat:
I resolved the conflict with `develop`. You'll need to `git pull` that change to your local repo.

valdaarhun:
Cmake fails on building docs. It says `/examples/simple/igraph_laplacian.c.xml` couldn't be loaded.

Also, `test::glpk_error` fails with SEGFAULT in the `static, x86_64` build.

How does one usually go about resolving these kind of errors?

ntamas:
Ignore the `glpk_error` failure; this happens all the time in the MinGW build, but it's non-deterministic.

As for the docs, I am refactoring the Laplacian functions in `develop` and something might have creeped in into this branch. I think we could also ignore this failure as it will be solved once we merge this PR into `develop`.

valdaarhun:
> Thanks for the contribution! The general approach is good, but this example is too slow. It runs for almost an entire minute on the CI, in debug mode. Can you get it down below 0.1 seconds?
> 
> https://dev.azure.com/igraph-team/igraph/_build/results?buildId=3052&view=logs&j=4c3a8381-b9de-5463-904f-918810769be8&t=a906eab8-d2a3-5e54-d199-606dbbfb0805&l=907
> 
> The reason is that this graph has a very large number of edges due to using high connection probabilities. Try using a smaller and much sparser graph.

The example now runs for 1-2 seconds.

https://dev.azure.com/igraph-team/igraph/_build/results?buildId=3116&view=logs&j=4c3a8381-b9de-5463-904f-918810769be8&t=a906eab8-d2a3-5e54-d199-606dbbfb0805&l=326

szhorvat:
Can you please try to reduce the timing to be comparable to most other tests, i.e. on the order of 0.01 seconds? This should be possible here by choosing the graph size and the inter-partition density appropriately.

I'll add some comments on the code to help with this.

examples/simple/igraph_assortativity_nominal.c
```diff
@@ -0,0 +1,44 @@
+#include <igraph.h>
+#include <stdio.h>
+
+int main() {
+    igraph_integer_t nodes = 250, types = 5;
```
szhorvat, line 5: Reduce graph size to speed up the calculation, e.g. 120 nodes, 4 types.

examples/simple/igraph_assortativity_nominal.c
```diff
@@ -0,0 +1,44 @@
...
+
+    igraph_matrix_t pref_matrix;
+    igraph_matrix_init(&pref_matrix, types, types);
+
+    igraph_rng_seed(igraph_rng_default(), 42);
+    printf("Randomly generated graph with 250 nodes and 5 vertex types\n\n");
```
szhorvat, line 11: Do not hard-code the numbers in this message. Use the variables defined above.

examples/simple/igraph_assortativity_nominal.c
```diff
@@ -0,0 +1,44 @@
...
+
+    /* Generate preference matrix giving connection probabilities for different vertex types */
+    for (igraph_integer_t i = 0; i < types; i++) {
+        for (igraph_integer_t j = i; j < types; j++) {
+            MATRIX(pref_matrix, i, j) = RNG_UNIF01();
+            MATRIX(pref_matrix, j, i) = MATRIX(pref_matrix, i, j);
+        }
+    }
```
szhorvat, line 19: Create a graph with high assortativity. Diagonal entries should be large, non-diagonal ones small. I suggest 0.1 for diagonal ones, 0.01 for the rest, and not using random numbers at all here.

examples/simple/igraph_assortativity_nominal.c
```diff
@@ -0,0 +1,44 @@
...
+
+    for (int i = 0; i < 5; i++) {
+        igraph_t g;
+        igraph_vector_int_t node_type_vec;
+        igraph_real_t assortativity;
+        igraph_vector_int_init(&node_type_vec, nodes);
```
szhorvat, line 25: Move the initialization/destruction of `node_type_vec` outside of the loop. This demonstrates that vectors can be re-used for a potential performance advantage.

examples/simple/igraph_get_laplacian_sparse.c
```diff
@@ -127,6 +127,7 @@ int test_normalized_laplacian(const igraph_vector_t *w, igraph_bool_t dir) {
     igraph_vector_int_t vec;
     igraph_vector_t *weights = 0;
     igraph_bool_t ok = 1;
+
```
szhorvat, line 130: Please revert changes that to files that are not related to this PR.

szhorvat:
@valdaarhun Please let us know if you're planning to complete this. No problem if you don't have time, I just want to know how to proceed.

valdaarhun:
Hi. I am sorry for not letting you know. I am actually in the middle of my end semester college exams right now. They get over on this Friday and then I would definitely like to get back to this issue.

szhorvat:
Good luck with your exams! There's no rush with this PR, we just needed to know how to proceed.

valdaarhun:
> Good luck with your exams!

Thank you!

valdaarhun:
Hi. My exams are done and I would like to get back to working on this PR.

examples/simple/igraph_get_laplacian_sparse.c
valdaarhun: I don't recall adding a blank line here. I think I did this accidentally.

ntamas:
Congrats and welcome back! Please go through the comments by @szhorvat first and address those that need to be addressed; mark the rest as resolved. We can then have another round of reviews.

valdaarhun:
> Congrats and welcome back! Please go through the comments by @szhorvat first and address those that need to be addressed; mark the rest as resolved. We can then have another round of reviews.

Ok, thanks. Will do.

examples/simple/igraph_get_laplacian_sparse.c
szhorvat: Probably just a small oversight while handling a merge conflict, no big deal.

examples/simple/igraph_assortativity_nominal.c
valdaarhun: Hi. I tried to move the declarations outside the loop to reduce the runtime. But this introduces a few memory leak issues such as [this one](https://dev.azure.com/igraph-team/igraph/_build/results?buildId=3360&view=logs&j=4c3a8381-b9de-5463-904f-918810769be8&t=e71559a0-54d3-55f3-64cc-74b67ad70398&l=389). The issues seem to originate in lines 28 and 35 of the program. Is this because it is expected to have an uninitialized data structure at the beginning of every iteration of the loop?

szhorvat: Only move the vector initialization/destruction outside (i.e. `node_type_vec`), not the graph.

`igraph_preference_game()` will create a new graph (i.e. it takes an uninitialized `igraph_t`), so it must stay inside of the loop.

szhorvat:
The `power_law_fit` failures on MinGW are known and are not related to what you are doing here. They can be ignored for now. We are investigating the cause of these (they may be due to an issue with GCC 12 in msys2 and not igraph).

examples/simple/igraph_assortativity_nominal.c
```diff
@@ -0,0 +1,44 @@
...
+#include <stdio.h>
+
+int main() {
+    igraph_integer_t nodes = 120, types = 4;
+
+    igraph_matrix_t pref_matrix;
+    igraph_matrix_init(&pref_matrix, types, types);
+
+    igraph_rng_seed(igraph_rng_default(), 42);
+    printf("Randomly generated graph with %ld nodes and %ld vertex types\n\n", (int)nodes, (int)types);
```
szhorvat, line 11: Use `IGRAPH_PRId` and remove the casts.

```suggestion
    printf("Randomly generated graph with %" IGRAPH_PRId " nodes and %" IGRAPH_PRId " vertex types\n\n", nodes, types);
```

examples/simple/igraph_assortativity_nominal.c
valdaarhun: Oh, I think I have got it. I moved the `igraph_t` part outside the loop as well which I shouldn't have done.

examples/simple/igraph_assortativity_nominal.c
valdaarhun: I couldn't find `IGRAPH_PRid` in the docs. I hope I didn't overlook it. It would be nice to have this in the docs as well, maybe in an API related miscellaneous section.

valdaarhun:
> Can you please try to reduce the timing to be comparable to most other tests, i.e. on the order of 0.01 seconds? 

The test now takes about as much time as the other tests running on the same platform.

examples/simple/igraph_assortativity_nominal.c
```diff
@@ -0,0 +1,44 @@
...
+    for (igraph_integer_t i = 0; i < types; i++) {
+        for (igraph_integer_t j = 0; j < types; j++) {
+            MATRIX(pref_matrix, i, j) = (i == j ? 0.1: 0.01);
+        }
+    }
+    
+    igraph_vector_int_t node_type_vec;
+    igraph_vector_int_init(&node_type_vec, nodes);
+
+    // igraph_matrix_print(&pref_matrix);
```
szhorvat, line 23:
```suggestion
```

examples/simple/igraph_assortativity_nominal.c
```diff
@@ -0,0 +1,44 @@
...
+    for (int i = 0; i < 5; i++) {
+        igraph_real_t assortativity;
+        igraph_t g;
+
+        /* Generate undirected graph with 1000 nodes and 50 vertex types */
+        igraph_preference_game(&g, nodes, types, /* type_dist= */ NULL, /* fixed_sizes= */ 1, &pref_matrix, &node_type_vec, IGRAPH_UNDIRECTED, IGRAPH_LOOPS);
```
szhorvat, line 29: Do not allow loops here, since `rewire` also called in a way that forbids loops.

```suggestion
        igraph_preference_game(&g, nodes, types, /* type_dist= */ NULL, /* fixed_sizes= */ 1, &pref_matrix, &node_type_vec, IGRAPH_UNDIRECTED, IGRAPH_NO_LOOPS);
```

szhorvat (commented):
Thank you. I'll resolve the two minor things I marked and will merge.

szhorvat:
@valdaarhun Could you please help and update the expected output after the switch from `LOOPS` to `NO_LOOPS`?

valdaarhun:
> @valdaarhun Could you please help and update the expected output after the switch from `LOOPS` to `NO_LOOPS`?

Sure. I'll do it right away.

valdaarhun:
I have updated the expected output.

valdaarhun:
`codecov/patch` seems to be failing. I haven't understood why that's the case.

valdaarhun:
Thank you. It was nice working on this issue. I got to learn a lot. If it's alright, I would like to also work on `igraph_realize_degree_sequence` as explained [here](https://github.com/igraph/igraph/pull/1946#issuecomment-1035974217).

szhorvat:
Thanks for the contribution! Feel free to go ahead.

PR #2048, opened 2022-05-01
tests: add write_graph_dot

GroteGnoom:
There is also an example, but this test also tests the attributes.

Fix one: remove undefined behaviour on NaN
Fix two: do not read out of bounds if orig parameter in
igraph_i_dot_escape has length 0

### DIFF
--- a/src/io/dot.c
+++ b/src/io/dot.c
@@ -73,7 +73,7 @@ static igraph_error_t igraph_i_dot_escape(const char *orig, char **result) {
             is_number = 0; need_quote = 1; newlen++;
         }
     }
-    if (is_number && orig[len - 1] == '.') {
+    if (is_number && len > 0 && orig[len - 1] == '.') {
         is_number = 0;
     }
     if (!is_number && isdigit(orig[0])) {
@@ -117,9 +117,8 @@ static igraph_error_t igraph_i_dot_escape(const char *orig, char **result) {
  * http://www.graphviz.org for details. The grammar of the DOT format
  * can be found here: http://www.graphviz.org/doc/info/lang.html
  *
- * </para><para>This is only a preliminary implementation, only the vertices
- * and the edges are written but not the attributes or any visualization
- * information.
+ * </para><para>This is only a preliminary implementation, no visualization
+ * information is written.
  *
  * \param graph The graph to write to the stream.
  * \param outstream The stream to write the file to.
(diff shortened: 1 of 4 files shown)


szhorvat:
It looks like GraphViz does not support a missing RHS of a `=` sign. It errors on `VAS=`:

```
dot igraph_write_graph_dot.dot
Error: igraph_write_graph_dot.dot: syntax error in line 16 near '='
```

The format specs are here: https://graphviz.org/doc/info/lang.html My reading is that we should just quote all strings for safety, and use `""` for empty strings.

It would be useful to test escape sequences. The spec I linked to above says that the only valid escape sequence is `\"`, but in some old versions `\n` was also allowed. Could you take a look and see if igraph handles these well?

ntamas:
SInce @GroteGnoom is working on igraph mostly during the weekends and I'd rather not let these fixes linger around in a PR, I'm inclined to merge this and we can fix the empty-string-RHS and the quotation issue in another PR. Any objections?

szhorvat:
It looks like `igraph_i_dot_escape()` only adds quotes when necessary, but I don't understand why. What is the disadvantage of adding quotes when not strictly needed?

Anyway, the simplest change might be to just add quotes when the string is empty.

Additionally, neither `\` nor newlines should be escaped according to the specs. Currently, both are escaped by igraph. However, if the string contains a newline, then quotes are necessary.

szhorvat:
> Any objections?

Make sure it doesn't get forgotten. If you don't have time to fix it now, I suggest opening an issue and assigning it to someone.

ntamas:
> It looks like igraph_i_dot_escape() only adds quotes when necessary, but I don't understand why. What is the disadvantage of adding quotes when not strictly needed?

Just started to write a comment about this :) It is for making it easier to do quick-and-dirty text processing stuff with `awk` and `sed` and stuff where you don't want to bother with un-escaping. It's mostly a philosophical argument but I tend to follow the don't-quote-unless-necessary principle everywhere. (E.g., CSV files are also much easier to parse if you don't quote every cell value).

ntamas:
Well, you'd still need to strip the quotes from around the item of interest so I'd rather leave it as is.

Note that making a fully specs-compliant GML parser is probably more work than what we'd like to commit ourselves to. For instance, there's the question of string encoding. GML assumes UTF-8 encoding, so if we really want to be fully specs-compliant, we must escape `"` _only_ if it is a glyph on its own and is not part of some _other_ glyph that happens to contain `"` in the encoded byte sequence. Also, to complicate things further, an input graph may specify `charset: ASCII` in a graph attribute, and _then_ we must parse the file as ASCII instead.

Also, this is also in the specs:

> Another way to avoid non-ASCII characters in labels is to use HTML entities for special characters. During label evaluation, these entities are translated into the underlying character. This [table](https://graphviz.org/doc/char.html) shows the supported entities, with their Unicode value, a typical glyph, and the HTML entity name.

This is also something that takes a lot of time if we want to get this fully correct.

szhorvat:
> Note that making a fully specs-compliant GML parser is probably more work than what we'd like to commit ourselves to.

Do you mean DOT parser? We only have a DOT writer, not a parser/reader. A DOT writer is important to have because it makes it possible to visualize igraph graph with GraphViz. Parsing is not important for us, as DOT isn't a common exchange format in my experience.

ntamas:
Okay, let me think.  You are probably right; I always mix up GML and DOT and I think I was mostly thinking about GML.

Also, I checked the UTF-8 encoding now, and luckily `"` cannot appear in any valid UTF-8 byte sequence except the one that encodes `"` itself, so we are okay, we don't need to go any more complicated with escaping as long as we only escape characters with an ASCII code less than 128.

ntamas:
Related bug filed in #2049

PR #2059, opened 2022-05-13
Basic character entity support for GML reader/writer

szhorvat:
`"` and `&` in string attributes values are now encoded to `&quot;` and `&amp;`, meaning that `"` no longer leads to corrupted (unreadable) GML files.

Decoding support is only added for [the five basic predefined entities](https://en.wikipedia.org/wiki/List_of_XML_and_HTML_character_entity_references#List_of_predefined_entities_in_XML).

Any other character entities or stray `&` characters are passed through unchanged, and left to be handled by the user. A warning is issued (at most once per read operation).

This means that if yEd writes an `&Eacute;` into a GML file, and this file is read in, then written back by igraph, we end up with `&amp;Eacute;` This is not ideal, but it's too complicated to deal with, and brings in a whole range of unpleasant issues, such as dealing with character encodings. What is the encoding of igraph string attribute values? We would need to make a decision about this before we can decode `&Eacute;`. (Yes, the GraphML reader is affected by this as well, and it produces consistent UTF-8: I documented this today.)

### DIFF
--- a/src/io/gml.c
+++ b/src/io/gml.c
@@ -62,7 +62,7 @@ static igraph_bool_t needs_coding(const char *str) {
 /* Encode & and " character in 'src' to &amp; and &quot;
  * '*dest' must be deallocated by the caller.
  */
-static igraph_error_t entity_encode(const char *src, char **dest) {
+static igraph_error_t entity_encode(const char *src, char **dest, igraph_bool_t only_quot) {
     igraph_integer_t destlen = 0;
     const char *s;
     char *d;
@@ -70,7 +70,10 @@ static igraph_error_t entity_encode(const char *src, char **dest) {
     for (s = src; *s != '\0'; s++, destlen++) {
         switch (*s) {
         case '&': /* &amp; */
-            destlen += 4; break;
+            if (! only_quot) {
+                destlen += 4;
+            }
+            break;
         case '"': /* &quot; */
             destlen += 5; break;
         }
@@ -80,7 +83,13 @@ static igraph_error_t entity_encode(const char *src, char **dest) {
     for (s = src, d = *dest; *s != '\0'; s++, d++) {
         switch (*s) {
         case '&':
-            strcpy(d, "&amp;"); d += 4; break;
+            if (! only_quot) {
+                strcpy(d, "&amp;");
+                d += 4;
+            } else {
+                *d = *s;
+            }
+            break;
         case '"':
             strcpy(d, "&quot;"); d += 5; break;
         default:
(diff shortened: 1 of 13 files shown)


szhorvat:
We could maybe add an option to the writer to disable encoding the `&` character, and only encode `"`. Then round-tripping through igraph will be fine: it will no longer break unsupported entities.

szhorvat:
It still bothers me that reading and re-exporting technically valid files with this would break them by substituting `&amp;aacute;` for `&aacute;`.

My preference is to add the option to _not_ encode `&` on export, and to do it with a bit-field solution for future extensibility, just as we do with `igraph_edge_type_sw_t`.

ntamas:
> What is the encoding of igraph string attribute value?

I'd say it's a _de facto_ UTF-8 in the sense that no one claimed officially that it's UTF-8, but we should probably treat it like that as it seems to be the only sensible choice.

> Instead of a boolean, this option could be implemented through a bit field for future extensibility.

I think we should have an `igraph_gml_options_t` and we should also move the `creator` string and the `id` vector there, the default being whatever we do now by default. In 99% of the cases I guess you only want to specify the graph and the output file, nothing else.

src/io/gml.c
```diff
@@ -762,6 +902,7 @@ static igraph_error_t igraph_i_gml_convert_to_key(const char *orig, char **key)
  * \param id Either <code>NULL</code> or a numeric vector with the vertex IDs.
  *        See details above.
  * \param creator An optional string to write to the stream in the creator line.
+ *        It must not contain the <code>"</code> character.
```
ntamas, line 905: Why not? Aren't we encoding it now as `&quot;`?

szhorvat: Because I was lazy. I can fix this, no problem.

szhorvat:
> I think we should have an `igraph_gml_options_t` and we should also move the `creator` string and the `id` vector there, the default being whatever we do now by default. In 99% of the cases I guess you only want to specify the graph and the output file, nothing else.

What would this achieve?

Using the function is not difficult at the moment. Both `creator` and `id` accept `NULL`. However, in several tests, `creator` must be set to `""` instead of `NULL` to avoid printing the date. I am concerned that introducing a `igraph_gml_options_t` would make some use cases significantly more complicated (`creator=""`) without making the typical usage much simpler.

It also wouldn't allow extensibility without breaking the ABI, so it doesn't really do much in the end. If we want some level of option-extensibility without breaking the ABI, we can use a bitfield (like libxml2), but I'd be happy with a simple boolean too, as I don't expect to want to add much more in the future.

ntamas:
As discussed on Zoom with @szhorvat : we'll go ahead with a bit field argument instead of a full `igraph_gml_options_t`.

@szhorvat I'm marking this as draft; let me know when you're done with it and then I'll merge (or you can do it).

szhorvat:
@ntamas Have a look at the stimulus part. I used the arbitrary value `DEFAULT`, but I don't recall where these are defined (intrerface-specific generator?) and whether the name should be prefixed, e.g. `WRITE_GML_DEFAULT`.

ntamas:
The values are defined in the high-level language specific parts, and there is no need for the prefix, so it's OK.

ntamas:
Awesome, thanks a lot!

PR #2067, opened 2022-05-14
return weighted_adjacency weights as a vector

### LINKED ISSUE #1978: weighted_adjacency() should work without attribute handler
`igraph_weighted_adjacency()` requires an attribute handler. Almost all other igraph functions will work without an attribute handler, and I think this one should too. igraph aims to be simpler to use than other graph libraries, even from C, and a beginner should not need to bother with learning the attribute handling interface.

I propose a function which stores the weights in a separate vector that does not need to be an attribute. We can either have only one such function, or we can keep the current function and add an extra one that returns the weights in a vector.

This is a very easy change. I only need feedback before I proceed with it.

GroteGnoom:
fixes #1978

We should of course still decide if we want this as a separate function. As far as I can tell, the current functionality just shouldn't exist.


szhorvat:
I just wanted to note that we haven't actually discussed whether we actually wanted to do what I proposed in #1978.  I never got any feedback on it. Needless to say, I'm in favour :-)

vtraag:
Completely in favour too! I'm fine with simply replacing this function, such that is accepts weights directly instead of attributes.

ntamas:
I'm in favour as well.

ntamas:
I must have caused conflicts for this PR by committing 1bc70e4d4 so it's my turn to fix it :)

ntamas:
Thanks!

PR #2070, opened 2022-05-21
Use parent vector of #1880 for get_shortest_paths

GroteGnoom:
get_shortest_paths_dijkstra too, because that passes results on to get_shortest_paths if there are no weights.


ntamas:
Thanks a lot! I've renamed some variables and will merge if the CI tests pass.

GroteGnoom:
I re-ran the tests, and they now mostly seem to pass :)

ntamas (approved):

ntamas:
Merged, thanks!

PR #2072, opened 2022-05-21
Add igraph_sparse_adjacency

### LINKED ISSUE #1838: Create graph based on `igraph_sparsemat_t`
Construct a graph based on a `sparsemat_t`. This could simply mimic existing functions `igraph_adjacency` and `igraph_weighted_adjacency`, but then `igraph_sparse_adjacency` and `igraph_sparse_weighted_adjacency`.

**Use cases for the feature**
Create graphs based on sparse matrix results from other `igraph` functions.

GroteGnoom:
includes the weighted counterpart.

Fixes #1838

issues:
- Some functions call igraph_sparsemat_index in the inner loop, and that seems like it will be slow.
- src/core/sparsemat.c defines CS_LONG 1 before including cs/cs.h. Should every source file that includes cs do this? Maybe this should be in igraph_sparsemat.h?

### DIFF
--- a/include/igraph_constants.h
+++ b/include/igraph_constants.h
@@ -60,7 +60,7 @@ typedef enum { IGRAPH_RECIPROCITY_DEFAULT = 0,
              } igraph_reciprocity_t;
 
 typedef enum { IGRAPH_ADJ_DIRECTED = 0,
-               IGRAPH_ADJ_UNDIRECTED = 1, IGRAPH_ADJ_MAX = 1,
+               IGRAPH_ADJ_UNDIRECTED, IGRAPH_ADJ_MAX,
                IGRAPH_ADJ_UPPER, IGRAPH_ADJ_LOWER, IGRAPH_ADJ_MIN,
                IGRAPH_ADJ_PLUS
              } igraph_adjacency_t;
--- a/include/igraph_constants.h
+++ b/include/igraph_constants.h
@@ -60,9 +60,10 @@ typedef enum { IGRAPH_RECIPROCITY_DEFAULT = 0,
              } igraph_reciprocity_t;
 
 typedef enum { IGRAPH_ADJ_DIRECTED = 0,
-               IGRAPH_ADJ_UNDIRECTED, IGRAPH_ADJ_MAX,
+               IGRAPH_ADJ_UNDIRECTED,
                IGRAPH_ADJ_UPPER, IGRAPH_ADJ_LOWER, IGRAPH_ADJ_MIN,
-               IGRAPH_ADJ_PLUS
+               IGRAPH_ADJ_PLUS,
+               IGRAPH_ADJ_MAX,
              } igraph_adjacency_t;
 
 typedef enum { IGRAPH_STAR_OUT = 0, IGRAPH_STAR_IN,
--- a/src/constructors/adjacency.c
+++ b/src/constructors/adjacency.c
@@ -1129,6 +1129,18 @@ static igraph_error_t igraph_i_sparse_adjacency_undirected(
     return IGRAPH_SUCCESS;
 }
 
+/**
+ * \ingroup generators
+ * \function igraph_sparse_adjacency
+ * \brief Creates a graph from a sparse adjacency matrix.
+ *
+ * This has the same functionality as \ref igraph_adjacency(), but uses
+ * a column-compressed adjacency matrix.
+ *
+ * Time complexity: O(|E|),
+ * where |E| is the number of edges in the graph.
+ */
+
 igraph_error_t igraph_sparse_adjacency(igraph_t *graph, igraph_sparsemat_t *adjmatrix,
         igraph_adjacency_t mode, igraph_loops_t loops) {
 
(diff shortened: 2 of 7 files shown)


include/igraph_constants.h
```diff
@@ -60,7 +60,7 @@ typedef enum { IGRAPH_RECIPROCITY_DEFAULT = 0,
              } igraph_reciprocity_t;
 
 typedef enum { IGRAPH_ADJ_DIRECTED = 0,
-               IGRAPH_ADJ_UNDIRECTED = 1, IGRAPH_ADJ_MAX = 1,
```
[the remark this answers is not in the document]
szhorvat: Renumbering only breaks ABI, not API, so it should have no drawbacks at all on `develop`.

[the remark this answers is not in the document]
ntamas: Not quite; if the caller happened to hardcode the enum value in the calling code (which shouldn't have happened, but most compilers only treat this as a warning), then this could break. This is not as far-fetched as it seems; take a look at `types-RR.yaml` in `igraph/rigraph`, which does this all the time (unfortunately). Luckily enough the code for `igraph_weighted_adjacency()` is not used there, but let's not get into the habit of randomly reassigning numeric values of enums just because it formally does not break the API.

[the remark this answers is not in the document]
szhorvat: Yes, but that's the caller's problem, and it _should_ really be fixed there ... In the Mathematica interface I also pass integers from Mma to C code, which represent these choices. But there is always an explicit `switch` which maps these integers (which are a concern of the Mma interface only) to proper named enum values. For example: https://github.com/szhorvat/IGraphM/blob/master/IGraphM/LibraryResources/Source/IG.h#L57

I'm not saying that we should just go ahead and renumber anything without a good justification, but this should really really be fixed on the R/Stimulus side. In the end, Stimulus needs to be made aware of the enum value names that we use in C. There's no other way that's reliable. If it operates with integers at the moment that's not okay.

src/constructors/adjacency.c
```diff
@@ -19,13 +19,17 @@
...
 */
+#include "igraph_sparsemat.h"
 
 #include "igraph_constructors.h"
 
 #include "igraph_adjlist.h"
 #include "igraph_attributes.h"
 #include "igraph_interface.h"
 
+#include <cs/cs.h> //TODO take a look at how this is done in /src/core/sparsemat.c. i didn't want to copy it all. Maybe we need a separate header?
```
ntamas, line 30: Is this TODO comment still relevant?

src/constructors/adjacency.c
```diff
@@ -19,13 +19,17 @@
    02110-1301 USA
 
 */
+#include "igraph_sparsemat.h"
```
ntamas, line 22: Move this #include to the second group of includes; the usual rule is that we first `#include` the igraph header that directly "belongs to" the file, then any additional headers that we need, in a separate group.

src/constructors/adjacency.c
```diff
@@ -508,6 +532,20 @@ static igraph_error_t igraph_i_weighted_adjacency_max(
...
 }
 
+static igraph_error_t igraph_i_weighted_adjacency_undirected(
+    const igraph_matrix_t *adjmatrix,
+    igraph_vector_int_t *edges,
+    igraph_vector_t *weights,
+    igraph_loops_t loops
+) {
+    if (!igraph_matrix_is_symmetric(adjmatrix)) {
+        IGRAPH_ERROR("Adjacency matrix should be symmetric for IGRAPH_ADJ_UNDIRECTED.",
```
ntamas, line 542: Can we also add a test case for this branch?

src/constructors/adjacency.c
```diff
@@ -858,3 +900,609 @@ igraph_error_t igraph_adjlist(igraph_t *graph, const igraph_adjlist_t *adjlist,
...
+        IGRAPH_CHECK(igraph_i_sparse_adjacency_upper(adjmatrix, &edges, loops));
+        break;
+    case IGRAPH_ADJ_LOWER:
+        IGRAPH_CHECK(igraph_i_sparse_adjacency_lower(adjmatrix, &edges, loops));
+        break;
+    case IGRAPH_ADJ_MIN:
+        IGRAPH_CHECK(igraph_i_sparse_adjacency_min(adjmatrix, &edges, loops));
+        break;
+    case IGRAPH_ADJ_PLUS:
+        IGRAPH_CHECK(igraph_i_sparse_adjacency_directed(adjmatrix, &edges, loops));
```
ntamas, line 1190: This should be `igraph_i_sparse_adjacency_plus()` -- but since this issue was not caught by the unit tests, I wonder whether this indicates that the tests are not thorough enough (or that in the sparse case the `_plus()` variant is actually not needed at all?).

ntamas (changes requested):
I've added a few comments inline, but my biggest concern right now is that we need to `#include <cs/cs.h>` and reach deep down into the internals of the sparse matrix in order to do our job. This probably indicates that the sparse matrix API is not good enough; if these functions were not implemented in igraph and an end user wanted to add something similar, he would not be able to do so without working with the CSparse internals either.

What is the primary thing that we are missing from the sparse matrix API that necessitated using the internal directly? Is it the iteration over nonzero entries? Can we use `igraph_sparsemat_iterator_t` for that, or can we simply get all nonzero entries with `igraph_sparsemat_getelements_sorted()` and just work from that?

src/constructors/adjacency.c
GroteGnoom: I think it was completely by accident, but yes they should be equal. I've removed _plus for the unweighted version.

PR #2079, opened 2022-05-26
Use eccentricity to find graph centres

### LINKED ISSUE #2027: Graph center

A function that computed the [graph center](https://en.wikipedia.org/wiki/Graph_center), i.e the set of vertices with lowest eccentricity.

**Use cases for the feature**

This is a standard concept in graph theory.

**References**

 - https://en.wikipedia.org/wiki/Graph_center
 - `igraph_eccentricity()`

pradkrish:
Partially fixes #2027


src/paths/distances.c
```diff
@@ -212,3 +212,34 @@ int igraph_radius(const igraph_t *graph, igraph_real_t *radius,
...
+
+    igraph_vector_t ecc;
+    igraph_vit_t vit;
+    igraph_vs_t vids = igraph_vss_all();
+
+    igraph_vector_init(&ecc, 0);
+
+    igraph_eccentricity(graph, &ecc, vids, mode);
+    IGRAPH_CHECK(igraph_vit_create(graph, vids, &vit));
+    if (!IGRAPH_VIT_END(vit)) {
```
szhorvat, line 232: This loop is unnecessary. We only need to find the vertices with maximum eccentricity, so we only need to loop over the eccentricity vector.

szhorvat (commented):
I added a few preliminary comments, but I think it will be easier for you to open a new PR against `develop` instead of rebasing this one. It's up to you which way to proceed though.

szhorvat:
FYI the test failures with MinGW are not related to your code, and have just been worked around `master` and `develop`. After you rebase this on `develop` (or open a new PR against `develop`), this annoyance will be gone.

src/paths/distances.c
```diff
@@ -212,3 +212,34 @@ int igraph_radius(const igraph_t *graph, igraph_real_t *radius,
...
+}
+
+int igraph_center(const igraph_t *graph, igraph_vector_t *res,
+                  igraph_neimode_t mode) {
+
+    igraph_vector_t ecc;
+    igraph_vit_t vit;
+    igraph_vs_t vids = igraph_vss_all();
+
+    igraph_vector_init(&ecc, 0);
```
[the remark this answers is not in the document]
pradkrish: Ack

src/paths/distances.c
pradkrish: The idea of adding the condition `if (!IGRAPH_VIT_END(vit))` is to check for Null graph. If this is not correct, is there another way of doing it?

src/paths/distances.c
```diff
@@ -212,3 +212,34 @@ int igraph_radius(const igraph_t *graph, igraph_real_t *radius,
 
     return 0;
 }
+
+long int min(long int a, long int b) 
+{
+    return (a > b) ? b : a;
+}
+
+int igraph_center(const igraph_t *graph, igraph_vector_t *res,
```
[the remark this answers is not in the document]
pradkrish: Ack

pradkrish:
> FYI the test failures with MinGW are not related to your code, and have just been worked around `master` and `develop`. After you rebase this on `develop` (or open a new PR against `develop`), this annoyance will be gone.

Yeah, it was an oversight on my part, I thought I was on develop branch. Also, it looks like the unit tests I have added is not building as part of the test suite.. I have added it to `tests/CMakeList.txt`. Am I missing something here?

src/paths/distances.c
szhorvat: Good point that we must check for the null graph, otherwise `vector_min` fails. Do this by checking the vertex count.

pradkrish:
That doesn't look good at all, best to abandon this and open a new PR.

Issue #1988, opened 2022-03-04
Proposal for refactoring the RNG interface


szhorvat:
Here is a concrete proposal for refactoring the RNG interface, aiming for simplicity instead of for flexibility.

Currently, we have this low-level interface:

```c
typedef struct igraph_rng_type_t {
    const char *name;
    unsigned long int min; /* 'min' must always be set to 0 */
    unsigned long int max;
    igraph_error_t (*init)(void **state);
    void (*destroy)(void *state);
    igraph_error_t (*seed)(void *state, unsigned long int seed);
    unsigned long int (*get)(void *state);
    igraph_real_t (*get_real)(void *state);
    igraph_real_t (*get_norm)(void *state);
    igraph_real_t (*get_geom)(void *state, igraph_real_t p);
    igraph_real_t (*get_binom)(void *state, long int n, igraph_real_t p);
    igraph_real_t (*get_exp)(void *state, igraph_real_t rate);
    igraph_real_t (*get_gamma)(void *state, igraph_real_t shape,
                               igraph_real_t scale);
} igraph_rng_type_t;
```

This would be changed like so:

```c
typedef struct igraph_rng_type_t {
    const char *name;

    /* No minimum or maximum. Minimum is always 0, maximum is always 2^31-1. */

    /* Init and destroy */
    igraph_error_t (*init)(void **state);
    void (*destroy)(void *state);

    igraph_error_t (*seed)(void *state, unsigned long int seed);

    /* Fundamental generators: at least one of these two MUST be provided */
    int31_t (*get)(void *state); /* returns 32-bit signed integer */
    igraph_real_t (*get_real)(void *state);

    /* Optional generators: */
    igraph_real_t (*get_norm)(void *state);
    igraph_real_t (*get_geom)(void *state, igraph_real_t p);
    igraph_real_t (*get_binom)(void *state, long int n, igraph_real_t p);
    igraph_real_t (*get_exp)(void *state, igraph_real_t rate);
    igraph_real_t (*get_gamma)(void *state, igraph_real_t shape, igraph_real_t scale);
    igraph_real_t (*get_pois)(void *state, igraph_real_t lambda); /* newly added Poisson generator */
} igraph_rng_type_t;
```

Principles:

 - Either an integer or a real generator must be provided.
 - The integer one must produce non-negative 32-bit signed integers, i.e. 31 random bits. 
 - Therefore, there is no need for `min` and `max` methods as these are always 0 and `2^31 - 1`.

[...]

Issue #1987, opened 2022-03-04
Support for large random integers


szhorvat:
`igraph_rng_get_integer()` should support large (larger than `2^31 - 1`) integers when building in 64-bit more, and should return `igraph_integer_t` instead of `long int`.

Once concern that was brought up before was: it is good to ensure that the RNG behaves identically on 32-bit and 64-bit platforms. This can be achieved: there can be two branches of the implementation depending on whether the `h` parameter (upper limit) exceeds the largest signed integer representable on 32 bits.

Notes:

 - I am not suggesting that the underlying RNG implementation should use 64 bit integers, only that the high-level `igraph_rng_get_integer()` function should be able to produce them.
 - We also have a `get_int31()` function. This is generating the RNG's largest possible native type. Perhaps we want to rename it to indicate that, in case we ever want to allow 64-bit RNGs.
 - `get_integer()` and `get_int31()` are the only two functions returning an integer type. All other integer-valued samplers, such as the geometric one, return an `igraph_real_t`. It is not an unreasonable possibility to let `get_integer()` return an `igraph_real_t` as well, but I am not in favour.

ntamas:
One problem that we might be running into when allowing 64-bit random numbers on 64-bit platforms is that igraph functions relying heavily on random numbers will give different results on 32-bit and 64-bit platforms, even in unit tests, and even if we seed the RNG first in the unit test.

So, something like this:

```python
g = Graph.GRG(100, 0.2)
comm = g.community_label_propagation()
assert comm.membership == [...some_expected_result...]
```

will fail on either 32-bit or on 64-bit platforms, unless we provide the expected result for both cases. This is not a vote for or against the proposed change, just something to be aware of.

On the other hand, ensuring reproducibility is a hard problem anyway: even if we provide an RNG that works the same way on both 32-bit and 64-bit platforms, the host language might use its own RNGs differently, or the behaviour of the host RNG might change between versions and we cannot do anything about that. So, it's not entirely up to us. I think igraph's job would be to ensure that if someone wants to achieve reproducible results, then igraph should not stand in its way. This could be achieved by letting the "default" RNG in igraph to use the full 64-bits on 64-bit platforms, but then we should provide a "reproducible" RNG for cases when reproducibility is more important than performance.

[...]

Issue #1986, opened 2022-03-04
Banish (long int) from random number generator interface


szhorvat:
We should eliminate `long int` from the igraph codebase as much as possible, including from the RNG interface. The size of `long int` is unpredictable and differs across platforms. If we want to use 32-bit integers in the RNG interface, we should use an explicitly 32-bit type.

(Using `long int` in specific RNG implementations may stay if necessary to ensure unchanged behaviour, but it should be banished from the general interface.)

ntamas:
This is now done in the `feat/rng-refactoring` branch, which I'l submit for a round of reviews as a PR soon(ish).

Issue #1981, opened 2022-03-03
Sparate R-specific definitions in random.c


szhorvat:
`random.c` contains a large number of R-specific functions/defined, such as these:

https://github.com/igraph/igraph/blob/develop/src/random/random.c#L1488

 - [ ] Unused ones should be removed
 - [ ] Those which are used should still be guarded by `#ifdef USING_R`. Currently, many aren't (see link above).
 - [ ] The file should be restructured so R-specific things are grouped together as much as possible. Function names should make it clear what is/isn't R-specific (igraph stuff should have an `igraph_i_` prefix).

ntamas:
`random.c` is now free of `USING_R`; R-specific functions were moved to igraph/rigraph

szhorvat:
There were a bunch of unused function declarations (without any definitions) that I remove now. They seem to be meant for linking to R.

ntamas:
I think this is now done, right? The only place where I see `USING_R` is in `igraph_random.h` and I just proposed on Slack to move this to an internal header. Once we have an internal header, we can remove the R-specific stuff from there and simply replace or patch this internal header in the source code of the R interface of igraph to use `GetRNGState()` and `PutRNGState()` where appropriate.

Or, we could do it like this:

```c
#ifndef RNG_BEGIN
#define RNG_BEGIN /* our standard implementation */
#endif
```

and then the R interface can override it by simply defining `RNG_BEGIN` before including `igraph_random_internal.h`

szhorvat:
Is it really necessary for the R interface to call `GetRNGState()` / `PutRNGState()` before/after every use of the RNG? Could it not just call these once before/after calling any igraph function, or after/before going into a callback?

PR #2081, opened 2022-05-27
RNG refactoring for 0.10

ntamas:
This PR makes significant modifications to igraph's RNG machinery to satisfy the requirements outlined in issues #1981, #1986, #1987 and #1988. I'd appreciate a careful review because this affects critical parts of igraph.

In particular:

* `long int` is banished from the public RNG interface, we use `igraph_integer_t` instead.
* `igraph_uint_t` was added to cater for RNG needs where we need a type that has the same size as `igraph_integer_t` but is unsigned.
* Care has been taken to ensure that the RNG behaves the same way on 32-bit and 64-bit platforms as long as the user does not start requesting integers with a range larger than what would fit in a 32-bit number.
* `min` and `max` were removed from the RNG interface, we now simply require RNG implementors to specify how many bits of randomness they can generate in a single round, and we derive `max` from that. (`min` is always zero).
* Casts to `double`s were removed from `igraph_rng_get_integer()` because doubles cannot represent integers exactly if the integer is larger than 2^53-1.
* 64-bit igraph now requires either support for `__uint128_t` (provided by `gcc` and `clang`) or access to the `_umul128` intrinsic (provided by MSVC) to ensure unbiased sampling from large integer ranges. I don't know whether ICC and other esoteric compilers support these or not so some help is needed with the testing there.
* PCG32 has been added as an alternative RNG; we might switch to PCG32 in future versions as it seems to be a better fit overall than Mersenne Twister.


ntamas:
I've requested an explicit review from @szhorvat because he is the owner of the issues that this PR is supposed to fix, but reviews and comments from @vtraag and @GroteGnoom are also appreciated.

szhorvat:
> * I don't know whether ICC and other esoteric compilers

No problem with ICC.

CHANGELOG.md
```diff
@@ -670,8 +670,13 @@ Some of the highlights are:
  - `igraph_write_graph_gml()` takes an additional bitfield parameter controlling some aspects of writing
    the GML file.
 
+ - `igraph_rng_seed()` now requires an `igraph_uint_t` as its seed arguments. RNG implementations are free to use only the lower bits of the seed if they do not support 64-bit seeds.
+
+ - `igraph_rngtype_rand` (i.e. the RNG that is based on BSD `rand()`) was removed due to poor statistical properties that sometimes resulted in weird artifacts like all-even "random" numbers when igraph's usage patterns happened to line up with the shortcomings of the `rand()` generator in a certain way.
+
 ### Added
 
+ - A new integer type, `igraph_uint_t` has been added. This is the unsigned pair of `igraph_integer_t` and th3ey are always consistent in size.
```
GroteGnoom, line 679:
```suggestion
 - A new integer type, `igraph_uint_t` has been added. This is the unsigned pair of `igraph_integer_t` and they are always consistent in size.
```

src/random/random.c
```diff
@@ -228,27 +256,54 @@ void igraph_rng_destroy(igraph_rng_t *rng) {
...
+ *
+ * @param rng The RNG.
+ * \return The number of random bits that can be generated in a single round
+ *         with the RNG.
+ *
+ * Time complexity: O(1).
+ */
+IGRAPH_EXPORT igraph_integer_t igraph_rng_bits(const igraph_rng_t* rng) {
+    const igraph_rng_type_t *type = rng->type;
+    return type->bits;
```
GroteGnoom, line 278: why not just return rng->type->bits?

szhorvat:
Is the function`igraph_i_rng_get_uint64()` kept intentionally, or is it a an accidental leftover? It is never used.

szhorvat:
Implementing a 64-bit PCG (or rather borrowing it from https://github.com/imneme/pcg-c) would be a nice test of the functionality, since now it's only tested with 32-bit generators.

szhorvat:
It would be useful to do a benchmark and see if we lost any performance with this method. For example, see how long it takes to generate `n` random integers in a range, or `n` random reals in a range, or test the function `igraph_tree_game()` with method `IGRAPH_RANDOM_LERW` for a realistic but RNG-heavy use case.

src/random/random.c
```diff
@@ -282,30 +487,49 @@ const char *igraph_rng_name(const igraph_rng_t *rng) {
...
-long int igraph_rng_get_integer(igraph_rng_t *rng,
-                                long int l, long int h) {
-    assert(h >= l);
-    const igraph_rng_type_t *type = rng->type;
+igraph_integer_t igraph_rng_get_integer(
+    igraph_rng_t *rng, igraph_integer_t l, igraph_integer_t h
+) {
+    igraph_uint_t range;
+
     /* We require the random integer to be in the range [l, h]. We do so by
```
szhorvat, line 495: The comments starting here appear to be outdated, and I think they should be removed.

src/random/random.c
```diff
@@ -373,11 +590,9 @@ igraph_real_t igraph_rng_get_unif01(igraph_rng_t *rng) {
     const igraph_rng_type_t *type = rng->type;
     if (type->get_real) {
         return type->get_real(rng->state);
-    } else if (type->get) {
-        unsigned long int max = type->max;
-        return type->get(rng->state) / ((double)max + 1);
+    } else {
+        return ldexp(type->get(rng->state), -igraph_rng_bits(rng));
```
szhorvat, line 594: This means that that if the underlying RNG provides only a few bits, we end up with very bad quality random reals. 

The interface makes it appears as if it were okay to supply only a few bits per generation round. This needs to be improved in some way, either in documentation or by changing the code.

I'm also somewhat bothered that we use only 32 bits for random reals with the default generator, even though `double` has 52 bits.  Within igraph code, we tend to assume that it's fine to move between integer and real representations of whole numbers, but random reals here are clearly worse than random integers.

szhorvat: I checked what assembly compilers generate for `ldexp` using Godbolt, and it seems that they always emit a call to the `ldexp` library function. It suspect that `ldexp` is one of those functions which has complex code paths for overflow/underflow detection (note that it sets `errno`) and is therefore much slower than a more naive multiplication approach.

I am thinking of `hypot(x,y)` which is much slower than `sqrt(x*x + y*y)`. 

My preliminary benchmark of random real generation seems to show that random reals are slower to create than random integers, which is surprising considering that the integer code is much more complex.

I'm not fully comfortable with `ldexp` at the moment ... could we come up with something better here? E.g. store `2^-bits` somewhere and just multiply it naively?

szhorvat: As a simple experiment, I replaced the `ldexp` code by `return type->get(rng->state) * 0x1.0p-32;` (knowing that the default RNG produces 32 bits). This took the benchmark timing from 2.8 to 2.0 seconds, a very significant gain. I'm wondering if a similar gain could be achieved by storing the `2^-bits` within the `igraph_rng_t` at seeding time and re-using it.

src/random/random.c
```diff
@@ -228,27 +255,54 @@ void igraph_rng_destroy(igraph_rng_t *rng) {
  * Time complexity: usually O(1), but may depend on the type of the
  * RNG.
  */
-igraph_error_t igraph_rng_seed(igraph_rng_t *rng, unsigned long int seed) {
+igraph_error_t igraph_rng_seed(igraph_rng_t *rng, igraph_uint_t seed) {
     const igraph_rng_type_t *type = rng->type;
-    rng->def = 0;
+    rng->is_seeded = 1;
```
szhorvat, line 260: Nitpick: I'd move this line after the next, so that in case the next line fails, `is_seeded` stays false.

src/random/rng_mt19937.c
```diff
@@ -198,3 +298,5 @@ IGRAPH_THREAD_LOCAL igraph_rng_t igraph_i_rng_default = {
 };
```
szhorvat, line 298: Need to change `def` to `is_seeded` here.

szhorvat:
I pushed a proof of concept of (mostly) getting rid of `ldexp()`. We can revert this if you disagree, this is just to show how it might work.

szhorvat:
An alternative, which I think I like better, is to tabulate the negative integer powers of 2 up to -64 and use them. This is just as fast as the solution I pushed, but doesn't involve that extra field in `igraph_rng_t`.

Yet another possibility is something like this:

```c
(igraph_real_t) type->get(rng->state) / ((igraph_uint_t) 2 << (type->bits - 1))
```

This is actually markedly slower, but not as slow as calling `ldexp` every time.

szhorvat:
Final suggestion (along with `neg_pow2` from above):

```c
        igraph_real_t r = 0.0;
        uint8_t b = 0;
        while (b < DBL_MANT_DIG-1) {
            r += type->get(rng->state);
            r *= neg_pow2[type->bits];
            b += type->bits;
        }
        return r;
```

This fills up all 52 digits of a `double` (instead of just 32), and will therefore make two calls to the RNG, and be slower. My benchmark runs in ~3.5 s with this compared to ~2.8 s with the original and 2.2 with the same solution as here but using only 32 bits for random reals.

This would also require an update of some test outputs.

szhorvat:
I pushed the above solution, again as a proof of concept. This one only has 32 bit enabled so the tests wouldn't need to be updated before we agree on which way to go.

CHANGELOG.md
```diff
@@ -701,6 +706,7 @@ Some of the highlights are:
  - `igraph_enter_safelocale()` and `igraph_exit_safelocale()` for temporarily setting the locale to C. Foreign format readers and writers require a locale which uses a decimal point instead of decimal comma.
  - `igraph_vertex_path_from_edge_path()` converts a sequence of edge IDs representing a path to an equivalent sequence of vertex IDs that represent the vertices the path travelled through.
  - `igraph_graph_count()` gives the number of unlabelled graphs on a given number of vertices. It is meant to find the maximum isoclass value.
+ - `igraph_rngtype_pcg32` implements the PCG random number generator, which can be used as an alternative source of randomness instead of the default Mersenne Twister. Future versions of igraph might switch to PCG as the default PNG.
```
szhorvat, line 709:
```suggestion
 - `igraph_rngtype_pcg32` implements the PCG random number generator, which can be used as an alternative source of randomness instead of the default Mersenne Twister. Future versions of igraph might switch to PCG as the default RNG.
```

src/random/random.c
ntamas: Thanks, I'll change that.

ntamas:
> Is the static function igraph_i_rng_get_uint64() kept intentionally, or is it a an accidental leftover? It is never used

It should have been used in `igraph_i_rng_get_uint64_bounded()`, well spotted. I'll fix that.

src/random/random.c
ntamas: > I'm also somewhat bothered that we use only 32 bits for random reals with the default generator, even though double has 52 bits.

This has always been the case, also with previous versions where we simply generated an integer and then divided it with the maximum integer that the RNG can generate. Since RNGs were 32-bit, we never managed to get better results than that.

I'm a bit wary about doing this:

```
        igraph_real_t r = 0.0;
        uint8_t b = 0;
        while (b < 32 /* DBL_MANT_DIG-1 */) {
            r += type->get(rng->state);
            r *= neg_pow2[type->bits];
            b += type->bits;
        }
```

It seems good in theory, and if we could work with floats in infinite precision, then of course this would be accurate. However, I don't know how the floating-point operations work out, especially in the second (and possibly further) iterations when you are adding a large-ish number from `type->get(rng->bits)` to a small-ish number obtained in the previous iteration in `r`. We might be doing extra work for nothing, or we might be introducing bias into the generated number in ways that we don't understand well enough.

Originally I went with the `ldexp` route because it was suggested [on the PCG page](https://www.pcg-random.org/using-pcg-c-basic.html#generating-doubles). If that's slow, I'm okay to use a single round of multiplication with`0x1.0p-32` but I'm not sure about multiple rounds.

Ultimately, we need to decide whether we want to be precise and unbiased (in which case a much more complex method is required that also caters for subnormal numbers, see `random_real()` [here](https://mumble.net/~campbell/tmp/random_real.c)), or we want to be fast and only reasonably precise and unbiased. I think that for the latter case, 32-bit precision is enough for most users, and those who need greater precision should probably roll their own RNG implementation anyway (and then they can get as fancy as they want in `rng->type->get_real()`).

szhorvat:
Here's another thought:

At this moment, we do not allowed the bounded integer generator to be overridden. _Maybe_ (I'm not sure yet myself) it would be useful to allow this. Here's why.

Our bounded integer generator is implemented in such a way that it will typically be faster for narrow ranges than wide ranges, simply because wide ranges may require more than one call to the RNG. It is plausible that the same is true for the bounded integer generators of some high-level languages.  Some high-level languages may not expose the raw RNG, only the bounded integer generator, meaning that that's the only thing we can hook up to igraph. If we use a wide range to supply the "raw" random values to igraph, then we lose performance even when igraph is generating numbers only for a narrow range.

We should take a careful look at some high-level languages (perhaps including those that don't have interfaces yet, such as Julia, Ruby or Perl), and see if they expose their raw random bit generator.

I checked Mathematica, and this is the performance of generating 100 million random integers in the range `[0, 2^bits - 1]`. As you can see, the larger the integers, the slower the function:

image

```mathematica
res = Table[
   {k, First@Timing[RandomInteger[{0, 2^k - 1}, 100000000];]},
   {k, 1, 63}
   ];

ListPlot[res, Frame -> True, FrameLabel -> {"bits", "timing (s)"}, PlotTheme -> "Detailed"]
```

Do you know that is the situation with Python and R, @ntamas?

szhorvat:
Something different:

If we are considering moving to PCG32, it might be best to do it for 0.10 (but only after this PR, so as not to introduce a lot of noise from updating test outputs). 0.10 will change the output from the RNG anyway. It's better to do this once for 0.10 rather than twice (also for 0.11).

szhorvat:
Yet another comment:

We have code for generating numbers from a Poisson distribution. This is used internally for the geometric distribution. I would suggest exposing the Poisson generator, and make it overridable (since R provides this). Poisson distributions are likely to be useful to combinatorial problems, which igraph deal with. Since we're making significant changes to the RNG infrastructure, now is the time to do this.

Once again, this can be done after this PR was merged.

szhorvat:
On the topic of overridable generators:

The geometric distribution generator is overridable. "Geometric distribution" can mean two slightly different things, [as explained on Wikipedia](https://en.wikipedia.org/wiki/Geometric_distribution). igraph and Mathematica use different definitions and this always confuses me. We need to expand the documentation and provide a clear definition and a warning (in case a user would override this generator with an incompatible one).

This is just a small doc TODO item that I wanted to write down before I forgot.

ntamas:
> Do you know what is the situation with Python and R, @ntamas?

Python has `random.randint()`, but the documentation says that it relies on `random.random()`, which generates a random 53-bit double in the range `[0, 1)`, using an underlying Mersenne Twister generator. We will be jumping through extra hoops anyway if we use the Python RNG, but it seems like we'd be using one less hoop if we allowed the user to override `igraph_random_get_integer()`.

ntamas:
PCG implementation replaced with the dual-licensed one. I have to go now but I'll continue later and add PCG-64.

szhorvat:
TODO:

We have a failure with the Intel compiler.

```
igraph/src/random/pcg/pcg_variants.h(49): error: #error directive: Nonstandard GNU inlining semantics. Compile with -std=c99 or better.
      #error Nonstandard GNU inlining semantics. Compile with -std=c99 or better.
```

`__GNUC_GNU_INLINE__` is defined by icc if compiling with `-std=gnu99` (which is CMake's default and possibly _needed_ by igraph) but not when compiling with `-std=c99`.

szhorvat:
Tried to add a workaround above. I did not use `defined()` in case ICC starts to define this macro to 0 instead of 1.

src/random/random.c
szhorvat: > We might be doing extra work for nothing

It is not for nothing: the second iteration runs only when not all 52 bits have been filled. Note that this is 52, as the 53rd is not represented and implicitly treated as 1.

> might be introducing bias into the generated number in ways that we don't understand well enough

Let's not be superstitious and instead understand what the code does: is there a bias or not.

In a single iteration, this is what happens:

 - Draw an integer `k` from `[0, MAX-1]` where MAX is `2^bits`.
 - Compute `k/MAX` to obtain a float in `[0, 1)`.
 
 In the second iteration, we take this number in `[0, 1)` and add it to an integer drawn from the range `[0, MAX-1]`. The result is still in `[0, MAX)`. Dividing by `MAX` still takes us into `[0, 1)`.
 
This is all correct with infinite precision. Now let's think about what can go wrong with finite precision.

We have 52 bits (the leading bit is considered to be 1). The first round filled out the first 32 bits, leaving the remaining 20 as zero. Note that multiplication by negative powers of 2 doesn't really do much: it only changes the exponent as we never get anywhere near [subnormal](https://en.wikipedia.org/wiki/Subnormal_number) territory.  There is no chance for rounding, which is what distinguishes floats from mathematical reals. The more interesting operation is the addition, which does indeed round.

Suppose that our first random integer is exactly `2^32 - 1`, i.e. all the bits are 1. Now what happens if the second is the same? We only have 52 bits to fill, so 64 - 52 = 12 will be "rounded off". But in which direction, up or down? If down, we have a number that is < 1.0. But if up, we get an exact 1.0. I checked this, and it rounded up (both in C++ and Mathematica with machine numbers). 

We don't exactly get any biases, but there is a risk of getting an exact 1.0, which should not be allowed.

I'll think about how to prevent this without affecting performance or making the code too complex. I suppose the easiest solution would be rejection sampling.

As for biases, I see none, and I do not believe that I missed anything.

Finally, something interesting to note: when I looked at how many bits Mathematica's real generator produces, I noticed that it always leaves the very last bit as zero. Possibly, it has to do with this rounding up issue.

szhorvat: Alright, I pushed a fix using rejection sampling, which is simple, and obviously won't produce any bias. Note that we restart the generation from scratch of we encounter an exact 1.0. I am confident about this method.

Note that this fix would have been necessary for the `ldexp` method as well. Here's a transcript of my cling (C++ inrterpreter) session checking this:

```
[cling]$ #include <cmath>
[cling]$ uint64_t a = -1;
[cling]$ a
(unsigned long long) 18446744073709551615
[cling]$ ldexp(a, -64)
(double) 1.0000000
[cling]$ ldexp(a, -64) == 1
(bool) true
```

szhorvat: Note regarding biases, just in case someone finds this confusing: 

The distribution of representable floating points numbers is not uniform. There are more (twice as many) representable numbers in [0, 1/2) than in [1/2, 1). Thus this method does not generate all representable floating point numbers with equal probability. But we wouldn't want that. We want our number to fall in [0, 1/2) with the same probability as in [1/2, 1), and this method achieves that.

szhorvat: There was in fact a bias: the lowest sampling bin would be underrepresented, as explained in the link you provided @ntamas. I pushed another update to use precisely as many bits as needed. This reduces performance somewhat, unfortunately, but eliminates this bias.

ntamas: I managed to convince myself that this is now probably correct (although it took me a while). I disagree with the rounding, though -- I think that the default IEEE floating-point rounding mode is the banker's rounding (i.e. round to nearest, ties to even), which is not supposed to introduce bias. But, this may still potentially end up rounding towards 1.0 so the rejection rule must still be used in that case.

ntamas:
I think all the comments / suggestions have been addressed. Time for another round of reviews?

src/random/random.c
szhorvat: I only saw this now. Let's discuss at next meeting.

tests/unit/random_sampling.c
```diff
@@ -175,6 +175,14 @@ void sample() {
     x = RNG_BINOM(5, 1);
     IGRAPH_ASSERT(x == 5);
 
+    x = RNG_BINOM((1LL << 31) - 1, 0.5);
+    IGRAPH_ASSERT(!igraph_is_nan(x));
+    IGRAPH_ASSERT(0 <= x && x <= (1LL << 31) - 1);
+
+    x = RNG_BINOM((1LL << 31), 0.5);
```
szhorvat, line 182: `(igraph_integer_t) (1LL << 31)` will create a negative value when `igraph_integer_t` is 32-bit.

szhorvat: Since this test is not printing, we could safely make it conditional on an `#if IGRAPH_INTEGER_SIZE > 32`

ntamas: oh, indeed, thanks.

szhorvat: Fixed.

PR #2084, opened 2022-05-27
Centre vertices of a graph

### LINKED ISSUE #2027: Graph center

pradkrish:
partially fixes #2027

### DIFF
--- a/src/paths/distances.c
+++ b/src/paths/distances.c
@@ -777,3 +777,50 @@ igraph_error_t igraph_pseudo_diameter_dijkstra(const igraph_t *graph,
 
     return IGRAPH_SUCCESS;
 }
+
+/**
+ * \function igraph_graph_center
+ * \brief Centre vertices of a graph.
+ *
+ * The centre vertices of a graph is calculated by finding the vertices
+ * with the minimum eccentricity
+ *
+ * \param graph The input graph, it can be directed or undirected.
+ * \param res Pointer to an initialized vector, the result is stored
+ *    here.
+ * \param mode What kind of paths to consider for the calculation:
+ *    \c IGRAPH_OUT, paths that follow edge directions;
+ *    \c IGRAPH_IN, paths that follow the opposite directions; and
+ *    \c IGRAPH_ALL, paths that ignore edge directions. This argument
+ *    is ignored for undirected graphs.
+ * \return Error code.
+ *
+ * Time complexity: O(v*(|V|+|E|)), where |V| is the number of
+ * vertices, |E| is the number of edges and v is the number of
+ * vertices for which eccentricity is calculated.
+ *
+ * \sa \ref igraph_eccentricity().
+ *
+ */
+igraph_error_t igraph_graph_center(const igraph_t *graph, igraph_vector_t *res,
+                  igraph_neimode_t mode) {
+
+    igraph_vector_t ecc;
+    igraph_vs_t vids = igraph_vss_all();
+
+    IGRAPH_CHECK(igraph_vector_init(&ecc, 0));
+    IGRAPH_CHECK(igraph_eccentricity(graph, &ecc, vids, mode));
+    if (igraph_vcount(graph)) {
+        igraph_integer_t n = igraph_vector_size(&ecc);
+        igraph_integer_t min_eccentricity = (igraph_integer_t) igraph_vector_min(&ecc);
+
+        for (int i = 0; i < n; i++) {
+            if (VECTOR(ecc)[i] == min_eccentricity) {
+                igraph_vector_push_back(res, i);
+            }
+        }
+    }
+
+    igraph_vector_destroy(&ecc);
+    return 0;
+}
--- a/src/paths/distances.c
+++ b/src/paths/distances.c
@@ -783,8 +783,9 @@ igraph_error_t igraph_pseudo_diameter_dijkstra(const igraph_t *graph,
  * \brief Central vertices of a graph.
  *
  * The central vertices of a graph are calculated by finding the vertices
- * with the minimum eccentricity. The calculation is done per connected component:
- * the central vertices of each component are returned.
+ * with the minimum eccentricity. This concept is typically applied to
+ * connected graphs. In undirected disconnected graphs, the calculation
+ * is effectively done per connected component.
  * 
  * \param graph The input graph, it can be directed or undirected.
  * \param res Pointer to an initialized vector, the result is stored
@@ -796,9 +797,8 @@ igraph_error_t igraph_pseudo_diameter_dijkstra(const igraph_t *graph,
  *    is ignored for undirected graphs.
  * \return Error code.
  *
- * Time complexity: O(v*(|V|+|E|)), where |V| is the number of
- * vertices, |E| is the number of edges and v is the number of
- * vertices for which eccentricity is calculated.
+ * Time complexity: O(|V| (|V|+|E|)), where |V| is the number of
+ * vertices and |E| is the number of edges.
  *
  * \sa \ref igraph_eccentricity().
  *
@@ -810,21 +810,25 @@ igraph_error_t igraph_graph_center(const igraph_t *graph,
     igraph_vector_t ecc;
 
     igraph_vector_clear(res);
+    if (igraph_vcount(igraph) == 0) {
+        return IGRAPH_SUCCESS;
+    }
+    
     IGRAPH_VECTOR_INIT_FINALLY(&ecc, 0);
-    // Assume igraph_eccentricity() does not return infinity or NaN 
     IGRAPH_CHECK(igraph_eccentricity(graph, &ecc, igraph_vss_all(), mode));
-    if (igraph_vcount(graph) > 0) {
-        igraph_real_t n = igraph_vector_size(&ecc);
-        igraph_real_t min_eccentricity = igraph_vector_min(&ecc);
 
-        for (igraph_integer_t i = 0; i < n; i++) {
-            if (VECTOR(ecc)[i] == min_eccentricity) {
-                IGRAPH_CHECK(igraph_vector_push_back(res, i));
-            }
+    /* igraph_eccentricity() does not return infinity or NaN, and the null graph
+     * case was handled above, therefore calling vector_min() is safe. */
+    igraph_real_t min_eccentricity = igraph_vector_min(&ecc);
+    igraph_real_t n = igraph_vector_size(&ecc);
+    for (igraph_integer_t i = 0; i < n; i++) {
+        if (VECTOR(ecc)[i] == min_eccentricity) {
+            IGRAPH_CHECK(igraph_vector_push_back(res, i));
         }
     }
 
     igraph_vector_destroy(&ecc);
     IGRAPH_FINALLY_CLEAN(1);
+
     return IGRAPH_SUCCESS;
 }
--- a/src/paths/distances.c
+++ b/src/paths/distances.c
@@ -805,23 +805,21 @@ igraph_error_t igraph_pseudo_diameter_dijkstra(const igraph_t *graph,
  */
 igraph_error_t igraph_graph_center(const igraph_t *graph, 
                     igraph_vector_t *res,
-                    igraph_vs_t vids,
                     igraph_neimode_t mode) {
 
     igraph_vector_t ecc;
 
     igraph_vector_clear(res);
-    IGRAPH_ASSERT(igraph_vector_empty(res));
     IGRAPH_VECTOR_INIT_FINALLY(&ecc, 0);
     // Assume igraph_eccentricity() does not return infinity or NaN 
-    IGRAPH_CHECK(igraph_eccentricity(graph, &ecc, vids, mode));
+    IGRAPH_CHECK(igraph_eccentricity(graph, &ecc, igraph_vss_all(), mode));
     if (igraph_vcount(graph) > 0) {
         igraph_real_t n = igraph_vector_size(&ecc);
         igraph_real_t min_eccentricity = igraph_vector_min(&ecc);
 
         for (igraph_integer_t i = 0; i < n; i++) {
             if (VECTOR(ecc)[i] == min_eccentricity) {
-                igraph_vector_push_back(res, i);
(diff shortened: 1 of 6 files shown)


pradkrish:
@szhorvat  this is ready for review.

szhorvat:
Next time you can use the "ready for review" button (merging is disabled until the PR is marked as a draft).

src/paths/distances.c
```diff
@@ -777,3 +777,50 @@ igraph_error_t igraph_pseudo_diameter_dijkstra(const igraph_t *graph,
 
     return IGRAPH_SUCCESS;
 }
+
+/**
+ * \function igraph_graph_center
+ * \brief Centre vertices of a graph.
```
szhorvat, line 783:
```suggestion
 * \brief The central vertices of a graph.
```

src/paths/distances.c
```diff
@@ -777,3 +777,50 @@ igraph_error_t igraph_pseudo_diameter_dijkstra(const igraph_t *graph,
...
+ *
+ * The centre vertices of a graph is calculated by finding the vertices
```
szhorvat, line 785:
```suggestion
 * The central vertices of a graph is calculated by finding the vertices
```

src/paths/distances.c
```diff
@@ -777,3 +777,50 @@ igraph_error_t igraph_pseudo_diameter_dijkstra(const igraph_t *graph,
...
+ * vertices for which eccentricity is calculated.
+ *
+ * \sa \ref igraph_eccentricity().
+ *
+ */
+igraph_error_t igraph_graph_center(const igraph_t *graph, igraph_vector_t *res,
+                  igraph_neimode_t mode) {
+
+    igraph_vector_t ecc;
+    igraph_vs_t vids = igraph_vss_all();
```
szhorvat, line 809: Please move this inline into the function call. There is no reason to have a variable for this purpose.

src/paths/distances.c
```diff
@@ -777,3 +777,50 @@ igraph_error_t igraph_pseudo_diameter_dijkstra(const igraph_t *graph,
...
+
+    IGRAPH_CHECK(igraph_vector_init(&ecc, 0));
+    IGRAPH_CHECK(igraph_eccentricity(graph, &ecc, vids, mode));
+    if (igraph_vcount(graph)) {
+        igraph_integer_t n = igraph_vector_size(&ecc);
+        igraph_integer_t min_eccentricity = (igraph_integer_t) igraph_vector_min(&ecc);
+
+        for (int i = 0; i < n; i++) {
+            if (VECTOR(ecc)[i] == min_eccentricity) {
+                igraph_vector_push_back(res, i);
```
szhorvat, line 819: `IGRAPH_CHECK` is missing.

src/paths/distances.c
```diff
@@ -777,3 +777,50 @@ igraph_error_t igraph_pseudo_diameter_dijkstra(const igraph_t *graph,
...
+    igraph_vector_t ecc;
+    igraph_vs_t vids = igraph_vss_all();
+
+    IGRAPH_CHECK(igraph_vector_init(&ecc, 0));
+    IGRAPH_CHECK(igraph_eccentricity(graph, &ecc, vids, mode));
+    if (igraph_vcount(graph)) {
+        igraph_integer_t n = igraph_vector_size(&ecc);
+        igraph_integer_t min_eccentricity = (igraph_integer_t) igraph_vector_min(&ecc);
+
+        for (int i = 0; i < n; i++) {
```
szhorvat, line 817: Never use `int` in igraph code, unless you have a good reason. What if the graph has more vertices than what is representable in an `int`? Use `igraph_integer_t`.

src/paths/distances.c
```diff
@@ -777,3 +777,50 @@ igraph_error_t igraph_pseudo_diameter_dijkstra(const igraph_t *graph,
...
+ * \sa \ref igraph_eccentricity().
+ *
+ */
+igraph_error_t igraph_graph_center(const igraph_t *graph, igraph_vector_t *res,
+                  igraph_neimode_t mode) {
+
+    igraph_vector_t ecc;
+    igraph_vs_t vids = igraph_vss_all();
+
+    IGRAPH_CHECK(igraph_vector_init(&ecc, 0));
```
szhorvat, line 811: This vector is used only for the lifetime of the function. We need to ensure that it gets deallocated if an error is raised in the middle of the function. Check the documentation on error handling especially section 5.3, on how to do this:

https://igraph.org/c/html/develop/igraph-Error.html

You can also look at `igraph_regular_tree()` for a very simple example of how this is done.

We put the allocated vector on the "finally stack", along with its destructor, using an `IGRAPH_FINALLY` call. There are macros to simplify this for vectors. When the vector is destroyed below, we remove it from the finally stack using an `IGRAPH_FINALLY_CLEAN` call.

src/paths/distances.c
```diff
@@ -777,3 +777,50 @@ igraph_error_t igraph_pseudo_diameter_dijkstra(const igraph_t *graph,
...
+    IGRAPH_CHECK(igraph_eccentricity(graph, &ecc, vids, mode));
+    if (igraph_vcount(graph)) {
```
szhorvat, line 813: Please use a `> 0` to make it explicit what is being checked here.

What if the input is a null graph? This case is not handled. The result should be an empty vector. Suppose I passed in a non-empty vector originally.

src/paths/distances.c
szhorvat: What is `res` was not empty when passed to this function? It should be cleared first.

src/paths/distances.c
```diff
@@ -777,3 +777,50 @@ igraph_error_t igraph_pseudo_diameter_dijkstra(const igraph_t *graph,
...
+
+        for (int i = 0; i < n; i++) {
+            if (VECTOR(ecc)[i] == min_eccentricity) {
+                igraph_vector_push_back(res, i);
+            }
+        }
+    }
+
+    igraph_vector_destroy(&ecc);
+    return 0;
```
szhorvat, line 825: Always use named enum values for error codes instead of numbers. `IGRAPH_SUCCESS`.

src/paths/distances.c
szhorvat: These are stylistic suggestions, but something of note is to use American spelling, as the rest of the documentation does. I also prefer British spelling in general, but I use American spelling for igraph documentation.

src/paths/distances.c
```diff
@@ -777,3 +777,50 @@ igraph_error_t igraph_pseudo_diameter_dijkstra(const igraph_t *graph,
...
+                  igraph_neimode_t mode) {
+
+    igraph_vector_t ecc;
+    igraph_vs_t vids = igraph_vss_all();
+
+    IGRAPH_CHECK(igraph_vector_init(&ecc, 0));
+    IGRAPH_CHECK(igraph_eccentricity(graph, &ecc, vids, mode));
+    if (igraph_vcount(graph)) {
+        igraph_integer_t n = igraph_vector_size(&ecc);
+        igraph_integer_t min_eccentricity = (igraph_integer_t) igraph_vector_min(&ecc);
```
szhorvat, line 815: Since `igraph_eccentricty` produces the result as `igraph_real_t` values, let's use that type throughout instead of converting to an integer type.

szhorvat (changes requested):
Thanks for the PR! I added a few technical comments.

Additionally, the test needs to be updated to check that the result is correct, not merely that the function doesn't error. Print out the result using the helper functions from `test_utilities.h`, and create an `.out` file with the expected results. Be sure to include cases where there is more than one central vertex (if they aren't present already).

Something important to think about is what to do with disconnected graphs (including non-strongly-connected directed graphs). This also brings up the question if `igraph_eccentricty()` might return values, such as infinity or NaN, that could cause trouble down the line, especially in `vector_min`. I checked that currently it does not. However, in the future this might be changed. For example, it would be reasonable to define the eccentricity of all vertices as infinity in disconnected graphs. Thus, please add a comment to the `igraph_eccentricity()` call mentioning that we assume that there are no infinite values returned.

For now, let's just document that this function computes the central vertices for each component separately (just as `igraph_eccentricity()` computes eccentricities separately). 

This will be something to keep in mind when we get to the implementation of the graph periphery, which should be consistent with the graph center.

tests/unit/igraph_graph_center.c
```diff
@@ -0,0 +1,86 @@
+/*
+   IGraph library.
+   Copyright (C) 2021  The igraph development team <igraph@igraph.org>
```
szhorvat, line 3:
```suggestion
   Copyright (C) 2022  The igraph development team <igraph@igraph.org>
```

src/paths/distances.c
pradkrish: If vids must go into the function call, does it mean the function should be able to calculate central vertices for user provided vertices? The current logic assumes for all vertices, then the logic must change.

src/paths/distances.c
szhorvat: `IGRAPH_CHECK` is still missing here.

src/paths/distances.c
```diff
@@ -777,3 +777,54 @@ igraph_error_t igraph_pseudo_diameter_dijkstra(const igraph_t *graph,
...
+ *
+ */
+igraph_error_t igraph_graph_center(const igraph_t *graph, 
+                    igraph_vector_t *res,
+                    igraph_vs_t vids,
+                    igraph_neimode_t mode) {
+
+    igraph_vector_t ecc;
+    igraph_vector_clear(res);
+    IGRAPH_ASSERT(igraph_vector_empty(res));
```
szhorvat, line 812: This assertion is not necessary since the vector is cleared on the previous line. Remove it.

src/paths/distances.c
```diff
@@ -777,3 +777,54 @@ igraph_error_t igraph_pseudo_diameter_dijkstra(const igraph_t *graph,
...
+ *
+ * \sa \ref igraph_eccentricity().
+ *
+ */
+igraph_error_t igraph_graph_center(const igraph_t *graph, 
+                    igraph_vector_t *res,
+                    igraph_vs_t vids,
+                    igraph_neimode_t mode) {
+
+    igraph_vector_t ecc;
```
szhorvat, line 810:
```suggestion
    igraph_vector_t ecc;

```

src/paths/distances.c
```diff
@@ -777,3 +777,54 @@ igraph_error_t igraph_pseudo_diameter_dijkstra(const igraph_t *graph,
 
     return IGRAPH_SUCCESS;
 }
+
+/**
+ * \function igraph_graph_center
+ * \brief Central vertices of a graph.
+ *
+ * The central vertices of a graph is calculated by finding the vertices
+ * with the minimum eccentricity.
```
szhorvat, line 786:
```suggestion
 * The central vertices of a graph are calculated by finding the vertices
 * with the minimum eccentricity. The calculation is done per connected component:
 * the central vertices of each component are returned.
```

src/paths/distances.c
```diff
@@ -777,3 +777,54 @@ igraph_error_t igraph_pseudo_diameter_dijkstra(const igraph_t *graph,
...
+ * Time complexity: O(v*(|V|+|E|)), where |V| is the number of
+ * vertices, |E| is the number of edges and v is the number of
+ * vertices for which eccentricity is calculated.
+ *
+ * \sa \ref igraph_eccentricity().
+ *
+ */
+igraph_error_t igraph_graph_center(const igraph_t *graph, 
+                    igraph_vector_t *res,
+                    igraph_vs_t vids,
```
szhorvat, line 807: Let's remove the `vids` argument. This is a simple convenience function which should work for the standard concept of graph center. We should not try to extend this concept here.

szhorvat (changes requested):
Thanks for the update, it looks better now.

Can you add a simple undirected test case where there is more than one central vertex, but not all vertices are central? For example, an even-length path graph. It would be useful to also add a disconnected undirected case.

szhorvat:
Can you please mark all of the addressed changes as "resolved"? Just use this button:

image

When you feel that the PR is ready, just write a comment and someone will take a look.

pradkrish:
> Can you please mark all of the addressed changes as "resolved"? Just use this button:
> 
> image
> 
> When you feel that the PR is ready, just write a comment and someone will take a look.

Thanks, will do. I thought it's up to the reviewer to "resolve conversation" based upon how convinced they are with changes. Now I know :)

pradkrish:
This PR is ready for review.

ntamas:
> I thought it's up to the reviewer to "resolve conversation" based upon how convinced they are with changes. Now I know :)

I think it's ultimately a matter of taste and habit. I tend to press the "resolve conversation" button myself if I resolved trivial things in the PR but leave it open if I'm not sure whether the reviewer will be satisfied with my solution.

pradkrish:
> Thanks, it looks good!
> 
> One little thing is to add the function to the docs in `doc/structural.xxml`, but I can take care of this later today and then merge.
> 
> Do you have any questions about the technical aspects of igraph so far? If not, you can go ahead with the Turán graph.

Thanks. Please go ahead and merge. I would like to implement `igraph_graph_periphery` next before moving on to Turan graph. It must be pretty much similar to the current PR.

pradkrish:
I have pushed another commit with the function added to the docs.

szhorvat:
It won't quite be the same. Here we made the choice to compute the central vertices per component. If we want the same behaviour from the periphery calculation, then we can't just select vertices with eccentricity equal to the maximum within the entire graph. Instead, we would need to find the maximum for each component, and make the comparison with that.

pradkrish:
by component, do you mean a connected subgraph?

szhorvat:
Yes, but I'm simply trying to find a way to make the periphery calculation consistent with the existing centre calculation for the disconnected case. Now I realize that this is not so simple for the directed case. We'll have to think this through carefully before proceeding.

PR #2085, opened 2022-05-28
add layout_kamada_kawai unit tests

GroteGnoom:
Based on the fruchterman_reingold tests. When I printed out the values they also seemed ok.

### DIFF
--- a/src/layout/kamada_kawai.c
+++ b/src/layout/kamada_kawai.c
@@ -114,39 +114,39 @@ igraph_error_t igraph_layout_kamada_kawai(const igraph_t *graph, igraph_matrix_t
 
     if (maxiter < 0) {
         IGRAPH_ERROR("Number of iterations must be non-negatice in "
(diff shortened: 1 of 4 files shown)


ntamas (approved):

ntamas:
Thanks! (And with this PR, coverage passes 80% for the first time ever).

szhorvat:
Could you please include one more test for a graph with two connected vertices? I fixed a bug recently for this case.

GroteGnoom:
https://github.com/igraph/igraph/commit/278904e027f4c2edd71385400043948c867d0e79

If I chose a ridiculous kkconst I could get the result out of the bounding box, but I can't really find any NaNs.

szhorvat:
Thanks for the test expansion! I wouldn't put extreme values for `kkconst`. The problem appeared with the default value as well. Let's just put a reasonable value there, and keep the test as is otherwise.

PR #2088, opened 2022-06-04
Complete multipartite generator

### LINKED ISSUE #1875: Generator for Turán graphs

Implement a generator for Turán graphs.

A Turán graph `T(n,r)` has `r` partitions, with `n` vertices divided as equally as possible between them. Two vertices are connected precisely when they are in different partitions. A Turán graph is a special case of a complete multipartite graph. See Wikipedia for a more precise definition.

Consider implementing a general complete multipartite generator first, then implement the Turán graph as a special case.

_The main purpose of this issue is to provide an easy way for would-be contributors to familiarize themselves with igraph._

**Use cases for the feature**

Turán graphs are a well studied graph type in graph theory, with applications to graph colouring.

**References**

 - https://en.wikipedia.org/wiki/Tur%C3%A1n_graph
 - SageMath: https://doc.sagemath.org/html/en/reference/graphs/sage/graphs/graph_generators.html#sage.graphs.graph_generators.GraphGenerators.TuranGraph

pradkrish:
fixes #1875


szhorvat:
Many of the CI tests runs are not passing. These should be fixed before marking the PR as ready for review. I have not yet looked at what's wrong. Let us know if you need help. Hint: the `linux_static_vendored` usually provides the most information.

pradkrish:
I managed to fix a few memory leak errors but there are still a few left and I am unable to figure out why :sob:  I can gladly do with some help.

ntamas:
These are "direct leaks", which means that you allocate some data structure and then you forget to destroy it. I think the errors are in the unit test itself. For instance, take a look at this one:

```
Direct leak of 8 byte(s) in 1 object(s) allocated from:
    #0 0x7fc2a1833a06 in __interceptor_calloc ../../../../src/libsanitizer/asan/asan_malloc_linux.cc:153
    #1 0x564f1cd1f736 in igraph_vector_int_init /home/vsts/work/1/s/src/core/vector.pmt:144
    #2 0x564f1ccf9d2c in main /home/vsts/work/1/s/tests/unit/igraph_full_multipartite.c:28
    #3 0x7fc2a0feb082 in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x24082)
```

This means that `igraph_vector_int_init()` is called from `tests/unit/igraph_full_multipartite.c`, line 28, and the vector allocated there is never destroyed. You are probably missing an `igraph_vector_int_destroy()` call at the point where you no longer need the vector.

This one is similar:

```
Direct leak of 8 byte(s) in 1 object(s) allocated from:
    #0 0x7fc2a1833a06 in __interceptor_calloc ../../../../src/libsanitizer/asan/asan_malloc_linux.cc:153
    #1 0x564f1cd1f736 in igraph_vector_int_init /home/vsts/work/1/s/src/core/vector.pmt:144
    #2 0x564f1cd5f81e in igraph_empty_attrs /home/vsts/work/1/s/src/graph/type_indexededgelist.c:96
    #3 0x564f1cd91e92 in igraph_empty /home/vsts/work/1/s/src/graph/type_common.c:70
    #4 0x564f1cd69324 in igraph_full_multipartite /home/vsts/work/1/s/src/constructors/full.c:187
    #5 0x564f1ccf9d5a in main /home/vsts/work/1/s/tests/unit/igraph_full_multipartite.c:30
    #6 0x7fc2a0feb082 in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x24082)
```

Here, you allocate an entire graph with `igraph_full_multipartite()` in line 30 of `tests/unit/igraph_full_multipartite.c`, and then you never call `igraph_destroy()` on it, hence the leak.

szhorvat:
Just a reminder to make sure that the documentation is correct, and include the function in `doc/generators.xxml`.

pradkrish:
I don't know why it failed in the Travis build. As far as I can see, there is no info on the raw job log either. Perhaps it requires a restart.

szhorvat:
Travis looks broken today. This has nothing to do with your code, don't worry about it.

I (or someone else) will take a loo at this PR next week.

Could you fix up the documentation text until then? Parts of it are still copied from `full_bipartite` and need to be updated.

szhorvat:
Also, if you want to go ahead with the Turán graph, feel free to do so within the same PR. The interface of `full_multipartite` looks good to me, so even if we make some changes to its internals, it should not affect any other functions you might add that use this one.

@ntamas and @vtraag: you had some ideas about consistency in argument ordering, but I forgot what the suggestion was. Does this interface look good to you or do we want to reorder the arguments? Note that if we do reorder them, we should do the same with `full_bipartite` as well.

src/constructors/full.c
```diff
@@ -131,6 +131,149 @@ igraph_error_t igraph_full(igraph_t *graph, igraph_integer_t n, igraph_bool_t di
...
+ * \return Error code.
+ *
+ * Time complexity: O(|V|+|E|), linear in the number of vertices and
+ * edges.
+ *
+ * \sa \ref igraph_full_bipartite() for bipartite full graphs.
+ */
+igraph_error_t igraph_full_multipartite(igraph_t *graph,
+                          igraph_vector_int_t *types,
+                          igraph_vector_int_t *partitions,
```
szhorvat, line 174: This argument should be made `const`.

ntamas:
I'm fine with the argument ordering (but indeed the `partitions` argument, which I'd rather call `sizes`, should be `const`).

This function (and `full_bipartite`) are odd ones out regarding the argument ordering anyway, because they are special in the sense that they both construct a graph _and_ a corresponding "types" vector at the same time. For graph constructors, the ordering is typically:

* the graph being constructed
* any additional data structures that we construct along with the graph
* input arguments for which there is no sensible default so the user will have to provide them for sure even in higher-level interfaces
* input arguments for which there _is_ a reasonable default

This function seems to fit the bill as it is now.

szhorvat:
I would call the `partitions` argument simply `n`, in analogy with `igraph_full` and `igraph_full_bipartite`. Both of these use `n` (or `n1` and `n2`) and the meaning is the same.

pradkrish:
Thanks for the feedback. I will change the `partitions` to `n` in order to be consistent with `igraph_full` and `igraph_full_bipartite` and continue working on Turan graphs.

pradkrish:
I have a question similar to the one I had before, where to put `igraph_turan`?

pradkrish:
What is the best way to compare two graphs in igraph? 

I have a working implementation of `igraph_turan` and when I ran the following snippet to check whether `igraph_full_multipartite` and `igraph_turan` are isomorphic,

```C
    igraph_t g_turan;
    igraph_t g_multipartite;
    igraph_vector_int_t partitions;
    igraph_vector_int_t types1;
    igraph_vector_int_t types2;
    igraph_bool_t res;

    igraph_vector_int_init(&partitions, 4);
    igraph_vector_int_init(&types1, 0);
    igraph_vector_int_init(&types2, 0);

    VECTOR(partitions)[0] = 3;
    VECTOR(partitions)[1] = 3;
    VECTOR(partitions)[2] = 3;
    VECTOR(partitions)[3] = 4;

    igraph_full_multipartite(&g_multipartite, &types1, &partitions, IGRAPH_DIRECTED, IGRAPH_OUT);
    igraph_turan(&g_turan, &types2, 13, 4, IGRAPH_DIRECTED, IGRAPH_OUT);
    igraph_isomorphic(&g_multipartite, &g_turan, &res);
```
The code did not compile. Here is part of the log.
```
/usr/bin/ld: /usr/local/lib/libigraph.a(bliss.cc.o): in function `(anonymous namespace)::bliss_from_igraph(igraph_s const*)':
/home/pradeep/repos/igraph/src/isomorphism/bliss.cc:83: undefined reference to `__cxa_allocate_exception'
/usr/bin/ld: /home/pradeep/repos/igraph/src/isomorphism/bliss.cc:83: undefined reference to `std::runtime_error::runtime_error(char const*)'
/usr/bin/ld: /home/pradeep/repos/igraph/src/isomorphism/bliss.cc:83: undefined reference to `std::runtime_error::~runtime_error()'
/usr/bin/ld: /home/pradeep/repos/igraph/src/isomorphism/bliss.cc:83: undefined reference to `typeinfo for std::runtime_error'
/usr/bin/ld: /home/pradeep/repos/igraph/src/isomorphism/bliss.cc:83: undefined reference to `__cxa_throw'
/usr/bin/ld: /home/pradeep/repos/igraph/src/isomorphism/bliss.cc:89: undefined reference to `operator new(unsigned long)'
/usr/bin/ld: /home/pradeep/repos/igraph/src/isomorphism/bliss.cc:91: undefined reference to `operator new(unsigned long)'
/usr/bin/ld: /home/pradeep/repos/igraph/src/isomorphism/bliss.cc:83: undefined reference to `__cxa_free_exception'
/usr/bin/ld: /home/pradeep/repos/igraph/src/isomorphism/bliss.cc:89: undefined reference to `operator delete(void*)'
/usr/bin/ld: /home/pradeep/repos/igraph/src/isomorphism/bliss.cc:91: undefined reference to `operator delete(void*)'
/usr/bin/ld: /usr/local/lib/libigraph.a(bliss.cc.o): in function `(anonymous namespace)::bliss_info_to_igraph(igraph_bliss_info_t*, bliss::Stats const&)':
/home/pradeep/repos/igraph/src/isomorphism/bliss.cc:173: undefined reference to `__gmpz_init'
/usr/bin/ld: /home/pradeep/repos/igraph/src/isomorphism/bliss.cc:175: undefined reference to `__gmpz_sizeinbase'
/usr/bin/ld: /home/pradeep/repos/igraph/src/isomorphism/bliss.cc:180: undefined reference to `__gmpz_get_str'
/usr/bin/ld: /home/pradeep/repos/igraph/src/isomorphism/bliss.cc:181: undefined reference to `__gmpz_clear'
/usr/bin/ld: /usr/local/lib/libigraph.a(bliss.cc.o): in function `(anonymous namespace)::AutCollector::operator()(unsigned int, unsigned int const*)':
/home/pradeep/repos/igraph/src/isomorphism/bliss.cc:219: undefined reference to `__cxa_allocate_exception'
```
When I manually inspected `g_multipartite` and `g_turan`, they looked the same. I don't think this is related to the implementation of `igraph_turan`. Replacing the last line with `igraph_isomorphic(&g_multipartite, &g_multipartite, &res)` gave the same compilation error.

szhorvat:
This is not a compilation error, but a linking error. If you first compile your .c source to .o object files, then link the object files to an executable, it will arise at the second stage.

It happens because you did not link to all the required libraries, notably the C++ standard library. You can resolve this by either using the C++ linker (e.g. compile with `gcc` but link with `g++`) or by explicitly linking to the C++ standard library. On a typical Linux distro that uses libstdc++, that can be done using `-lstdc++`. Note that some platforms, most notably macOS, use `libc++`, not `libstdc++`.

To avoid all these complications, I recommend that you build your program with CMake, which will figure out the required options automatically. This is described in the tutorial.  If you don't want to use CMake, you can still use `pkg-config --libs --cflags igraph`, after adding the installation location to `PKG_CONFIG_PATH`. This prints the required options, and _should_ include the flag to link to the C++ standard library.

Or, a simple solution is to build igraph as a shared library (not static library) which means that you won't need to link to igraph's dependencies yourself. The shared library knows its own dependencies.

szhorvat:
> I have a question similar to the one I had before, where to put `igraph_turan`?

You can put it in the same file as it's closely related.

> Regarding the interface, I am currently thinking about `igraph_turan(igraph_t *graph, igraph_vector_int_t *types, igraph_integer_t n, igraph_integer_t r, igraph_bool_t directed, igraph_neimode_t mode)`, any thoughts?

I thought a bit about this and I think we should not support directed graphs at all with this function. What is normally understood as a Turán graph is an undirected graph. For those who want a directed version, `full_multipartite` already provides that functionality, or they can convert to a directed one with reciprocal edges later.

pradkrish:
On a related note, I wish `igraph` had some kind of discord channel where we can have discussions related to build questions and such things. I just like the idea of bringing contributors of a project together in some kind of IRC sort of of medium.

pradkrish:
A build related question. I build igraph on my Ubuntu 20.04 LTS machine in the following way 
```
mkdir build
cd build
cmake ..
cmake --build .
cmake --install .
```
This is part of my build configuration
```
-- Version number from Git: 0.10.0-dev+e314ea2c
--  
-- -----[ Build configuration ]----
-- Version:     0.10.0-dev+e314ea2c
-- CMake build type:        Release
-- Library type:             static
-- igraph_integer_t size:   64 bits
-- Compiler cache:           ccache

```
When I load a test script using gdb, I get `No Debugging symbols in ..`. How can I build igraph with all debug symbols?

szhorvat:
The easiest way to explore CMake options is to run `ccmake .` in the build directory after you have configured the project. This gives you an interactive interface to edit options.

You can toggle the `CMAKE_BUILD_TYPE` option between several values, such as `Debug` or `RelWithDebInfo`, both of which use the `-g` option to include debug symbols. This is not igraph specific, but a standard CMake setting, see https://cmake.org/cmake/help/latest/variable/CMAKE_BUILD_TYPE.html

Other useful tips:

 - Install `ccache` to speed up your build (but I see you already have it).
 - Use the `Ninja` generator instead of `Unix Makefiles`. Pass the `-GNinja` option to CMake when you first configure the project. Ninja uses all your CPU cores by default, and makes both configuration and build slightly faster. But it prevents some colouring in the output. To help with this, change the `FORCE_COLORED_OUTPUT` CMake option. To use Ninja by default, set the `CMAKE_GENERATOR` environment variable to the value `Ninja` in your profile.
 - If you want to add custom compiler options, add them to `CMAKE_C_FLAGS` and `CMAKE_CXX_FLAGS`. In the `ccmake` GUI, you can reveal these by pressing `t`. These are also auto-filled from the `CFLAGS` and `CXXFLAGS` environment variables.

ntamas:
What's the status of this PR? Is it ready to merge @szhorvat ? I went through the commits quickly and it seemed good to me.

src/constructors/full.c
```diff
@@ -131,6 +131,202 @@ igraph_error_t igraph_full(igraph_t *graph, igraph_integer_t n, igraph_bool_t di
...
+ *   directed graphs. If \c IGRAPH_OUT, then edges point from vertices
+ *   of low-index vertices to high-index vertices; if \c
+ *   IGRAPH_IN, then the opposite direction is realized; if \c
+ *   IGRAPH_ALL, then mutual edges will be created.
+ * \return Error code.
+ *
+ * Time complexity: O(|V|+|E|), linear in the number of vertices and
+ * edges.
+ *
+ * \sa \ref igraph_full_bipartite() for full bipartite graphs.
```
szhorvat, line 157:
```suggestion
 * \brief Create a full multipartite graph.
 *
 * A multipartite graph contains two or more types of vertices and connections
 * are only possible between two vertices of different types. This function
 * creates a complete multipartite graph.
 *
 * \param graph Pointer to an igraph_t object, the graph will be
 *   created here.
 * \param types Pointer to an integer vector. If not a null pointer,
 *   the type of each vertex will be stored here.
 * \param n Pointer to an integer vector, the number of vertices
 *   of each type.
 * \param directed Boolean, whether to create a directed graph.
 * \param mode A constant that gives the type of connections for
 *   directed graphs. If \c IGRAPH_OUT, then edges point from vertices
 *   of low-index vertices to high-index vertices; if \c
 *   IGRAPH_IN, then the opposite direction is realized; if \c
 *   IGRAPH_ALL, then mutual edges will be created.
 * \return Error code.
 *
 * Time complexity: O(|V|+|E|), linear in the number of vertices and
 * edges.
 *
 * \sa \ref igraph_full_bipartite() for full bipartite graphs.
```

src/constructors/full.c
```diff
@@ -131,6 +131,202 @@ igraph_error_t igraph_full(igraph_t *graph, igraph_integer_t n, igraph_bool_t di
...
+
+    if (nn == 1) {
+        igraph_integer_t num = VECTOR(*n)[0];
+        igraph_empty(graph, num, directed);
+        if (types) {
+            IGRAPH_CHECK(igraph_vector_int_resize(types, num));
+            igraph_vector_int_null(types);
+            for (igraph_integer_t i = 0; i < num; i++) {
+                VECTOR(*types)[0] = 0;
+            }
```
szhorvat, line 189: This is redundant since `vetor_int_null()` was already called above.

```suggestion
```

src/constructors/full.c
```diff
@@ -131,6 +131,202 @@ igraph_error_t igraph_full(igraph_t *graph, igraph_integer_t n, igraph_bool_t di
...
+    igraph_integer_t no_of_edges = 0;
+    igraph_integer_t no_of_nodes;
+    igraph_vector_int_t edges;
+    igraph_vector_int_t n_acc;
+    igraph_integer_t ptr = 0;
+
+    igraph_integer_t nn = igraph_vector_int_size(n);
+
+    if (nn == 0) {
+        igraph_empty(graph, 0, directed);
```
szhorvat, line 174:
```suggestion
        IGRAPH_CHECK(igraph_empty(graph, 0, directed));
```

All functions that can fail need an `IGRAPH_CHECK`

src/constructors/full.c
```diff
@@ -131,6 +131,202 @@ igraph_error_t igraph_full(igraph_t *graph, igraph_integer_t n, igraph_bool_t di
...
+        if (types) {
+            igraph_vector_int_clear(types);
+        }
+        return IGRAPH_SUCCESS;
+    }
+
+    if (nn == 1) {
+        igraph_integer_t num = VECTOR(*n)[0];
+        igraph_empty(graph, num, directed);
```
szhorvat, line 183:
```suggestion
        IGRAPH_CHECK(igraph_empty(graph, num, directed));
```

szhorvat (changes requested):
I took a quick look at `full_multipartite()`, but I did not yet have time to test it. In addition to the comments I added, a few questions:

Could the code be simplified somewhat by making `n_acc` one element longer and adding `n_acc[nn] = no_of_nodes` to the end? For example this would eliminate `v < nn` on line 249 and line 200 as well.

Can the loop on line 248 start from `i=0`, eliminating line 247?

Is it really necessary to special-case `nn==1`, does the implementation for `nn > 1` not handle this well?

A small comment: The four nested loops could be replaced by only two in a manner similar to how the `types` vector is filled out. But the current way is also intuitive.

szhorvat:
In the tests, can you add one where all partitions have zero size, and one where some (but not all) partitions have zero size?

Note: Repeated initialization and destruction of the `types` vector is unnecessary, but it doesn't hurt either, so let's not spend time on that. For as long as the test does the job, it's fine.

szhorvat:
I forgot to say: don't spend too much time on polishing the docs, just make sure that they're complete (include all information) and correct (match what the function does). I'll polish them a bit before merging.

szhorvat:
Regarding line numbers, you could look at the commit which was current at the time of the review:

https://github.com/igraph/igraph/blob/e314ea2c93cfe4d651455e32018c44bf056bdc9f/src/constructors/full.c

pradkrish:
> I took a quick look at `full_multipartite()`, but I did not yet have time to test it. In addition to the comments I added, a few questions:
> 
> Could the code be simplified somewhat by making `n_acc` one element longer and adding `n_acc[nn] = no_of_nodes` to the end? For example this would eliminate `v < nn` on line 249 and line 200 as well.
> 
> Can the loop on line 248 start from `i=0`, eliminating line 247?
> 
> Is it really necessary to special-case `nn==1`, does the implementation for `nn > 1` not handle this well?
> 
> A small comment: The four nested loops could be replaced by only two in a manner similar to how the `types` vector is filled out. But the current way is also intuitive.

I have addressed everything except the four nested loop comment. Like you said, it feels pretty intuitive, so I will leave it as is.

src/constructors/full.c
```diff
@@ -131,6 +131,196 @@ igraph_error_t igraph_full(igraph_t *graph, igraph_integer_t n, igraph_bool_t di
...
+    if (no_of_types == 0) {
+        IGRAPH_CHECK(igraph_empty(graph, 0, directed));
+        if (types) {
+            igraph_vector_int_clear(types);
+        }
+        return IGRAPH_SUCCESS;
+    }
+
+    if (igraph_vector_int_min(n) < 0) {
+        IGRAPH_ERROR("Invalid number of vertices in type vector.", IGRAPH_EINVAL);
```
szhorvat, line 181:
```suggestion
        IGRAPH_ERROR("Number of vertices must not be negative in any partition.", IGRAPH_EINVAL);
```

src/constructors/full.c
```diff
@@ -131,6 +131,196 @@ igraph_error_t igraph_full(igraph_t *graph, igraph_integer_t n, igraph_bool_t di
...
+igraph_error_t igraph_turan(igraph_t *graph,
+                            igraph_vector_int_t *types,
+                            igraph_integer_t n,
+                            igraph_integer_t r) {
+    igraph_integer_t quotient;
+    igraph_integer_t remainder;
+    igraph_vector_int_t subsets;
+
+    if (n < 0) {
+        IGRAPH_ERROR("Invalid number of vertices.", IGRAPH_EINVAL);
```
szhorvat, line 289:
```suggestion
        IGRAPH_ERRORF("Number of vertices must not be negative, got %" IGRAPH_PRId ".", IGRAPH_EINVAL, n);
```

ntamas:
> I have addressed everything except the four nested loop comment. Like you said, it feels pretty intuitive, so I will leave it as is.

I'm okay with the current construction.

src/constructors/full.c
```diff
@@ -131,6 +131,196 @@ igraph_error_t igraph_full(igraph_t *graph, igraph_integer_t n, igraph_bool_t di
...
+    }
+
+    if (r <= 0) {
+        IGRAPH_ERROR("Invalid number of partitions.", IGRAPH_EINVAL);
```
szhorvat, line 293:
```suggestion
        IGRAPH_ERRORF("Number of partitions must be positive, got %" IGRAPH_PRId ".", IGRAPH_EINVAL, r);
```

src/constructors/full.c
```diff
@@ -131,6 +131,197 @@ igraph_error_t igraph_full(igraph_t *graph, igraph_integer_t n, igraph_bool_t di
...
+ * returned if the number of partitions is greater than the number of
+ * vertices.
+ *
+ * \param graph Pointer to an igraph_t object, the graph will be
+ *   created here.
+ * \param types Pointer to an integer vector. If not a null pointer,
+ *   the type (partition index) of each vertex will be stored here.
+ * \param n Integer, the number of vertices in the graph.
+ * \param r Integer, the number of partitions of the graph.
+ * \return Error code.
```
szhorvat, line 273:
```suggestion
 * Turán graphs are complete multipartite graphs with the property
 * that the sizes of the partitions are as close to equal as possible.
 *
 * This function generates undirected graphs. The null graph is
 * returned when the number of vertices is zero. A complete graph is
 * returned if the number of partitions is greater than the number of
 * vertices.
 *
 * \param graph Pointer to an igraph_t object, the graph will be
 *   created here.
 * \param types Pointer to an integer vector. If not a null pointer,
 *   the type (partition index) of each vertex will be stored here.
 * \param n Integer, the number of vertices in the graph.
 * \param r Integer, the number of partitions of the graph, must be
 *   positive.
 * \return Error code.
```

ntamas:
Thank you! This is now merged.

PR #2091, opened 2022-06-11
Add functions to functions.yaml

### LINKED ISSUE #1828: functions.yaml is incomplete
Not all public functions are listed in `functions.yaml`. This issue is to:

 - Identify what is missing.
 - Add all missing functions to `functions.yaml`

GroteGnoom:
To fix #1828

### DIFF
--- a/interfaces/functions.yaml
+++ b/interfaces/functions.yaml
@@ -1555,6 +1555,9 @@ igraph_hrg_create:
     PARAMS: OUT HRG hrg, GRAPH graph, VECTOR prob
     DEPS: prob ON graph
 
+igraph_hrg_resize:
+    PARAMS: INOUT HRG hrg, INTEGER newsize
+
 #######################################
 # Conversion
 #######################################
--- a/interfaces/functions.yaml
+++ b/interfaces/functions.yaml
@@ -604,6 +604,19 @@ igraph_personalized_pagerank:
         vids ON graph, weights ON graph, vector ON graph vids,
         options ON algo
 
+igraph_personalized_pagerank_vs:
+    PARAMS: |-
+        GRAPH graph, PAGERANKALGO algo=PRPACK,
+        OUT VERTEX_QTY vector, OUT REAL value,
+        VERTEXSET vids=ALL, BOOLEAN directed=True,
+        REAL damping=0.85,
+        VERTEXSET reset_vids,
+        OPTIONAL EDGEWEIGHTS weights,
+        INOUT PAGERANKOPT options=NULL
+    DEPS: |-
+        vids ON graph, weights ON graph, vector ON graph vids,
+        options ON algo
+
 igraph_rewire:
     PARAMS: INOUT GRAPH rewire, INTEGER n, REWIRING_MODE mode=SIMPLE
 
--- a/interfaces/functions.yaml
+++ b/interfaces/functions.yaml
@@ -2404,3 +2404,11 @@ igraph_strerror:
 
 igraph_expand_path_to_pairs:
     PARAMS: INOUT VERTEX_INDICES path
+
+#######################################
+# Meta info
+#######################################
+
+igraph_version:
+    PARAMS: CSTRING_P version_string, OPTIONAL OUT INT major, OPTIONAL OUT INT minor, OPTIONAL OUT INT subminor
+    RETURN: VOID
--- a/interfaces/functions.yaml
+++ b/interfaces/functions.yaml
@@ -615,7 +615,7 @@ igraph_personalized_pagerank_vs:
         VERTEXSET vids=ALL, BOOLEAN directed=True,
         REAL damping=0.85,
         VERTEXSET reset_vids,
-        OPTIONAL EDGEWEIGHTS weights,
+        OPTIONAL EDGEWEIGHTS weights=NULL,
         INOUT PAGERANKOPT options=NULL
     DEPS: |-
         vids ON graph, weights ON graph, vector ON graph vids,
@@ -969,7 +969,8 @@ igraph_average_local_efficiency:
 igraph_sort_vertex_ids_by_degree:
     PARAMS: |-
         GRAPH graph, OUT VERTEX_INDICES outvids, VERTEXSET vids,
-        NEIMODE mode=OUT, BOOLEAN loops=False, ORDER order, BOOLEAN only_indices=False
+        NEIMODE mode=OUT, BOOLEAN loops=False, ORDER order=ASCENDING,
+        BOOLEAN only_indices=False
     DEPS:
         vids ON graph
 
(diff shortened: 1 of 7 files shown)


interfaces/functions.yaml
```diff
@@ -171,6 +174,13 @@ igraph_realize_degree_sequence:
         OUT GRAPH graph, VECTOR_INT out_deg, OPTIONAL VECTOR_INT in_deg=NULL,
         EDGE_TYPE_SW allowed_edge_types=SIMPLE, REALIZE_DEGSEQ_METHOD method=SMALLEST
 
+igraph_circulant:
+    PARAMS: OUT GRAPH graph, INTEGER n, VECTOR_INT shifts, BOOLEAN directed
+
+igraph_weighted_sparsemat:
+    PARAMS:
+        OUT GRAPH graph, SPARSEMAT A, BOOLEAN directed, CSTRING attr, BOOLEAN loops=False
```
ntamas, line 182: Note to ourselves: this reflects the current state correctly, but we will probably have to update `igraph_weighted_sparsemat` soon to be consistent with `igraph_weighted_adjacency` in the sense that it does not take an attribute _name_ but a vector (in which the weights will be stored). See `CHANGELOG.md` for the reasons.

interfaces/functions.yaml
```diff
@@ -572,6 +608,19 @@ igraph_personalized_pagerank:
...
         options ON algo
 
+igraph_personalized_pagerank_vs:
+    PARAMS: |-
+        GRAPH graph, PAGERANKALGO algo=PRPACK,
+        OUT VERTEX_QTY vector, OUT REAL value,
+        VERTEXSET vids=ALL, BOOLEAN directed=True,
+        REAL damping=0.85,
+        VERTEXSET reset_vids,
+        OPTIONAL EDGEWEIGHTS weights,
```
ntamas, line 618: Default value of `weights` should be `NULL`

interfaces/functions.yaml
```diff
@@ -917,6 +966,19 @@ igraph_average_local_efficiency:
         BOOLEAN directed=True, NEIMODE mode=ALL
     DEPS: weights ON graph
 
+igraph_sort_vertex_ids_by_degree:
+    PARAMS: |-
+        GRAPH graph, OUT VERTEX_INDICES outvids, VERTEXSET vids,
+        NEIMODE mode=OUT, BOOLEAN loops=False, ORDER order, BOOLEAN only_indices=False
```
ntamas, line 972: I think that we can reasonably assume ascending order as the default value for `order`

interfaces/functions.yaml
```diff
@@ -1011,6 +1073,13 @@ igraph_bipartite_game_gnm:
         INTEGER n1, INTEGER n2, INTEGER m, BOOLEAN directed,
         NEIMODE mode
 
+igraph_bipartite_game:
+    PARAMS: |-
+        OUT GRAPH graph, OPTIONAL OUT VECTOR_BOOL types,
+        ERDOS_RENYI_TYPE type, INTEGER n1, INTEGER n2, REAL p,
```
ntamas, line 1079: Let's add defaults for `p` and `m` here; `p` is ignored for G(n,m) graphs and `m` is ignored for G(n,p) graphs so it does not make sense to ask the user to specify both.

Similarly, we should provde defaults for `directed` and `mode`

interfaces/functions.yaml
```diff
@@ -1090,6 +1159,12 @@ igraph_maximal_cliques:
     PARAMS: GRAPH graph, OUT VERTEXSET_LIST res, INTEGER min_size=0, INTEGER max_size=0
     DEPS: res ON graph
 
+igraph_maximal_cliques_subset:
+    PARAMS: |-
+        GRAPH graph, VECTOR_INT subset, OUT VECTOR_INT_LIST res,
```
ntamas, line 1164: `subset` should be of type `VERTEX_INDICES` to make sure that R translates between 0-based and 1-based indices when generating code.

Similarly, `res` should be `VERTEXSET_LIST`

interfaces/functions.yaml
```diff
@@ -1090,6 +1159,12 @@ igraph_maximal_cliques:
...
+        OUT INTEGER no, OUTFILE outfile, INTEGER min_size, INTEGER max_size
+    DEPS: res ON graph
```
ntamas, line 1166: `subset` needs to depend on `graph` if we change its type to `VERTEX_INDICES`

interfaces/functions.yaml
```diff
@@ -1742,10 +1842,18 @@ igraph_compose:
         OUT INDEX_VECTOR edge_map1, OUT INDEX_VECTOR edge_map2
     DEPS: edge_map1 ON g1, edge_map2 ON g2
 
+igraph_induced_subgraph_map:
+    PARAMS: |-
+        GRAPH graph, OUT GRAPH res, VERTEXSET vids, SUBGRAPH_IMPL impl, OPTIONAL OUT VECTOR_INT map,
+        OPTIONAL OUT VECTOR_INT invmap
```
ntamas, line 1848: `vids` need to depend on `res` here

interfaces/functions.yaml
```diff
@@ -1766,6 +1874,14 @@ igraph_maxflow_value:
 igraph_mincut_value:
     PARAMS: GRAPH graph, OUT REAL res, EDGE_CAPACITY capacity
 
+igraph_st_mincut:
+    PARAMS: |-
+        GRAPH graph, OUT REAL value, OUT EDGESET cut, OUT VERTEXSET partition1, OUT VERTEXSET partion2,
```
ntamas, line 1879: Typo here: `partion2` should be `partition2`

interfaces/functions.yaml
```diff
@@ -1766,6 +1874,14 @@ igraph_maxflow_value:
...
+        VERTEX source, VERTEX target,
+        EDGE_CAPACITY capacity
+    DEPS:
+        source ON graph, target ON graph
```
ntamas, line 1883: `capacity` needs to depend on `graph`

interfaces/types.yaml
```diff
@@ -46,6 +46,10 @@ CSTRING:
     # A null-terminated immutable C string
     CTYPE: const char*
 
+CSTRING_P:
```
ntamas, line 49: This is not needed; declaring a parameter as `OUT CSTRING` will add an extra level of pointer indirection automatically.

interfaces/functions.yaml
```diff
@@ -2174,9 +2343,76 @@ igraph_vertex_coloring_greedy:
...
+
+igraph_expand_path_to_pairs:
+    PARAMS: INOUT VERTEX_INDICES path
+
+#######################################
+# Meta info
+#######################################
+
+igraph_version:
+    PARAMS: CSTRING_P version_string, OPTIONAL OUT INT major, OPTIONAL OUT INT minor, OPTIONAL OUT INT subminor
```
ntamas, line 2417: `version_string` needs to be an `OPTIONAL OUT CSTRING`

interfaces/functions.yaml
ntamas: Nope, actually, it needs to depend on `graph`

szhorvat:
There were some here that should probably never be exposed to a high-level language such as `igraph_has_attribute_table`, but I guess they don't hurt either.

PR #2097, opened 2022-06-25
Make random generators overflow safe

GroteGnoom:
Part of #51


GroteGnoom:
In [#51](https://github.com/igraph/igraph/issues/51#issuecomment-1058187413) it says that grg is already checked, but I don't understand how. It seems to me that the maximum number of edges is `nodes * nodes`, and that can overflow.

I added checks.

szhorvat:
> In [#51](https://github.com/igraph/igraph/issues/51#issuecomment-1058187413) it says that grg is already checked, but I don't understand how. It seems to me that the maximum number of edges is ~`nodes * nodes`, and that can overflow.
> 
> I added checks.

`igraph_vector_int_push_back` is overflow-safe. If a vector is only expanded using `push_back`, no checks are needed.

szhorvat:
I have a some not-yet-push work on my computer related to converting reals to integers, which is not arithmetic, but can also cause overflow (e.g. a positive real may convert to a negative integer due to overflow). These affect some of the Erdős-Rényi type functions. Before you start working on related issues, can you ping me please?

ntamas:
@szhorvat Shall we consider this as a blocker for this PR, or can this PR be merged (in theory, when it's complete) without those?

szhorvat:
If you are referring to the float-to-int kind of overflow, no, that will not go into this PR.

If you are referring to the unnecessary checks, yes, I would like those removed before merging this PR (unless I missed something and there's a reason why they're necessary).

src/games/callaway_traits.c
```diff
@@ -88,6 +88,10 @@ igraph_error_t igraph_callaway_traits_game(igraph_t *graph, igraph_integer_t nod
         IGRAPH_ERROR("The number of vertices must be non-negative.", IGRAPH_EINVAL);
     }
 
+    if(edges_per_step < 0){
```
szhorvat, line 91:
```suggestion
    if (edges_per_step < 0) {
```

szhorvat:
I looked a bit more, and it seems that most of the changes are checking edge number before `push_back()`. Can you please remove these, so the code doesn't get overcomplicated, and so that we don't do redundant checks?

There are some checks on edge number input parameters. I have not looked at these carefully, but again: are they necessary? Is there any situation where overflow would occur if the check is not present? Looking e.g. at `igraph_lastcit_game()`, it's not clear to me how overflow would occur if `nodes * edges_per_node > IGRAPH_ECOUNT_MAX`. All subsequent operations seem to be already protected. If this is not the case, there should be a comment indicating where overflow can occur.

To be clear, overflow does not refer to passing in a too-large number to an igraph function, but an arithmetic operation giving incorrect result, e.g. `a + b` becoming negative when `a` and `b` are large positive.

So there do not need to be checks on every single edge count input parameter. All these would eventually use `igraph_create()` or `igraph_empty()` and `igraph_add_edges()`, which are already protected.

Note: Checking that edge counts are not negative is a different matter: that should indeed be done, as negative values will cause trouble in most algorithms, even before `igraph_create()` is reached.

szhorvat:
> The checks like `no_of_edges > IGRAPH_ECOUNT_MAX` now just mean that no_of_edges gets multiplied by 2 afterwards

You are right, this is a cleaner way than doing a safe multiplication.

At first my thinking was that `IGRAPH_ECOUNT_MAX` may change in the future, so there's a mismatch in semantics. But the entire justification for the current value of `IGRAPH_ECOUNT_MAX` is to allow multiplication by two without concerns. Could you paste a comment in front of these checks though? I'm sure eventually we (or at least I) will forget the purpose of these checks.

ntamas:
Maybe do we want to add a macro for the following code snippet?

```c
    /* To ensure the size of the edges vector will not overflow. */
    if (no_of_edges > IGRAPH_ECOUNT_MAX) {
        IGRAPH_ERROR("Overflow in number of edges.", IGRAPH_EOVERFLOW);
    }
```

It seems to occur frequently here and we probably have other places in the code so we could save us some typing by adding something like `IGRAPH_CHECK_EDGE_COUNT(no_of_edges)`, and we could also unify the error message in one single place.

Other than that, I'm happy with the PR.

szhorvat:
Regarding `simple_interconnected_islands_game()`, as I proposed a couple of weeks ago in chat, I think this function should be deprecated and removed.

It is a very specific concept. While some people might need similar random graphs, it is unlikely that anyone would need exactly this one. The function was not very well written (was not written by Gábor or Tamás), and we already had to make multiple fixes. I still has problems: for example, it can create multi-edges between islands, but not within islands. I expect that this was an oversight and not intentional. In the end, such functions are a drain on our time, and probably no-one uses them anyway.

Any objections to deprecating (and eventually removing) this one?

A while ago I have already looked at making Erdős-Rényi $G(n,p)$ style functions overflow safe and have a half-done solution for this function on my computer. It goes further than the fix here (e.g. it is careful not to over-estimate the edge count by too much). It was while working on this that I came to the conclusion that this function is best removed.

This is another lesson on being more careful when taking external contributions. We need to make sure that they are up to a standard where they won't be a drain on our time later.

szhorvat:
In other functions, I used `no_of_edges2` to denote twice the number of edges. I don't mind what name we use (and I agree that `no_of_edge_endpoints` is clearer) but it would be useful to standardize on one name for readability.

szhorvat:
`preference_game` is one of the $G(n,p)$-style functions, along which "islands game", which I mentioned needs special attention. Let's discuss this here then, as the same issues come up with many other functions as well, which somehow incorporate a $G(n,p)$ component in their definitions.

Generally, $G(n,p)$-style functions consider the entries of the adjacency matrix as a linearly-indexable flat vector. They differ on the details of how the adj. mat. entries are mapped to the vector depending on whether the graph is directed and whether self-loops are allowed, but generally they use the same idea: generate flat indices by adding up geometrically distributed random numbers (the gaps between entries), and store them into an `s` vector. Then decode these into `(u, v)` vertex index pairs and add them as edges of the graph.

These flat indices are represented as floating point numbers in order to allow large graphs on 32-bit systems. Otherwise we would have a vertex count limit of `sqrt(2^31 - 1) ~ 46340`.  Some of this was broken during the 64-bit transition, as some values were converted to integer types at least temporarily. I fixed this a few months ago. 

The steps for $G(n,p)$ generation are usually as follows:

 1. Compute the maximum possible flat index, which is equal to the maximum possible edge count, `maxedges`. This is stored as a floating point value.
 2. Estimate how many edges there will be. On average there are `p*maxedges`, but there may be more, so we typically reserve `1.1 * p * maxedges` storage.
 3. Generate flat indices and collect them into an `s` float vector using `push_back()`. Note that `push_back()` itself is overflow-safe.
 4. Convert the flat indices (floating point numbers) back to vertex indices (integer). This is another potential overflow situation. 

Issues to consider:

 - Float-to-integer conversion may overflow (and produce negative values). Relevant when reserving `1.1 * p * maxedges` storage. On my machine, I have a macro to protect against this, but I wasn't yet happy with the final design.
 - Do we want to hard-fail when trying to _reserve_ storage in advance, in a situation when we do not know how much storage will actually be used? Or can we better estimate the needed storage? I do have a better estimator than multiplying by `1.1` but in the end this doesn't really matter. Reducing the largest possible graph sizes by 10% is no big deal IMO.
 - A more subtle issue is that floats may not be able to represent all integers that integer types can. We have a limit of 2^53 on most platforms. For larger values, we get rounding errors. Can there be truly bad consequences? Well, some precision is lost, but that's probably not a big deal. But can worse things happen, such as two numbers, which the code assumes to be different, becoming identical, and somehow causing a crash? Sure, we will never have a vector with 2^53 elements, but this is precisely the situation where we _don't_ allocate a `maxedges`-long vector yet we still need to represent `maxedges` accurately. Otherwise, it could happen that we store two identical flat-indices in the `s` vector, which will lead to multi-edges. This can _potentially_ happen with a graph having `sqrt(2^53) ~ 94906266`, 94 million vertices. Sure, that's very large, but not inconceivably large. It's comfortable representable even on 32-bit machines.
 - When converting the flat indices back to integers, can we get overflow? I think not, as `maxedges` itself was computed for a representable vertex count.

szhorvat:
I laid out some of the issues with functions that deal with $G(n,p)$, but not any proposals for solutions. As I said, I was not happy with what I had locally. Suggestions are welcome.

----

Let me give a starting point. It might be useful to have a real-to-integer conversion macro.

```
/* Overflow-safe conversion of real to integer. Checks whether the value is too large
 * (positive or negative), but does not check if the value is a whole number. */
#define IGRAPH_SAFE_REAL_TO_INT(r, i) \
    do { \
        igraph_real_t _safe_r = (r); \
        if (IGRAPH_INTEGER_MIN <= _safe_r && _safe_r <= IGRAPH_INTEGER_MAX) { \
            *(i) = r; \
        } \
        else { \
            IGRAPH_ERRORF("Cannot convert %g to integer, outside of representable range.", \
                IGRAPH_EOVERFLOW, _safe_r); \
        } \
    } while (0)
```

Let's have a function that estimates the number of edges based on `maxedges` and `p`, _and_ raises an error on overflow at the same time. This can be used in all similar functions. If we want a better estimation than multiplying by `1.1`, we can easily do that then, but this is not very important. The bigger convenience of such a function is the overflow check.

----

Regarding integers that are non-representable by `double`s, here's a macro to help with checking for them:

```
/* Largest positive value for igraph_real_t that can safely represent integers. */
#define IGRAPH_MAX_EXACT_REAL ((double)(1LL << DBL_MANT_DIG))
```

This assumes that: (1) we use IEEE doubles, which we probably assume elsewhere anyway (2) `long long` is 64-bit, which is true on all 32-bit systems we are currently dealing with.

That said, while thing _can_ go wrong in graphs with more than 94 million vertices, is it not perhaps exceedingly unlikely that they will go wrong? Suppose that `maxedges` is large enough that `maxedges - 1 == maxedges` due to roundoff. Isn't the probability for a geometrically distributed gap to be as small as 1 exceedingly unlikely for a reasonable number of edges?

Let's make a small calculation. Suppose that `maxedges = 2 * 2^53`, which happens with a vertex count of `V = 2^27` = 134 million assuming loopy directed graphs. Now let's choose `p` to give the same number `E = 2^27` of edges on average, i.e. `p = E / maxedges`. The probability of getting a gap of 1 is `p`. What is the change that this never happens for `E / 2` edges (which is the expected number to have a flat index >= 2^53)?  That's `(1 - p)^(E/2)`, which is approximately `1 - E*p/2` for small `p`, i.e. `1 - E^2 / maxedges`, which is not close to 1 with our choice of `E ~ sqrt(maxedges)`!! The exact value is about 60% here.

In other words, there is a 40% chance that a $G(n,p)$ graph with 134 million vertices and mean edge count of the same will incorrectly get at least one multi-edge!

So yes, this is a real problem.

ntamas:
> Regarding `simple_interconnected_islands_game()`, as I proposed a couple of weeks ago in chat, I think this function should be deprecated and removed. [...] In the end, such functions are a drain on our time, and probably no-one uses them anyway.

Agree with the first part (it's a drain) but not with the second one. It's exposed via the R interface (see `sample_islands`), and I think we don't have a clear picture about what parts of igraph are being used and what parts are not being used (and since there's no telemetry built into igraph, we'll probably never know). `simple_interconnected_islands_game()` might be used to generate small graphs for presentational purposes, or it might already be a part of a textbook or university course about igraph, or God knows what, so I wouldn't remove it from the library just because it has subtle issues regarding multi-edges. If there are such bugs, and we don't want to fix them (which I totally understand - we must prioritize things), then let's document the shortcomings, put a big warning in the documentation, maybe move the function into a separate section, and see if anyone complains. If people complain, it means that the function _is_ being used and we should put some effort into fixing it.

I wish we had some way to ask the community whether these problematic functions are actually being used, though. Asking it on the forums is probably not going to reach enough people so we could tell the difference.

Regarding the questions raised about representing "flat" indices as floats:

> Do we want to hard-fail when trying to reserve storage in advance, in a situation when we do not know how much storage will actually be used?

Yes, I think we should. Again, if people hit this limit and start to complain, then we can start thinking about better ways to estimate the required storage, or just allocate smaller space and go ahead, hoping that the generation will not hit the limit in the end. But until people start requesting this, let's focus our efforts on places where it pays off faster.

> A more subtle issue is that floats may not be able to represent all integers that integer types can. We have a limit of 2^53 on most platforms. For larger values, we get rounding errors.

This is indeed a problem if this somehow results in multi-edges being generated. The total number of edges should be limited to `IGRAPH_MAX_EXACT_REAL` if we use the flat index trick during generation.

szhorvat:
Re G(n,p), I'm happy with making it a hard error. I wanted to note that the `if (maxedges > IGRAPH_MAX_EXACT_REAL)` check is itself approximate and affected by roundoff errors in the calculation of `maxedges`. However, I think it's good enough: the chance of multi-edges with this check is exceedingly small.

ntamas:
Yes, I guess  that's the best we can do; if we want to be pedantic, we can adjust the 1.1 factor in a way that we calculate the probability of having more than `maxedges` edges, and calibrate `maxedges` in a way that the probability of exceeding this limit is small enough (say, 1e-7). But this may be an overly conservative estimate that prevents the algorithm from running to completion.

szhorvat:
Regarding interconnected islands game:

How about a more gradual path to removal then? Step 1 is update docs (what you suggested @ntamas), step 2 explicit deprecation, step 3 removal. In the end, this is basically a kind of SBM where G(n,p) is used within block and G(n,m) between blocks. So there are suitable replacements.

If people do indeed use this function, I am not comfortable leaving it in its current buggy state. Regarding _only_ updating the docs: people rarely read the docs. If someone is already using this function, they won't check the docs for updates ...

ntamas:
> How about a more gradual path to removal then? Step 1 is update docs (what you suggested @ntamas), step 2 explicit deprecation, step 3 removal. In the end, this is basically a kind of SBM where G(n,p) is used within block and G(n,m) between blocks. So there are suitable replacements.

Hmmm, I wonder whether we could simply re-write it then in terms of `igraph_erdos_renyi_game()` and / or `igraph_sbm_game()` and get rid of all the issues. I mean, we could generate the islands with `igraph_sbm_game()`, and then generate multiple Erdos-Renyi subgraphs with G(n, m) and overlay them on the original graph by rewriting the vertex IDs (the islands have consecutive vertex IDs, right?). I don't really care about performance here if it's indeed not used that frequently.

ntamas:
`igraph_simple_interconnected_islands_game()` is now fixed in b702036e6 so it does not generate multi-edges any more.

include/igraph_types.h
```diff
@@ -87,6 +88,9 @@ typedef int    igraph_bool_t;
 #  define IGRAPH_UINT_MIN UINT32_MIN
 #endif
 
+/* Largest positive value for igraph_real_t that can safely represent integers. */
+#define IGRAPH_MAX_EXACT_REAL ((double)(1LL << DBL_MANT_DIG))
+
```
szhorvat, line 93: Let's put this into `safe_intop.h`. It does not need to be public.

src/games/preference.c
```diff
@@ -263,6 +264,10 @@ igraph_error_t igraph_preference_game(igraph_t *graph, igraph_integer_t nodes,
                 }
             }
 
+            if (maxedges > IGRAPH_MAX_EXACT_REAL) {
+                IGRAPH_ERROR("Overflow in number of edges.", IGRAPH_EOVERFLOW);
```
szhorvat, line 268: This is technically not an overflow in the number of edges, but in the maximum number of edges. What would be a better error message here that also communicates to the user what they need to do (i.e. reduce the vertex count)?

Maybe just be as explicit as possible and write

```
Too many vertices, overflow in maximum number of edges.
```

What do you think @GroteGnoom ?

GroteGnoom: I completely agree 🙂

szhorvat: There were some other instances of this, which can be found by searching for `IGRAPH_MAX_EXACT_REAL`. Can you fix those too?

Also, the same type of fix will be necessary for all functions that are "G(n,p)-like". Off the top of my head, `erdos_renyi_gnp`, `sbm`, `hsbm`, `islands`. Grepping for `maxedges` will be helpful in finding these. Can you please fix these too?

GroteGnoom: sbm seems to be a different situation, but I still think the check is correct.

Don't we have 2 problems? One for maxedges > MAX_EXACT_REAL, but also one where igraph_integer_t is 32 bits and maxedges overflows it when calling igraph_vector_reserve(.., maxedges) ?

szhorvat: In G(n,p)-style generators, we never reserve a vector of size maxedges. This value is easily much larger than available memory. Where exactly do you see `maxedges` storage being reserved? Can you please post a direct link to the line?

include/igraph_types.h
ntamas: It seems like this is done now.

src/games/preference.c
ntamas: All instances related to `IGRAPH_MAX_EXACT_REAL` seem to be fixed in the PR.

I have searched the codebase for `igraph_.*_reserve.*maxedges` (regex) and there are four instances where we try to reserve space for a vector with `maxedges * p * 1.1` elements. On 32-bit machines this can potentially overflow if `maxedges * p * 1.1 > INT32_MAX`, but it's even worse because converting a double to an integer type when the double is out of bounds for the integer type is undefined behaviour. So I'm adding an `igraph_i_safe_ceil()` and `igraph_i_safe_floor()` to truncate the double first using `ceil()` or `floor()` (to make the intent explicit), and then do a range check before storing the result in an integer.

ntamas: Resolved in [e844432](https://github.com/igraph/igraph/pull/2097/commits/e844432ca048e47d0833fc47f51637bb94445349)

ntamas:
I'd say there are plenty of improvements in this PR now so it would be worth merging it. We can keep on adding overflow checks later on in separate PRs. Any objections?

GroteGnoom:
No objections from me.

ntamas:
Thanks!

PR #2101, opened 2022-07-01
Caching graph properties

ntamas:
This is a joint work of @szhorvat and me and is meant to resolve https://github.com/igraph/igraph/issues/1219

I think the branch is ready for a merge now but I've opened a PR anyway for a final round of reviews before we merge it.


szhorvat:
Can you give me til Monday to look at this?

include/igraph_datatype.h
```diff
@@ -30,6 +30,41 @@
...
+    IGRAPH_PROP_HAS_LOOP = 0,
+
+    /* Stores whether the graph has at least one multi-edge, taking into account
+     * edge directions in directed graphs. In other words, this property should
+     * be false for a directed graph with edges (a, b) and (b, a). */
+    IGRAPH_PROP_HAS_MULTI,
+
+    /* Stores whether the graph has at least one multi-edge, ignoring edge
+     * directions in directed graphs. In other words, this property should
+     * be true for a directed graph with edges (a, b) and (b, a). */
```
szhorvat, line 47: It might be better to interpret this precisely as the name would suggest, i.e. make it true when there are reciprocal edges. Examples:

```
a -> b, b -> a

MULTI=false, RECIPROCAL=true
```

```
a -> b, a -> b

MULTI=true RECIPROCAL=false
```

```
a -> b, b -> a, b->a

MULTI=true, RECIPROCAL=true
```

In _undirected_ graphs, `RECIPROCAL` would just be ignored.

I think that there are legitimate reasons to have a public function that checks for reciprocal edges, and it would be nice if the meaning of this property matched. That function, however, is outside of the scope of this PR.

ntamas:
> Can you give me til Monday to look at this?

Sure, there's no rush.

include/igraph_datatype.h
ntamas: Fixed in 3d2c043

ntamas:
@szhorvat Did you have a chance to take a look at this PR?

src/connectivity/components.c
```diff
@@ -384,39 +390,48 @@ static igraph_error_t igraph_is_connected_weak(const igraph_t *graph, igraph_boo
...
+
+        case IGRAPH_STRONG:
+            prop = IGRAPH_PROP_IS_STRONGLY_CONNECTED;
+            break;
+
+        default:
+            IGRAPH_ERROR("Invalid connectedness mode.", IGRAPH_EINVAL);
+    }
+
+    IGRAPH_RETURN_IF_CACHED_BOOL(graph, prop, res);
```
szhorvat, line 411: If the graph is undirected, it should always consider `IGRAPH_PROP_IS_WEAKLY_CONNECTED`, no?

Why not do the caching directly in `igraph_is_connected_weak` and `igraph_i_connected_components_strong`?

include/igraph_datatype.h
```diff
@@ -30,6 +30,43 @@
 
 __BEGIN_DECLS
 
+struct igraph_i_property_cache_t;
+typedef struct igraph_i_property_cache_t igraph_i_property_cache_t;
+
+typedef enum {
+    /* Stores whether the graph has at least one self-loop */
```
szhorvat, line 37:
```suggestion
    /* Stores whether the graph has at least one self-loop. */
```

include/igraph_datatype.h
```diff
@@ -30,6 +30,43 @@
...
+    /* Stores whether the graph has at least one reciprocal edge pair. Ignored
+     * in undirected graphs. This property should be true for a directed graph
+     * with edges (a, b) and (b, a), and false for a directed graph with
+     * edges (a, b) and (a, b) again. */
+    IGRAPH_PROP_HAS_RECIPROCAL,
+
+    /* Stores whether the graph is weakly connected */
+    IGRAPH_PROP_IS_WEAKLY_CONNECTED,
+
+    /* Stores whether the graph is strongly connected */
```
szhorvat, line 55:
```suggestion
    /* Stores whether the graph is strongly connected. Ignored in undirected graphs. */
```

include/igraph_datatype.h
```diff
@@ -30,6 +30,43 @@
...
+     * for a directed graph with edges (a, b) and (a, b) again. */
+    IGRAPH_PROP_HAS_MULTI,
+
+    /* Stores whether the graph has at least one reciprocal edge pair. Ignored
+     * in undirected graphs. This property should be true for a directed graph
+     * with edges (a, b) and (b, a), and false for a directed graph with
+     * edges (a, b) and (a, b) again. */
+    IGRAPH_PROP_HAS_RECIPROCAL,
+
+    /* Stores whether the graph is weakly connected */
```
szhorvat, line 52:
```suggestion
    /* Stores whether the graph is weakly connected. */
```

include/igraph_datatype.h
```diff
@@ -30,6 +30,43 @@
...
+    IGRAPH_PROP_IS_WEAKLY_CONNECTED,
+
+    /* Stores whether the graph is strongly connected */
+    IGRAPH_PROP_IS_STRONGLY_CONNECTED,
+
+    /* Stores whether the graph is a directed acyclic graph. Not used for
+     * undirected graphs */
```
szhorvat, line 59:
```suggestion
    /* Stores whether the graph is a directed acyclic graph. Not used for
     * undirected graphs. */
```

include/igraph_datatype.h
```diff
@@ -30,6 +30,43 @@
...
+    IGRAPH_PROP_IS_DAG,
+
+    /* Stores whether the graph is a forest, i.e. an undirected or directed
+     * graph that is cycle-free even if we ignore edge directions */
```
szhorvat, line 63:
```suggestion
    /* Stores whether the graph is a forest, i.e. an undirected or directed
     * graph that is cycle-free even if we ignore edge directions. */
```

include/igraph_datatype.h
```diff
@@ -30,6 +30,43 @@
...
+     * undirected graphs. */
+    IGRAPH_PROP_IS_DAG,
+
+    /* Stores whether the graph is a forest, i.e. an undirected or directed
+     * graph that is cycle-free even if we ignore edge directions. */
+    IGRAPH_PROP_IS_FOREST,
+
+    /* Dummy value used to count enum values */
+    IGRAPH_PROP_I_SIZE
+} igraph_cached_property_t;
```
szhorvat, line 68: Since we decided to leave the door open to adding non-bool properties in the future, should this type not have `bool` in the name as well? Right now the system works by assuming that these enum values can be used to index into a Boolean array.

I'm not sure what the best way is here, but if we want to design the API to allow this kind of extension in the future, it's best to think about it now.

Personally, I don't _think_ we'll ever need non-bool properties though (but as we've learned, these things are hard to predict).

src/graph/caching.c
```diff
@@ -0,0 +1,126 @@
...
+    assert(graph->cache != NULL);
+    /* Even though graph is const, updating the cache is not considered modification.
+     * Functions that merely compute graph properties, and thus leave the graph structure
+     * intact, will often update the cache. */
+    graph->cache->value[prop] = value;
+    graph->cache->known[prop] = 1;
+}
+
+/**
+ * \brief Clears the cached value of a property in a graph.
```
szhorvat, line 104:
```suggestion
 * \brief Invalidates the cached value of a property in a graph.
```

src/graph/caching.c
```diff
@@ -0,0 +1,126 @@
...
+ *
+ * \param graph  the graph whose cache is to be modified
+ * \param prop   the property to clear from the cache
```
szhorvat, line 107:
```suggestion
 * \param prop   the property to invalidate in the cache
```

szhorvat:
Let's be smarter about cache invalidation upon modification.

Adding edges does not invalidate:

 - strong / weak connectivity

Deleting edges does not invalidate:

 - is_dag, is_forest

Adding vertices does not invalidate:

 - has_loop, has_multi, has_reciprocal, is_dag, is_forest

Deleting vertices does not invalidate:

 - is_dag, is_forest

The safe way to do this is not to explicitly invalidate a specific set of properties, but to selectively preserve a set. This way if we add new properties in the future, there is a smaller chance of breaking things.

szhorvat:
`is_tree` can make use of the cache for _undirected_ graphs. A tree is a connected forest, so it can return quickly if IS_FOREST && IS_WEAKLY_CONNECTED.

szhorvat:
There are some tests within the library which can be used to _set_ cached properties, even if they can't make use of cached values. Some that I can think of:

 - If the graph is a _directed_ in- or out-tree or forest, then set `is_dag`, `!has_reciprocal`, `!has_loop`
 - If the graph is a DAG, then set `!has_reciprocal`, `!has_loop`
 - If the graph is an _undirected_ forest, then set `!has_multi` and `!has_loop`.
 - If the graph has loops, then set `!is_forest` and in the directed case also `!is_dag`
 - If the graph has multi-edges, then set `!is_forest`
 - If the graph has reciprocal edges and is directed, then set `!is_dag`
 - ... ??

Tricky ones we need to be careful about:

 - If the graph has more than one vertex (!!) then being a DAG implies `!strongly_connected` and being strongly connected implies `!is_dag`. But note that the singleton graph is a strongly connected DAG, hence the need for the vertex count check!

szhorvat:
As we discussed, some checks should be performed during graph construction. Even if not done within this PR, I want to have this written down.

 - `igraph_empty_attrs` should set every property. Here we need to be careful with the special case of the null graph and singleton graph. For example, `n=0` is disconnected, `n=1` is connected, `n=2` is disconnected again.
 - `igraph_add_edges` should check if it is adding new loop edges and update `has_loop` accordingly (if this property was already known)
 - ... ??

include/igraph_datatype.h
ntamas: I think that the properties can be stored in a common enum, and if we start introducing properties with different types, we could have an array that maps property names into their storage types. Otherwise, if we started introducing different property enums for different cached property types, we'd need to have three variants of the `igraph_i_property_cache_has()` and the `igraph_i_property_cache_invalidate()` functions, one for each property enum.

ntamas:
I'm getting increasingly worried about adding all sorts of custom logic that preemptively sets or uses cache bits (apart from the most trivial cases, like caching `IGRAPH_PROP_IS_FOREST` in `igraph_is_forest()` of course). I believe it's hard to draw the line between simple cases and "things that look simple but have subtle edge cases where things are not that simple any more". We should be more conservative and add more complex caching logic only in cases where there seems to be a specific use-case that suffers from performance loss. Gaining a few microseconds with caching over just computing things straight away does not seem to be a benefit that's large enough to justify the complexity of the logic we need to add to achieve it.

That being said, I find the idea of not invalidating everything when adding / deleting vertices or edges easy enough to start thinking a bit more about it. I'd probably add four "bitmasks" for these four operations; the mask is set to 1 for cached properties that should be _kept_ after the operation. We could then use a similar bitmask to store which properties are actually being cached, and then the selective cache invalidation is essentially an AND operation with the bitmask of the operation that was performed.

ntamas:
> Adding edges does not invalidate: strong / weak connectivity

Actually, that's not entirely true. Adding edges does not change strong / weak connectivity if we know already that the graph is strongly / weakly connected. However, adding edges may make a disconnected graph strongly / weakly connected, so we need to invalidate these properties if they are currently cached as "false", and keep them if they are currently cached as "true".

> Deleting edges does not invalidate: is_dag, is_forest

Same here, but the opposite way, we need to keep these cached properties if they are cached as "false" but invalidate them if they are cached as "true".

> Adding vertices does not invalidate: has_loop, has_multi, has_reciprocal, is_dag, is_forest

I think this is okay, we can keep these cached properties unconditionally after a vertex addition.

> Deleting vertices does not invalidate: is_dag, is_forest

Deleting a vertex may turn a directed graph with one or more cycles into a DAG if all cycles were broken with the deletion. Similarly, deleting more than vertex from an undirected graph with one or more cycles may turn the undirected graph into a forest if all cycles contained at least one of the the deleted vertices.

ntamas:
Added a commit that implements conditional cache invalidation for graph mutation operations; please _check this commit carefully_ as I may have mixed up things accidentally.

szhorvat:
Regarding the use of bitmasks and the implementation of the cache:

The system is designed in such a way that none of the property cache implementation details are exposed. We can rewrite the internal implementation without breaking either the API or ABI. So we don't even have to anticipate how many properties we may have in the future. I think it's even fine to use `uint32_t`, as we likely won't have as many, and if we do anyway, we can safely make the change.

The one sensitive point is the `igraph_i_property_cache_invalidate_conditionally` function which also uses bitmasks and is "semi-public" at the moment. I said that I would like to use some of the functions from the Mathematica interface, but this is not one of them. I only need `igraph_i_property_cache_set_bool()` so that I can transfer cached values from Mathematica's own graph format. The rest might as well go in `caching.h` instead of `igraph_interface.h`.  We should think carefully if we want to expose these functions. One argument for doing so is that whatever function is implemented _in_ igraph should also be implementable _with_ igraph.

ntamas:
I'm not at all against hiding `igraph_i_property_cache_invalidate_conditionally()` (and in fact, all other invalidation functions) from the general public. Would you need it in the Mathematica interface?

The current implementation of the conditional invalidation function uses bitmasks in the public API just to make it easier to supply a set of properties, but we could also opt for varargs. I think that using bitmasks will allow us to use bit operations in the internal implementation so it's probably going to be faster than using varargs. The current implementation does not use bit operations internally, but I'm about to rewrite the cache data structure to use an `uint64_t` or `uint32_t` to store the "valid" bits and I'll check if it simplifies the code or not.

src/connectivity/components.c
ntamas: Good point about the undirected case, I've fixed that in 3189a7229

As for the second question, I wanted to have a top-level function that deals with the caching and then delegates the grunt work to lower-level functions that then do _not_ need to be concerned with caching.

src/connectivity/components.c
```diff
@@ -384,39 +390,52 @@ static igraph_error_t igraph_is_connected_weak(const igraph_t *graph, igraph_boo
...
+    } else if (mode == IGRAPH_WEAK) {
+        retval = igraph_is_connected_weak(graph, res);
+    } else {   /* mode == IGRAPH_STRONG */
         /* A strongly connected graph has at least as many edges as vertices,
          * except for the singleton graph, which is handled above. */
         if (igraph_ecount(graph) < no_of_nodes) {
             *res = 0;
-            return IGRAPH_SUCCESS;
+        } else {
+            retval = igraph_i_connected_components_strong(graph, NULL, NULL, &no);
```
szhorvat, line 431: Is there any reason you are not using `IGRAPH_CHECK` instead of saving `retval`? Suppose `retval` is an error. Then the value of `*res` will be invalid, but it seems it's still added to the cache.

ntamas: Good point, I have no idea why this was done this way. Just pushed a fix in 676e75ff6

src/graph/caching.c
```diff
@@ -0,0 +1,186 @@
...
+
+#include <assert.h>
+
+/****** Strictly internal functions ******/
+
+/**
+ * \brief Initializes a property cache, ensuring that all values are unknown.
+ */
+igraph_error_t igraph_i_property_cache_init(igraph_i_property_cache_t *cache) {
+    STATIC_ASSERT(IGRAPH_PROP_I_SIZE <= 32);
```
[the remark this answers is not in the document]
ntamas: It's still part of an internal header only (`internal/hacks.h`) and I plan to keep it internal, so there's probably no need to rename it. Are you worried that some other public header that we include from an external package _also_ defines `STATIC_ASSERT`?

src/graph/type_common.c
```diff
@@ -179,3 +175,29 @@ igraph_error_t igraph_edges(const igraph_t *graph, igraph_es_t eids, igraph_vect
...
+ * data structure. This function invalidates the contents of the cache and
+ * forces a recalculation of the cached properties the next time they are
+ * needed.
+ *
+ * </para><para>
+ * You should not need to call this function during normal usage; however, we
+ * might ask you to call this function explicitly if we suspect that you are
+ * running into a bug in igraph's cache handling. A tell-tale sign of an invalid
+ * cache entry is that the result of a cached igraph function (such as
+ * \ref igraph_is_dag() or \ref igraph_is_simple() ) is different before and
```
szhorvat, line 194: Just curious: is the space in `igraph_is_siple() )` needed for the doc tools to work correctly?

src/graph/caching.h
```diff
@@ -0,0 +1,52 @@
...
+
+    /** Bit field that stores which of the properties are cached at the moment */
+    uint32_t known;
+};
+
+igraph_error_t igraph_i_property_cache_init(igraph_i_property_cache_t *cache);
+igraph_error_t igraph_i_property_cache_copy(
+        igraph_i_property_cache_t *cache,
+        const igraph_i_property_cache_t *other_cache);
+void igraph_i_property_cache_destroy(igraph_i_property_cache_t *cache);
```
szhorvat, line 44: As I remember, `init`/`copy`/`destroy` was my design originally, but looking at it again, it seems that they are never used without `calloc` and `free`. Should we have included memory allocation / freeing directly in these functions?

Anyway, refactoring this wouldn't bring any gains, so let's leave as is.

src/graph/type_indexededgelist.c
```diff
@@ -373,6 +415,37 @@ igraph_error_t igraph_add_vertices(igraph_t *graph, igraph_integer_t nv, void *a
...
+     * - IGRAPH_PROP_HAS_RECIPROCAL
+     * - IGRAPH_PROP_IS_DAG (adding a node does not create/destroy cycles)
+     * - IGRAPH_PROP_IS_FOREST (same)
+     *
+     * Adding one or more nodes without any edges incident on them is sure to
+     * make the graph disconnected (weakly or strongly), so we can keep the
+     * connectivity-related properties if they are currently cached as false.
+     * (Actually, even if they weren't cached as false, we could still set them
+     * to false, but we don't have that functionality yet).
+     */
```
szhorvat, line 433: There is a special case when `is_connected` is not `keep_when_false`.

If the graph has no vertices, we consider it disconnected. Adding a single vertex to it will make it connected.

To keep things as simple as possible, we can invalidate the cache unconditionally when the final vertex count is less than 2. There is little gain from caching properties for the 0- or 1-vertex graph.

src/graph/type_indexededgelist.c
```diff
@@ -483,8 +554,30 @@ igraph_error_t igraph_delete_edges(igraph_t *graph, igraph_es_t edges) {
...
+    /* modification successful, clear the cached properties of the graph.
+     *
+     * Deleting one or more edges cannot make a directed acyclic graph cyclic,
+     * or an undirected forest into a cyclic graph, so we keep those flags if
+     * they are cached as true
+     *
+     * Also, if the graph had no loop edges before the deletion, it will have
+     * no loop edges after the deletion either. The same applies to reciprocal
+     * edges or multiple edges as well.
+     */
```
szhorvat, line 569: I think we can also have a `keep_when_false` for `is_strongly_connected` and `is_weakly_connected`.

Is this correct or am I missing something?

src/properties/trees.c
```diff
@@ -254,27 +254,39 @@ static igraph_error_t igraph_i_is_tree_visitor(const igraph_t *graph, igraph_int
...
+    igraph_bool_t is_tree = 0;
     igraph_integer_t iroot = 0;
     igraph_integer_t visited_count;
     igraph_integer_t vcount, ecount;
 
     vcount = igraph_vcount(graph);
     ecount = igraph_ecount(graph);
 
+    /* For undirected graphs, we can return early if we know from the cache
+     * that the graph is weakly connected and is a forest */
```
szhorvat, line 266: Not just for undirected graphs, but also when `mode == IGRAPH_ALL`. So below it should be `(!igraph_is_directed(graph) || mode == IGRAPH_ALL)`. Am I missing something?

szhorvat:
I made another pass though this. It seems that it's very easy to make mistakes here. Let's give this a couple more days before we merge it. I'll take one final look before then. Hopefully I did not make any mistakes this time, please double-check everything.

src/graph/type_indexededgelist.c
ntamas: Aaaargh, these are _exactly_ the special cases because of which I'm scared to add any smart logic to cached properties. I'll fix this.

src/graph/caching.c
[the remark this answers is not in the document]
ntamas: Renamed to `IGRAPH_STATIC_ASSERT` then.

src/graph/type_indexededgelist.c
ntamas: Hmmm, right, I was thinking like removing all the edges of all connected components but one, but then we still have the vertices there so we are okay.

src/properties/trees.c
```diff
@@ -346,15 +358,27 @@ igraph_error_t igraph_is_tree(const igraph_t *graph, igraph_bool_t *res, igraph_
...
     if (root) {
         *root = iroot;
     }
 
+    if (is_tree && !igraph_is_directed(graph)) {
+        /* For undirected graphs, a tree is weakly connected and is a forest, so
+         * we can cache this */
+        igraph_i_property_cache_set_bool(graph, IGRAPH_PROP_IS_FOREST, 1);
+        igraph_i_property_cache_set_bool(graph, IGRAPH_PROP_IS_WEAKLY_CONNECTED, 1);
+    }
```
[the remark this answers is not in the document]
ntamas: I guess you meant `igraph_bool_t directed = igraph_is_directed(graph) && mode != IGRAPH_ALL;`, right? (Or the variable should have been called `undirected` or `treat_as_undirected`, which I actually like better).

src/graph/caching.h
szhorvat: Yes, I think that may be the reason why I did it this way.

src/properties/trees.c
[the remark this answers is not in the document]
ntamas: Anyhow, this is done in ce7925c9f

src/graph/type_indexededgelist.c
szhorvat: Yes, removing vertices is different, and we can't keep any cached connectedness properties there. But removing edges is simpler: it can never make a disconnected graph be connected.

src/connectivity/components.c
```diff
@@ -384,39 +390,51 @@ static igraph_error_t igraph_is_connected_weak(const igraph_t *graph, igraph_boo
...
+            *res = (no == 1);
         }
-
-        retval = igraph_i_connected_components_strong(graph, NULL, NULL, &no);
-        *res = (no == 1);
-        return retval;
     }
 
-    IGRAPH_ERROR("Invalid connectedness mode.", IGRAPH_EINVAL);
+    igraph_i_property_cache_set_bool(graph, prop, *res);
```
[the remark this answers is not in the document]
ntamas: Done in c14eb72a6

szhorvat:
Should we set the cached values in `igraph_empty_attr()`? Right now, it wouldn't do much, as this function is almost certainly follows by `igraph_add_vertices()` and `igraph_add_edges()`, which will invalidate everything. But it would get the groundwork to gradually adding some checks to `add_vertices()` and `add_edges()` at a later stage, such as adding a self-loop check to the latter.

szhorvat:
Do we want to add some tests for the special cases that are most likely to be broken in the future?

E.g., create a null graph, check if it is connected, add a vertex, check again.

Or more generally, create a graph, do some checks (which trigger caching), make a modification to the graph, ensure that the checks still return correct values.

szhorvat:
While on the topic of tests, we should check if some tests need to call `igraph_invalidate_cache()` to ensure that they really run all code they're supposed to be running.

ntamas:
> Should we set the cached values in igraph_empty_attr()? 

It can be done later if we want to, it's not necessarily for merging this PR.

> Do we want to add some tests for the special cases that are most likely to be broken in the future? E.g., create a null graph, check if it is connected, add a vertex, check again.

I don't see it yet which cases are the ones likely to be broken in the future, but we should definitely have some tests that exercise the caching mechanism, even if the code paths seems to be covered by the current unit tests. On the other hand, I don't see yet what's the best way to do it. After call, caching should be a mechanism that does not change the visible behaviour of the library, and if the tests are indeed testing _behaviour_ and not _implementation_, then they should not be concerned with what's being cached and what's not being cached. If we want to test the caching mechanism itself, we must start writing tests that know about the internal implementation (i.e. where are the parts that are being cached and how the cache logic works), and I'm not quite fond of that idea.

How about this? We should come up with a few tests that are organized like this:

* do some complex sequence of operation on a graph that are supposed to trigger caching
* check some property of the graph that we know that it can potentially work with cached values, test whether the returned value is correct
* invalidate all caches
* request the same property again, check whether the value remains the same

The last three points can probably be wrapped up nicely in a utility macro in `test_utilities.c` so we can make these tests reasonably concise. What remains is to come up with a sequence of operations that exercises certain parts of the cache.

ntamas:
I don't have strong opinions about the `igraph_is_mutual()` / `igraph_has_mutual()` function. No one seemed to miss it so far, and it's not in the scope of this PR so I think we can defer it to later.

> I do not like how `igraph_is_mutual()` treats self-loops. It does not fit my personal use cases very well.

Well, you cannot always please everyone ;) What  you want to do is probably achievable with `igraph_is_mutual()` and `igraph_is_loop()`, right? In that case I don't think there's a real need for another boolean argument.

include/igraph_datatype.h
```diff
@@ -30,6 +30,43 @@
...
+    /* Stores whether the graph has at least one multi-edge, taking into account
+     * edge directions in directed graphs. In other words, this property should
+     * be false for a directed graph with edges (a, b) and (b, a), and true
+     * for a directed graph with edges (a, b) and (a, b) again. */
+    IGRAPH_PROP_HAS_MULTI,
+
+    /* Stores whether the graph has at least one reciprocal edge pair. Ignored
+     * in undirected graphs. This property should be true for a directed graph
+     * with edges (a, b) and (b, a), and false for a directed graph with
+     * edges (a, b) and (a, b) again. */
```
szhorvat, line 49:
```suggestion
    /* Stores whether the graph has at least one reciprocal edge pair. Ignored
     * in undirected graphs. This property should be true for a directed graph
     * with edges (a, b) and (b, a), and false for a directed graph with
     * edges (a, b) and (a, b) again. Self-loops (a, a) are not considered
     * reciprocal. */
```

This is so as to keep this useful for determining whether a directed graph would have multi-edges when ignoring edge directions.

ntamas:
I've added unit tests and I think this PR is ready to be merged now. There are a few remaining ideas like pre-populating the cache in certain cases after constructing a graph, or the `igraph_is_mutual()` and `igraph_has_mutual()` functions, but these are unrelated to this PR and are best discussed in separate issues. Any objections to merging this?

(Surely, the unit tests can be improved further if we want to test more things, but this can also be done later).

PR #2102, opened 2022-07-03
Add local_scan_neighborhood_ecount tests

GroteGnoom:
part of #1592

Right now I get inconsistent results. When I use a neighborhood of vertices {0,1,2} while looking at vertex 3, I get the count of edges in the subgraph induced by {0,1,2,3}. But when I look at an empty neighborhood at vertex 1, the loop at vertex 1 isn't counted. 

Also, is this used? The documentation is not too elaborate, but the R documentation for local scan (https://igraph.org/r/doc/local_scan.html) says about the neighborhoods parameter:

> In theory this could be useful if the same graph.us graph is used for multiple graph.them arguments. Then the neighborhoods can be calculated on graph.us and used with multiple graphs. In practice, this is currently slower than simply using graph.them multiple times.

### DIFF
--- a/src/misc/scan.c
+++ b/src/misc/scan.c
@@ -795,28 +795,26 @@ igraph_error_t igraph_local_scan_k_ecount_them(const igraph_t *us, const igraph_
 
     return IGRAPH_SUCCESS;
 }
-
 /**
- * \function igraph_local_scan_neighborhood_ecount
- * Local scan-statistics with pre-calculated neighborhoods
+ * \function igraph_local_scan_subset_ecount
+ * Local scan-statistics of subgraphs induced by subsets of vertices 
  *
  * Count the number of edges, or sum the edge weights in
- * neighborhoods given as a parameter.
+ * induced subgraphs given as a parameter.
  *
  * \param graph The graph to perform the counting/summing in.
  * \param res Initialized vector, the result is stored here.
  * \param weights Weight vector for weighted graphs, null pointer for
  *        unweighted graphs.
- * \param neighborhoods List of <code>igraph_vector_int_t</code>
- *        objects, the neighborhoods, one for each vertex in the
- *        graph.
+ * \param subsets List of <code>igraph_vector_int_t</code>
+ *        objects, the vertex subsets.
  * \return Error code.
  */
 
-igraph_error_t igraph_local_scan_neighborhood_ecount(const igraph_t *graph,
+igraph_error_t igraph_local_scan_subset_ecount(const igraph_t *graph,
         igraph_vector_t *res,
         const igraph_vector_t *weights,
-        const igraph_vector_int_list_t *neighborhoods) {
+        const igraph_vector_int_list_t *subsets) {
 
     igraph_integer_t node, no_of_nodes = igraph_vcount(graph);
     igraph_inclist_t incs;
@@ -825,10 +823,6 @@ igraph_error_t igraph_local_scan_neighborhood_ecount(const igraph_t *graph,
 
     if (weights && igraph_vector_size(weights) != igraph_ecount(graph)) {
         IGRAPH_ERROR("Invalid weight vector length in local scan", IGRAPH_EINVAL);
-    }
-    if (igraph_vector_int_list_size(neighborhoods) != no_of_nodes) {
-        IGRAPH_ERROR("Invalid neighborhood list length in local scan",
-                     IGRAPH_EINVAL);
     }
 
     IGRAPH_VECTOR_INT_INIT_FINALLY(&marked, no_of_nodes);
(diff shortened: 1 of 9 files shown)


ntamas:
> Right now I get inconsistent results. When I use a neighborhood of vertices {0,1,2} while looking at vertex 3, I get the count of edges in the subgraph induced by {0,1,2,3}. But when I look at an empty neighborhood at vertex 1, the loop at vertex 1 isn't counted.

This is because the function marks the vertices in the neighborhood vector _plus_ the seed vertex first, but then iterates over the incidence list of the vertices in the neighborhood vector _only_.  (Note the `VECTOR(marked)[node] = node + 1;` line, which marks `node` itself as being in the neighborhood, but then later on the iteration is only over the members of `igraph_vector_int_list_get_ptr(neighborhoods, node)`). I don't know whether it's intentional or not -- I'm not familiar with the whole terminology that the entire "local scan" section of the codebase uses. It may well be the case that this is intentional and it's how it is supposed to work.

I suspect that the whole concept is related to [this paper](https://wires.onlinelibrary.wiley.com/doi/pdf/10.1002/wics.1217), but I cannot get through the paywall from home.

szhorvat:
> I suspect that the whole concept is related to [this paper](https://wires.onlinelibrary.wiley.com/doi/pdf/10.1002/wics.1217), but I cannot get through the paywall from home.

It says:

> This article is a U.S. Government work, and as such, is in the public domain in the United States of America.

But non-Americans still have to pay, I guess.

I'm emailing the PDF to both of you.

szhorvat:
Could you please expand the C docs within this PR?

GroteGnoom:
I doubt this function should exist. Its purpose seems to be to speed things up in certain cases, but the R docs say that doesn't work. 

One other use of it could be to be more versatile, and instead of using actual neighborhoods around vertices, using arbitrary subgraphs for scan statistics. But then we shouldn't use `neighborhoods` around vertices but just some vector of `subgraphs`. This is not mentioned in the article, but just some idea I could think of.

GroteGnoom:
I also wonder whether our generalization to weighted graphs makes sense. I think for local scans you should take the weights into account when determining the neighborhood.

szhorvat:
Note: https://github.com/igraph/igraph/issues/782

szhorvat:
- [ ] Verify that this function simply takes the subgraph induced by each vertex set ("neighbourhood") and countes edges / sums edge weights within it.
- [ ] Rename the function from `neighborhood` to something more sensible, e.g. `subgraph` or `subset`.
- [ ] Relax restrictions that no. of subsets must match no. of vertices, and each vertex must be contained within a corresponding subset.
- [ ] Update documentation to reflect the above.

GroteGnoom:
Yes!

Get Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

src/misc/scan.c
```diff
@@ -795,60 +795,58 @@ igraph_error_t igraph_local_scan_k_ecount_them(const igraph_t *us, const igraph_
 
     return IGRAPH_SUCCESS;
 }
-
 /**
- * \function igraph_local_scan_neighborhood_ecount
- * Local scan-statistics with pre-calculated neighborhoods
+ * \function igraph_local_scan_subset_ecount
+ * Local scan-statistics of subgraphs induced by subsets of vertices
```
szhorvat, line 800:
```suggestion
 * \brief Local scan-statistics of subgraphs induced by subsets of vertices.
```

src/misc/scan.c
```diff
@@ -795,60 +795,58 @@ igraph_error_t igraph_local_scan_k_ecount_them(const igraph_t *us, const igraph_
...
  *
  * \param graph The graph to perform the counting/summing in.
  * \param res Initialized vector, the result is stored here.
  * \param weights Weight vector for weighted graphs, null pointer for
  *        unweighted graphs.
- * \param neighborhoods List of <code>igraph_vector_int_t</code>
- *        objects, the neighborhoods, one for each vertex in the
- *        graph.
+ * \param subsets List of <code>igraph_vector_int_t</code>
+ *        objects, the vertex subsets.
```
szhorvat, line 810:
```suggestion
 * \param subsets List of \type igraph_vector_int_t
 *        objects, the vertex subsets.
```

src/misc/scan.c
```diff
@@ -875,3 +873,37 @@ igraph_error_t igraph_local_scan_neighborhood_ecount(const igraph_t *graph,
...
+ *
+ * \deprecated-by igraph_local_scan_subset_ecount 0.10.0
+ *
+ * \param graph The graph to perform the counting/summing in.
+ * \param res Initialized vector, the result is stored here.
+ * \param weights Weight vector for weighted graphs, null pointer for
+ *        unweighted graphs.
+ * \param neighborhoods List of <code>igraph_vector_int_t</code>
+ *        objects, the neighborhoods, one for each vertex in the
+ *        graph.
```
szhorvat, line 892:
```suggestion
 * \param neighborhoods List of \type igraph_vector_int_t
 *        objects, the neighborhoods, one for each vertex in the
 *        graph.
```

src/misc/scan.c
```diff
@@ -795,60 +795,58 @@ igraph_error_t igraph_local_scan_k_ecount_them(const igraph_t *us, const igraph_
...
+        IGRAPH_ERROR("Invalid weight vector length in local scan.", IGRAPH_EINVAL);
     }
 
     IGRAPH_VECTOR_INT_INIT_FINALLY(&marked, no_of_nodes);
-    IGRAPH_CHECK(igraph_inclist_init(graph, &incs, IGRAPH_OUT, IGRAPH_LOOPS_ONCE));
+    if (directed) {
+        IGRAPH_CHECK(igraph_inclist_init(graph, &incs, IGRAPH_OUT, IGRAPH_LOOPS_ONCE));
+    } else {
+        IGRAPH_CHECK(igraph_inclist_init(graph, &incs, IGRAPH_OUT, IGRAPH_LOOPS_TWICE));
+    }
```
szhorvat, line 834: It is fine to use `IGRAPH_LOOPS_TWICE` with directed graphs. It won't duplicate loops (except when using `mode=IGRAPH_ALL`). So this conditional is not necessary.

szhorvat (commented):
This is not a full review, just a quick look.

Regarding `loops=ONCE` or `loops=TWICE`: we need to make a decision whether to count loops once or twice in local scan functions, and ensure that all these functions are consistent with each other. If you have time, can you check what the other functions do?

I seem to recall that at some point @ntamas changed some of these to `ONCE`, but I didn't pay attention at that time. Any comments @ntamas ? Perhaps I am misremembering.

The description of these functions states that they count edges. Thus, it makes sense to me that in the final result, each edge should be counted once, including self-loops.

We multiply their counts by two in other contexts, such as when computing degrees, precisely so that the edge count would be correct when using the simple formula $2|E| = \sum_i d_i$. We count them twice in the adjacency matrix so that the sum of all adjacency matrix elements would be $\sum_{i,j} A_{ij} = 2|E|$. Since what we are dealing in the local scan statistics is a plain edge count, self-loops should be counted once.

Note that I am talking about the final result. I did not look at the code, and how the construction of the `inclist` and its `loops` option influences the final result.

szhorvat:
Sorry, I just looked at the code more carefully, and I see that in order to obtain the correct _final result_, counting each self-loop once, we need `LOOPS_TWICE` in `inclist_init`. This is because every edge is counted twice, once for each endpoint. Then the final count is divided by two.

So, the current code is correct.

szhorvat:
Final comment is that right now the function looks a bit unusual: why this it take a list of subsets instead of a single subset (other than needing to stay compatible with the `neighbors` version)? Actually, it makes sense to do this as this function creates an inclist. It will improve performance to create that inclist only once even when handling many subsets. So this is good in its current form.

src/misc/scan.c
GroteGnoom: fix subset_ecount brief typos

szhorvat (approved):
Once this is merged (feel free to go ahead @GroteGnoom), the `ONCE -> TWICE` fix should be backported to `master`, and added to the 0.9 section of the changelog.

szhorvat:
> should be backported to `master`, and added to the 0.9 section of the changelog.

Done.

PR #2103, opened 2022-07-03
Add constraint unit tests

GroteGnoom:
Part of #1592 

I'm not sure `constraint` handles weights correctly. There's a call to igraph_strength, but it doesn't pass the weights.

Then, when taking the `/* squared sum of the contributions */`, edge weights are also not taken into account, so a line graph and a graph where a line is carved out with 1's and 0's will still give different results. I'm not sure if that's supposed to happen.

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -285,6 +285,7 @@ add_legacy_tests(
   igraph_betweenness
   igraph_betweenness_subset
   igraph_closeness
+  igraph_constraint
   igraph_convergence_degree
   igraph_count_multiple
   igraph_density
--- a/tests/unit/igraph_constraint.c
+++ b/tests/unit/igraph_constraint.c
@@ -1,0 +1,84 @@
+/* IGraph library.
+   Copyright (C) 2021  The igraph development team <igraph@igraph.org>
+
+   This program is free software; you can redistribute it and/or modify
+   it under the terms of the GNU General Public License as published by
+   the Free Software Foundation; either version 2 of the License, or
+   (at your option) any later version.
+
+   This program is distributed in the hope that it will be useful,
+   but WITHOUT ANY WARRANTY; without even the implied warranty of
+   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+   GNU General Public License for more details.
+
+   You should have received a copy of the GNU General Public License
+   along with this program.  If not, see <https://www.gnu.org/licenses/>.
+*/
+
+#include <igraph.h>
+#include "test_utilities.h"
+
+void call_and_print(igraph_t *graph, igraph_vs_t vids, igraph_vector_t *weights) {
+    igraph_vector_t result;
+    igraph_vector_init(&result, 0);
+    IGRAPH_ASSERT(igraph_constraint(graph, &result, vids, weights) == IGRAPH_SUCCESS);
+    print_vector(&result);
+    igraph_vector_destroy(&result);
+    printf("\n");
+}
+
+
+int main() {
+    igraph_t g_0, g_1, g_lmu, g_lm, g_4_full, g_4_line, g_hole;
+    igraph_vector_t weights_full, weights_line, result;
+
+    igraph_vector_init_real(&weights_line, 6, 1., 0., 0., 1., 0., 1.);
+    igraph_vector_init_real(&weights_full, 6, 1., 1., 1., 1., 1., 1.);
+    igraph_vector_init(&result, 0);
+    igraph_small(&g_0, 0, 0, -1);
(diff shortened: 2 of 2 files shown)


szhorvat:
Someone should look at the original paper and decide if the implementation is wrong. We would also need to check if the concept makes sense for:

 - directed graphs
 - graphs with multi-edges
 - graphs with self-loops

If someone offered to implement such a concept today, I do not think we should accept it unless it came with a detailed writeup addressing all these questions. I had a quick look at the original paper. It's 52 pages, talking mostly about economics. Extracting a mathematically rigorous definition is going to be a time-consuming and difficult task. I did not notice any formulas, so the idea is probably described only in words, and I'm not sure where the description is found among those 52 pages ...

@vtraag As you're closer to social networks, do you happen to be familiar with this concept already?

Finally, shall we rename this to `igraph_constraint_score` and deprecate the current name? "Constraint" is extremely general, and the concept that this function implements seems to be a very niche one.

GroteGnoom:
On page 15/362 there is a footnote with formula's, apparently coming from a book chapter, starting with "The network constraint index begins with the".

ntamas:
> I'm not sure constraint handles weights correctly. There's a call to igraph_strength, but it doesn't pass the weights.

If you `git blame` the file, you can see that this line was changed in a commit that migrated lots of `igraph_vector_t` instances to `igraph_vector_int_t`, and that the original call was this:

```
    if (weights == 0) {
        IGRAPH_CHECK(igraph_degree(graph, &degree, igraph_vss_all(),
                                   IGRAPH_ALL, IGRAPH_NO_LOOPS));
    } else {
        for (a = 0; a < no_of_edges; a++) {
            igraph_edge(graph, a, &from, &to);
            if (from != to) {
                VECTOR(degree)[from] += VECTOR(*weights)[a];
                VECTOR(degree)[to  ] += VECTOR(*weights)[a];
            }
        }
    }
```

so I'm pretty sure that the original intention was to use the weights in the `igraph_strength()` call.

ntamas:
> shall we rename this to `igraph_constraint_score` and deprecate the current name?

I think it's called "Burt's constraint" in the literature, not "Burt's constraint score", even though the latter seems more natural to me. If we want to rename and avoid the genericity of `igraph_constraint()`, we could call it `igraph_burts_constraint()` instead, but only if it causes confusion for the end users. As long as we don't have any other "constraint" measure in igraph, I don't think it hurts anyone.

szhorvat:
> If we want to rename and avoid the genericity of `igraph_constraint()`, we could call it `igraph_burts_constraint()` instead, but only if it causes confusion for the end users. As long as we don't have any other "constraint" measure in igraph, I don't think it hurts anyone.

In igraph 1.0 we want to have a stable API. We won't be able to remove functions until 2.0. It makes sense to make this change now and not wait until it becomes an inconvenience. `burts_constrain` is also a much more informative name, and makes it clear what this might be even if you're not a sociologist.

ntamas:
In that case the new name should be added but the old name should be kept until it goes through a proper deprecation cycle.

szhorvat:
> In that case the new name should be added but the old name should be kept until it goes through a proper deprecation cycle.

Obviously, same as always :-)

GroteGnoom:
I think the remaining issue is similar to the one for local scans: https://github.com/igraph/igraph/pull/2102#issuecomment-1186116515. Should you take weights into account when looking in the neighborhood, and for example not treat a vertex connected with edge weight 0 as a neighbor?

This then also depends on if you see weights as distances or strengths.

szhorvat:
I would avoid making non-trivial generalizations of the published network metric. If we do that, we're in research territory. While I'm not against that, it makes this a rather large project ...

GroteGnoom:
> I would avoid making non-trivial generalizations of the published network metric. If we do that, we're in research territory. While I'm not against that in general, it makes this a rather large project ...

So we'll leave the current implementation as a trivial weighted generalization?

szhorvat:
I'm sorry, I didn't look carefully at this, and didn't realize that there was already a generalization.

I'll get back to it later, but for the time being leaving as-is sounds like the safest option ...

GroteGnoom:
I have read the article more carefully, and I don't think we have a generalization, but we're just doing what the article says. It's more a question of what a 'contact' is. Right now if someone is connected with weight 0 it's still counted as a contact, which is fine, and seems to be what the article wants.

Multiple edges between two contacts seem to work correctly.

With loops I have no idea. They seem to be ignored:
```
            if (i != j) {     /* excluding loops */
```
, which is also what I notice in the tests. Other options would be to add a warning or give an error, but maybe it's just fine to talk to yourself, and it shouldn't influence the interpretation of connections with others.

ntamas:
Thanks!

PR #2111, opened 2022-07-09
Fast way to reverse all edges of a directed graph

### LINKED ISSUE #1477: Wishlist: Reverse the direction of edges while preserving attributes
There should be a function that takes a directed graph and reverses the direction of each edge while preserving all graph attributes.

szhorvat:
@ntamas Have a look, and if good, we can close #1477.

### DIFF
--- a/src/graph/internal.h
+++ b/src/graph/internal.h
@@ -16,8 +16,8 @@
    along with this program.  If not, see <https://www.gnu.org/licenses/>.
 */
 
-#ifndef IGRAPH_NEIGHBORS_H
-#define IGRAPH_NEIGHBORS_H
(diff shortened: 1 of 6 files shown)


ntamas:
Okay, so, this basically means that we are committing ourselves that all future graph data types that we might implement will provide an `igraph_i_reverse()` method as a special case for reversing all edges, right?

Can we call this function `igraph_i_reverse_all_edges()` instead to make it clear what it's for?

ntamas:
If we rename `neighbors.h`, let's also rename the `#include` guard at the top from `IGRAPH_NEIGHBORS_H` to follow up the new name of the file. Otherwise I'm fine with the PR.

PR #2113, opened 2022-07-09
fix: permute_vertices() now verifies that the permutation vector is valid

szhorvat:
This ensures that the permutation vector passed to `permute_vertices()` is valid. Recently I added a check for the vector not containing invalid indices. Now we also check that it does not contain duplicates.

This check is necessary because `permute_vertices()` is accessible from high-level interfaces, and passing in duplicates in the permutation vector will produce garbage output. It's unclear if a crash is possible, or whether it corrupts internal state, but it does provide a loophole to introduce duplicate vertex.

The check is done through a newly added helper function that inverts permutations. This should be useful in the future as we deal with the inconsistent representation of permutations, which is being discussed [here](https://github.com/igraph/igraph/issues/1930#issuecomment-1176079877). Using the terminology from there, `permute_vertices` takes an R2 type input, but it uses `igraph_i_attribute_permute_vertices()` which requires an R1 type input. Converting between R2 and R1 is equivalent to inversion, and the function already had to do this before this refactoring. This PR simply couples validating permutations with inverting them, as doing them together is more efficient: both would use the same algorithm anyway.

Is there a performance cost? Yes, there is. Previously, inversion was only done when an attribute handler was set up. Now it is done even if attributes are not used.  One possible solution would be to introduce an internal, unchecked version of `permute_vertices()`. However, I would leave that for a later time (and for `develop`).

On the topic of performance, the check could be done slightly more efficiently if we allocated `inverse` directly within `igraph_i_invert_permutation()`, as we could exploit that it is already filled with zeros.

I have not yet added tests. I will do that only on `develop` as the tools to easily print graph attributes (and check that they're properly permuted) are only available there.

### DIFF
--- a/src/operators/permute.c
+++ b/src/operators/permute.c
@@ -28,6 +28,38 @@
 #include "graph/attributes.h"
 
 /**
+ * \brief Inverts a permutation.
+ *
+ * Produces the inverse of \p permutation into \p inverse and at the same time it checks
+ * that the permutation vector is valid, i.e. all indices are within range and there are
+ * no duplicate entries.
+ *
+ * \param permutation A permutation vector containing 0-based integer indices.
+ * \param inverse An initialized vector. The inverse of \p permutation will be stored here.
+ * \return Error code.
+ */
+static int igraph_i_invert_permutation(const igraph_vector_t *permutation, igraph_vector_t *inverse) {
+    const long int n = igraph_vector_size(permutation);
+
+    IGRAPH_CHECK(igraph_vector_resize(inverse, n));
+    igraph_vector_fill(inverse, -1);
+
+    for (long int i=0; i < n; i++) {
+        long int j = VECTOR(*permutation)[i];
+        if (j < 0 || j >= n) {
+            IGRAPH_ERROR("Invalid index in permutation vector.", IGRAPH_EINVAL);
+        }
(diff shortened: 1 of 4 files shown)


src/operators/permute.c
```diff
@@ -27,6 +27,38 @@
...
+ *
+ * Produces the inverse of \p permutation into \p inverse and at the same time it checks
+ * that the permutation vector is valid, i.e. all indices are within range and there are
+ * no duplicate entries.
+ *
+ * \param permutation A permutation vector containing 0-based integer indices.
+ * \param inverse An initialized vector. The inverse of \p permutation will be stored here.
+ * \return Error code.
+ */
+static int igraph_i_invert_permutation(const igraph_vector_t *permutation, igraph_vector_t *inverse) {
```
ntamas, line 41: I wonder whether we should expose this function in public. It seems useful enough for end-users who might have an R1-style permutation for an R2-style igraph function or vice versa.

ntamas (approved):

src/operators/permute.c
szhorvat: Yes, that's the intention, but not in `master`.

ntamas: `igraph_invert_permutation()` is now exposed in `develop`.

PR #2114, opened 2022-07-09
Make functions overflow-safe 

GroteGnoom:
Fixes part of #51
#2097 handles random generators

I just did a grep for "no_of_edges \* 2", and even_tarjan_reduction was the first.

### DIFF
--- a/src/flow/st-cuts.c
+++ b/src/flow/st-cuts.c
@@ -37,6 +37,7 @@
 #include "core/marked_queue.h"
 #include "flow/flow_internal.h"
 #include "graph/attributes.h"
+#include "math/safe_intop.h"
 
 typedef igraph_error_t igraph_provan_shier_pivot_t(const igraph_t *graph,
                                                    const igraph_marked_queue_int_t *S,
@@ -86,12 +87,21 @@ igraph_error_t igraph_even_tarjan_reduction(const igraph_t *graph, igraph_t *gra
     igraph_integer_t no_of_nodes = igraph_vcount(graph);
     igraph_integer_t no_of_edges = igraph_ecount(graph);
 
-    igraph_integer_t new_no_of_nodes = no_of_nodes * 2;
-    igraph_integer_t new_no_of_edges = no_of_nodes + no_of_edges * 2;
+    igraph_integer_t new_no_of_nodes;
+    igraph_integer_t new_no_of_edges;
 
     igraph_vector_int_t edges;
     igraph_integer_t edgeptr = 0, capptr = 0;
     igraph_integer_t i;
+
+    IGRAPH_SAFE_MULT(no_of_nodes, 2, &new_no_of_nodes);
+    IGRAPH_SAFE_MULT(no_of_edges, 2, &new_no_of_edges);
+    IGRAPH_SAFE_ADD(new_no_of_edges, no_of_nodes, &new_no_of_edges);
+
+    /* To ensure the size of the edges vector will not overflow. */
+    if (new_no_of_edges > IGRAPH_ECOUNT_MAX) {
+        IGRAPH_ERROR("Overflow in number of edges.", IGRAPH_EOVERFLOW);
+    }
 
     IGRAPH_VECTOR_INT_INIT_FINALLY(&edges, new_no_of_edges * 2);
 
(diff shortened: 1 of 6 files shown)


szhorvat:
Note that when `no_of_edges` is initialized with an `igraph_ecount(graph)`, it is always safe to multiply it by 2. Almost all `no_of_edges * 2` you will find in functions (except in constructors) is already safe.

Instead of grepping for `no_of_edges * 2`, I would look for functions which create a very large work area, whose size has a real risk of overflowing.  Problem functions are those that require storage that grows faster than linearly with the graph size. One example that I fixed already was the ILP-based feedback arc set implementation.

src/flow/st-cuts.c
```diff
@@ -86,13 +87,22 @@ igraph_error_t igraph_even_tarjan_reduction(const igraph_t *graph, igraph_t *gra
...
-    igraph_integer_t new_no_of_edges = no_of_nodes + no_of_edges * 2;
+    igraph_integer_t new_no_of_nodes;
+    igraph_integer_t new_no_of_edges;
 
     igraph_vector_int_t edges;
     igraph_integer_t edgeptr = 0, capptr = 0;
     igraph_integer_t i;
 
+    IGRAPH_SAFE_MULT(no_of_nodes, 2, &new_no_of_nodes);
+    IGRAPH_SAFE_MULT(no_of_edges, 2, &new_no_of_edges);
```
szhorvat, line 98: So this check is not necessary, the others you added are fine.

GroteGnoom:
> look for functions which create a very large work area, whose size has a real risk of overflowing. Problem functions are those that require storage that grows faster than linearly with the graph size.

Alright, but that's very hard for me to estimate 🙂

GroteGnoom:
Maybe I should just hand all the overflow checks over to you, @szhorvat?

szhorvat:
It'd be great if you could continue, but since you have several other draft PRs already, perhaps it's a good idea to finish those first. In the meantime, I can look at this a bit more carefully.

GroteGnoom:
Just to check if I got this correctly:
in `igraph_community_fastgreedy()`, now at line 772 of `src/community/fast_modularity.c`, there's this line:

>  dq[j] = 2 * (1.0 / (no_of_edges * 2.0) - VECTOR(a)[from] * VECTOR(a)[to] / (4.0 * no_of_edges * no_of_edges));

even though there's a `no_of_edges * no_of_edges` in there, it doesn't overflow, right? Because the first no_of_edges will first be converted to a double, and then the second too.

szhorvat:
Yes, that's right.

szhorvat:
Can you explain why you are changing `igraph_adjacency()` from allocating an appropriate-size vector an filling it up to allocating a zero-size one and pushing to it? This certainly has a performance impact. What am I missing?

GroteGnoom:
It's not igraph_adjacency(), but `igraph_sparse_adjacency()`, right? igraph_adjacency already starts at 0 edges, and then does push_backs. the sparsemat version didn't, because I thought it would use less allocations:
```
 igraph_integer_t no_of_edges = igraph_sparsemat_count_nonzero(adjmatrix);
    if (no_of_edges) {
        no_of_edges *= igraph_sparsemat_max(adjmatrix);
    }
```
But of course this is an upper bound and can be a huge overestimation. I was looking at making this overflow-safe, but that also meant that we would give errors when nothing was wrong. So then I decided to let the sparse version do the same as the non-sparse version and use push_back.

Maybe I should leave the count_nonzero in as a starting point?

szhorvat:
> But of course this is an upper bound and can be a huge overestimation.

Thanks for the clarification!

szhorvat:
> Maybe I should leave the count_nonzero in as a starting point? That makes sense as a minimum amount of edges.

Yes, I would do that. Reserve `count_nonzero` in the directed case and half as many in the undirected case. This will cover the most common use cases.

szhorvat:
I went through these changes and it looks all good to me, except that it'd be good to reserve storage in `igraph_sparse_adjacency()` as I mentioned above.

ntamas (approved):
I'll commit that small change about reserving edge storage in `igraph_sparse_adjacency()` (although note that igraph_sparsemat_count_nonzero()` has an unspecified runtime cost because it needs to filter duplicate edges) and then I'll merge.

PR #2115, opened 2022-07-09
fix: sequence edge/vertex selectors now consistently include "to"

### LINKED ISSUE #2110: eit_create() and vit_create() must check that IDs are valid
`igraph_eit_create()` and `igraph_vit_create()` create an iterator (ties to a specific graph) from a selector (which is an abstract object that can contain any edge/vertex IDs).  This issue is a TODO item to make sure that the iterator creation functions verify that all edge/vertex IDs coming from the selectors are valid for the given graph. Note that there are many code paths, only some of which can receive arbitrary IDs. All these should be verified.

szhorvat:
This fixes #2110.

 - Ranges are checked when creating iterators for sequence selectors
 - The meaning of `from` / `to` fields in the `vs_t` and `es_t` structs is clarified and made consistent. Previously it was different between `es_t` and `vs_t`
 - It is now made clear and respected that the `to` index is included when creating sequence edge/vertex selectors

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@
  - `igraph_permute_vertices()` now checks for out-of-range indices in the permutation vector.
  - `igraph_create()` now checks for non-finite vertex indices in the edges vector.
  - `igraph_eigenvector_centrality()` would return incorrect scores when some weights were negative.
+ - `igraph_es_seq()` and `igraph_ess_seq()` did not include the `to` vertex in the sequence.
+ - `igraph_eit_create()` and `igraph_vit_create()` now checks that all edge/vertex indices are in range when creating iterators from sequence-type selectors.
 
 ### Other
 
--- a/include/igraph_iterators.h
+++ b/include/igraph_iterators.h
@@ -52,9 +52,9 @@ typedef struct igraph_vs_t {
             igraph_neimode_t mode;
         } adj;                  /* adjacent vertices  */
         struct {
-            igraph_integer_t from;
-            igraph_integer_t to;
-        } seq;                              /* sequence of vertices from:to */
+            igraph_integer_t from;          /* first index */
+            igraph_integer_t to;            /* last index */
+        } seq;                  /* sequence of vertices from:to */
     } data;
 } igraph_vs_t;
 
@@ -108,8 +108,8 @@ IGRAPH_EXPORT int igraph_vs_type(const igraph_vs_t *vs);
 typedef struct igraph_vit_t {
     int type;
     long int pos;
-    long int start;
-    long int end;
+    long int start; /* first index */
+    long int end;   /* one past last index */
     const igraph_vector_t *vec;
 } igraph_vit_t;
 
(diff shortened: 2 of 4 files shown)


GroteGnoom:
I don't like the inclusive to's, just as I don't like the other functions with inclusive endpoints
```
480:    IGRAPH_CHECK(igraph_vector_int_init_seq(&label, 0, no_of_nodes - 1));

src/paths/sparsifier.c
214:    IGRAPH_CHECK(igraph_vector_int_init_seq(&clustering, 0, no_of_nodes - 1));

src/misc/graphicality.c
631:    IGRAPH_CHECK(igraph_vector_int_init_seq(&index_array, 0, vcount - 1));

src/linalg/arpack.c
686:    IGRAPH_CHECK(igraph_vector_init_seq(&order, 0, nconv - 1));
779:    IGRAPH_CHECK(igraph_vector_init_seq(&order, 0, nconv - 1));

[etc., etc., etc.]
```
all those `-1` should not be there. If we keep it like this we don't have to inconvenience our users with another API change, but it will forever not be what you'd expect of such a function. Exclusive endpoints is what you expect from C and C++, so you don't have to look up the docs anymore if we consistently do that too.

szhorvat:
I agree that exclusive endpoints are a better design, and clearly what a C++ programmer would expect, as that's what the C++ standard library uses. Not sure sure the "typical C programmer" would expect the same, but it's true that in `for` loops one usually uses `< n` instead of `<= last`.

Transitioning to exclusive endpoints would be a big breaking change though ... I don't think now (i.e. 0.10) is the best time to do it.

The only thing I insist on is consistency across all functions. 

Note that this PR is for `master` where the behaviour of `igraph_es_seq()` was clearly a bug: it's the only function that excluded the endpoint. Not even `igraph_vs_seq()` did. So I think making this fix for 0.9.10 is warranted.

GroteGnoom:
> Transitioning to exclusive endpoints would be a big breaking change though ... I don't think now (i.e. 0.10) is the best time to do it.

I thought we specifically wanted to do big breaking API changes in 0.10?

I do agree this is a bug.

ntamas:
@GroteGnoom has a point here and I don't like the inclusive endpoints too much either in the C core. We shouldn't be designing the lower-level C API around specifics of a higher-level interface, and I can clearly see the R influence here.

But -- if we want to fix this for 0.9.10, we clearly cannot change everything to exclusive endpoints there as that would be a breaking change, and since `vs_seq` already uses inclusive endpoints in 0.9.9, the PR is okay and good to go as is.

We can revisit the question of inclusive vs exclusive endpoints in 0.10 or 0.11.

ntamas:
My proposal would be to keep and deprecate `vs_seq` and `es_seq` in 0.10 and replace them with `vs_range` and `es_range` with C-style semantics (start point inclusive, end point exclusive). Then we remove `vs_seq` and `es_seq` in 0.11. We could do the same for `vector_init_seq`.

ntamas (approved):

szhorvat:
Sounds good to me, can you open an issue for that so that it doesn't slip off the radar? In the meantime I'll port the fix to `develop`.

PR #2116, opened 2022-07-10
Control whether self-loops are considered mutual

szhorvat:
Currently, `igraph_is_mutual()` considers all directed self-loops to be mutual with the reasoning that edge (A, B) is mutual if edge (B, A) is present as well. I think this is a bit naive and does not fit practical use-cases well.

Let's think about conversion between directed and undirected graphs. A "mutual edge _pair_" is the directed equivalent of an undirected edge. We went to great lengths to ensure that even undirected self-loops are interpreted as the equivalent of not one, but _two_ directed loops. We called this "counting loops twice" in undirected graphs.  This is inconsistent with counting a single directed self-loop as "mutual".

Another way to think about it is that a directed graph with no mutual edges is one that can be obtained from a multi-edge-free undirected graph by assigning a direction to each edge. Again, this would mean that a single directed self-loop does not fit the idea of a "mutual edge" well.

Finally, a practically useful application for checking for mutual edges is to verify if converting a multi-edge-free directed graph to undirected (by ignoring edge directions) would keep the graph multi-edge-free. Again, this does not fit with interpreting directed self-loops as mutual.

I believe this makes a good case that _not_ treating self-loops are "mutual" is more useful than the current behaviour. At least it's application-dependent which behaviour one may want.

This PR adds an option to control how directed self-loops are treated. In high-level interfaces the default will be as before. In C, having this option forces users to think about the special case of loops, and avoids surprises due to misunderstandings (we don't want surprised in C, right?). So it's a win no every front. It has the added benefit that the `HAS_RECIPROCAL` property can be hooked up to this function in the caching PR. 

----

**Update:** A `has_mutual` function is now also added. Since we have a corresponding cached property, it makes sense to have a function to hook it up to. With `loops=false` this function is now usable to check whether ignoring edge directions in a directed graph effectively creates undirected multi-edges.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -695,6 +695,8 @@ Some of the highlights are:
 
  - `igraph_matrix_minmax()`, `igraph_matrix_which_minmax()`, `igraph_matrix_which_min()` and `igraph_matrix_which_max()` no longer return an error code. The return type is now `void`. These functions never fail.
 
+ - `igraph_is_mutual()` has an additional parameter which controls whether directed self-loops are considered mutual.
+
 ### Added
 
  - A new integer type, `igraph_uint_t` has been added. This is the unsigned pair of `igraph_integer_t` and they are always consistent in size.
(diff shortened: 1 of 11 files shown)


ntamas:
Changed the argument ordering in `igraph_has_mutual()`, but otherwise I think it's fine.
I've also added caching for the case when `loops` is `0`.

PR #2117, opened 2022-07-10
add hsbm tests

GroteGnoom:
part of #1592

### DIFF
--- a/tests/unit/igraph_hsbm_game.c
+++ b/tests/unit/igraph_hsbm_game.c
@@ -21,7 +21,7 @@
 
 void call_and_print(igraph_integer_t n, igraph_integer_t m, igraph_vector_t *rho, igraph_matrix_t *pref_matrix, igraph_real_t p) {
     igraph_t result;
-    IGRAPH_ASSERT(igraph_hsbm_game(&result, n, m, rho, pref_matrix, p) == IGRAPH_SUCCESS);
+    igraph_hsbm_game(&result, n, m, rho, pref_matrix, p);
     print_graph_canon(&result);
     printf("\n");
     igraph_destroy(&result);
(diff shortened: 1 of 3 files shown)


tests/unit/igraph_hsbm_game.c
```diff
@@ -0,0 +1,74 @@
+/*
+   IGraph library.
+   Copyright (C) 2021  The igraph development team <igraph@igraph.org>
```
szhorvat, line 3:
```suggestion
   Copyright (C) 2022  The igraph development team <igraph@igraph.org>
```

tests/unit/igraph_hsbm_game.c
```diff
@@ -0,0 +1,74 @@
...
+   You should have received a copy of the GNU General Public License
+   along with this program.  If not, see <https://www.gnu.org/licenses/>.
+*/
+
+#include <igraph.h>
+#include "test_utilities.h"
+
+void call_and_print(igraph_integer_t n, igraph_integer_t m, igraph_vector_t *rho, igraph_matrix_t *pref_matrix, igraph_real_t p) {
+    igraph_t result;
+    IGRAPH_ASSERT(igraph_hsbm_game(&result, n, m, rho, pref_matrix, p) == IGRAPH_SUCCESS);
```
szhorvat, line 24: It is not necessary to assert that the return code is `IGRAPH_SUCCESS`. If it is not, the program aborts with an error and the test fails, even before the `IGRAPH_ASSERT` code could be reached.

We used to use this when the error handler was set to "ignore" so we could test invalid input. But a while ago you added a convenient macro to check bad input, so we almost never need to use a non-aborting error handler anymore.

szhorvat:
Thanks! I left two small comment, please go ahead and merge when done with those.

ntamas:
Thanks!

PR #2121, opened 2022-07-16
add weighted eccentricity

### LINKED ISSUE #1995: Wishlist: weighted eccentricity

A weighted version of `igraph_eccentricity()`.

**Use cases for the feature**

Generally useful vertex measure. We already have an unweighted version, and internal functions which _may_ be useful for the weighted one (`igraph_i_eccentricity_dijkstra()`).

GroteGnoom:
Fixes #1995

I quickly prototyped this by using the unweighted version, including the tests.

Is this the plan? Should it be a separate function? Do we deal with multi-edges by taking the smallest weight?


szhorvat:
I suggest a separate function named `igraph_eccentricity_dijkstra()`. Not a pretty name but it's what `diameter` and `average_path_length` use as well, so at least it's consistent.

Yes, we take the min weight from parallel edges. This is what the Dijsktra algorithm does, effectively. No need to special case multigraphs if you use inclists.

Note: I have not looked at the code except the header.

GroteGnoom:
I forgot this existed: https://github.com/igraph/igraph/pull/1736

I now just loop over vids, maybe things can be faster when the loop is in i_eccentricity_dijkstra itself.

ntamas:
> maybe things can be faster when the loop is in i_eccentricity_dijkstra itself

either that, or you could move the construction / destruction of objects used in `igraph_i_eccentricity_dijkstra()` out so they are allocated only once instead of as many times as the number of vertices you are querying.

GroteGnoom:
> either that, or you could move the construction / destruction of objects used in `igraph_i_eccentricity_dijkstra()` out so they are allocated only once instead of as many times as the number of vertices you are querying.

Which objects? For the inclist it's easy, and already done. But for the 2wheap and distance vector it doesn't really add much performance, and with the current algorithm they need to be refreshed for every starting vertex I think. It seems hard to really reuse the previous distance information without rewriting the algorithm.

I see that igraph_distances_dijkstra also clears its 2wheap for every starting vertex. So the best I can do is just save some allocations by not destroying and re-initializing? It does make the interface worse.

ntamas:
> So the best I can do is just save some allocations by not destroying and re-initializing? It does make the interface worse.

Indeed, this is the only benefit and it indeed makes the interface worse.
Let's just leave it as is.

Why is this targeted at 0.10, though? No API breakage here, this could go in any later 0.10.x release, right?

szhorvat:
Retargeted. I wanted a clean slate of PRs before the release.

src/paths/distances.c
```diff
@@ -170,6 +273,84 @@ igraph_error_t igraph_eccentricity(const igraph_t *graph,
...
+                        igraph_vs_t vids,
+                        igraph_neimode_t mode) {
+    igraph_lazy_inclist_t inclist;
+    igraph_vit_t vit;
+    igraph_integer_t dump;
+    igraph_real_t ecc;
+    igraph_integer_t no_of_edges = igraph_ecount(graph);
+
+    if (weights == NULL) {
+        return(igraph_eccentricity(graph, res, vids, mode));
```
szhorvat, line 318:
```suggestion
        return igraph_eccentricity(graph, res, vids, mode);
```

src/paths/distances.c
```diff
@@ -121,6 +121,109 @@ static igraph_error_t igraph_i_eccentricity(const igraph_t *graph,
...
 }
 
+/**
+ * This function finds the weighted eccentricity and returns it via \p ecc.
+ * It's used for igraph_pseudo_diameter_dijkstra. \p vid_ecc returns the vertex
+ * id of the ecc with the greatest
+ * distance from \p vid_start. If two vertices have the same greatest distance,
+ * the one with the lowest degree is chosen.
+ * When the graph is not (strongly) connected and \p unconn is false, then \p ecc
+ * wil be set to infinity, and \p vid_ecc to -1;
```
szhorvat, line 131:
```suggestion
 * This function finds the weighted eccentricity and returns it via \p ecc.
 * It's used for igraph_pseudo_diameter_dijkstra() and igraph_eccentricity_dijkstra().
 * \p vid_ecc returns the vertex id of the ecc with the greatest
 * distance from \p vid_start. If two vertices have the same greatest distance,
 * the one with the lowest degree is chosen.
 * When the graph is not (strongly) connected and \p unconn is false, then \p ecc
 * wil be set to infinity, and \p vid_ecc to -1;
```

src/paths/distances.c
```diff
@@ -170,6 +273,84 @@ igraph_error_t igraph_eccentricity(const igraph_t *graph,
...
+                                          IGRAPH_NO_LOOPS));
+    IGRAPH_FINALLY(igraph_lazy_inclist_destroy, &inclist);
+
+    IGRAPH_CHECK(igraph_vector_resize(res, 0));
+    IGRAPH_CHECK(igraph_vit_create(graph, vids, &vit));
+
+    for (IGRAPH_VIT_RESET(vit);
+            !IGRAPH_VIT_END(vit);
+            IGRAPH_VIT_NEXT(vit)) {
+        IGRAPH_CHECK(igraph_i_eccentricity_dijkstra(graph, weights, &ecc, IGRAPH_VIT_GET(vit), /*vid_ecc*/&dump, /*unconn*/ 1, &inclist));
```
szhorvat, line 346:
```suggestion
        IGRAPH_CHECK(igraph_i_eccentricity_dijkstra(graph, weights, &ecc, IGRAPH_VIT_GET(vit), /*vid_ecc*/ &dump, /*unconn*/ 1, &inclist));
```

szhorvat (approved):
LGTM.

szhorvat:
Actually, one thing is missing: update `functions.yaml`.

GroteGnoom:
Yes, fixing it!

interfaces/functions.yaml
```diff
@@ -975,6 +975,13 @@ igraph_eccentricity:
         NEIMODE mode=ALL
     DEPS: vids ON graph, res ON graph vids
 
+igraph_eccentricity_dijkstra:
+    PARAMS: |-
+        GRAPH graph, EDGEWEIGHTS weights=NUL,
```
szhorvat, line 980:
```suggestion
        GRAPH graph, EDGEWEIGHTS weights=NULL, 
```

interfaces/functions.yaml
```diff
@@ -975,6 +975,13 @@ igraph_eccentricity:
         NEIMODE mode=ALL
     DEPS: vids ON graph, res ON graph vids
 
+igraph_eccentricity_dijkstra:
+    PARAMS: |-
+        GRAPH graph, EDGEWEIGHTS weights=NUL, 
+        LOUT VERTEX_QTY res, VERTEXSET vids=ALL,
```
szhorvat, line 981:
```suggestion
        OUT VERTEX_QTY res, VERTEXSET vids=ALL,
```

interfaces/functions.yaml
```diff
@@ -975,6 +975,13 @@ igraph_eccentricity:
...
+        NEIMODE mode=ALL
```
szhorvat, line 982:
```suggestion
        GRAPH graph, EDGEWEIGHTS weights=NULL, 
        OUT VERTEX_QTY res, VERTEXSET vids=ALL,
        NEIMODE mode=ALL
```

PR #2143, opened 2022-07-31
Add hsbm list game tests

GroteGnoom:
Part of #1592

There was a heap-buffer-overflow because the wrong variable name was used.


szhorvat:
As far as I can tell, this bug was not present on `master`, and introduced on `develop` only. This shows how important it is to have these tests.

szhorvat (approved):

PR #2156, opened 2022-08-07
provide missing terminating newlines it out[puts]

jgmbenoit:
Description: upstream: examples: missing No Newline at EOF
 Some output files against which outputs of simple examples
 can be checked against miss a terminating newline ('\n').
 The terminatiing newline is expected by Un*x tools.
 This patch corrects this issue by appending the missing
 newline wherever appropriate. (Note that we can visualise
 newlines with cat(1): $ cat -A <DATAFILE>). This patch is
 meant to be submitted to the upstream maintainer.
Origin: debian
Author: Jerome Benoit
Last-Update: 2022-08-07


ntamas:
Thanks again!

PR #2157, opened 2022-08-07
redirect expected Warnings to <stdout> in examples

jgmbenoit:
Description: upstream: examples: reserve stderr for errors
 Some simple examples emit (expected) Warning messages to
 <stderr> (others to <stdout>). As <stderr> is a simple mean
 used by some test machineries to determinate test FAILUREs,
 expected warning message might be instead emitted to <stdout>
 (ideally to a <stdwar>). This patch redirects expected Warnings
 to <stdout> where applicable; it also provides the expected
 outputs. This patch is meant to be submitted to the upstream
 maintainer.
Origin: debian
Author: Jerome Benoit
Last-Update: 2022-08-07

### DIFF
--- a/examples/simple/graphml.c
+++ b/examples/simple/graphml.c
@@ -23,11 +23,19 @@
 #include <stdio.h>
 #include <unistd.h>     /* unlink */
 
+void igraph_stdwar_print_warning(const char *reason, const char *file,
+                                 int line) {
+    fprintf(stdout,
+        "STDWAR: Warning at %s:%i : %s\n", file, line, reason);
+}
+
 int main() {
     igraph_t g;
-    igraph_error_handler_t* oldhandler;
+    igraph_error_handler_t* olderrorhandler;
     int result;
     FILE *ifile, *ofile;
+
+    igraph_set_warning_handler(igraph_stdwar_print_warning);
 
     igraph_set_attribute_table(&igraph_cattribute_table);
 
@@ -39,7 +47,7 @@ int main() {
 
     /* GraphML support is an optional feature so we disable the error handler
      * temporarily to handle the case when it is not implemented */
-    oldhandler = igraph_set_error_handler(igraph_error_handler_ignore);
+    olderrorhandler = igraph_set_error_handler(igraph_error_handler_ignore);
     if ((result = igraph_read_graph_graphml(&g, ifile, 0))) {
         /* maybe it is simply disabled at compile-time */
         if (result == IGRAPH_UNIMPLEMENTED) {
@@ -47,7 +55,7 @@ int main() {
         }
         return 1;
     }
-    igraph_set_error_handler(oldhandler);
+    igraph_set_error_handler(olderrorhandler);
 
     fclose(ifile);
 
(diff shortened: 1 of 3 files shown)


szhorvat:
I don't understand the description of this PR. Can you clarify what you are trying to do?

I would rather not include warnings in the expected output of an example. Examples should be kept as simple as possible.

The GraphML example treats a return value as a Boolean. This should be fixed, and replaced by a comparison with `IGRAPH_SUCCESS`.

jgmbenoit:
The two simple examples `graphml` and `safelocale` write to stderr even when every works as expected.

In details, `graphml` reads `test.graphml` which contains a flaw at line 20 as explained in line 17:
> d3 is incorrect attribute key for nodes, should issue a warning
And indeed a warning is issued to stderr. This patch does not treat this Warning as an error but rather and an expected Warning by writing it to stdout so that stderr remains clear. This example looks actually more a test than an example.

As concerns `safelocale`, it always write a message to stderr. The patch only emits an error when the desired locale language can not be  set up as expected.

Having a clear stderr can be used by autopkgtest(1) to claim that an autotest is successful.

szhorvat:
It is by design that warnings are issued to stderr. Output to stderr does not mean that the test failed. Can the Debian test system not handle this?

These two are not the only tests that output warnings to stderr (see e.g. `tests/unit/gml.c`), and changing this would be rather inconvenient. Outputting to stdout is not a reasonable option. We have several tests which read a graph from one format and output to another (or the same) format. Warnings go to stderr because they should not be interspersed with machine-readable output. Warnings also change frequently and the specific message may change across platforms (as you can see in the test failures of this PR). I do not want to create a situation where we need to update test outputs each time the phrasing of an error or warning message is updated—we've been there and it's not tenable. At the same time, I do want to see these warnings when I run these tests manually. I do so regularly when working on the graph readers, to ensure that warnings are reasonable. So disabling warnings entirely is also not something I am in favour of.

In summary, making these changes across the test suite (which includes more than these two cases) would create a burden for day-to-day igraph development.

Is there any other way we can make this work conveniently with Debian's setup?

jgmbenoit:
I was dealing with the examples in examples. The Debian CI autotests just check regularly on different arch that those examples build and execute as expected. The tests in tests are only checked at build time. I will remove this patch from Debian and I will refine my scripts.

jgmbenoit:
So I am closing the pull request.

szhorvat:
OK, so you only need this for examples, not for proper tests?

In that case we can move the GraphML example into tests (as you say, it's close to a test than an example) and create a simpler example in place of it that shows no warnings. I can look at the `setlocale()` thing and see if I can make it behave. I don't remember why I added that printing to stderr, and it does seem avoidable.

Would this work for you?

jgmbenoit:
This would be great. And along the way it might simplify further the examples.

szhorvat:
OK, this is easy enough so I'll do it before the 0.10 release.

Whether a no-stderr policy can be maintained for examples in the long term remains to be seen. No hard promises about that right now, let's see how it goes.

PR #2164, opened 2022-08-13
add hrg_create tests

GroteGnoom:
I think `igraph_hrg_init` with n=0 used to work, but this passed -1 to `igraph_vector_init`, and assertions were added to `igraph_vector_init` in 4d72c2fa24 to prevent that.


src/hrg/hrg.cc
```diff
@@ -285,6 +285,12 @@ static igraph_error_t igraph_i_hrg_getsimplegraph(const igraph_t *igraph,
  */
 
 igraph_error_t igraph_hrg_init(igraph_hrg_t *hrg, igraph_integer_t n) {
+    if (n < 0) {
+        IGRAPH_ERRORF("Number of vertices should be non-negative, found %" IGRAPH_PRId ".", IGRAPH_EINVAL, n);
```
szhorvat, line 289: Nitpick: Let use the friendlier phrasing "should not be negative" for these cases.

szhorvat, line 289:
```suggestion
        IGRAPH_ERRORF("Number of vertices should not be negative, got %" IGRAPH_PRId ".", IGRAPH_EINVAL, n);
```

ntamas:
Looks good so far, but if you are already going over the HRG functions, can you also go through the source of `igraph_hrg_create()`? I can see at least one place where we missed an `IGRAPH_CHECK()` around an `igraph_hrg_resize()` call.

szhorvat:
Yes, the HRG functions are the last remaining pieces of igraph that have not yet been fixed up properly. 

There are almost certainly some memory management issues where functions that can fail are not checked. But be careful: we can't use `IGRAPH_ERROR` from functions that don't return `igraph_error_t`. I fixed such a bug today:

c8f811cf860a04d4d05f08e5b7a960a165017f61

GroteGnoom:
All right, I'll put some more work into it than just testing. I was confused by the documentation, so I'll look into that too. It does take some time. I'm reading the cited articles to know what's going on.

szhorvat:
This is the original code written by the paper's author.

It might help to experiment with the R interface, which is the only one that exposes this functionality.

GroteGnoom:
The interface is very simple, and I'm bad with R, so I'm choosing C for now. Maybe in general I should try to work with R and Python a bit more so I can more easily explore problems.

`hrg_create` now forces the user to give a probability for all nodes, including the leaf nodes, and then ignores them. Is that really what we want? Why not just give probabilities for all the internal nodes? I find It very confusing right now.

ntamas:
Indeed it does not make much sense; leaf nodes don't need probabilities.

szhorvat:
This PR contains a breaking change. Do we include it in 0.10?

GroteGnoom:
The problem with the non-tree test is that it passes through all the checks at the start of `igraph_hrg_create()`. These are checks like 'HRG nodes must have in-degree one, except for the root vertex.". What these checks _should_ check is that the graph is a directed [full](https://en.wikipedia.org/wiki/Binary_tree#Types_of_binary_trees) binary tree, but they don't. 

Should we check if it's a full binary tree? Is my analysis correct that it should be? Maybe just checking if it's a binary tree would be good enough.

What would be the best way to check if a graph is a (full) binary tree? Maybe a variant of the `is_tree` algorithm?

ntamas:
> Should we check if it's a full binary tree? Is my analysis correct that it should be?

Indeed, it should be a full binary tree.

ntamas:
> What would be the best way to check if a graph is a (full) binary tree?

Can we do it recursively?

1. Check if the graph has N nodes and N-1 edges.
2. Then, if N = 1, we are done. If not, then pick a vertex, test that it has 2 neighbors, and do the test recursively for both neighbors after removing (masking) the vertex that we picked in this round.

GroteGnoom:
With a recursive solution I'm a little worried about slowness and stack overflows on big graphs.

ntamas:
It wouldn't be "real" recursion, you could do it with a stack where you push the nodes to be checked and popping them off one by one.

But probably here's another way to do it:

1. Check whether the graph has N nodes and N-1 edges.
2. Check whether all degrees are less than 4 (you cannot have a node with 4 edges in a full binary tree).
3. Pick a node with degree 2, remove it along with its incident edges. Rinse and repeat until there are no more such nodes.
4. At this point, all nodes should be isolates iff the tree was a full binary tree.

Haven't though this fully through, and I hope the process is not screwed up by multiple/loop edges (you should check for loops in advance, that's easy to do), but I think it could work.

GroteGnoom:
I'm now checking for loops and multiple edges, and combine this with the old checks which do something a little bit similar to your suggestion.

GroteGnoom:
errors are external issues like
```
E: Failed to fetch http://azure.archive.ubuntu.com/ubuntu/pool/main/l/language-pack-de-base/language-pack-de-base_20.04+20220211_all.deb  404  Not Found [IP: 20.54.144.51 80]
E: Failed to fetch http://azure.archive.ubuntu.com/ubuntu/pool/main/l/language-pack-de/language-pack-de_20.04+20220211_all.deb  404  Not Found [IP: 20.54.144.51 80]
E: Unable to fetch some archives, maybe run apt-get update or try with --fix-missing?
##[error]Bash exited with code '100'.
```
(https://dev.azure.com/igraph-team/igraph/_build/results?buildId=4134&view=logs&jobId=4c3a8381-b9de-5463-904f-918810769be8&j=4c3a8381-b9de-5463-904f-918810769be8&t=f2c0ddb1-19a2-562d-d3bb-66fc4854897f)

ntamas:
Will merge when the CI tests pass (I just started them again).

PR #2168, opened 2022-08-18
Fix ieee754_endianness_check for node.js

flange-ipb:
Identified in [python-igraph#561](https://github.com/igraph/python-igraph/pull/561)

When cross-compiling to WebAssembly with the Emscripten SDK, ieee754_endianness_check is invoked by node.js, which requires a final newline in stdout in order to flush.


flange-ipb:
Do we need \r\n in Windows? :see_no_evil:

szhorvat:
> node.js, which requires a final newline in stdout in order to flush.

This explains why two tests failed.

This is a very problematic and non-standard (actually: plainly incorrect) behaviour. It is not reasonable for igraph to adapt to this. Is there no way to make emscripten behave correctly?

Sure, this one instance can be fixed now, but we can't commit to the extra mental overhead of remembering to always add final newlines.

szhorvat:
Here's the FAQ item:

https://emscripten.org/docs/getting_started/FAQ.html#what-does-exiting-the-runtime-mean-why-don-t-atexit-s-run

If I understand correctly:

 - This non-standard behaviour is default on purpose.
 - This affects only executable, not libraries, as it only concerns the `main()` function
 - The standard behaviour can be requested using `EXIT_RUNTIME=1`, so there is a clean solution.

If so, then `EXIT_RUNTIME` should be set while configuring and building igraph. This is necessary for the test suite as well as checks during configuration time. However, if some other code links to the igraph library later, that would not be affected, and can choose it's own setting for `EXIT_RUNTIME`.

Does my understanding sound accurate @flange-ipb ?

ntamas:
Nice catch @szhorvat , this is indeed what seems to be going on behind the scenes. `EXIT_RUNTIME=1` is indeed not suitable for the web environment where everything is running asynchronously in callbacks to a Giant JavaScript Event Loop that you have no control over, and there isn't really any point in time where the loop "exits" (it never does).

What I don't see clearly is whether it affects the code that is _built_, i.e. is the code the same with and without `EXIT_RUNTIME=1` (except for the `main()` function)?

szhorvat:
@hoodmane Would you be able to answer the last question above?

flange-ipb:
Indeed, having `EXIT_RUNTIME=1` for the whole project might be a bad option. `-s EXIT_RUNTIME` is an option of the `emcc` compiler and I'm not sure how to "transport" it in `ieee754_endianness.cmake` as cmake's [`try_run`](https://cmake.org/cmake/help/latest/command/try_run.html) doesn't seem to have any parameter for that.

Also, calling `fflush(stdout);` after the `printf` [is not a solution](https://github.com/emscripten-core/emscripten/issues/10990#issuecomment-862470770).

ntamas:
Okay, let's try to be a bit more pragmatic here :) Clearly, not having a newline there is a problem for `emcc` by default, so what if we

1) add a newline, and
2) only check whehther the text printed by the program _starts with_ "OK"? That way we don't have to deal with the different newline conventions.

If we worry that the program might accidentally print "OK" due to some debugging output coming from the cross-compiler emulator, let's change "OK" to "OMG it works" or something.

szhorvat:
I was not suggesting to define `EXIT_RUNTIME` specifically for the endianness test, but have it defined globally while configuring and compiling igraph. Presumably, this does not affect the library code, only executable programs.

ntamas:
I think it ultimately depends on whether `emcmake` allows us a convenient way to define it globally. If I understand correctly, this has to be passed as `emcc -sEXIT_RUNTIME=1` to every `emcc` invocation. `emcmake` pre-configures CMake with a toolchain file that makes CMake aware of `emcc` and teaches it how to call `emcc` to compile things. If it can be told easily to invoke `emcc` with `-sEXIT_RUNTIME=1` for the linking phase only (I think that's the only phase where `EXIT_RUNTIME=1` has any effect), I think we're done.

flange-ipb:
> emcmake pre-configures CMake with a toolchain file that makes CMake aware of emcc and teaches it how to call emcc to compile things. If it can be told easily to invoke emcc with -sEXIT_RUNTIME=1 for the linking phase only (I think that's the only phase where EXIT_RUNTIME=1 has any effect), I think we're done.

That would be something like `emcmake cmake .. -DIGRAPH_WARNINGS_AS_ERRORS:BOOL=OFF -DCMAKE_C_FLAGS:STRING="-s EXIT_RUNTIME=1"`?

szhorvat:
Yes, I think that might work, but you will need to add the same for `CMAKE_CXX_FLAGS` as well.

hoodmane:
`-sEXIT_RUNTIME` is only for the main module, I think it has no effect if added to a shared library. I also think it is likely that Emscripten will print the newline even with `-sEXIT_RUNTIME`. What you really need is to fix the stream handling. I'm not sure how you are running compiled Emscripten modules, but you need to add something like this `setupStreams` function:
https://github.com/hoodmane/pyodide/blob/cmdline-runner/tools/python.js#L39
For instance, you could make a file called `pre.js` with the following contents:

<details>
<summary>pre.js</summary>

```js
function make_tty_ops(stream) {
    let newline = false;
    return {
        // get_char has 3 particular return values:
        // a.) the next character represented as an integer
        // b.) undefined to signal that no data is currently available
        // c.) null to signal an EOF
        get_char(tty) {
            if (newline) {
                newline = false;
                return null;
            }
            if (!tty.input.length) {
                var result = null;
                var BUFSIZE = 256;
                var buf = Buffer.alloc(BUFSIZE);
                var bytesRead = fs.readSync(0, buf, 0, BUFSIZE, -1);
                if (bytesRead === 0) {
                    return null;
                }
                result = buf.slice(0, bytesRead);
                tty.input = Array.from(result);
            }
            let res = tty.input.shift();
            newline = true;
            return res;
        },
        put_char(tty, val) {
            stream.write(Buffer.from([val]));
        },
        flush(tty) {},
    };
}

function setupNodeStreams(FS, TTY) {
    let mytty = FS.makedev(FS.createDevice.major++, 0);
    let myttyerr = FS.makedev(FS.createDevice.major++, 0);
    TTY.register(mytty, make_tty_ops(process.stdout));
    TTY.register(myttyerr, make_tty_ops(process.stderr));
    FS.mkdev("/dev/mytty", mytty);
    FS.mkdev("/dev/myttyerr", myttyerr);
    FS.unlink("/dev/stdin");
    FS.unlink("/dev/stdout");
    FS.unlink("/dev/stderr");
    FS.symlink("/dev/mytty", "/dev/stdin");
    FS.symlink("/dev/mytty", "/dev/stdout");
    FS.symlink("/dev/myttyerr", "/dev/stderr");
    FS.closeStream(0);
    FS.closeStream(1);
    FS.closeStream(2);
    FS.open("/dev/stdin", 0);
    FS.open("/dev/stdout", 1);
    FS.open("/dev/stderr", 1);
}

if(!Module.preRun){
	Module.preRun = [];
}

const IN_NODE =
  typeof process !== "undefined" &&
  process.release &&
  process.release.name === "node" &&
  typeof process.browser === "undefined";

if (IN_NODE) {
	Module.preRun.push(() => setupNodeStreams(FS, TTY));
}
```

</details>

and then add `--pre-js pre.js` to your LDFLAGS.

hoodmane:
> What I don't see clearly is whether it affects the code that is built, i.e. is the code the same with and without EXIT_RUNTIME=1 (except for the main() function)?

Not exactly, but it is quite similar. `EXIT_RUNTIME` adds a bunch of stuff to keep track of tasks to do `ATEXIT` and a `runtimeKeepaliveCounter` and stuff like this. I don't think the code for the `main` function is changed at all actually.

ntamas:
>  I'm not sure how you are running compiled Emscripten modules

We aren't either :) My understanding is that we simply prepend `emcmake`  to our `cmake` invocation, which sets up CMake nicely to work with Emscripten (i.e. tells CMake that it's now cross-compiling, asks CMake to invoke `emcc` instead of `cc` and use `node` to run the compiled code etc). All the magic seems to happen in the [CMake toolchain file](https://github.com/emscripten-core/emscripten/blob/main/cmake/Modules/Platform/Emscripten.cmake) that `emcmake` uses to configure the build.

flange-ipb:
> We aren't either :) My understanding is that we simply prepend emcmake to our cmake invocation, which sets up CMake nicely to work with Emscripten (i.e. tells CMake that it's now cross-compiling, asks CMake to invoke emcc instead of cc and use node to run the compiled code etc). All the magic seems to happen in the [CMake toolchain file](https://github.com/emscripten-core/emscripten/blob/main/cmake/Modules/Platform/Emscripten.cmake) that emcmake uses to configure the build.

Exactly, `node` is registered as `CMAKE_CROSSCOMPILING_EMULATOR`. This is used to execute `ieee754_endianness_check` and the tests.

How to proceed here now? Adding `-DCMAKE_C_FLAGS:STRING="-sEXIT_RUNTIME" -DCMAKE_CXX_FLAGS:STRING="-sEXIT_RUNTIME"` to the initial `emcmake cmake ..` doesn't resolve the issue with `ieee754_endianness_check` because the flags are not passed to the compiler. I'd suggest to merge the fix done in this PR (add newline) and keep the flags in mind for test build and execution.

- Should the flags be added to the python wheel build? I'd tend not to do it because igraph is used as a library there.
- @ntamas @szhorvat Can you add a CI for the WebAssembly build here? (I'm not familiar with the Azure CI, but I'll contribute the Wheel-build Action in the python project.) Should be easy: Take the existing Linux CI, install and activate the Emscripten SDK (version 3.1.14) and run `emcmake cmake .. -DIGRAPH_WARNINGS_AS_ERRORS:BOOL=OFF -DCMAKE_C_FLAGS:STRING="-sEXIT_RUNTIME" -DCMAKE_CXX_FLAGS:STRING="-sEXIT_RUNTIME"`

Speaking of tests:
Adding `-sEXIT_RUNTIME` doesn't change the set of failed tests. 34 fail at the moment.

flange-ipb:
@hoodmane 

> For instance, you could make a file called pre.js with the following contents:
pre.js
and then add --pre-js pre.js to your LDFLAGS.

I'm trying to elaborate your suggestion for [ieee754_endianness_check.c](https://github.com/igraph/igraph/blob/develop/etc/cmake/ieee754_endianness_check.c):
```
igraph@xxx:~/igraph/etc/cmake$ emcc --pre-js pre.js ieee754_endianness_check.c 
igraph@xxx:~/igraph/etc/cmake$ node a.out.js 
/home/igraph/igraph/etc/cmake/a.out.js:213
      throw ex;
      ^

ReferenceError: TTY is not defined
    at /home/igraph/igraph/etc/cmake/a.out.js:88:51
    at callRuntimeCallbacks (/home/igraph/igraph/etc/cmake/a.out.js:1777:11)
    at preRun (/home/igraph/igraph/etc/cmake/a.out.js:1333:3)
    at run (/home/igraph/igraph/etc/cmake/a.out.js:2355:3)
    at runCaller (/home/igraph/igraph/etc/cmake/a.out.js:2297:19)
    at removeRunDependency (/home/igraph/igraph/etc/cmake/a.out.js:1478:7)
    at receiveInstance (/home/igraph/igraph/etc/cmake/a.out.js:1666:5)
    at receiveInstantiationResult (/home/igraph/igraph/etc/cmake/a.out.js:1684:5)
igraph@xxx:~/igraph/etc/cmake$ emcc ieee754_endianness_check.c 
igraph@xxx:~/igraph/etc/cmake$ node a.out.js 
stdio streams had content in them that was not flushed. you should set EXIT_RUNTIME to 1 (see the FAQ), or make sure to emit a newline when you printf etc.
(this may also be due to not including full filesystem support - try building with -sFORCE_FILESYSTEM)
igraph@xxx:~/igraph/etc/cmake$ emcc -sEXIT_RUNTIME ieee754_endianness_check.c 
igraph@xxx:~/igraph/etc/cmake$ node a.out.js 
OK
```

hoodmane:
Hmm try `Module.TTY`? Or perhaps dead code elimination is removing `TTY` and you have to pass another flag to force it to keep it. I will look into it and get back to you...

szhorvat:
@ntamas I suggest we merge this now as-is to make progress.

szhorvat:
> 2\. only check whehther the text printed by the program _starts with_ "OK"? That way we don't have to deal with the different newline conventions.

Why not print `"OK\n"` and test for exactly that using `STREQUAL`, as before? Is there any reason this wouldn't work?

hoodmane:
Okay well the following works:
```
emcc ieee754_endianness_check.c --pre-js pre.js -sFORCE_FILESYSTEM
```
and `pre.js` needs an extra `FS.init()`:

<details>
<summary>`pre.js`</summary>

```js
function make_tty_ops(stream) {
    let newline = false;
    return {
        // get_char has 3 particular return values:
        // a.) the next character represented as an integer
        // b.) undefined to signal that no data is currently available
        // c.) null to signal an EOF
        get_char(tty) {
            if (newline) {
                newline = false;
                return null;
            }
            if (!tty.input.length) {
                var result = null;
                var BUFSIZE = 256;
                var buf = Buffer.alloc(BUFSIZE);
                var bytesRead = fs.readSync(0, buf, 0, BUFSIZE, -1);
                if (bytesRead === 0) {
                    return null;
                }
                result = buf.slice(0, bytesRead);
                tty.input = Array.from(result);
            }
            let res = tty.input.shift();
            newline = true;
            return res;
        },
        put_char(tty, val) {
            stream.write(Buffer.from([val]));
        },
        flush(tty) {},
    };
}

function setupNodeStreams(FS, TTY) {
    let mytty = FS.makedev(FS.createDevice.major++, 0);
    let myttyerr = FS.makedev(FS.createDevice.major++, 0);
    TTY.register(mytty, make_tty_ops(process.stdout));
    TTY.register(myttyerr, make_tty_ops(process.stderr));
    FS.mkdev("/dev/mytty", mytty);
    FS.mkdev("/dev/myttyerr", myttyerr);
    FS.unlink("/dev/stdin");
    FS.unlink("/dev/stdout");
    FS.unlink("/dev/stderr");
    FS.symlink("/dev/mytty", "/dev/stdin");
    FS.symlink("/dev/mytty", "/dev/stdout");
    FS.symlink("/dev/myttyerr", "/dev/stderr");
    FS.closeStream(0);
    FS.closeStream(1);
    FS.closeStream(2);
    FS.open("/dev/stdin", 0);
    FS.open("/dev/stdout", 1);
    FS.open("/dev/stderr", 1);
}

if(!Module.preRun){
	Module.preRun = [];
}

const IN_NODE =
  typeof process !== "undefined" &&
  process.release &&
  process.release.name === "node" &&
  typeof process.browser === "undefined";

if (IN_NODE) {
	Module.preRun.push(() => {
        FS.init();
        setupNodeStreams(FS, TTY);
    });
}
```

</details>

hoodmane:
Anyways the solution here of adding a trailing newline looks perfectly reasonable. If you run into this problem more, the solution I gave above fixes stdin, stdout, and stderr to behave the way you expect.

szhorvat:
Sorry about going ahead with merging this quickly. I was trying to figure out why the same test fails on Windows UWP, and the newline gave a bit of extra security that it's not related to that.

hoodmane:
Well the change set in this PR is simpler so it's probably better.

flange-ipb:
NB:
@hoodmane's suggested pre.js would fix the original ieee754_endianness_check, which did not add \n. But again, there would be the problem of telling cmake's try_run to use these flags.
I also ran the test suite with the develop branch with these additions but didn't see any improvements. That would be something like
```
emcmake cmake .. -DIGRAPH_WARNINGS_AS_ERRORS:BOOL=OFF -DCMAKE_C_FLAGS:STRING="--pre-js /path/to/pre.js -sFORCE_FILESYSTEM" -DCMAKE_CXX_FLAGS:STRING="--pre-js /path/to/pre.js -sFORCE_FILESYSTEM"
cmake --build . --target check
```

Issue #2165, opened 2022-08-15
Many examples use int for error codes


szhorvat:
There are many examples that use `int` instead of `igraph_error_t` for error codes. These should be fixed so that they don't confuse newcomers. A simple way to find them is to grep the `examples` directory:

```bash
grep 'int\s.*;' -R .
```

However, most of these are not actually examples but tests that attempt to verify that error codes are as expected. These should be moved to the `tests` directory and replaced with proper examples.

szhorvat:
@GroteGnoom Unless you're aware of other examples of using `int` instead of `igraph_error_t`, please close this.

PR #2171, opened 2022-08-20
Move error checks from examples to tests

GroteGnoom:
Part of  #2165

igraph_degree was failing. 
__lsan_unregister_root_region(): region at 0x7fb7e31fe400 of size 512 has not been registered

It was a vector that was not initialized but was destroyed ([db6c81c](https://github.com/igraph/igraph/pull/2171/commits/db6c81c77c0af1a7cb204537298b91acf0487c92)) Pretty confusing situation, because it worked on some systems, and the sanitizers didn't complain on those.


ntamas:
Looks good so far! For the non-trivial cases (`igraph_bipartite_create()` and `igraph_subisomorphic_lad()`), maybe we should add a comment above the `CHECK_ERROR()` lines explaining why we expect the given error code (i.e. what's wrong with the input).

ntamas (approved):

ntamas:
Thanks a lot!

PR #2173, opened 2022-08-27
use igraph_matrix_print in examples

GroteGnoom:
I'd like examples to use igraph's matrix printing functionality, instead of repeatedly defining a separate matrix printing function in different examples.

There's two problems:
1 - Printing of nan and inf is not done correctly in old visual studio compilers. We use the output of the examples as checks, so we want consistent outputs.
2 - We want the matrix to be readable, so it would be nice if the rows are aligned.

I've tried to make a quick example of a possible solution, but maybe there's an easier way.


src/core/printing.c
```diff
@@ -79,6 +79,23 @@ int igraph_real_printf(igraph_real_t val) {
         return printf("%g", val);
     }
 }
+
+int igraph_real_printf_aligned(igraph_real_t val) {
```
szhorvat, line 83: Some ideas:

 - Strings like `"Inf"` can also be printed in an aligned way using `%8s`
 - If we use `*` instead of a number in `printf`, then an additional `int` argument can specify the field width. This makes it possible to add an extra argument for this function that controls the field width instead of hard-coding 8.

src/core/printing.c
```diff
@@ -79,6 +79,23 @@ int igraph_real_printf(igraph_real_t val) {
...
+        return printf("%8g", val);
+    } else if (igraph_is_nan(val)) {
+        return printf("     NaN");
+    } else if (igraph_is_inf(val)) {
+        if (val < 0) {
+            return printf("    -Inf");
+        } else {
+            return printf("     Inf");
+        }
+    } else {
```
szhorvat, line 94: But this cannot possibly happen, right? If so, it's better to put an `IGRAPH_ASSERT` or `IGRAPH_FATAL` here so we become aware of any problems instead of blindly trying to work around them, without even knowing if they exist.

szhorvat:
This PR involves a "slightly breaking" change as it modifies the output format of `print_matrix()`. I think it would make sense to handle this before 0.10.

@GroteGnoom Did you see my comments above, and what do you think?

src/core/printing.c
GroteGnoom: Yes, I just copied it, but it should just give an error.

src/core/printing.c
GroteGnoom: Very good point! Somehow I didn't think about using the `*`, even though I implemented it myself as practice last year 😬

szhorvat:
Note to self: We need to also consider the complex printing functions.

szhorvat:
I'm not sure what the best field width would be for printing matrices. `%g` will produce as many as 13 characters, which means that we need a field width of at least 13, plus a leading space, to make printing safe. Alternatively we can reduce the number of digits that is printed, but I don't like that.

```
-1.23457e+123  <-- 13 chatacters
-1.23457       <-- 8 characters
```

szhorvat:
@ntamas @GroteGnoom 

I am going to stop here and first ask you to take a look, and decide whether we want this in 0.10.

Reasons to include in 0.10:

 - This is a "semi-breaking" change in that it modifies the way that the matrix is printed

Changes in this PR:

 - The matrix printing function now aligns columns, and the column width is calculated dynamically for each column. If we don't do this, each entry would need to take up 14 characters which is too much. This way we get nice readable output without making columns unnecessarily wide.
 - Inf/NaN are printed consistently, even on non-standards-compliant MSVCRT. Note that: (1) UCRT is standards compliant (2) MSVC uses UCRT now, only some flavours of MinGW are stuck on MSVCRT
 - Default format changes from `%G` to `%g`, meaning that `e` is printed instead of `E`. But most places already had `e`, so I don't know if this makes a difference at all.
 - Some new functions are added for aligned printing.

Concerns:

 - The change turned out to be quite a bit more complex than I expected, so I'm not keen to rush it
 - I am also not keen on exposing the aligned printing functions just yet
 - Right now alignment width is calculated separately, per-column. This means that if a value changes in a column, increasing the column width, then the printing of all rows of that matrix will change. This will make it harder to identify what changes in a diff.
 - There are additional exposed functions

ntamas:
There's no need to rush this into 0.10. I wouldn't consider it a breaking change if the format of the printing functions changes; the outputs of these functions are meant to be human-readable but I think we never made promises that we will never change the format of the output. So, take your time, there's no rush, but there's no need to overcomplicate things either. Anything that suits our goals of using these functions in examples would do. (I probably wouldn't even have bothered with the alignment of the columns).

szhorvat:
@ntamas It's done now, I'll leave it to you do decide what to do with it. I am not very comfortable with rushing it either, so we can let it sit for a few days.

ntamas:
This seems fine for me; if there are any remaining bugs in there, that's what patch releases are for :)

PR #2174, opened 2022-08-27
export attribute printing

GroteGnoom:
attribute examples that want to print results now define their own attribute printing function. This PR exports attribute printing so it's available to everyone and can be used in examples.


szhorvat:
Some concerns:

 - Do we really want to increase the API surface with a non-critical function in an area which will very likely see a significant overhaul?
 - It feels like a debugging function, do we really need it in the API? The output is human-readable, but not machine readable. The only applications I see are debugging and examples.
 - This works only with the C attribute handler, so it should not be called `attribute` but at least `cattribute`. The comment should not make the impression that it works with all attribute handlers.
 - Anything that uses `printf` needs to be guarded with `USING_R`, unfortunately. 
 - The printing function itself made a reasonably good example. Having it in the example sources was perhaps not a bad thing.

GroteGnoom:
I thought I would just reduce some code duplication. I don't think increasing the API surface is a big deal. Even when internals would change, maybe this output would stay the same, it just prints out values assigned to keys. If everything changes so drastically that this function's output needs changes, this would only be a minor concern compared to all the other changes.

It's just a function that does something useful so people don't have to write it themselves. I think the amount of people that will be using this is ~1, so it's all not a big deal anyway. It's just so we don't have _almost_ the same code in two places. It makes the example 'short and simple' (https://github.com/igraph/igraph/wiki/How-to-write-unit-tests%3F). It makes it easier to change the function because we don't have to do it twice.

szhorvat:
> I thought I would just reduce some code duplication.

Was there any code duplication? It was removed from only one example. It wasn't shared between multiple.

GroteGnoom:
The same function already existed in the test utilities.

szhorvat:
Yes, but that's for testing. This is an example demonstrating how to to use the C attribute handler. The printing function itself is a good way to do that.

ntamas:
Okay, so to sum it all up:

* Printing attributes this way is mostly for testing and debugging purposes.
* We already have canonical functions to do that in `test_utilities.c`
* The attribute printing code was mostly duplicated in `examples/simple/cattributes.c` because examples have to be self-contained so we cannot use `test_utilities.c` there. I guess this was the original reason that triggered this PR.
* I agree with @szhorvat that this should not be part of the public API as its usage is limited - it only works with our C attribute handler.

I think the conclusion is not to merge this PR, given the discussion above. I'm closing it but feel free to reopen if you disagree and you want to continue the discussion further.

Issue #1574, opened 2020-12-17
Separate tests from examples


szhorvat:
Right now the `examples/simple` directory contains both tests and examples linked from the documentation. Many of these "examples" are not actually useful as sample code. These should be identified and moves to `examples/tests`.

Additional;y, files in `examples/tests` should be moved to `tests/unit`.

ntamas:
`examples/tests` is no more, all the relevant files are now in `tests/unit`

ntamas:
I went through the files in `examples/simple` and moved those that were not explicitly referenced from the docs to `tests/unit`. I don't know when I will have time to continue this so I'll just put it back to the common stack.

szhorvat:
Related: #1585.

ntamas:
Phew, I started thinking about this, but I had to realize that this is going to be pretty subjective. I propose sorting the `.c` files in `examples/simple` by size in descending order, and then to start processing the list from the top. The reasoning is that files that are large are more likely to be tests than real examples. Ideally, we should limit ourselves to including small examples in the documentation only, preferably not more than, say, 15-20 lines of "real" code (not counting comments, empty lines and such).

Jeroenvh99:
I would like to work on this, is this task still available?

ntamas:
Thanks for your interest in this issue! Yes, this is still something that needs to be done; @GroteGnoom was working on tidying up the examples in PR #2175, so please take a look at that PR as well, but other than that you are free to work on it. If you do, please open a draft PR so we can follow the progress and others can see that you are already working on it.

The ultimate goal is to ensure that `tests/unit` contains unit tests only (i.e. code that is meant to be executed to test that the library is working correctly) and `examples/simple` contains illustrative examples only (i.e. code that is presented in a way that it can be embedded in the user manual). Right now it's a mixed bag; historically, the unit tests were also in `examples/simple` and many of the examples that we still have there are more like unit tests than illustrative examples.

Jeroenvh99:
Hi @ntamas,
I'm going to work on this issue, could you assign me to this issue then?

szhorvat:
Don't forget that you need to work together and keep in sync with @GroteGnoom who has already done part of the job in #2175. It might make sense to merge #2175 first, or otherwise be careful to avoid conflict.

Jeroenvh99:
I think it's easiest then that I open a PR with @GroteGnoom's repo, to ensure we stay in sync.

GroteGnoom:
[...]

PR #2175, opened 2022-08-28
Smaller examples

GroteGnoom:
Part of https://github.com/igraph/igraph/issues/1574


szhorvat:
@GroteGnoom @ntamas Can we get this into shape and merge before 0.10? I think there's great value in having good and educational examples.

szhorvat:
I'll do a quick review right now to help with getting this ready.

A slight concern I have is that this duplicates a lot of the code, effectively creating many more tests than we had before. This will blow up the running time of the test suite.

Still, I think this approach of "forking" earlier examples into an example part and a test part will be a good approach for the long term.

szhorvat:
@GroteGnoom Is there anything in particular that you'd want me to take a closer look in this PR?

GroteGnoom:
At least 
[lapack_dgeevx](https://github.com/igraph/igraph/pull/2175/commits/552256da02032ea15514788c504d33e8c4f5b623) and
[lapack_dsyevr](https://github.com/igraph/igraph/pull/2175/commits/96cbda6aaac7993e59b6658d5eacce4d189e1df6)

ntamas:
> Can we get this into shape and merge before 0.10?

It's not a tragedy if this does not get into 0.10 but only to 0.10.1 :) If you manage to mark this ready for review before I resume the 0.10 release process (not today, most likely tomorrow), it can get into 0.10. Otherwise it will get into 0.10.1. If you say that you need a bit more time to finish this (meaning a day or two, not a week or more), I am willing to wait a bit with 0.10.0, but as I said, one needs to draw a line somewhere, otherwise we could keep on delaying 0.10.0 indefinitely because there will always be "just one more thing" that we can squeeze into 0.10.0.

szhorvat:
@GroteGnoom Can you merge develop into this and fix whatever is broken by the updated matrix printing?

Actually it's probably better to do what Tamás said and leave this for 0.10.1. It shouldn't block 0.10, and let's not give ourselves too much pressure.

szhorvat (approved):
I did not look at each test individually, but the general approach seems fine. If there are concerns about the lapack tests, can we leave them for later?

szhorvat:
Just a note that after discussion in chat with @GroteGnoom, we decided that the `dgeevx` (and possibly `dsyevr`) example needs a bit more attention. Please do not merge until then.

szhorvat:
`dsyevr` example looks good.

Nitpicks:

 - I think we don't need variables for `il`, `iu`, `vl, `vu`.
 - Docs could be explicit about what the "first" eigenvalue is (the largest, I presume)

GroteGnoom:
> Optionally, the example can check the info value to see if we got convergence.
With an input of 1, errors are generated when there's a problem, so this check isn't necessary.

GroteGnoom:
> Docs could be explicit about what the "first" eigenvalue is (the largest, I presume)

Yes, that would be nice. But I can't find it in the original docs either.

dsyevr:
https://netlib.org/lapack/explore-html/d2/d8a/group__double_s_yeigen_gaeed8a131adf56eaa2a9e5b1e0cce5718.html

dsyevr refers to dstemr:
https://netlib.org/lapack/explore-html/da/dba/group__double_o_t_h_e_rcomputational_ga14daa3ac4e7b5d3712244f54ce40cc92.html

GroteGnoom:
Now the dgeevx example does not show the general way to print out the eigenvectors, but to do it right we need something like arpack_unpack_complex.

szhorvat:
> Now the dgeevx example does not show the general way to print out the eigenvectors, but to do it right we need something like arpack_unpack_complex.

Do you mean that you would like to introduce a function that converts the "compressed" representation that LAPACK returns into a proper complex matrix? Yes, that would be nice, but perhaps it's better to leave it for a followup PR, so this one can be merged. There are a very large number of files touched in this PR, so the risk of conflicts is high.

GroteGnoom:
> Do you mean that you would like to introduce a function that converts the "compressed" representation that LAPACK returns into a proper complex matrix? Yes, that would be nice, but perhaps it's better to leave it for a followup PR, so this one can be merged. There are a very large number of files touched in this PR, so the risk of conflicts is high.

I agree, I want this merged as soon as possible.

szhorvat:
> I agree, I want this merged as soon as possible.

Are you still planning to work on it, or is it good to go?

szhorvat:
Great! Now if anyone has time to change all those `main()` to `main(void)` and remove https://github.com/igraph/igraph/blob/master/etc/cmake/compilers.cmake#L54 that would be great!

szhorvat:
> Great! Now if anyone has time to change all those `main()` to `main(void)` in tests/examples and remove https://github.com/igraph/igraph/blob/master/etc/cmake/compilers.cmake#L54 that would be great!

This is done.

PR #2177, opened 2022-08-30
Including iterations in Leiden algorithm

### LINKED ISSUE #1423: Leiden algorithm: add iteration scheme to C core
@vtraag I noticed that both the [Python](https://github.com/igraph/python-igraph/pull/304) and [R](https://github.com/igraph/rigraph/pull/3990) bindings you contributed support iterating the algorithm multiple times, potentially until nothing changes.

So far, the philosophy that igraph followed was to implement as much as possible in the C core. It would be useful to include the iteration scheme in the C core as well, even if that seems trivial. It would both reduce the burden for the implementors of high-level interfaces, and provide some guidance on how to use the Leiden function.

vtraag:
This PR fixes #1423, where it was requested to add iteration to the Leiden algorithm in the C core, instead of having to reimplement it in higher-level languages.

### DIFF
--- a/interfaces/functions.yaml
+++ b/interfaces/functions.yaml
@@ -1601,7 +1601,7 @@ igraph_community_leiden:
     PARAMS: |-
         GRAPH graph, OPTIONAL EDGEWEIGHTS weights,
         OPTIONAL VERTEXWEIGHTS vertex_weights,
-        REAL resolution, REAL beta, BOOLEAN start, INTEGER n_iterations=2, 
+        REAL resolution, REAL beta=0.01, BOOLEAN start, INTEGER n_iterations=2, 
         OPTIONAL INOUT VECTOR_INT membership,
         OUT INTEGER nb_clusters, OUT REAL quality
     DEPS: weights ON graph, vertex_weights ON graph
--- a/interfaces/functions.yaml
+++ b/interfaces/functions.yaml
@@ -1601,7 +1601,7 @@ igraph_community_leiden:
     PARAMS: |-
         GRAPH graph, OPTIONAL EDGEWEIGHTS weights,
         OPTIONAL VERTEXWEIGHTS vertex_weights,
-        REAL resolution, REAL beta, BOOLEAN start, INTEGER n_iterations, 
+        REAL resolution, REAL beta, BOOLEAN start, INTEGER n_iterations=2, 
         OPTIONAL INOUT VECTOR_INT membership,
         OUT INTEGER nb_clusters, OUT REAL quality
     DEPS: weights ON graph, vertex_weights ON graph
--- a/interfaces/functions.yaml
+++ b/interfaces/functions.yaml
@@ -1601,7 +1601,8 @@ igraph_community_leiden:
     PARAMS: |-
         GRAPH graph, OPTIONAL EDGEWEIGHTS weights,
         OPTIONAL VERTEXWEIGHTS vertex_weights,
-        REAL resolution, REAL beta, BOOLEAN start, OPTIONAL INOUT VECTOR_INT membership,
+        REAL resolution, REAL beta, BOOLEAN start, INTEGER n_iterations, 
+        OPTIONAL INOUT VECTOR_INT membership,
         OUT INTEGER nb_clusters, OUT REAL quality
     DEPS: weights ON graph, vertex_weights ON graph
 
--- a/src/community/leiden.c
+++ b/src/community/leiden.c
@@ -56,7 +56,8 @@ static igraph_error_t igraph_i_community_leiden_fastmovenodes(
         const igraph_vector_t *edge_weights, const igraph_vector_t *node_weights,
         const igraph_real_t resolution_parameter,
         igraph_integer_t *nb_clusters,
-        igraph_vector_int_t *membership) {
+        igraph_vector_int_t *membership,
+        igraph_bool_t *changed) {
 
     igraph_dqueue_int_t unstable_nodes;
     igraph_real_t max_diff = 0.0, diff = 0.0;
@@ -179,6 +180,7 @@ static igraph_error_t igraph_i_community_leiden_fastmovenodes(
 
         /* Add stable neighbours that are not part of the new cluster to the queue */
         if (best_cluster != current_cluster) {
+            *changed = true;
             VECTOR(*membership)[v] = best_cluster;
 
             for (i = 0; i < degree; i++) {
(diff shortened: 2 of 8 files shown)


src/community/leiden.c
```diff
@@ -960,6 +960,9 @@ static igraph_error_t igraph_i_community_leiden(
  * \param start Start from membership vector. If this is true, the optimization
  *    will start from the provided membership vector. If this is false, the
  *    optimization will start from a singleton partition.
+ * \param n_iterations Iterate the core Leiden algorithm for the indicated number 
+ *    of times. If this is a negative number, it will continue iterating until 
+ *    the quality no longer changed.
```
szhorvat, line 965:
```suggestion
 *    the quality no longer changes.
```

src/community/leiden.c
```diff
@@ -1036,10 +1040,29 @@ igraph_error_t igraph_community_leiden(const igraph_t *graph,
...
 
-    /* Perform actual Leiden algorithm */
-    IGRAPH_CHECK(igraph_i_community_leiden(graph, i_edge_weights, i_node_weights,
-                                           resolution_parameter, beta,
-                                           membership, nb_clusters, quality));
+    /* Perform actual Leiden algorithm iteratively. We either
+     * perform a fixed number of iterations, or we perform
+     * iterations until the quality remains unchanged. Note that
+     * the Leiden algorithm can still change in a subsequent 
+     * iteration.
```
szhorvat, line 1047: I don't understand the last sentence. Do you mean that even if we use `n_iterations < 0`, another application of the function to the result may produce a change just because the algorithm is randomized?

Could you clarify the comment?

src/community/leiden.c
```diff
@@ -1036,10 +1040,29 @@ igraph_error_t igraph_community_leiden(const igraph_t *graph,
...
+        IGRAPH_CHECK(igraph_i_community_leiden(graph, i_edge_weights, i_node_weights,
+                                               resolution_parameter, beta,
+                                               membership, nb_clusters, quality));
+        if (n_iterations >= 0) {
+            itr++;
+            if (itr >= n_iterations)
+                stop = true;
+        }
+        else if (prev_quality == *quality)
+            stop = true;
```
szhorvat, line 1062:
```suggestion
        else if (prev_quality == *quality) {
            stop = true;
        }
```

szhorvat:
Question: Is the iteration always guaranteed to stop? Could the quality ever be computed as NaN, which might cause problems?

interfaces/functions.yaml
```diff
@@ -1601,7 +1601,8 @@ igraph_community_leiden:
     PARAMS: |-
         GRAPH graph, OPTIONAL EDGEWEIGHTS weights,
         OPTIONAL VERTEXWEIGHTS vertex_weights,
-        REAL resolution, REAL beta, BOOLEAN start, OPTIONAL INOUT VECTOR_INT membership,
+        REAL resolution, REAL beta, BOOLEAN start, INTEGER n_iterations,
```
szhorvat, line 1604: Should `n_interations` have a default value here?

For `beta`, the docs recommend 0.01. Should it have a default value here?

szhorvat:
Ah, just one more thing: don't forget the changelog. I mentioned everything I noticed, it can go ahead as far as I'm concerned.

src/community/leiden.c
vtraag: This is intentionally past tense: it is only the last iteration that didn't change. A next iteration might still change. Perhaps I can better rephrase this to "improved" though.

src/community/leiden.c
vtraag: Yes, that is correct. Each iteration might find some possible improvement of the quality function. It is possible that a single iteration didn't find an improvement, but that the next iteration finds an improvement again. This is indeed because the algorithm is randomised. Each iteration considers (possibly) slightly different subsets of nodes to consider for moving around to other communities in order to improve the quality function. 

I will clarify the comment.

interfaces/functions.yaml
vtraag: Yes, good point. The recommended default is actually 2 iterations. Not sure if I applied that recommendation currently in the higher level interfaces.

vtraag:
> Question: Is the iteration always guaranteed to stop? Could the quality ever be computed as NaN, which might cause problems?

Good question! In principle, yes, the iteration is always guaranteed to stop, but good to double check how this works with possible NaN results.

src/community/leiden.c
szhorvat: "It continues iterating until the quality stops changing" then? Can a simple past ("changed") follow a future tense ("will continue")?

vtraag:
I've now refactored the implementation, so that it does not rely on the actual quality being calculated, but by simply tracking whether there was an actual changed.

This also solves a bug with the optional quality argument, which otherwise would have required me to create an additional internal variable. Additionally, it also would have required to always calculate the actual quality, which is now no longer the case.

szhorvat:
It also solves the potential issue with NaN quality.

However, it is now not quite as obvious that the iteration is guaranteed to stop. Quality is bounded from above, so if we rely on it, eventually it will stop growing. But can the iteration start cycling over multiple same-quality partitionings? I guess that technically stopping is "guaranteed" in the sense that the algorithm is randomized, so there is some probability that no change will occur. But are you sure that this probability is not small?

vtraag:
> However, it is now not quite as obvious that the iteration is guaranteed to stop. Quality is bounded from above, so if we rely on it, eventually it will stop growing. But can the iteration start cycling over multiple same-quality partitionings? I guess that technically stopping is "guaranteed" in the sense that the algorithm is randomized, so there is some probability that no change will occur. But are you sure that this probability is not small?

No, only strictly improving moves are made. This means that it is not possible for the algorithm to keep on changing the partition in equally good ways. Hence, it is guaranteed to stop.

szhorvat:
> No, only strictly improving moves are made.

This would actually be useful to mention not only in a comment, but directly in the docs.

vtraag:
> This would actually be useful to mention not only in a comment, but directly in the docs.

Sure, let me state that more clearly.

> Any reason that you didn't set defaults for gamma and beta in functions.yaml?

Nah, not really. For beta it's easy to set a default, let me do that. 

For gamma it's more difficult, see https://github.com/igraph/rigraph/issues/529. FWIW, the discussed solution cannot be implemented in C directly, because the "objective function" choice isn't made directly in the C core. Following your reasoning, you might also want to include that in the C core, but I'm not sure what your preferences are.

szhorvat (approved):

PR #2178, opened 2022-08-31
Grid parameters for umap fit_ab function, it now converges correctly

iosonofabio:
The *experimental* support for UMAP has a function that computes the smoothing parameters for the low-dimensional distance metric.

Finding the parameters is a Newton search we implemented ourselves using a grid of points. Turns out that the actual resolution of the grid matters to ensure convergence. I now use the same grid that the original UMAP implementation uses and indeed converge onto the same numbers over a wide range of input values (`min_dist`).

Starting guesses for `a` and `b` are also improved to lie close to the final solution.

We could add a test function for this, but it's not part of the API. Do we have a way to test internal functions in unittests?


szhorvat:
> Do we have a way to test internal functions in unittests?

See `igraph_i_layout_sphere_2d()` for an example.

Add `IGRAPH_PRIVATE_EXPORT` to the function, and include the private header in the test.

It's not ideal to have to access private functions like this, so it's best to keep it to a minimum. But this is how it's done now.

szhorvat:
I won't have time to read up on UMAP in enough detail to review this, but I'm sure you and Daniel know what you're doing 🙂

iosonofabio:
Thank you @szhorvat . What do you recommend in this case? I've tested the function by hand on all relevant cases and it works, so if we are trying to keep those functions to a minimum I should perhaps just not test it?

szhorvat:
Use your judgement. If you think that there's value in having this test in the test suite for the future, just add it.

szhorvat:
Important: Make sure the header is also included in the implementation file, i.e. also in the file where the function is defined, not just where the function is used.

tests/unit/igraph_i_umap_fit_ab.c
```diff
@@ -31,10 +31,10 @@
 
 int main () {
     int i;
```
szhorvat, line 33: Can you make this a `size_t` instead of `int`?

There are still some `int`s in tests/examples that should really be `igraph_integer_t`. Here it's not a problem to have an `int` but the fewer we have the least distraction there will be when looking for inappropriate uses. `size_t` is appropriate for indexing into raw arrays, like here, and won't cause noise when `grep`ping for `int`s.

tests/unit/igraph_i_umap_fit_ab.c
```diff
@@ -0,0 +1,53 @@
...
+    igraph_real_t a, b;
+
+    igraph_rng_seed(igraph_rng_default(), 42); /* make tests deterministic */
+    igraph_real_t min_dists[7] = {0.001, 0.003, 0.01, 0.03, 0.1, 0.3, 1.0};
+
+    RNG_BEGIN();
+
+    /* test with various typical min_dist values. Originally there is a scaling sigma
+     * factor, but it's 1.0 in all default cases so we fix it for now */
+    for (i = 0; i < 7; i++) {
```
szhorvat, line 43: You can get the size as `sizeof(min_dists) / sizeof(min_dists[0])`. Then you don't repeat `7` in two places.

tests/unit/igraph_i_umap_fit_ab.c
```diff
@@ -0,0 +1,53 @@
...
+        igraph_i_umap_fit_ab(min_dists[i], &a, &b);
+        printf("%f, %.1f, %.1f\n", min_dists[i], a, b);
```
szhorvat, line 45: Generally, prefer `%g` instead of `%f` (unless you have a reason for `%f`). It won't print unneeded digits.

iosonofabio:
thanks @szhorvat - I made those changes and added one more test case for `min_dist=0`. If this passes CI I'm going to merge.

@ntamas is this going to make it into 0.10?

ntamas:
Yes, it can get into 0.10; let me go through the PR real quick.

ntamas:
Seems fine to me - feel free to merge. I'm going through the open issues for 0.10 now to see if there's anything to fix after 0.10-rc2. If there isn't, I'd like to make an official 0.10 release tomorrow.

PR #2181, opened 2022-09-02
Simple Cycle Search

### LINKED ISSUE #1398: Find all simple cycles
This is a feature request for functionality to list all simple cycles, i.e. cycles with no repeating vertices.

GenieTim:
This could one day fix #1398, but still requires lots of work.

As of writing this (after first commit), I guess Johnson's algorithm is sketched out, but the results are never returned as I have not yet agreed on what structure I want to use to persist the search state between different calls to the function.
Various other todos are noted in the code.


szhorvat:
`develop` is already merged back into `master`, so you should rebase onto `master` now.

ntamas:
There are quite a few unrelated changes in this PR that should ideally be reverted in order to make it easier to review later. These unrelated changes can roughly be grouped as follows:

* Code reformatting probably done by your IDE, like putting braces on separate lines. These should simply be reverted.
* Replacing `()` with `(void)` in C++ files or headers that are solely included in C++ files; these are not needed and should also be reverted.
* Changes in `vendor/f2c`; here you also replaced `()` with `(void)` but in places that are not used (they are guarded by an `#ifdef`).

Can you revert these so we can see better what the relevant changes are?

GenieTim:
Yes, sure, thanks for the feedback, will do. Currently, there is lots other things wrong in my code that I might focus on first; I suppose you will want to squash merge this PR in the end anyway, if at all

ntamas:
Also, what will you need `igraph_vector_pop_front()` for? If you want to be able to pop items off the front or the back of a vector, you can probably use `igraph_dqueue_t`; `igraph_dqueue_pop()` will pop from the front and `igraph_dqueue_pop_back()` will pop from the back.

GenieTim:
> If you want to be able to pop items off the front or the back of a vector, you can probably use igraph_dqueue_t; igraph_dqueue_pop() will pop from the front and igraph_dqueue_pop_back() will pop from the back.

Yes, I actually have a comment where I use the `igraph_vector_pop_front` reminding me to replace it with dqueue in the end. It was just easier to start with the adjlist which has already all the dimensionality handled for me for now.

GenieTim:
For the two simple cases of the directed star and ring, the algorithm succeeds. I _think_ what causes yet troubles in more complicated/undirected systems is to fulfill the assumption of the adjlist to be ordered and "directed" from smaller to larger indices (i.e., the simple adjlist is not yet a "adjacency structure of strong components with least vertex in subgraph induced by ...").

GenieTim:
The issue seems to have been the directness of the graph and its consequences on the adjacency list. My tests pass, and I would consider them complex enough to assume the algorithm works. 

I suppose the remaining todos are (a) finding out whether there is a more efficient way to do the undirected graphs, and (b) adjust the implementation to follow the igraph guidelines (i.e., IGRAPH_CHECK etc.), and finally (c) implement any adjustments your reviews will ask for.

GenieTim:
I cannot reproduce the "Non-matching number of IGRAPH_FINALLY and IGRAPH_FINALLY_CLEAN." locally. 
Any tips, dear @szhorvat ?

szhorvat:
This error is only issued if the `IGRAPH_VERIFY_FINALLY_STACK` was set to `ON` when configuring igraph with CMake. This check is not on by default, not even in Debug mode, because it has a severe performance cost. It is enabled for the CI tests though.

It means that when a function returns, the size of the "finally stack" is not the same as when the function was called. This stack is used to keep track of initializations (`IGRAPH_FINALLY`) and is used to run destructors on error. When destructors are run normally (no error), the corresponding stack element is removed using `IGRAPH_FINALLY_CLEAN()`. There must be a mismatch between the number of `IGRAPH_FINALLY` calls the `IGRAPH_FINALLY_CLEAN()`.

That's a short explanation of where this might come from. I'll take a look at the code to see if I can find the specific cause at a later time.

src/cycles/simple_cycles.c
```diff
@@ -0,0 +1,314 @@
...
+    IGRAPH_FINALLY(igraph_vector_bool_destroy, &state->blocked);
+    IGRAPH_CHECK(igraph_adjlist_init(graph, &state->AK, IGRAPH_OUT, IGRAPH_LOOPS_ONCE, IGRAPH_MULTIPLE)); // TODO: understand what we actually want to include
+    IGRAPH_FINALLY(igraph_adjlist_destroy, &state->AK);
+    igraph_adjlist_sort(&state->AK);
+    state->directed = igraph_is_directed(graph);
+    IGRAPH_CHECK(igraph_adjlist_init_empty(&state->B, N));
+    IGRAPH_FINALLY(igraph_adjlist_destroy, &state->B);
+
+    return IGRAPH_SUCCESS;
+}
```
szhorvat, line 199: There appears to be a problem here. When a constructor like this returns, it must clean up the finally stack. Then, at the point where the constructor is called, the `igraph_simple_cycle_search_state_t` structure must be placed on the finally stack with its destructor.

GenieTim:
Thank you for these useful insights and pointing to the right CMake variable.

src/cycles/simple_cycles.c
```diff
@@ -0,0 +1,316 @@
...
+                // copy output: from stack to vector
+                igraph_vector_int_t res;
+                IGRAPH_CHECK(igraph_vector_int_init_copy(&res, &state->stack));
+                IGRAPH_FINALLY(igraph_vector_int_destroy, &res);
+                IGRAPH_CHECK(igraph_vector_int_reverse(&res));
+                // undirected graphs lead to every cycle being found twice.
+                // this is our naïve filter for now
+                igraph_bool_t persist_result = true;
+                if (!state->directed && search_mode == IGRAPH_UNDIRECTED_CYCLE_SEARCH_ONE) {
+                    igraph_vector_int_sort(&res);
```
ntamas, line 123: This line here kinda bothers me because it means that for undirected graphs with `search_mode == IGRAPH_UNDIRECTED_CYCLE_SEARCH_ONE`, the algorithm will return the vertices in the cycle ordered by their vertex IDs, not in the order they are to be found in the cycle -- but for all other cases (other search modes, directed graphs) it will return the vertices in order.

I'll probably refactor this part of the code to something that does not involve sorting the vector.

GenieTim: I agree, it is a bad solution. Ideally, there would be an easy way to compute a hash for each result, and just check whether that hash was found already. 

The straight-forward way to circumvent the problem you mention would be to copy the vector and keep both the sorted and unsorted versions, but given that this algorithm might lead to memory issues for large/lattice like graphs anyway, I would be somewhat cautios with that.

szhorvat:
I am wondering:

 - Does the code handle graphs with multi-edges?
 - How many cycles are returned for this graph? One or two? There are two cycles, but they are made up of the same vertices.

    image

 - What about the undirected version of the above?

    image

    This graph now has not two, but three cycles.

 - Is it possible to return cycles in terms of edge IDs instead of vertex IDs?
 - Are self-loops handled?  E.g. the graph `1 -> 1` would have one cycle of size 1.

After a cursory look, I can see that Johnson discusses cycles in terms of vertices, but actually cycles are made up of edges, not vertices, and many practical applications _require_ this view.  Consider the following structure:

image

You can imagine this to be made of write, in which electric current might flow, limited by the natural resistance of wires. So it's a real physical problem.

There are clearly three cycles.

We may formalize this as a graph in multiple ways. We may choose vertices to be branching points:

image

Then this is a multigraph and all three cycles contain only the two vertices A and B. Writing them in terms of vertices is not useful as they are not distinguishable this way.

Or we may choose to add more vertices along the wires:

image

In this case the three cycles are ACB, ABD and ACBD.

Written this way they are distinguishable, but if I want to make calculations on this electric circuit, I still need to have a reference to the edges, not the vertices, as it's the edges (wires) that carry resistance. If we store this graph on a computer, we would assign resistance values as edge attributes.

This example is meant to illustrate that:

 - Many (most?) practical applications _require_ multigraph support. There is no one correct way to choose where to put the vertices on the structure above. When working with such data one will inevitably end up with multigraphs.
 - Once we have multigraphs, it only makes sense to talk about cycles in terms of edges.
 - Even without multigraphs, most applications of cycles or paths require edges, not vertices. Think of shortest paths: the path length is the sum of edge weights along the path. The cycle length is the same. Any weighted application requires treating cycles in terms of edges.

I am wondering how difficult it would be for us to support these use cases.

src/cycles/simple_cycles.c
szhorvat: Just noting that we have a function that orders edges (not vertices!) in cycle order https://github.com/igraph/igraph/blob/master/src/misc/order_cycle.cpp#L34

szhorvat:
As for how to proceed with this PR:

@GenieTim, now you have a good understanding of Johnson's algorithm. In your opinion, does it generalize to producing cycles in terms of edges?

If yes, and if this is quite straightforward to implement, it would be great to do it.  

If this is not straightforward, then this set of functions should remain as it is and return cycles in terms of vertices. However, it's still important to understand how it handles (1) multi-edges and (2) self-loops and ensure that the behaviour is reasonable. If it can't handle such graphs, the function should reject them.  We should also make sure that the naming of the function is such that it leaves room for introducing an edge-based alternative in the future.

Either way, the documentation of all functions in this PR should include `\experimental` before the first paragraph.

GenieTim:
With these tests, I can answer some of the questions asked:

> Does the code handle graphs with multi-edges?

I guess, in the directed case clearly yes, in the undirected one not so much, there, some more investigation is needed.

> How many cycles are returned for this graph? One or two? There are two cycles, but they are made up of the same vertices.

Two are returned.

> What about the undirected version of the above?

That's where the problems start. With some luck, it is just the igraph_adjlist_init() call that needs to be adjusted, but I will need to invest some time for furhter investigations. In any case, the check to make sure that we do not find one cycle twice, the way it is implemented now, clearly would prevent the (in vertex terms) double cycle from being found.

> Are self-loops handled? E.g. the graph 1 -> 1 would have one cycle of size 1.

Yes, they are.

> Is it possible to return cycles in terms of edge IDs instead of vertex IDs?

I have not yet found a function in igraph which would reliably provide the desired mapping between the entry in the adjacency matrix and its corresponding edge id. It is possible that the incidence list is compliant, but I need to check the code, though possibly you can also tell me whether if it is, it would stay compliant or if this may break some time and I should not rely on it.

GenieTim:
> provide the desired mapping between the entry in the adjacency matrix and its corresponding edge id

In my current understanding, the trick is to go the other way round, use `igraph_inclist_init` and then `igraph_adjlist_init_from_inclist` for the adjlist. This should make things work for the directed case. I will try to implement this, at best sometime later this week.

Still thinking about better approaches to undirected graphs.

GenieTim:
While I do support querying the edges now, I had to realise that the behaviour of the undirected graphs is even worse this way, as it is not even the case that every cycle is detected exactly twice (i.e., when using `IGRAPH_UNDIRECTED_CYCLE_SEARCH_BOTH`, certain cycles are still only found once). 

I do not see a straight-forward way to get said parameter to behave consistently and would therefore suggest to get rid of it, and always filter duplicates.

GenieTim:
To filter duplicates in the undirected case, I would anticipate the most performant way to be hashing the cycles (with a combination of sum, product and xor of its edges) and then deciding whether to keep it based on it being present in such a hashmap or not.

@szhorvat or @ntamas : I did not find any hashmap kind of structure implemented in igraph. In C++, one could use e.g. std::map. For C-only code, there are already existing implementations out there, e.g. https://github.com/tidwall/hashmap.c ... what is your policy, should I include such an external library or come up with my own, or did I just not see an existing implementation in igraph?

GenieTim:
It's a bit sad that this PR is stale. As I see it, the main reason not to merge this is the potentially unclear behavior of undirected graphs.
Do you have a different view, are there more things to clear up? 
Is this PR still salvageable, or am I falling for the sunken cost fallacy?

szhorvat:
This work will definitely be merged in some form, in the worst case as restricted to simple directed graphs, but hopefully in a more general version 

I'm travelling now, and will be busy for at least the next 2 weeks, but we'll get back to this PR eventually.

Apologies for the delay

GenieTim:
I think to have solved the first issue you mention, the edge id vector is (in my tests) always the length of the vertex ids.

GenieTim:
In general, for large lattice-like graphs, the number of cycles might be gigantic and could potentially lead to memory issues, not just stack overflow – we might want to introduce a `max_cycle_length` parameter or similar?

szhorvat:
@GenieTim Would you have time to come back to this PR and see what we need to do to bring it into a form that's not perfect, but good for merging?  As I recall, the biggest difficulty was handling undirected multigraphs.  Is this right?  If so, we can restrict undirected graphs to simple graphs for the moment. 

There seem to be some failing tests, which need to be cleaned up.

> we might want to introduce a `max_cycle_length` parameter or similar?

This absolutely makes sense. We don't need to complete this before merging, as the function will be marked as "experimental" anyway.

GenieTim:
@szhorvat sure, I will try. If I remember correctly, all types of graphs are handled quite fine currently (well, apart from slight performance differences and the issue with the stack overflow possibility).

The failing tests are two: one that I wrote in order to understand an issue I had with `igraph_delete_edges`. While I don't remember whether I figured out the issue at some point or not, it is clear that it has nothing to do in this PR anyway, so I will remove it.
The other seems to be a stack overflow, but I cannot reproduce it locally currently. Will see whether I have a machine somewhere that also encounters the stack overflow.

szhorvat:
Thanks so much for working on this, and apologies that this took so long! Let try to push it through. I'm re-targeting this for igraph 1.0, which is coming out this summer, but we'll still keep the functionality "experimental". This way we won't have too much pressure to make everything perfect, we'll have the opportunity to revise the functionality later.

Let us know if you'd like us to look into anything specific (such as the stack overflow) right away.

At some point we can also have a Zoom call if this helps sort out the remaining issues more efficiently.

szhorvat:
I just ran tests locally, and saw no failure.

szhorvat:
@GenieTim Can you please merge `master` into this branch again (the Update branch button in this PR should work) so that CI tests can start running? I don't want to step on your toes so didn't do it myself.

GenieTim:
How do you run the test independently, @szhorvat ? Both `./build/tests/test_igraph_simple_cycles` and `ctest -R ".*simple_cycle.*"` pass on the device I have been on so far.

szhorvat:
Sorry, to be clear, I had AddressSanitizer enabled when I ran it, which increases stack use, and helps detect issues.

GenieTim:
Or, let's do it this way, my own example: 

```c
static igraph_error_t igraph_i_simple_cycles_unblock(
    igraph_simple_cycle_search_state_t *state, igraph_integer_t u
) {
    igraph_vector_int_t* neis;
    igraph_integer_t w;
    igraph_stack_int_t u_stack;
    igraph_stack_int_init(&u_stack, 0);
    igraph_stack_int_push(&u_stack, u);

    while(igraph_stack_int_size(&u_stack) > 0) {
        igraph_integer_t current_u = igraph_stack_int_top(&u_stack);
        VECTOR(state->blocked)[current_u] = false;

        neis = igraph_adjlist_get(&state->B, current_u);
        bool recurse_deeper = false;
        while (!igraph_vector_int_empty(neis) && !recurse_deeper) {
            w = igraph_vector_int_pop_back(neis);
            if (VECTOR(state->blocked)[w]) {
                igraph_stack_int_push(&u_stack, w);
                recurse_deeper = true;
            }
        }

        if (!recurse_deeper) {
            igraph_stack_int_pop(&u_stack);
        }
    }

    return IGRAPH_SUCCESS;
}
```

Quite inefficient due to the repeated iteration of the same neighbours (might need a stack of igraph_vector_int, right?), but otherwise, this is the way, right?

szhorvat:
We're having a holiday here, but I'll get back to you Friday or in a couple of days.

GenieTim:
@szhorvat I am happy to report that the code is now non-recursive and seems to pass (at least) all automatically approved tests.

If you merge, when you merge, please squash my commits together, I was a bit careless and simplified me working on different OSs by commiting more than just successes.

Currently, most of the variable names are the ones from the Johnson paper, and they use a completely different naming scheme than the ones I introduced. I believe it makes sense since that simplifies comparing the code to the one proposed by Johnson, but I would understand if you ask me to rename those variables to names that are a bit more explanatory.

szhorvat:
Sorry about being slow with the feedback about how to go from the recursive to the non-recursive version. Thank you so much for making this work! I'll start slowly testing it, and I hope we can release it with the next version.

Are you find with being credited using both your real name and GitHub handle?

GenieTim:
All good. Yes, sure, you may credit me hoewever you are used to doing it, real name, handle, all good

GenieTim:
Any news, @szhorvat , anything I can help with (does the documentation need more work, rename variables to be descriptive rather than follow the Johnson paper, or anthing like that)?

szhorvat:
I need to integrate this into the Mathematica interface of igraph so I can do some more empirical testing.

szhorvat:
I'm sorry, I have not forgotten about this, but some things came up. I am hoping to get back to it between roughly Aug 10-15.  This is the top priority PR to merge, and must be merged before version 1.0, tentatively scheduled for the end of September.

szhorvat:
Could you please let me know if the code can handle length-2 cycles in undirected graphs such as this (one cycle):

image

Or this (three cycles):

image

If we can't make this work, I would document it as a temporary limitation for now, and not block merging on this. But in the long term we will need to sort these out. The reason is that this is equivalent to a graph like this once the degree-2 vertices are "smoothened" out:

image

szhorvat:
Also, what about self-loops? These should be found as well. Luckily these are easy to handle as a special case, but I'm not sure if it's easy to include them with the iterator design like here. For example, this graph should ideally result in two cycles:

image

Namely, `1-1` and `1-2-1`.

GenieTim:
Dear @szhorvat 

Thanks for pointing those out to me, I totally forgot about that.
I have now implemented a fix, as well as the three tests you requested, which seem to pass now on my device.

Best regards,

szhorvat:
I did a very brief test, but I did not check the code yet.

The first potential issue is that if I pass `NULL` to the 2nd parameter of `igraph_simple_cycles_search_all()` but an initialized vector list to the 3rd, the problem crashes with a `!= NULL` assertion failure. Is this something you expect? Again, I didn't have time to look.

I went ahead by passing a vector list both to the 2nd and 3rd parameter, even though I only wanted the edges, not the vertices.

I did some fuzzing, and found this graph where the "outer" cycle is not returned:

image

There are 6 cycles:

 - 3 of length 3, this is fine
 - 2 of length 4, this is fine
 - 1 of length 5, this is not returned

Could you please have a look at this?

szhorvat:
This is another example:

image

There should be 4 cycles of length 5, but it only gives me 3 of them, one is missing.

As I try larger and denser graphs, more and more cycles are missing.

szhorvat:
I did a bit of reformatting with astyle and removed some headers that didn't belong.

src/cycles/simple_cycles.c
```diff
@@ -0,0 +1,506 @@
...
+ *
+ */
+
+static igraph_error_t igraph_i_simple_cycles_unblock(
+    igraph_simple_cycle_search_state_t *state, igraph_integer_t u) {
+    // TODO: introduce stack for w & neis in order to reduce the number of iterations.
+    igraph_vector_int_t *neis;
+    igraph_integer_t w;
+    igraph_stack_int_t u_stack;
+    igraph_stack_int_init(&u_stack, 0);
```
szhorvat, line 138: Please use `IGRAPH_STACK_INT_INIT_FINALLY` here and add the corresponding `FINALLY_CLEAN` after destroying the stack.

There might be other places where this is missing, I won't leave additional comments now.

src/cycles/simple_cycles.c
```diff
@@ -0,0 +1,506 @@
...
+    igraph_stack_int_push(&u_stack, u);
```
szhorvat, line 139:
```suggestion
    IGRAPH_CHECK(igraph_stack_int_push(&u_stack, u));
```

src/cycles/simple_cycles.c
```diff
@@ -0,0 +1,506 @@
...
+ *
+ * \ref igraph_simple_cycle_search_state_destroy
+ */
+igraph_error_t igraph_simple_cycle_search_state_init(
+    igraph_simple_cycle_search_state_t *state, const igraph_t *graph) {
+    state->N = igraph_vcount(graph);
+
+    IGRAPH_CHECK(igraph_vector_int_init(&state->vertex_stack, 0));
+    IGRAPH_CHECK(igraph_vector_int_reserve(&state->vertex_stack, 8));
+    IGRAPH_FINALLY(igraph_vector_int_destroy, &state->vertex_stack);
```
szhorvat, line 369: The pattern here is slight incorrect. The `IGRAPH_FINALLY` should come directly after the `init()`, and before the `reserve()`. We can combine the `init` and `FINALLY` into a single `IGRAPH_VECTOR_INT_INIT_FINALLY`.

```suggestion
    IGRAPH_VECTOR_INT_INIT_FINALLY(&state->vertex_stack, 0);
    IGRAPH_CHECK(igraph_vector_int_reserve(&state->vertex_stack, 8));
```

Before merging these will need to be cleaned up.

szhorvat (commented):
I pointed out some minor things that caught my eye, and which will need to be resolved before merging. These are not the first priority though. It's better to make sure that the result is correct and not stress about error checking and memory management until after that.

GenieTim:
Thank you for your excellent answers and tests. First investigations indicate that the problem is actually that the hashes collide, and that that's the reason why some cycles in the undirected case are not detected. Would have been too easy I suppose if my extension would have just worked. I will see what I can come up with.

GenieTim:
I have found the reason for the hash coliisions and eliminated it. The two examples (I call them "envelope" and "boat") are implemented as tests and pass on my device.

I will continue with the cleanup of how to do things the igraph internal way, as you suggest and correctly point out, later these days, @szhorvat , and am further grateful for any other examples you find that do not behave as expected.

szhorvat:
I quickly re-ran my fuzzing code, and still found missing cycles for these 5-vertex graphs:

image

And these 6-vertex graphs:

image

Some of these are the same as before. I wanted to check the new tests you added, to see what's going wrong, but won't get to that for a while.

The vertices are labelled with 1-based indexing here, so translating these to C requires subtracting one from each vertex index.

GenieTim:
Would you mind adding a number to those so I know how many cycles you are expecting to see?

szhorvat:
For the size 5 graphs, the first one has 3 of length 3, 2 of length 4, 1 of length 5.

The second has 1 of length 3, 1 of length 4, 1 of length 4.

I'm about to travel now, I can add detailed numbers in a few days.

GenieTim:
Ok, I seem to have found the culprit, interestingly a code line that I once had (see e.g. here: https://github.com/igraph/igraph/pull/2181#issuecomment-2075417276), but I guess it's not worth it to investigate when or where it got lost.

szhorvat:
I ran the tests again, and did not find problems with 5 and 6 vertex graphs, but cycles start going missing at 7 vertices. This is one of the smaller examples, with only a single cycle missing, highlighted in red:

image

Again, the labelling is one-based, but here's the zero-based edge list:

```
{{0, 1}, {0, 4}, {0, 5}, {0, 6}, {1, 2}, {1, 3}, {1, 5}, {1, 6}, {2, 3}, {2, 6}, {3, 4}, {3, 5}, {3, 6}, {4, 5}}
```

Another example:

image

```
{{0, 1}, {0, 4}, {0, 5}, {0, 6}, {1, 2}, {1, 3}, {1, 5}, {2, 3}, {2, 4}, {2, 6}, {3, 4}, {3, 5}, {3, 6}, {4, 5}, {5, 6}}
```

Another:

image

```
{{0, 1}, {0, 2}, {0, 3}, {0, 4}, {0, 6}, {1, 2}, {1, 3}, {1, 4}, {1, 5}, {1, 6}, {2, 3}, {2, 4}, {2, 6}, {3, 4}, {3, 5}, {4, 6}, {5, 6}}
```

It's not always a single cycle that's missing, these are just the simpler examples I could generate.

Do you think it's a hash collision again?

szhorvat:
I forgot to add cycle size distributions. For the last one it is: `<|3 -> 18, 4 -> 43, 5 -> 78, 6 -> 96, 7 -> 60|>`, i.e. 18 cycles of size 3, etc. As we go to larger graphs, there will be a very large number of cycles.

GenieTim:
Thank you, I will check. If it is indeed a hash collision, I will simply get rid of the whole hashing, without a hash-map type the performance gain is negligible anyway

GenieTim:
Ok, the subset of your examples that I implemented as tests passes, now that I stopped using explicit hashes.

GenieTim:
If I understand correctly, @szhorvat , the tests that fail now fail because of the runtime of the test, is that possible? Should I split the test into an undirected and directed one, or remove the tests of the larger structures?

szhorvat:
It seems that most tests fail because of a memory leak. You can see this here:

https://dev.azure.com/igraph-team/igraph/_build/results?buildId=7896&view=logs&j=7343ad18-517e-5e8c-b935-0dc003a2c137&t=7f4c1165-ffb0-5515-a744-9dbde53a15cc&l=902

The Windows/MSVC tests fail for some other reason. There's a linking error and it's unclear to me what is going on. I'll have a look at that later.

szhorvat:
The MSVC failure is because of the explicit use of `__CHAR_BIT__`. `CHAR_BIT` would work, but if you're looking for the number of bits in an `igraph_integer_t`, please just use `IGRAPH_INTEGER_SIZE`.

GenieTim:
Thank you so much, @szhorvat , and sorry for missing that.

Now just one job is failing, and I somhow don't see any output in the job that could help me figure out what's going wrong...?!?

szhorvat:
The Travis failure is a fluke, don't worry about it.

szhorvat:
Thanks for all your hard work on this! I just wanted to let you know that I'm working on the review, and so far I haven't found an issue by doing automated checks on undirected/directed simple graphs, and manual checks on undirected/directed multigraphs.

So far I've only tested `igraph_simple_cycles_search_all()`.

I've been doing comparisons with Mathematica, which only supports simple graphs. Performance is considerably worse than Mathematica at the moment, but not unusably so. I don't want to worry about performance for the moment. Let's get it correct first, and we can look at speedups later, perhaps after merging the initial version. 

Do you mind if I push small changes to this PR (such as missing error checks) that are easier to just do myself than to ask you?

szhorvat:
Please note the very last commit I made. Adjacency lists are already sorted. We can make use of this for as long as we don't add new vertices to them. If the code ever changes to add new vertices to the adjlist, this won't be valid anymore.

GenieTim:
Thank you very much, @szhorvat . Is the performance much worse for the directed case as well? 
I would expect the directed case to be much more efficient compared to the undirected one. 

Because, in the undirected case, I believe there to be a more efficient way to prevent the duplicates. In fact, I would expect there to be an "analytic" one (in the sense that there is no dependence on previously found cycles, just something compareable to having an extra constraint, such as [incorrect example] that the id of the second vertex in the loop is higher than the first), but have not found it so far. Will try again.

szhorvat:
In the directed case, the performance is slightly better.

In the undirected case, the bottleneck is the duplicate removal.

Would the following approach work for preventing duplicates? In the directed case we start with a vertex $S$, and find simple paths that starting from $S$ that also end at $S$.  In the undirected case, this approach gives every cycle twice, in both directions. How about starting with an edge $T - S$, and finding simple paths starting in $S$ that ends not at $S$, but $T$? We would never move from $S$ to $T$, as that's the wrong direction along the cycle. [NetworkX](https://networkx.org/documentation/stable/_modules/networkx/algorithms/cycles.html#simple_cycles) seems to do something like that.

GenieTim:
Your input is great; blocking edges could work.

I could imagine in the directed case that the choice of origins to check for cycles is too eager and a better choice could lead to performance improvements, e.g. by doing clustering and in each cluster, only search for loops from the vertex 0 as well as all vertices with a degree > 2.

How do you see the future of this PR, given that the current state works: should I continue to improve it here, e.g. with the things described above, and an argument to specify the maximum length of the cycles, or is it better to leave this PR here and continue in a new one with additions that then have to be tested and reviewed again?

szhorvat:
Let's try to handle the undirected case (i.e. efficient duplicate avoidance) properly before merging.

szhorvat:
Shall I give this another spin?

GenieTim:
Yes, please, if you may

szhorvat:
I only managed to do a little bit of testing, but so far it looks both correct and performant. :-)

szhorvat:
It also survived some libfuzzer-based stress-testing, in the sense that it did not crash.

src/cycles/simple_cycles.c
```diff
@@ -0,0 +1,459 @@
...
+
+    IGRAPH_VECTOR_INT_INIT_FINALLY(&state->vertex_stack, 0);
+    IGRAPH_CHECK(igraph_vector_int_reserve(&state->vertex_stack, 8));
+    IGRAPH_VECTOR_INT_INIT_FINALLY(&state->edge_stack, 0);
+    IGRAPH_CHECK(igraph_vector_int_reserve(&state->edge_stack, 8));
+    IGRAPH_CHECK(igraph_vector_bool_init(&state->v_blocked, state->N));
+    IGRAPH_FINALLY(igraph_vector_bool_destroy, &state->v_blocked);
+    IGRAPH_CHECK(igraph_inclist_init(
+                     graph, &state->IK, IGRAPH_OUT,
+                     IGRAPH_LOOPS_ONCE)); // TODO: understand what we actually want to include
```
szhorvat, line 335:
```suggestion
    IGRAPH_CHECK(igraph_inclist_init(
                     graph, &state->IK, IGRAPH_OUT,
                     IGRAPH_LOOPS_ONCE)); // each self-loop counts as a single cycle in directed graphs
```

I think this is correct, and the TODO note can be removed.

src/cycles/simple_cycles.c
```diff
@@ -0,0 +1,459 @@
...
+                                igraph_vector_int_list_t *v_result,
+                                igraph_vector_int_list_t *e_result) {
+
+    igraph_simple_cycle_search_state_t state;
+
+    IGRAPH_CHECK(igraph_simple_cycle_search_state_init(&state, graph));
+    IGRAPH_FINALLY(igraph_simple_cycle_search_state_destroy, &state);
+
+    // TODO: depending on the graph, it is rather unreasonable to search cycles
+    // from each and every node
```
szhorvat, line 447: As I understand, duplicates are already taken care of by Johnson's procedure. So avoiding the search from every vertex is probably a minor benefit only.

If I got this right, we need to search from the lowest-index vertex of each cycle. It is not sufficient to search from at least one vertex in each cycle, but we need to search from specifically the lowest index one. Can you confirm that I got this right @GenieTim ? If yes, it seems difficult to reduce the number of starting vertices _efficiently_ (other than skipping those with degree zero and one), and probably not worth bothering. Let me know if you have other ideas.

If we want to optimize more, we can avoid edges that are not in strongly connected components (directed) or biconnected components (undirected), but that's a different optimization, and one that I don't think we need to jump into right away.

szhorvat:
I haven't been able to go through the entire algorithm, and I took a mostly empirical approach to the review so far. But I think it looks great, functionally, and we can now get the more boring things ready. Here's what I think would be nice to do before merging:

 1. Complete the documentation. I will do this myself.
 2. Finalize the names of public functions. I will do this myself as a last step just before merging, after we agreed on names.
 3. `igraph_simple_cycles_search_all()` should work when passing `NULL` for `v_result`, as it is a common use case to only want edges. It already works when passing `NULL` for `e_result`. Could you please take a look at this?
 4. I'm thinking not to expose `igraph_simple_cycle_search_state_t` (including its init/destroy) and `igraph_simple_cycles_search_from_one_vertex` as public functions at this stage, as they are not in a state that is easy to use safely by end users yet. I will take care of this myself just before merging. We don't need to remove anything from headers. Simply using an `igraph_i_` naming will be enough to indicate that these are not for public use.
 5. Optional: Add interruption support, if easy enough. This may require some care given the cycle search state object. This won't be a blocker for merging, but it's useful to have since there may be a very large number of cycles, so it's easy to accidentally start a search that won't terminate in a reasonable amount of time.
 6. Given the large number of cycles, we should try to support filtering the results:
     - It seems to me that it should be possible to cut off the search at a certain depth, i.e. only return cycles no longer than a cutoff length. @GenieTim, how difficult do you think this feature would be to add? A cutoff length can be a parameter to all public functions.
     - It would be nice to add a callback version of the search function, similar to `igraph_cliques_callback()`, `igraph_motifs_randesu_callback()` and `igraph_get_isomorphisms_vf2_callback()`. Instead of collecting results into a vector list, a user-defined function would be called for each result, as they are found. The callback can request stopping the search by returning `IGRAPH_STOP`. The function that collects results can then be implemented in terms of the callback.  This is a fairly straightforward refactoring, and I can help with this, but if you have time to do it, it would be much appreciated.

Please also let me know how we should credit you for this contribution. Normally we use both the real name and GitHub handle, but if you'd like to skip one of the two, let me know.

Thank you so much for taking on this fairly large project, and for the patience with the slow reviews. We are really close to the finish line now.

src/cycles/simple_cycles.c
```diff
@@ -0,0 +1,459 @@
...
+
+    /* Whether the graph is directed */
+    igraph_bool_t directed;
+
+    /* Hashes of the vertices of found cycles (for filtering in undirected graphs)
+     */
+    igraph_vector_int_list_t found_cycles_vertex_hashes;
+
+    /* Hashes of the edges of found cycles (for filtering in undirected graphs) */
+    igraph_vector_int_list_t found_cycles_edge_hashes;
```
szhorvat, line 76:
```suggestion
```

I believe these can now be removed.

GenieTim:
Thank you so much!

5. Makes sense. I will see whether I can find a good example in the igraph codebase, how this is usually achieved, and implement it.
6.1.: I think adding a cut-off should not be a problem at all. I will try to do that over the weekend.
6.2.: I guess from a user standpoint that makes absolute sense. I anticipate that this too will require some care regarding the search state, but accept your view that it is straight-forward and will therefore attack it as well, assuming I succeed fast enough with the other two objectives.

Best regards,

Tim

___________

szhorvat:
I would suggest starting with 6.1, and doing another round of checks once that's done. The rest is quite straightforward. I can help with 5, let's leave to the end.

GenieTim:
What's not clear to me is how to clean up after an interruption – if I understood other code correctly, there is no cleanup, is that correct?

szhorvat:
Interruption is treated the same as an error, so if an error wouldn't cause trouble at that point, then interruption won't either. I'll take a look at the interruption later, after testing the cycle length limit.

szhorvat:
Don't worry about Travis, the fault is on their side: https://travis-ci.community/t/ppc64-s390x-failing-on-startup/14358 I'm sure they'll fix it.

I'll be on the road for a couple of days and will try to get back to this next week. I will be a bit slow as next week is the first week of teaching.

szhorvat:
This is just a heads up that we are planning a new release early next week. I would like to include this work in it. I will try to refactor the interface a bit this weekend, and expose a subset of the functionality as "experimental", just enough so we can feel confident about including it. We can then continue making improvements in a new PR.

I'll also do some testing to make sure it's producing correct results after the last few changes. Based on the last testing round, the PR is in great shape, so hopefully no new issues come to light :-) If some bugs appear and I can't fix them during the weekend, then let's bump this to the next release, but hopefully that doesn't happen.

Thank you for all the hard work and the patience so far @GenieTim! I don't need any input from you right now (this is a very short notice), just letting you know the plans.

@ntamas, I hope to have this ready by Monday morning.

GenieTim:
Good point, thank you as always @szhorvat for your excellent tests. I have added your example to the tests and fixed the issue in the algorithm to make the test pass.

GenieTim:
Hm. It _could_ be that there is some pathological case where the algorithm fails due to a dependency on the order of the edges and/or vertices, but I currently don't see it. 

Assuming your test still fails while mine passes, I would be thankful for knowing the order in which you add your edges and vertices.

szhorvat:
Sorry about being so slow about this. I just wanted to give a small update and say that something must have gone wrong on my machine last time, as I no longer see problems. I still need to do a bit more thorough testing.

szhorvat:
Unfortunately, the mistake was in the code I used today. There are indeed problems. I am pushing a commit with some test graphs, which I also plot below, along with their cycles.

In the graphs below, both vertices and edges are indexed from 1, not from 0. In C they would be indexed from 0. Just keep this in mind when reading the figures. All expected cycles are highlighted.

image

```
0, 2, 0, 3, 0, 4, 1, 0, 2, 3, 3, 4, 4, 1, 4, 3
```

Graph 2:

image

```
0, 2, 0, 3, 1, 2, 1, 4, 2, 3, 2, 4, 3, 1, 4, 0, 4, 2
```

Graph 3:

image

```
0, 2, 0, 4, 1, 3, 2, 5, 3, 0, 4, 1, 5, 4
```

GenieTim:
Thank you for these additional tests. I am sorry for all the issues caused, I hope it gets harder and harder to find such tests. 
In any case, they should be passing now as well.

szhorvat:
Thank you, great! I can't find any more problems, let's hope I didn't make a mistake again :-)

szhorvat:
Added a `mode` parameter and made sure that non-public functions have `_i_` in the name. Removed the unused function. Tested with function wit libFuzzer.

szhorvat:
Thank you for your hard work, and patience on this @GenieTim! If all goes well, it will be released later this week.

I'd love to have this in the Python interface, to actually get it out to users. If you have time to help with that, let me know, and I ca give you the basic steps. https://github.com/igraph/python-igraph/  Exposing `igraph_simple_cycles()` should be easy, `igraph_simple_cycles_callback()` a bit more work, but that can come later.  I'll look into exposing it in R and finishing the interface (which I used for testing) in Mathematica.

We can continue making improvements, and expose more functions which are now internal, at a later stage. It might also be good to look into the performance on large cycles (I guess this relates to not running the search from all vertices).

GenieTim:
Well, thank you too.

Progress with the Python binding will be handled in https://github.com/igraph/python-igraph/pull/803

ntamas (approved):

PR #2195, opened 2022-09-11
Adjlist cache

### LINKED ISSUE #2125: adjlist_init() should make use of the cache
As I recall, using `IGRAPH_NO_MULTIPLE` in `igraph_(lazy_)adjlist_init()` has a noticeable performance impact on some functions. Do you remember the details @GroteGnoom ?

This function is an excellent candidate for using the cached "multiedges" property. If there are no multi-edges, there is no need to run the multi-edge removal code.

Conversely, since `igraph_adjlist_init()` with `IGRAPH_NO_MULTIPLE` effectively checks for multi-edges, this is a good place to set the cache.

I suggest doing this in 0.10.x for x > 0 after the initial cache implementation has been battle tested.

GroteGnoom:
Fixes #2125, also does loops.

Is there any policy on what should happen if a user first initializes a lazy adjlist, then changes the graph, and then calls igraph_lazy_adjlist_get? Undefined behaviour?


GroteGnoom:
Setting the cache because of igraph_i_simplify_sorted_int_adjacency_vector_in_place
is surprisingly hard. It's just a few `if`s, but very easy to get wrong.

`loops == IGRAPH_LOOPS_TWICE && multiple == IGRAPH_NO_MULTIPLE` is also a strange situation. If you find two loops, but no other multi-edges, should you cache IGRAPH_MULTIPLE to false? Probably not.

 IGRAPH_LOOPS_TWICE is handled like the second loop always exists. I wonder if that's correct:
```
if (mode == IGRAPH_ALL) {
       VECTOR(*v)[p] = VECTOR(*v)[i];
        p++;
}
```

ntamas:
> Is there any policy on what should happen if a user first initializes a lazy adjlist, then changes the graph, and then calls `igraph_lazy_adjlist_get`? Undefined behaviour?

I'd say the behaviour of the lazy adjacency list is undefined if the user changes the graph while querying the lazy adjacency list. Right now we return "old" values for cases when the vertex was queried already before the mutation, and "new" values otherwise, but there are no official guarantees; for instance, we could implement a LRU cache and "forget" old values if we want to keep the memory usage in check.

> Setting the cache because of igraph_i_simplify_sorted_int_adjacency_vector_in_place
is surprisingly hard. It's just a few ifs, but very easy to get wrong.

These cases are _exactly_ why I am in general against any sort of preemptive cache initialization; it is sometimes easy to make mistakes that then lead to hard-to-debug issues later. And it's very hard to test caching-related issues because caching is supposed to be transparent by definition, so the tests are not supposed to know whether they are working with cached values or not. The only reliable way I've found so far to test caching at least to some extent is to perform a query, invalidate the cache, then perform the operation again, and see if the results are identical (they should be).

We have a few caching-related tests in `prop_caching.c`; if you feel like something should be added there (especially in relation to this PR), let me know.

szhorvat:
I would not update the cache from the _lazy_ adjlist/inclist functions. The chance of something going wrong is too great. 

What I meant to say in #2125 was that it would be nice to set the cache from the plain adjlist init function. (I also meant to say that this would benefit future calls of the plain/lazy adjlist init, and I do see how that can be confusing.)

GroteGnoom:
> I would not update the cache from the lazy adjlist/inclist functions. The chance of something going wrong is too great.

Setting the cache in the lazy_adjlist_get_real is only done for the simple cases. Checking that there are _no_ loops is hard, checking for _a_ loop is much simpler, because it doesn't matter if you missed some cases. If at some point a vertex is its own neighbor, then the loops cache is just set to true.

But maybe I should just remove it anyway because it probably also doesn't gain you much speed.

ntamas:
I would also vote for the removal just to keep the code clean; it's unlikely that we will gain much speed by messing around with the cache from here.

ntamas:
Will merge upon CI pass.

tests/unit/adjlist.c
```diff
@@ -272,7 +272,7 @@ int test_multiedge_elimination_for_directed_graph(void) {
     return 0;
 }
 
-int test_caching() {
+int test_caching(void) {
```
szhorvat, line 275: Hm, why did the lack of that `void` not cause a CI failure with the Clang 15 test?

PR #2209, opened 2022-09-27
Fix compilation on Clang 15+ (unused vars, prototype functions)

GenieTim:
MacOS started shipping Clang 15.0.0, which has some stronger defaults regarding detection of set-but-not-used variables as well as prohibiting prototype functions.

This PR fixes these issues on my device, the project compiles again. Tests pass.


szhorvat:
After the release of 0.10 we moved back to working on `master` instead of `develop`. I believe I fixed all Clang 15 issues on `master` already. Can you try and see if there are any remaining problems on `master`?

I'm sorry that you went to all the trouble to make these fixes because of the misunderstanding about `master` vs `develop`. We'll look into how make it clearer what the current active branch is.

GenieTim:
I see, sorry, I guess I should have remembered as this was already a topic in #2181. Sorry, and thanks for the fixes.

Issue #996, opened 2017-05-03
Out-of-bounds in igraph_lcf_vector with 0 nodes


szhorvat:
`igraph_lcf_vector()` may crash due to an out-of-bounds array access when the graph that gets created has zero nodes.

This can be the case when e.g. nodes=0, shifts={} (i.e. empty vector) and repeats=1.  The R interface allows this, but I did not check if it special cases it.

The crash does not always happen, there is a luck element. But the error is there and Address Sanitizer catches it.

The problem is on this line:

https://github.com/igraph/igraph/blob/master/src/structure_generators.c#L1636

    VECTOR(edges)[ptr-1]=0;

because `ptr` never got incremented.

I am simply going to special case this in IGraph/M, but I wanted to notify you anyway.

PR #2210, opened 2022-09-27
moved this file to the tests folder and made a comment about some cod…

Jeroenvh99:
I think this is a unit test because it is tested if the value of res satisfies a few conditions. While looking at the file I found it a bit unclear how the formula in line 61 worked, so I made a comment about that, I would suggest putting it in brackets so it's more clear how it's calculated.


ntamas:
Yes, it seems more like a unit test to me and not an educational example so it belongs to `tests/unit`. Ideally, we should have a replacement in `examples/simple` that simply illustrates how the function should be used on a generic, non-trivial and non-degenerate graph.

I think the parentheses are not necessary where you indicated with the comment; the precedence rules in C are pretty clear here, i.e. 2.0/3.0 is executed first and then it is subtracted from `res`. People who are unfamiliar with this should read up on the operator precedence rules in C. We can probably help a bit by omitting the spaces around `/`, though.

Jeroenvh99:
@ntamas is that why all those tests for this PR failed?

szhorvat:
> @ntamas is that why all those tests for this PR failed?

Yes. You can look at individual CI runs to see what the specific failure is. https://dev.azure.com/igraph-team/igraph/_build/results?buildId=4428&view=logs&j=4c3a8381-b9de-5463-904f-918810769be8&t=257e73c3-f9f2-55cb-d0c8-231f3a50fe79&l=122

Jeroenvh99:
it looks like in igraph_lcf this piece of code:
// [3, -2]^4, n=8
    igraph_lcf(&g, 8, 3, -2, 4, 0);

    if (igraph_ecount(&g) != 16) {
        printf("Failure: [3, -2]^4, n=8\n");
        return 1;
    }

    igraph_destroy(&g);

    // [2, -2]^2, n=2
    igraph_lcf(&g, 2, 2, -2, 2, 0);

    if (igraph_ecount(&g) != 1) {
        printf("Failure: [2, -2]^2, n=2\n");
        return 1;
    }

    igraph_destroy(&g);

    // [2]^2, n=2
    igraph_lcf(&g, 2, 2, 2, 0);

    if (igraph_ecount(&g) != 1) {
        printf("Failure: [2]^2, n=2\n");
        return 1;
    }

    igraph_destroy(&g);

    // Regression test for bug #996
    igraph_lcf(&g, 0, 0);
    if (igraph_vcount(&g) != 0 || igraph_ecount(&g) != 0) {
        printf("Failure: regression test for #996\n");
        return 1;
    }

    igraph_destroy(&g);

is meant to be used like this:
/*/ [3, -2]^4, n=8
    igraph_lcf(&g, 8, 3, -2, 4, 0);

    if (igraph_ecount(&g) != 16) {
        printf("Failure: [3, -2]^4, n=8\n");
        return 1;
    }

    igraph_destroy(&g);*/

    // [2, -2]^2, n=2
    igraph_lcf(&g, 2, 2, -2, 2, 0);

    if (igraph_ecount(&g) != 1) {
        printf("Failure: [2, -2]^2, n=2\n");
        return 1;
    }

    igraph_destroy(&g);

    /*/ [2]^2, n=2
    igraph_lcf(&g, 2, 2, 2, 0);

    if (igraph_ecount(&g) != 1) {
        printf("Failure: [2]^2, n=2\n");
        return 1;
    }

    igraph_destroy(&g);

    // Regression test for bug #996
    igraph_lcf(&g, 0, 0);
    if (igraph_vcount(&g) != 0 || igraph_ecount(&g) != 0) {
        printf("Failure: regression test for #996\n");
        return 1;
    }

    igraph_destroy(&g);*/
to test for specific cases each time, I think this must be this way because you can't use a variable after it's been destroyed I would think. Can somebody tell me if my analysis here is correct?

Jeroenvh99:
The documentation tests still fail, can someone please give me as to where to start looking for a solution?

ntamas:
> I think this must be this way because you can't use a variable after it's been destroyed I would think
It's totally fine to use a graph again after `igraph_destroy()` was called on it as long as the new call is a graph constructor (like here). C is not as strict as C++; `igraph_destroy()` just frees all memory related to the graph, but you can re-allocate new chunks of memory with a graph constructor without problems.

As for the CI failure, the doc build complains because the examples that you have just moved to `tests/unit` were included in the documentation, and the documentation chunks that refer to them can't find the file any more. These failures will remain this way until you implement a replacement example in place of the files that you have just moved to  `tests/unit`. I think that usually a stripped-down version of the unit test that just contains the bare minimum with an illustrative simple graph is usually suitable as an example (although you can make it more educational if you want to).

ntamas:
It's worth moving forward with this PR but [this comment](https://github.com/igraph/igraph/pull/2210#issuecomment-1260085012) needs to be addressed. @Jeroenvh99 are you able to work on this or shall we get this done ourselves?

Jeroenvh99:
@ntamas I don't have time to do this in the coming period, so it's probably best if you get this done yourselves, sorry

PR #2215, opened 2022-09-29
feat: igraph_ecc() computes the edge clustering coefficient

szhorvat:


szhorvat:
I decided to expose the parameters that control normalization and offsetting the cycle count by one.

src/properties/ecc.c
```diff
@@ -0,0 +1,344 @@
...
+        IGRAPH_ALLOW_INTERRUPTION();
+
+        if (v1 == v2) {
+            /* A self-loop isn't, and cannot be part of any triangles. */
+            z = 0.0;
+            s = 0.0;
+        } else {
+            igraph_vector_int_t *a1 = igraph_lazy_adjlist_get(&al, v1);
+            if (! VECTOR(b)[edge]) {
+                IGRAPH_CHECK_OOM(a1, "Not enough memory for edge clustering coefficient.");
```
ntamas, line 156: Shouldn't we check for OOM conditions outside the `if (...) { ... }` block here?

src/properties/ecc.c
```diff
@@ -0,0 +1,344 @@
...
+                igraph_vector_int_sort(a1);
+                VECTOR(b)[edge] = true;
+            }
+
+            igraph_vector_int_t *a2 = igraph_lazy_adjlist_get(&al, v2);
+            if (! VECTOR(b)[edge]) {
+                IGRAPH_CHECK_OOM(a2, "Not enough memory for edge clustering coefficient.");
```
ntamas, line 163: Same question here re `IGRAPH_CHECK_OOM`

src/properties/ecc.c
```diff
@@ -0,0 +1,344 @@
...
+        const igraph_t *graph, igraph_vector_t *res,
+        const igraph_es_t eids, igraph_bool_t offset, igraph_bool_t normalize) {
+
+    igraph_integer_t no_of_nodes = igraph_vcount(graph);
+    igraph_vector_int_t degree;
+    igraph_adjlist_t al;
+    igraph_eit_t eit;
+    igraph_real_t c = offset ? 1.0 : 0.0;
+
+    IGRAPH_CHECK(igraph_adjlist_init(graph, &al, IGRAPH_ALL, IGRAPH_NO_LOOPS, IGRAPH_NO_MULTIPLE));
```
ntamas, line 197: The adjacency list construction here is the same as in `igraph_i_ecc3()`, right? Wouldn't it be better to move it out to a separate static function so we don't need to repeat it twice? (I mean both the construction and the sorting of its elements).

ntamas (commented):
Seems good to me in general but I've added a few comments.

src/properties/ecc.c
szhorvat: When the condition of the `if` block is false, then the neighbour list has already been retrieved, and the `igraph_lazy_adjlist_get()` call cannot fail. It is not necessary to check for OOM in this case.

I'll add comments at the appropriate places.

src/properties/ecc.c
szhorvat: First I'll do a benchmark to see if there is even a point in keeping separate implementations for doing the computation for _all_ edges vs _some_ edges.

szhorvat:
> Finally, when not considering all edges, it would be helpful to get degrees on demand. Would it make sense to get an `igraph_degree_1` function, which gets the degree of a single vertex, similar to how `igraph_neighbors()` get the neighbours of a single vertex only?

Any opinions on this @ntamas ? I mentioned this the other day at a meeting.

src/properties/ecc.c
szhorvat: Let me be clear.

Here I am relying on the fact that if the first call to `igraph_lazy_adjlist_get()` did not fail (with a given argument), then the second one won't fail either. I think this behaviour is reasonable to commit to, but if you disagree, I will move it outside of the `if`.

ntamas: No, that's fine, but it needs an explanation as a comment.

src/properties/ecc.c
szhorvat: I didn't look carefully and I misunderstood. I thought you were asking to share code between the two versions of `igraph_ecc3`, but you are asking to share code between `ecc3_1` and `ecc4`.

I consider the implementations of the `k=3` and `k=4` cases to be entirely separate and mostly unrelated. I worked on them separately and will continue to work on them separately. Unless we embark on implementing this for arbitrarily large `k` (which is problematic, so unlikely) I would rather not share code between them.

szhorvat:
Preliminary benchmark results:

```
|> Benchmark file: /Users/szhorvat/Repos/igraph-main/igraph/tests/benchmarks/igraph_ecc.c
Erdos-Renyi GNM:

|  1 vc=100, ec=5000, k=3, all, 100x                                               0.214s  0.213s      0s
|  2 vc=100, ec=5000, k=3, subset: all, 100x                                       0.221s   0.22s      0s
|  3 vc=100, ec=5000, k=3, subset: 100, 1000x                                      0.248s  0.248s      0s
|  4 vc=100, ec=5000, k=4, subset: all, 100x                                        13.4s   13.4s  0.006s
|  5 vc=100, ec=5000, k=4, subset: 100, 1000x                                       2.94s   2.94s  0.002s

|  1 vc=1000, ec=5000, k=3, all, 100x                                              0.085s  0.085s      0s
|  2 vc=1000, ec=5000, k=3, subset: all, 100x                                      0.114s  0.113s      0s
|  3 vc=1000, ec=5000, k=3, subset: 100, 1000x                                     0.124s  0.124s      0s
|  4 vc=1000, ec=5000, k=4, subset: all, 100x                                      0.465s  0.464s      0s
|  5 vc=1000, ec=5000, k=4, subset: 100, 1000x                                     0.464s  0.464s      0s

|  1 vc=10000, ec=10000, k=3, all, 100x                                             0.21s  0.209s      0s
|  2 vc=10000, ec=10000, k=3, subset: all, 100x                                    0.328s  0.327s  0.001s
|  3 vc=10000, ec=10000, k=3, subset: 100, 1000x                                   0.115s  0.114s  0.001s
|  4 vc=10000, ec=10000, k=4, subset: all, 100x                                    0.248s  0.248s      0s
|  5 vc=10000, ec=10000, k=4, subset: 100, 1000x                                    1.73s   1.72s  0.002s

|  1 vc=10000, ec=50000, k=3, all, 100x                                            0.903s  0.901s  0.001s
|  2 vc=10000, ec=50000, k=3, subset: all, 100x                                     1.19s   1.18s  0.006s
|  3 vc=10000, ec=50000, k=3, subset: 100, 1000x                                   0.175s  0.174s  0.001s
|  4 vc=10000, ec=50000, k=4, subset: all, 100x                                     5.22s    5.2s  0.009s
|  5 vc=10000, ec=50000, k=4, subset: 100, 1000x                                    4.38s   4.35s  0.015s

Barabasi:

|  1 vc=10000, ec=50000, k=3, all, 100x                                             1.29s   1.28s  0.004s
|  2 vc=10000, ec=50000, k=3, subset: all, 100x                                     1.54s   1.53s  0.009s
|  3 vc=10000, ec=50000, k=3, subset: 100, 1000x                                   0.165s  0.164s  0.001s
|  4 vc=10000, ec=50000, k=4, subset: all, 100x                                       11s   10.9s   0.01s
|  5 vc=10000, ec=50000, k=4, subset: 100, 1000x                                    3.18s   3.16s  0.005s
```

I think this is good enough for now. Further improvements require careful profiling. It would be nice to look into a lazy-adjlist version of `k=4`, but I expect that this function will rarely be used with small edge subsets. It's also nice to look into `igraph_degree_1()`, but once again it requires careful profiling, which requires more time than what I have for this right now.

src/properties/ecc.c
```diff
@@ -0,0 +1,349 @@
...
+}
+
+
+/**
+ * \function igraph_ecc
+ * \brief Edge clustering coefficient of some edges.
+ *
+ * \experimental
+ *
+ * The edge clustering coefficient <code>C^(k)_ij</code> of an edge $(i, j)$
```
ntamas, line 278: I think that the `$(i, j)$` LaTeX syntax won't work in the documentation, right?

szhorvat:
The tests added in this PR currently have a workaround for #2224.

ntamas:
#2224 is now fixed so you can remove the workaround (if needed).

szhorvat:
In this function, I need to get sorted neighbour lists from a lazy adjlist. I.e., the first time a new neighbour list is retrieved, it must immediately be sorted. To do this, I use a separate boolean vector to keep track of whether the neighbour list has already been computed. But of course the lazy adjlist already keeps track of this, it just doesn't expose this information.

To make things simpler here, and to spare the allocation of a boolean vector, I propose introducing `igraph_lazy_adjlist_has()` to query whether a given element of the adjlist has already been computed. We should also document that if it has, then `igraph_lazy_adjlist_get()` cannot fail.

Right now we have:

```c
#define igraph_lazy_adjlist_get(al,no) \
    ((al)->adjs[(igraph_integer_t)(no)] != 0 ? ((al)->adjs[(igraph_integer_t)(no)]) : \
     (igraph_i_lazy_adjlist_get_real(al, no)))
```

I propose:

```c
#define igraph_lazy_adjlist_has(al,no) ((igraph_bool_t) (al)->adjs[(igraph_integer_t)(no)])

#define igraph_lazy_adjlist_get(al,no) \
    (igraph_lazy_adjlist_has(al,no) ? ((al)->adjs[(igraph_integer_t)(no)]) \
                                    : (igraph_i_lazy_adjlist_get_real(al, no)))
```

And of course the same for lazy inclists.

Any objectsions @ntamas? If not, should I keep this new macro undocumented, or can I document it?

szhorvat:
The above suggestion can also be implemented later.

For now I consider this PR to be ready to merge.

ntamas (approved):

ntamas:
Looks good to me; it's okay for me if you add an `igraph_lazy_adjlist_has()`, but let's do that in a separate PR.

szhorvat:
Please don't delete this branch _yet_. @vtraag might want to give some feedback here at a later time. I'll delete it myself at some point.

szhorvat:
> it's okay for me if you add an igraph_lazy_adjlist_has(), but let's do that in a separate PR.

@ntamas Added in above-mentioned commit.

ntamas:
Branch restored as requested by @szhorvat

PR #2216, opened 2022-09-29
feat: graph distances with cutoff

### LINKED ISSUE #2126: Wishlist: Distance matrix with cutoff

A generalization of `igraph_distances()` with a distance cutoff.

```c
igraph_distances(const igraph_t *graph, igraph_matrix_t *res,
                          const igraph_vs_t from, const igraph_vs_t to,
                          igraph_neimode_t mode, igraph_real_t cutoff);
```

The distance between `u` and `v` would be considered infinity if it is larger than `cutoff`.

**Use cases for the feature**

In many cases, one needs to process only a local neighbourhood of vertices. This can be done much more efficiently in large graphs if the shortest path algorithm stops once it reaches `cutoff`.

szhorvat:
This PR adds graph distance calculations with a cutoff.

Since we can't break API in 0.10, I introduced a separate functions. We can discuss merging the two together in 0.11 if you like. I need this for 0.10.2. A quick review would be appreciated so that I can build on top of this.

Fixes #2126


ntamas:
Thanks!

szhorvat:
I forgot to update `functions.yaml`. Will do that later.

szhorvat:
@ntamas Do you think it makes sense to mark these as `\experimental` to make it easier to merge `igraph_distances_cutoff()` back into `igraph_distances()` for 0.11? I think it would be good to do that, as the two functions differ only in a single argument, but I don't have a strong opinion.

ntamas:
Yes, let's do it.

PR #2218, opened 2022-10-01
add mode parameter to distances_johnson

### LINKED ISSUE #2193: Johnson shortest path functions should have a mode argument

`igraph_distances_johnson()` and `igraph_shortest_paths_johnson()` should have a `mode` argument.

**Use cases for the feature**

The Dijkstra, Bellman-Ford and unweighted variants already have this argument. Ideally, a high-level interface would have a single distances() function, where the algorithm and the mode would be separate parameters. Right now, the Johnson algorithm cannot handle the mode.

GroteGnoom:
Fixes #2193


GroteGnoom:
The biggest slowdown I see in the benchmarks is about 0.1%, so that's not an issue 🙂

src/paths/johnson.c
```diff
@@ -115,6 +121,11 @@ igraph_error_t igraph_distances_johnson(const igraph_t *graph,
                      IGRAPH_EINVAL);
     }
 
+    if (mode == IGRAPH_ALL) {
+        IGRAPH_ERROR("Mode IGRAPH_ALL used with negative weight. This will cause a negative loop.",
+                     IGRAPH_EINVAL);
```
szhorvat, line 126: - Let's make this return `IGRAPH_ENEGLOOP`
 - Let's avoid mentioning C specific things in error messages such as the `IGRAPH_ALL` constant. Instead, I would just say "Negative weight in undirected graph", as mode=ALL implies treating the graph as undirected. Additionally, at the beginning of the function, I would put something like `if (! igraph_is_directed(graph)) mode = IGRAPH_ALL;`

tests/unit/igraph_distances_johnson.c
```diff
@@ -34,35 +34,42 @@ int main(void) {
...
+    IGRAPH_ASSERT(igraph_distances_johnson(&g_lm, &result, igraph_vss_1(0), igraph_vss_1(2), &weights_lm, IGRAPH_OUT) == IGRAPH_SUCCESS);
     print_matrix(&result);
 
     printf("Directed graph with loops and multi-edges, select none:\n");
-    IGRAPH_ASSERT(igraph_distances_johnson(&g_lm, &result, igraph_vss_none(), igraph_vss_none(), &weights_lm) == IGRAPH_SUCCESS);
+    IGRAPH_ASSERT(igraph_distances_johnson(&g_lm, &result, igraph_vss_none(), igraph_vss_none(), &weights_lm, IGRAPH_OUT) == IGRAPH_SUCCESS);
+    print_matrix(&result);
+
+    printf("Directed graph with loops and multi-edges, IGRAPH_IN:\n");
+    IGRAPH_ASSERT(igraph_distances_johnson(&g_lm, &result, vids, vids, &weights_lm, IGRAPH_IN) == IGRAPH_SUCCESS);
```
szhorvat, line 62: Could you please add a comparison with the result of Belman-Ford here, once for both OUT and IN, for a bit of extra security? We might need to use `igraph_matrix_all_almost_e()` and add a reasonable relative tolerance (start with `1e-14` and increase the exponent until it succeeds).

szhorvat: Sorry, scratch that. The weight are all integers. No need for tolerances. Just use `igraph_matrix_all_e()`.

tests/unit/igraph_distances_johnson.c
```diff
@@ -34,35 +34,42 @@ int main(void) {
...
     print_matrix(&result);
 
+    VERIFY_FINALLY_STACK();
+
+    printf("Checking error for directed graph with loops and multi-edges with negative loop.\n");
+    CHECK_ERROR(igraph_distances_johnson(&g_lm, &result, vids, vids, &weights_lm_neg_loop, IGRAPH_OUT), IGRAPH_ENEGLOOP);
+
+    printf("Directed graph with loops and multi-edges, IGRAPH_ALL:\n");
+    CHECK_ERROR(igraph_distances_johnson(&g_lm, &result, vids, vids, &weights_lm, IGRAPH_ALL), IGRAPH_EINVAL);
```
szhorvat, line 71: This'll need to be updated to `IGRAPH_ENEGLOOP` as well.

szhorvat:
Left a few comments, looks good otherwise.

CHANGELOG.md
```diff
@@ -18,6 +18,7 @@
    igraph's own RNG if needed instead of relying on LAPACK or ARPACK to do so.
    This makes sure that the results obtained from these functions remain the
    same if igraph's RNG is seeded with the same value.
+ - `igraph_distances_johnson()` now takes a mode parameter to determine in which direction paths should be followed.
```
szhorvat, line 21: Ah yes, one more thing:

This line shouldn't be added here. Instead, the changelog should be separated to two parts, one for 0.10.x and one for 0.11. This should be added to the latter.

We need to come up with different headers for these. I would actually suggest just writing `[0.10.2] - Unreleased` and `[0.11.0] - Unreleased`, i.e. use "Unreleased" in place of the date instead of in place of the version. @ntamas Are you fine with this notation?

The current header should be updated on `master` first.

ntamas: Having two headers is OK, but note that the bracketed parts are actually links to be resolved by the Markdown renderer -- there's a list at the bottom of the changelog that tells which Github URLs the bracketed stuff refers to. We need to add [0.10.2] and [0.11.0] there, one referring to `master`, the other to `develop`. But if we do that, we could even call the branches by names and not by upcoming version numbers.

szhorvat: @ntamas Can you just update the `CHANGELOG` on `master` with your preferred phrasing? Then we can adapt this PR to it.

ntamas: Added the `master` section on the branch, and I'll add the `develop` section here

src/paths/johnson.c
GroteGnoom: I fixed the first two.

For the third: There's already an error for undirected graphs.

szhorvat:
Please ping me when you think this is ready.

szhorvat (approved):

ntamas (approved):

PR #2219, opened 2022-10-01
make layout_umap interruptible

### LINKED ISSUE #2211: Make UMAP layout interruptible
UMAP layout should be interruptible

**Use cases for the feature**
To make it easier for higher-level interfaces to stop the process.

GroteGnoom:
Fixes #2211

I use an interrupt period of 10000 loops over neighbor repulsions. The counter is not static, so will be reset every epoch. This makes sure interruptions are allowed at least once every epoch (if there are any nodes).

I did not test this, and the `10000` is pretty random. Any suggestions on timing this better are welcome.

### DIFF
--- a/src/layout/umap.c
+++ b/src/layout/umap.c
@@ -532,11 +532,8 @@ static igraph_error_t igraph_i_umap_compute_cross_entropy(const igraph_t *graph,
 
 
 /* clip forces to avoid too rapid shifts */
-static igraph_error_t igraph_i_umap_clip_force(igraph_real_t *force, igraph_real_t limit) {
-
+static void igraph_i_umap_clip_force(igraph_real_t *force, igraph_real_t limit) {
     *force  = fmax(fmin(*force, limit), -limit);
-
-    return IGRAPH_SUCCESS;
 }
 
 /*xd is difference in x direction, mu is a weight */
--- a/src/layout/umap.c
+++ b/src/layout/umap.c
@@ -617,8 +617,6 @@ static igraph_error_t igraph_i_umap_apply_forces(
      * not be doing UMAP.
      * */
     igraph_vector_int_t neis, negative_vertices;
-    igraph_integer_t loop_count = 0;
-    igraph_integer_t interrupt_period = 10000;
 
     /* Initialize vectors */
     IGRAPH_VECTOR_INIT_FINALLY(&from_emb, ndim);
@@ -669,10 +667,8 @@ static igraph_error_t igraph_i_umap_apply_forces(
         /* Random other nodes are repelled from one (the first) vertex */
         IGRAPH_CHECK(igraph_random_sample(&negative_vertices, 0, no_of_nodes - 2, n_random_vertices));
         for (igraph_integer_t j = 0; j < n_random_vertices; j++) {
-            loop_count++;
-            if (loop_count % interrupt_period == 0) {
-                IGRAPH_ALLOW_INTERRUPTION();
-            }
+            IGRAPH_ALLOW_INTERRUPTION();
+
             /* Get random neighbor */
             to = VECTOR(negative_vertices)[j];
             /* obviously you cannot repel yourself */
--- a/src/layout/umap.c
+++ b/src/layout/umap.c
@@ -26,6 +26,7 @@
 #include "igraph_random.h"
 
 #include "layout/layout_internal.h"
+#include "core/interruption.h"
 
 #include <math.h>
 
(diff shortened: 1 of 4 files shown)


szhorvat:
A good guidelines for where NOT to insert `IGRAPH_ALLOW_INTERRUPTION()` is inner loops which have no non-inlineable function calls.

This loop does indeed have only minimal non-inlineable calls when `avoid_neighbor_repulsion == false`. There are some instances of `fmin`/`fmax`. However, the loop is fairly complex, so in this case `IGRAPH_ALLOW_INTERRUPTION()` may not cause a slowdown.

It is probable fine to get rid of the counter.

Is 10000 a good number? I don't know. I would not use this number without testing interruptability on very large graphs (e..g using the Python interface), and making sure it's responsive.  I'd rather just remove the counter entirely.

szhorvat:
Something I noticed while looking at this is the unnecessary complexity of the `igraph_i_umap_clip_force` function. It does not need an error code, and getting rid of `fmax`/`fmin` may increase performance. Also check if you expect NaN to appear here and whether the way `fmin`/`fmax` handles that suits your use case.

iosonofabio:
@GroteGnoom you could time it on a few graphs of different sizes in Python and get a sense of scaling. Or just insert a timer. Either way, interrupting every quarter of a second is more than enough and more important than losing 5% speed.

src/layout/umap.c
```diff
@@ -666,6 +669,10 @@
         /* Random other nodes are repelled from one (the first) vertex */
         IGRAPH_CHECK(igraph_random_sample(&negative_vertices, 0, no_of_nodes - 2, n_random_vertices));
         for (igraph_integer_t j = 0; j < n_random_vertices; j++) {
+            loop_count++;
+            if (loop_count % interrupt_period == 0) {
```
vtraag, line 673: Just my two cents: In terms of speed it typically is faster to do something like `loop_count > interrupt_period` and reset `loop_count = 0` inside the `if`. The modulo is relatively expensive.

GroteGnoom:
With interruption for every repel:

| Small graph, epochs: 5000, repetitions: 40                                          11s   10.9s  0.032s
| Small graph, epochs: 500, repetitions: 400                                        11.2s   11.1s  0.035s
| Larger graph, epochs: 60, repetitions: 1, vertices: 10000, density: 0.001         13.6s   13.6s  0.004s

No interruption:
|> Benchmark file: /Users/dnoom/Documents/igraph/tests/benchmarks/igraph_layout_umap.c
| Small graph, epochs: 5000, repetitions: 40                                        10.8s   10.8s  0.033s
| Small graph, epochs: 500, repetitions: 400                                        11.1s     11s  0.037s
| Larger graph, epochs: 60, repetitions: 1, vertices: 10000, density: 0.001         13.3s   13.3s  0.005s

No clip_force:
|> Benchmark file: /Users/dnoom/Documents/igraph/tests/benchmarks/igraph_layout_umap.c
| Small graph, epochs: 5000, repetitions: 40                                        10.6s   10.6s  0.032s
| Small graph, epochs: 500, repetitions: 400                                        10.8s   10.8s  0.036s
| Larger graph, epochs: 60, repetitions: 1, vertices: 10000, density: 0.001         12.6s   12.6s  0.005s

GroteGnoom:
> Something I noticed while looking at this is the unnecessary complexity of the igraph_i_umap_clip_force function. It does not need an error code, and getting rid of fmax/fmin may increase performance. Also check if you expect NaN to appear here and whether the way fmin/fmax handles that suits your use case.

I think it could be possible to have NaN's, and they will be set to the maximum positive force, which doesn't seem horrible. I don't know how to further simplify it. It does seem possible to gain a few % speed without clipping.

src/layout/umap.c
```diff
@@ -531,11 +532,8 @@ static igraph_error_t igraph_i_umap_compute_cross_entropy(const igraph_t *graph,
 
 
 /* clip forces to avoid too rapid shifts */
-static igraph_error_t igraph_i_umap_clip_force(igraph_real_t *force, igraph_real_t limit) {
-
+static void igraph_i_umap_clip_force(igraph_real_t *force, igraph_real_t limit) {
```
szhorvat, line 535: Why not just return the result directly instead of through a pointer argument?

szhorvat:
```
static igraph_real_t igraph_i_umap_clip_force(igraph_real_t force, igraph_real_t limit) {
    if (force < -limit) return -limit;
    if (force > limit) return limit;
    return force;
}
```

This will also pass through any NaNs while being more efficient than the `fmin`/`fmax` version.

GroteGnoom: Because there was no reason to change it. Now I did change it.

szhorvat (approved):

szhorvat:
It's nice that you remembered to use `main(void)`, @GroteGnoom, even though the I wouldn't fail without it. I still keep forgetting this. It looks great to me, I'll leave it to you to merge.

GroteGnoom:
> It's nice that you remembered to use main(void), @GroteGnoom, even though the I wouldn't fail without it. I still keep forgetting this. It looks great to me, I'll leave it to you to merge.

That was completely by accident. I always copy stuff 😬

iosonofabio:
Thank you!

PR #2220, opened 2022-10-01
feat: igraph_voronoi() for Voronoi partitioning of graphs

szhorvat:
See the included docs for an explanation, and let me know if anything is unclear.

Notes:

  - I chose to specify the generator points as an integer vector instead of a vertex selector. This is the interface igraph functions typically use when only a small number of vertices are selected, e.g. when choosing roots for tree operations.


interfaces/functions.yaml
```diff
@@ -550,6 +550,11 @@ igraph_distances_floyd_warshall:
         GRAPH graph, OUT MATRIX res, EDGEWEIGHTS weights=NULL, NEIMODE mode=OUT
     DEPS: weights ON graph
 
+igraph_voronoi:
+    PARAMS: |-
+        GRAPH graph, OUT VECTOR_INT membership, OUT VECTOR distances,
+        VECTOR_INT generators, VECTOR weights, NEIMODE mode=OUT, VORONOI_TIEBREAKER tiebreaker=RANDOM
+
```
GroteGnoom, line 557: Should you not have some DEPS here?

szhorvat: Thanks, added the dependency for weights.

GroteGnoom:
A voronoi partitioning is also defined for graphs with negative edge weights, right? Is that never useful? Or just too much work? (Yes, of course I ask because of working on Johnson's 🙂  )

src/paths/voronoi.c
```diff
@@ -0,0 +1,423 @@
...
+ *    when it is at equal distance from/to multiple generator points.
+ *    \clist
+ *    \cli IGRAPH_VORONOI_FIRST assign the vertex to the first generator point.
+ *    \cli IGRAPH_VORONOI_LAST assign the vertex to the last generator point.
+ *    \cli IGRAPH_VORONOI_RANDOM assign the vertex to a random generator point.
+ *    \endclist
+ *    Note that \c IGRAPH_VORONOI_RANDOM does not guarantee that all vertices
+ *    of the subgraphs induced by the partitions will be reachable from the
+ *    generator. For example, if 1 and 2 are chosen as genrators for the graph
+ *    <code>4-3-2, 2-1</code>, then 3 and 4 may be randomly assigned to different
```
GroteGnoom, line 357: I'm not sure I understand this line, should it be
`<code>4-3, 2-1</code>`
maybe?

szhorvat:
That's an interesting topic to think about, but I don't have the energy for it right now. I do have a personal use case for positive-weighted graphs, and I wanted to implement that here. Of course, generalizations could be possible in the future. This function should be generalizable to negative weights without having to change the interface at all, but I did add `\experimental` just in case there's something I missed.

src/paths/voronoi.c
szhorvat: That was meant to be shorthand for `4-3, 3-2`, but the example is wrong. I will fix it, thanks for noticing.

It should be `1-3, 2-3, 3-4`, for this graph:

image

If the generators are 1 and 2, then 3 and 4 are at equal distance from both of them. The random assignment may cause 3 to be assigned to 2 and 1 to 4. Like this:

image

That means that the partition that belongs to 1 is no longer contiguous. If we take the corresponding subgraph, it will not be connected. This is "not nice", and I though about possible solutions, but it seems far from trivial, so I decided to leave it for later.

szhorvat: I tried to clarify this paragraph, hopefully it's better now.

GroteGnoom:
Thanks! That was very nice to read. I feel like I learned a lot in short time. The comments really help 🙂 
I didn't re-check the results of your tests, because I assume you've made your own checker for it.
I can't think of any more edge cases.

It seems like in theory there could be a faster algorithm, right? Because the first generators check distances to lots of vertices that will be thrown away later. If you keep looping over generators and let them all take one step at a time, you wouldn't have that problem. (Maybe very hard to implement and unimportant)

szhorvat:
Thanks for the comments! Simultaneous growth of the partitions is something interesting to explore. I'm a bit concerned about the memory requirements, as it would require as many `dqueue`s or `2wheap`s as the number of generator points. It might be feasible though.

I'm quite happy with the current implementation though, which computed an entire distance matrix with as many rows as the generators and as many columns as the vertices. This saves on that my not holding the entire matrix in memory and still handling random tie breaking, as well as exploring smaller and smaller regions with subsequent shortest path runs.

szhorvat:
I'm still thinking whether to add a `cutoff` argument, but for now I consider this ready to merge.

interfaces/functions.yaml
```diff
@@ -550,6 +550,12 @@ igraph_distances_floyd_warshall:
         GRAPH graph, OUT MATRIX res, EDGEWEIGHTS weights=NULL, NEIMODE mode=OUT
     DEPS: weights ON graph
 
+igraph_voronoi:
+    PARAMS: |-
+        GRAPH graph, OUT VECTOR_INT membership, OUT VECTOR distances,
+        VECTOR_INT generators, VECTOR weights, NEIMODE mode=OUT, VORONOI_TIEBREAKER tiebreaker=RANDOM
```
ntamas, line 556: * `generators` should be of type `VERTEX_INDICES`.
* `weights` should be of type `EDGEWEIGHTS` with a default of `NULL`

interfaces/functions.yaml
```diff
@@ -550,6 +550,12 @@ igraph_distances_floyd_warshall:
...
+    DEPS: weights ON graph
```
ntamas, line 557: We also need `generators ON graph` here

src/paths/voronoi.c
```diff
@@ -0,0 +1,423 @@
...
+    igraph_dqueue_int_t q;
+
+    igraph_vector_int_t already_counted;
+
+    /* tie_count[vid] is the number of generators that vid is an equal distance from.
+     * This value is needed to pick one of these generators uniformly at random
+     * without needing to store all of them. */
+    igraph_vector_int_t tie_count;
+
+    igraph_adjlist_init(graph, &al, mode, IGRAPH_LOOPS, IGRAPH_MULTIPLE);
```
ntamas, line 50: This needs an `IGRAPH_CHECK`

src/paths/voronoi.c
```diff
@@ -0,0 +1,423 @@
...
+                case IGRAPH_VORONOI_LAST:
+                    /* Always replace existing generator assignment. */
+                    VECTOR(*membership)[vid] = i;
+                    break;
+                case IGRAPH_VORONOI_RANDOM:
+                    /* We replace the membership assignment with probability 1/k upon
+                     * encountering the kth same-distance generator. This ensures
+                     * that one of these generators is selected uniformly at random. */
+                    VECTOR(tie_count)[vid]++;
+                    RNG_BEGIN();
```
ntamas, line 123: Let's move these out of the loop; in R, this operation involves copying the entire state vector of the RNG from R-space to C-space (or the other way round).

src/paths/voronoi.c
```diff
@@ -0,0 +1,423 @@
...
+        /* Weighted shortest path implementation using Dijkstra's algorithm */
+
+        IGRAPH_ALLOW_INTERRUPTION();
+
+        igraph_2wheap_clear(&q);
+
+        /* We store negative distances in the maximum heap. Since some systems
+         * distinguish between -0.0 and +0.0, we need -0.0 to ensure +0.0 as
+         * the final result. */
+        igraph_2wheap_push_with_index(&q, g, -0.0);
```
ntamas, line 228: Don't we need an `IGRAPH_CHECK` here?

src/paths/voronoi.c
```diff
@@ -0,0 +1,423 @@
...
+                case IGRAPH_VORONOI_LAST:
+                    /* Always replace existing generator assignment. */
+                    VECTOR(*membership)[vid] = i;
+                    break;
+                case IGRAPH_VORONOI_RANDOM:
+                    /* We replace the membership assignment with probability 1/k upon
+                     * encountering the kth same-distance generator. This ensures
+                     * that one of these generators is selected uniformly at random. */
+                    VECTOR(tie_count)[vid]++;
+                    RNG_BEGIN();
```
ntamas, line 268: This should also be moved out of the loop

ntamas (changes requested):

src/paths/voronoi.c
szhorvat: I'm assuming we don't care if `RNG_END()` is not called upon error, right?

szhorvat:
Thanks for catching these, @ntamas. See if you like where I put the RNG_BEGIN/END calls.

src/paths/voronoi.c
ntamas: RIght, that's a mess, ideally it should be called, but the only use-case for `RNG_BEGIN()` and `RNG_END()` is R's peculiarity and the worst thing that can happen there is that the R-side RNG does not get updated with the state changes on the C side.

In the long term, `RNG_BEGIN()` and `RNG_END()` should be eliminated from the C core and moved entirely to the R side, after annotating our functions in `functions.yaml` so we know which ones use the RNG and which ones do not.

szhorvat:
This one should be merged using "squash and merge".

ntamas (approved):

Issue #1979, opened 2022-03-03
Several data structures need documentation


szhorvat:
There are several specialized data structures in `src/core` which need at least some rudimentary documentation. This may be just a set of raw comments, but it should still exists.

 - [x] `vector_t`, public
 - [x] `strvector_t`, public
 - [x] `vector_list_t`, public
 - [x] `vector_ptr_t`, public
 - [x] `stack_t`, public
 - [ ] `array_t`, public
 - [x] `buckets_t` 
 - [ ] `cutheap_t`
 - [ ] `dqueue_t`
 - [ ] `estack_t`
 - [ ] `fixed_vectorlist_t`
 - [ ] `2dgrid_t`
 - [ ] `hashtable_t`
 - [x] `heap_t`, public
 - [ ] `indheap_t`
 - [ ] `marked_queue_t`, mostly documented in .h file, but may need improvements
 - [x] `matrix_t`, public
 - [x] `psumtree_t`, public 
 - [x] `set_t`, consider making it public?
 - [x] `sparsemat`
 - [x] `spmatrix` 
 - [x] `trie_t`

ntamas:
Rudimentary documentation of the two-way heap was added in 274784d2a

PR #2221, opened 2022-10-01
refactor: simplify Dijkstra distances implementation

szhorvat:
igraph has many copies of Dijkstra's algorithm, most copied from `igraph_distances_dijkstra()` and modified for some purpose. This original implementation includes a note about some "dirty tricks" that are used in the implementation. This note is currently incomplete and inaccurate.

This PR cleans up the implementation, simplifies it, and slightly improves its performance. It is likely that this code will continue to get copied and we don't want to base new work on outdated code that does not have ideal readability.

The original note states the following:

```
    /* Implementation details. This is the basic Dijkstra algorithm,
       with a binary heap. The heap is indexed, i.e. it stores not only
       the distances, but also which vertex they belong to.
       From now on we use a 2-way heap, so the distances can be queried
       directly from the heap.
       Dirty tricks:
       - the opposite of the distance is stored in the heap, as it is a
         maximum heap and we need a minimum heap.
       - we don't use IGRAPH_INFINITY in the res matrix during the
         computation, as IGRAPH_FINITE() might involve a function call
         and we want to spare that. -1 will denote infinity instead.
    */
```

Regarding the note about checking for infinities:

 - There aren't actually any checks for infinities
 - If there were, it would not be necessary to use a function call for the check. `x == IGRAPH_INFINITY` is safe on any platform we can reasonably support. We already practically requires IEEE-754 compliance.

The notes did not mention an important point: It was not `-distance`, but `-(distances + 1)` that was stored in the heap. This does not appear necessary and I removed this hack.

I also added a small optimization: skip infinite-weight edges. Some other algorithms set the weight of edges to infinity to effectively remove them, so this is a relevant use case.

The performance improvement from this PR is minor but noticeable. E.g. in the distances benchmark, the first Dijkstra timing drops from about 0.82 seconds to about 0.78, roughly a 10% improvement.

----

I do not quite understand the concept of "deactivation" in the 2wheap and why deactivation is necessary instead of deletion here. @ntamas Do you understand this? It would be good to document some of these special-feature data structures. The issue for that is #1979.

### DIFF
--- a/src/paths/dijkstra.c
+++ b/src/paths/dijkstra.c
@@ -92,12 +92,9 @@ igraph_error_t igraph_distances_dijkstra_cutoff(const igraph_t *graph,
        From now on we use a 2-way heap, so the distances can be queried
        directly from the heap.
 
-       Dirty tricks:
-       - the opposite of the distance is stored in the heap, as it is a
+       Tricks:
+       - The opposite of the distance is stored in the heap, as it is a
          maximum heap and we need a minimum heap.
-       - we don't use IGRAPH_INFINITY in the res matrix during the
-         computation, as IGRAPH_FINITE() might involve a function call
-         and we want to spare that. -1 will denote infinity instead.
     */
 
     igraph_integer_t no_of_nodes = igraph_vcount(graph);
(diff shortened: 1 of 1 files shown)


GroteGnoom:
A little more on deactivation of the 2wheap: https://github.com/igraph/igraph/pull/1893#issuecomment-997477633

ntamas (approved):

PR #2233, opened 2022-10-10
Kneser Graph Generator

### LINKED ISSUE #1876: Generator for Kneser graphs and related graphs

The vertices of the Kneser graph `K(n,k)` are the `k`-element subsets of `n` elements. Two vertices are connected precisely when the subsets are disjoint.

_This task is a bit more involved than other "good first issue" tasks. Work out the math before you start coding._

**Implementation**

One possible implementation is the direct one: represent subsets as Boolean vectors `b` of length `n`, where `b[i]` is true if element `i` is part of the subset. Directly generate all `k`-element subsets and compute their intersections. Question: Is there a more efficient way to generate these graphs?

If using the direct implementation, first implement a `k`-subset iterator. It may be useful for other parts of igraph as well.

The function should return the subsets as well (e.g. as a Boolean matrix).

**Related graph families which may be implemented at the same time**

 - Generalized Kneser graph `K(n, k, s)`: two `k`-subsets are connected when they intersect in no more than `s` elements.
 - Johnson graph `J(n, k)`: two `k`-subsets are connected when they intersect in precisely `k-1` elements.

**Use cases for the feature**

This is a well-studied graph family in graph theory.

**References**

 - https://en.wikipedia.org/wiki/Kneser_graph
 - SageMath: https://doc.sagemath.org/html/en/reference/graphs/sage/graphs/generators/families.html#sage.graphs.generators.families.KneserGraph

valdaarhun:
New feature: Generation of kneser graphs.

Resolves #1876.

So far, I have only implemented functions to calculate $\binom{n}{k}$ and generate k-size subsets. I have made both functions static and put them in `kneser.c`. Please let me know if I should integrate them into `iterators.c` instead.

I think it'll be best to finalize the design and implementation of the above functions before implementing the actual graph generator.

### DIFF
--- a/src/constructors/kneser.c
+++ b/src/constructors/kneser.c
@@ -1,6 +1,6 @@
 /*
    IGraph library.
-   Copyright (C) 2022  The igraph development team
+   Copyright (C) 2025  The igraph development team
 
    This program is free software; you can redistribute it and/or modify
    it under the terms of the GNU General Public License as published by
@@ -20,9 +20,8 @@
 
 #include "igraph_vector_list.h"
 
-static igraph_integer_t igraph_i_number_of_subsets(igraph_integer_t n, igraph_integer_t k){
-    igraph_integer_t r = (k < n - k ? k : n - k);
-    igraph_integer_t number_of_subsets = 1;
+static igraph_integer_t igraph_i_number_of_subsets(igraph_integer_t n, igraph_integer_t k) {
+    igraph_integer_t r = (k < n - k ? k : n - k), number_of_subsets = 1;
 
     for(igraph_integer_t i = 1 ; i <= r ; i++) {
         number_of_subsets *= n - i + 1;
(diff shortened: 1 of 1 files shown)


ntamas:
* `igraph_i_generate_subsets` does not have a return type; it should be `igraph_error_t` and you should wrap the inner `igraph_vector_list_...` call in an `IGRAPH_CHECK()` to make sure that error codes are handled and propagated correctly. You should return `IGRAPH_SUCCESS` from the function when you wish to terminate the recursion.
* Global variables should not be used; `igraph_i_generate_subsets()` should receive the vector list in which the subsets should be filled as a parameter.
* The `nCk` variable is redundant; you can just return the corresponding expression directly.
* `genSubset` is not defined; you probably meant `igraph_i_generate_subsets()`, right?

valdaarhun:
> * `genSubset` is not defined; you probably meant `igraph_i_generate_subsets()`, right?

Hi. You're right, my bad. I forgot to change the name while refactoring.

szhorvat:
Just a note that it's strongly preferred to use non-recursive implementations. Recursive functions are prone to stack overflow. [This has happened in the past.](https://github.com/igraph/python-igraph/issues/563)

valdaarhun:
Hi. I thought I would just give an update here. I have made `igraph_i_generate_subsets` iterative. I am still working on making `igraph_i_number_of_subsets` iterative.

src/constructors/kneser.c
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    return numerator;
+}
+
+static igraph_error_t igraph_i_generate_subsets(igraph_integer_t n, igraph_integer_t k, 
+                              igraph_integer_t subsets, igraph_vector_int_list_t *vertices) {
+
+    igraph_vector_int_t *temp = igraph_vector_list_get_ptr(vertices, 0);
+    IGRAPH_CHECK(igraph_vector_init(temp, k));
```
ntamas, line 54: Okay, so there's one significant problem with this. `igraph_vector_int_list_t` _manages_ a list of `igraph_vector_int_t` objects such that if the list contains _k_ elements, then it is guaranteed that all _k_ elements are already initialized. If you get a pointer to the 0-th element (like you do here), the vector that you get is already initialized, so you cannot call `igraph_vector_init()` on it. (Well, you can, but then you lose a reference to the old vector, leaking memory).

The bottom line is: don't call `igraph_vector_init()` here. If you want to make sure that the vector is empty, call `igraph_vector_int_clear()` instead.

src/constructors/kneser.c
```diff
@@ -0,0 +1,80 @@
...
+        denominator *= r;
+        igraph_integer_t gcd = igraph_i_get_gcd(numerator, denominator);
+        numerator /= gcd;
+        denominator /= gcd;
+        --r;
+        --n;
+    }
+
+    return numerator;
+}
```
ntamas, line 48: I think there's a much easier approach that does not involve getting common denominators and does not risk overflow either. In pseudo-code:

```
result = 1
for i in 1..r:
    result = result * (n-i+1)
    result = result / i
```

Note that the order of operations is important; doing the multiplication and the division separately ensures that by the time you divide with `i` the result itself is already divisible by `i` due to previous multiplications.

src/constructors/kneser.c
valdaarhun: Oh, I understand this now. I think I got confused between an "empty" vector and an "uninitialised" vector. I thought `igraph_vector_list_init()` was only initialising the list itself and not the actual items in the list.

GroteGnoom:
What's the status of this PR? @valdaarhun are you waiting for us before you continue?

valdaarhun:
Hi. Yes, I was waiting for feedback. I apologize for not sending a reminder, I got caught up with other work. I would like to resume working on this now.

src/constructors/kneser.c
```diff
@@ -0,0 +1,65 @@
...
+    }
+
+    return number_of_subsets;
+}
+
+static igraph_error_t igraph_i_generate_subsets(igraph_integer_t n, igraph_integer_t k,
+                              igraph_integer_t subsets, igraph_vector_int_list_t *vertices) {
+
+    igraph_vector_int_t *temp = igraph_vector_list_get_ptr(vertices, 0);
+    IGRAPH_CHECK(igraph_vector_resize(temp, k));
```
ntamas, line 39: Since `temp` is an `igraph_vector_int_t`, you need to use the functions that take an `igraph_vector_int_t`. For instance, `igraph_vector_int_resize()` instead of `igraph_vector_resize()`, `igraph_vector_int_update()` instead of `igraph_vector_update()` and so on. I haven't checked the correctness of the algorithm yet, this just simply struck my eye.

I haven't checked the correctness of this function, this just simply struck my eye.

GroteGnoom:
> Hi. Yes, I was waiting for feedback. I apologize for not sending a reminder, I got caught up with other work. I would like to resume working on this now.

No apologies necessary :slightly_smiling_face: 
As Tamás noticed, your types are not completely correct.

If you add your file to src/CMakeLists.txt you can use CMake to build your code and see the warnings.
(You can also maybe set up your editor so you can see these warnings while coding.)

If you add the actual generator function you can also add some tests in `/tests/unit/` and add it in `tests/CMakeLists`, so your code gets run. That's one basic way of getting some feedback on your code :)

If you have any further questions (like about using CMake) please ask.

src/constructors/kneser.c
valdaarhun: Oh, I must have missed this. I'll fix it.

valdaarhun:
> If you add your file to src/CMakeLists.txt you can use CMake to build your code and see the warnings. (You can also maybe set up your editor so you can see these warnings while coding.)
> 
> If you add the actual generator function (or maybe for now just something that tests your subset generator) you can also add some tests in `/tests/unit/` and add it in `tests/CMakeLists`, so your code gets run. That's one basic way of getting some feedback on your code :)

Oh, that's nice. I'll definitely do this.

> If you have any further questions (like about using CMake) don't hesitate to ask.

Sure thing. Thanks.

Issue #1842, opened 2021-11-14
Create a triangular grid graph


szhorvat:
_This is a good first issue to work on for a new contributor interested in math._

**What is the feature or improvement you would like to see?**

Functionality to create various kinds of triangle grid graphs. Since we have a clean slate to work with, we should take the opportunity to think about how we can create a more general and nicer triangle grid generator than what is available on other libraries. I would like igraph to be able to generate at least the following graphs:

<img width="270" alt="image" src="https://user-images.githubusercontent.com/1212871/141675218-9c73bf6d-a90b-41aa-931b-4b83f52aa658.png">

Part of this task is to think about the best way to accomplish this. What interface would be both simple and general? We should think about this in terms of igraph's flexible higher-level interfaces, and then decide on an appropriate C interface late.

----

The following proposal is meant as a starting point for a discussion.

The function will be able to operate in three modes, taking either a single number, a pair of numbers or a triplet as input. The result will be a triangle shape, a rectangle shape or a general hexagon shape, respectively.

For the input `(5)` we get:

<img width="360" alt="image" src="https://user-images.githubusercontent.com/1212871/141675472-cf137ea2-23a8-4cd5-a0e0-8c81bb957176.png">

(5 vertices on each side.)

For `(4, 6)` we get:

<img width="360" alt="image" src="https://user-images.githubusercontent.com/1212871/141675499-d4bb994a-b7b8-4bc8-a1ca-d3542f220ecd.png">

(The rectangle is 6 vertices wide and 4 vertices tall.)

For `(5, 4, 3)` we get:

<img width="360" alt="image" src="https://user-images.githubusercontent.com/1212871/141675528-0fa7cb49-6304-4a7f-b409-1a5ae3bca3a0.png">

(The three sides of the hexagon are 5, 4 and 3 vertices in length.)

For `(4, 4, 4)`:

<img width="360" alt="image" src="https://user-images.githubusercontent.com/1212871/141675546-535ad789-943e-4721-8f46-c9e995d7fbed.png">

(All sides of the hexagon have 4 vertices along them.)

**Vertex coordinates**

The function should also be able to compute vertex coordinates for producing nice graph layouts, as in the illustrations above.

**Use cases for the feature**

Triangle grids are commonly used, and are provided by many other graph libraries. igraph should have them too.

**See also**

 - #1843 for hexagonal grids.

GroteGnoom:
[...]

PR #2235, opened 2022-10-11
Feature/triangular grid graph

rfulekjames:
Implemented #1842.

- Createad a function `triangle_lattice` that produces a triangulation of a piece of 2D-lattice whose each vertex can be associated with $(i, j) \in \mathbb{Z}_{\ge 0}^2$. The following arguments determine the *shape* of the grid:
  - `row_lengths_vector` the number of vertices in rows, and
  - `row_start_vector` the $x$-coordinate of rows' left-most vertices
- Used `triangle_lattice` by the exported `igraph_triangle_lattice` function for creating the three type of meshes (triangular, rectanglar and hexagonal) 

Tried to test it thoroughly, e.g., edge cases with 1s in the shape vector `dims`.

A *nice layout* with the triangles being equilateral could be obtained by a simple linear tranformation taking the vertex coordinates in $\mathbb{Z}_{\ge 0}^2$.


szhorvat:
A few more notes:

 - The test program should go into `tests/unit`. We can write a _simple_ example later, but now the priority is thorough testing.
 - The function name will be `igraph_triangle_lattice()`, in analogy with the existing `igraph_square_lattice()`.
 - Names of functions that won't be part of the public API should not be of the form `igraph_something`. Function which are used within the same translation unit only can have any name (except `igraph_something`) and must be `static`. Internal functions which are used form multiple translation units should have a name of the form `igraph_i_something`.
 - It's really great that you wrote up the documentation in advance!

rfulekjames:
> - please take care of error handling; in particular, calling any igraph function that returns an igraph_error_t error code must be wrapped in an IGRAPH_CHECK() call
> - when initializing a data structure inside a function, you should queue up its destructor in the FINALLY stack so it gets called when some internal igraph function throws an error, and then also call IGRAPH_FINALLY_CLEAN(...) later if you called the destructor on your own and it is not needed any more

Thanks for the comment!
Tried to revise the code accordingly. Let me please know if there is still something to correct. @ntamas

rfulekjames:
> A few more notes:
> 
> * The test program should go into `tests/unit`. We can write a _simple_ example later, but now the priority is thorough testing.
> * The function name will be `igraph_triangle_lattice()`, in analogy with the existing `igraph_square_lattice()`.
> * Names of functions that won't be part of the public API should _not_ be of the form `igraph_something`. Function which are used within the same translation unit only can have any name (except `igraph_something`) and must be `static`. Internal functions which are used form multiple translation units should have a name of the form `igraph_i_something`.
> * It's really great that you wrote up the documentation in advance!  This makes it easier for us to follow what you are doing.

Thanks for the comment! Removed the simple example for now and renamed the functions as suggested. I will write a unit test shortly. @szhorvat

src/constructors/tessellation.c
```diff
@@ -0,0 +1,304 @@
...
+ *         at least 1.
+ *
+ * Time complexity:  O(|V|), where |V| is the number of vertices in the generated graph.
+ *
+ */
+igraph_error_t igraph_triangle_lattice(igraph_t *graph, const igraph_vector_int_t *dims, igraph_bool_t directed, igraph_bool_t mutual)
+{
+    if (dims->stor_begin == NULL)
+    {
+        IGRAPH_ERRORF("dims vector must be initialized  (%" IGRAPH_PRId ").", IGRAPH_EINVAL, (igraph_integer_t) 0);
```
rfulekjames, line 283: I am throwing an error in this way since using `IGRAPH_ERROR` didn't work for 
`IGRAPH_ASSERT(igraph_triangle_lattice(&graph, &dimvector, true, true) == IGRAPH_EINVAL);` when `dimvector` was not initialized.

src/constructors/tessellation.c
```diff
@@ -0,0 +1,304 @@
...
+ * \return Error code:
+ *         \c IGRAPH_EINVAL: The size of \p dims must be either 1, 2, or 3 with all the components
+ *         at least 1.
+ *
+ * Time complexity:  O(|V|), where |V| is the number of vertices in the generated graph.
+ *
+ */
+igraph_error_t igraph_triangle_lattice(igraph_t *graph, const igraph_vector_int_t *dims, igraph_bool_t directed, igraph_bool_t mutual)
+{
+    if (dims->stor_begin == NULL)
```
rfulekjames, line 281: Not sure what is a better way for checking whether `dimvector` is initialized.

tests/CMakeLists.txt
```diff
@@ -967,3 +967,9 @@ add_legacy_tests(
   igraph_sample_dirichlet
   igraph_sample_sphere
 )
+
+# tessellation
+add_legacy_tests(
```
rfulekjames, line 972: is adding a new `add_legacy_tests` Ok?

rfulekjames:
Not sure why is the new unit test failing. It runs fine locally.

src/constructors/tessellation.c
```diff
@@ -0,0 +1,305 @@
...
+        if (VECTOR(row_lengths_vector)[i] < 0)
+        {
+            IGRAPH_ERRORF(
+                "row_lengths_vector vector must have non-negative coordinates "
+                "(%" IGRAPH_PRId ").",
+                IGRAPH_EINVAL, VECTOR(row_lengths_vector)[i]);
+        }
+    }
+
+    IGRAPH_VECTOR_INT_INIT_FINALLY(&edges, 0);
```
rfulekjames, line 83: I copied this from https://github.com/igraph/igraph/blob/2c0cebbaa6e8179f1c2ed79f9d189eb22dc08f38/src/constructors/regular.c#L398 
Perhaps, this has something to do with failed tests.

src/constructors/tessellation.c
szhorvat: Don't try to check this. There is no reliable check for this in C. Just use the vector.

src/CMakeLists.txt
```diff
@@ -106,6 +106,7 @@ add_library(
   constructors/linegraph.c
   constructors/prufer.c
   constructors/regular.c
+  constructors/tessellation.c
```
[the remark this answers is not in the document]
szhorvat: I suggest using "lattices"

tests/unit/igraph_triangle_lattice.c
```diff
@@ -0,0 +1,82 @@
...
+
+    igraph_triangle_lattice(&graph, &dimvector, false, true);
+    number_of_edges = igraph_vector_int_size(&graph.from);
+    IGRAPH_ASSERT(number_of_edges == 87);
+
+    igraph_vector_int_destroy(&dimvector);
+
+    /*Erroneous calls*/
+    igraph_set_error_handler(igraph_error_handler_ignore);
+    IGRAPH_ASSERT(igraph_triangle_lattice(&graph, &dimvector, true, true) == IGRAPH_EINVAL);
```
szhorvat, line 71: For verifying error codes, please use the `CHECK_ERROR` macro from `test_utilities.h`. Then `igraph_set_error_handler()` is no longer necessary.

tests/unit/igraph_triangle_lattice.c
```diff
@@ -0,0 +1,82 @@
+/*
+   IGraph library.
+   Copyright (C) 2021  The igraph development team <igraph@igraph.org>
```
szhorvat, line 3:
```suggestion
   Copyright (C) 2022  The igraph development team <igraph@igraph.org>
```

src/constructors/tessellation.c
```diff
@@ -0,0 +1,305 @@
...
+   This program is distributed in the hope that it will be useful,
+   but WITHOUT ANY WARRANTY; without even the implied warranty of
+   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+   GNU General Public License for more details.
+
+   You should have received a copy of the GNU General Public License
+   along with this program; if not, write to the Free Software
+   Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
+   02110-1301 USA
+*/
```
szhorvat, line 20:
```suggestion
/*
   IGraph library.
   Copyright (C) 2022  The igraph development team <igraph@igraph.org>
   This program is free software; you can redistribute it and/or modify
   it under the terms of the GNU General Public License as published by
   the Free Software Foundation; either version 2 of the License, or
   (at your option) any later version.
   This program is distributed in the hope that it will be useful,
   but WITHOUT ANY WARRANTY; without even the implied warranty of
   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
   GNU General Public License for more details.
   You should have received a copy of the GNU General Public License
   along with this program.  If not, see <https://www.gnu.org/licenses/>.
*/
```

src/constructors/tessellation.c
rfulekjames: Ok. I used the pattern from `igraph_square_lattice`  function https://github.com/igraph/igraph/blob/bc7c160008ffe3d8dbc352c77add1571e282b519/src/constructors/regular.c#L354

szhorvat:
> Not sure why is the new unit test test_igraph_triangle_lattice failing. It runs fine locally.

Some of the CI runs have various sanitizers and runtime checks enabled. Check their output to see what's going on. In this case it's a memory leak.

https://dev.azure.com/igraph-team/igraph/_build/results?buildId=4649&view=logs&j=4c3a8381-b9de-5463-904f-918810769be8&t=e71559a0-54d3-55f3-64cc-74b67ad70398&l=1061

szhorvat:
Thanks for the contribution! I (or someone else) will only be able to take a look a bit later, as there are several other outstanding PRs that I'd long owed a review for.

tests/CMakeLists.txt
szhorvat: Yes, just change the heading to "lattices".

szhorvat:
In the meantime you can fix the memory leak and add the missing `IGRAPH_CHECK`s.

There's also a checklist here I use in order not to forget things: https://github.com/igraph/igraph/wiki/Checklist-for-new-%28and-old%29-functions

szhorvat:
I noticed that you keep force pushing changes. You can certainly do so if you prefer, but it is likely inconvenient for you and I wanted to let you know that it's not necessary. We can always squash all commits before merging the PR.

tests/unit/igraph_triangle_lattice.c
szhorvat: When a review commit is addressed, please click the "Resolve conversation" button, so we know it's done. I'll do so for all of them now.

rfulekjames:
> I noticed that you keep force pushing changes. You can certainly do so if you prefer, but it is likely inconvenient for you and I wanted to let you know that it's not necessary. We can always squash all commits before merging the PR.

Ok.
Those (force pushed) amended commits are just small corrections and do not cause any inconvenience on my side.

src/constructors/lattices.c
```diff
@@ -0,0 +1,308 @@
...
+        IGRAPH_SAFE_ADD(VECTOR(row_lengths_prefix_sum_vector)[i - 1], VECTOR(*row_lengths_vector)[i - 1], &(VECTOR(row_lengths_prefix_sum_vector)[i]));
+    }
+    no_of_nodes = VECTOR(row_lengths_prefix_sum_vector)[row_count];
+
+#define VERTEX_INDEX(i, j) (VECTOR(row_lengths_prefix_sum_vector)[j] + i - VECTOR(*row_start_vector)[j])
+#define ROW_END(j) (VECTOR(*row_start_vector)[j] + VECTOR(*row_lengths_vector)[j] - 1)
+#define ADD_EDGE_IJ_KL_IF_EXISTS(i, j, k, l)                                                                                                      \
+    if (VECTOR(*row_start_vector)[l] <= k && k <= ROW_END(l) && 0 <= l && l <= row_count - 1)                                                     \
+    {                                                                                                                                             \
+        igraph_vector_int_push_back(&edges, VERTEX_INDEX((i), (j)));                                                                              \
```
ntamas, line 98: Please add a comment either here or above the `ADD_EDGE_IJ_KL_IF_EXISTS` macro to explain that `&edges` is guaranteed to have enough space for the new items so that's why we don't need an `IGRAPH_CHECK(...)` call. We typically do this by adding `/* reserved */` or `/* already reserved */` to the end of the line calling `igraph_vector_int_push_back(...)`

src/constructors/lattices.c
```diff
@@ -0,0 +1,308 @@
...
+        VECTOR(row_lengths_vector)[i] = size - i;
+        VECTOR(row_start_vector)[i] = 0;
+    }
+
+    IGRAPH_CHECK(triangle_lattice(graph, directed, mutual, &row_lengths_vector, &row_start_vector));
+
+    igraph_vector_int_destroy(&row_lengths_vector);
+    IGRAPH_FINALLY_CLEAN(1);
+    igraph_vector_int_destroy(&row_start_vector);
+    IGRAPH_FINALLY_CLEAN(1);
```
ntamas, line 170: The first `IGRAPH_FINALLY_CLEAN(...)` is not necessary and the second one should be `IGRAPH_FINALLY_CLEAN(2)`.

src/constructors/lattices.c
```diff
@@ -0,0 +1,308 @@
...
+            IGRAPH_ERRORF(
+                "row_lengths_vector vector must have non-negative coordinates, "
+                "was (%" IGRAPH_PRId ") for the (%" IGRAPH_PRId ")-th row.",
+                IGRAPH_EINVAL, VECTOR(*row_lengths_vector)[i], i);
+        }
+    }
+
+    IGRAPH_VECTOR_INT_INIT_FINALLY(&edges, 0);
+
+    IGRAPH_CHECK(igraph_vector_int_init(&row_lengths_prefix_sum_vector, row_count + 1));
```
ntamas, line 83: We have a shortcut macro for this _and_ the next line:

```
IGRAPH_VECTOR_INT_INIT_FINALLY(&row_lengths_prefix_sum_vector, row_count + 1);
```

src/constructors/lattices.c
```diff
@@ -0,0 +1,308 @@
...
+{
+    igraph_integer_t row_count = size;
+    igraph_vector_int_t row_lengths_vector;
+    igraph_vector_int_t row_start_vector;
+    igraph_integer_t i;
+
+    IGRAPH_CHECK(igraph_vector_int_init(&row_lengths_vector, row_count));
+    IGRAPH_FINALLY(igraph_vector_int_destroy, &row_lengths_vector);
+    IGRAPH_CHECK(igraph_vector_int_init(&row_start_vector, row_count));
+    IGRAPH_FINALLY(igraph_vector_int_destroy, &row_start_vector);
```
ntamas, line 157: Use `IGRAPH_VECTOR_INT_INIT_FINALLY(...)` here as well.

src/constructors/lattices.c
```diff
@@ -0,0 +1,308 @@
...
+{
+    igraph_integer_t row_count = size_y;
+    igraph_vector_int_t row_lengths_vector;
+    igraph_vector_int_t row_start_vector;
+    igraph_integer_t i;
+
+    IGRAPH_CHECK(igraph_vector_int_init(&row_lengths_vector, row_count));
+    IGRAPH_FINALLY(igraph_vector_int_destroy, &row_lengths_vector);
+    IGRAPH_CHECK(igraph_vector_int_init(&row_start_vector, row_count));
+    IGRAPH_FINALLY(igraph_vector_int_destroy, &row_start_vector);
```
ntamas, line 186: Use `IGRAPH_VECTOR_INT_INIT_FINALLY(...)` here as well.

src/constructors/lattices.c
```diff
@@ -0,0 +1,308 @@
...
+        VECTOR(row_lengths_vector)[i] = size_x;
+        VECTOR(row_start_vector)[i] = (row_count - i) / 2;
+    }
+
+    IGRAPH_CHECK(triangle_lattice(graph, directed, mutual, &row_lengths_vector, &row_start_vector));
+
+    igraph_vector_int_destroy(&row_lengths_vector);
+    IGRAPH_FINALLY_CLEAN(1);
+    igraph_vector_int_destroy(&row_start_vector);
+    IGRAPH_FINALLY_CLEAN(1);
```
ntamas, line 199: The first `IGRAPH_FINALLY_CLEAN(...)` is not necessary and the second one should be `IGRAPH_FINALLY_CLEAN(2)`.

src/constructors/lattices.c
```diff
@@ -0,0 +1,308 @@
...
+{
+    igraph_integer_t row_count = size_y + size_z - 1;
+    igraph_vector_int_t row_lengths_vector;
+    igraph_vector_int_t row_start_vector;
+    igraph_integer_t i;
+
+    IGRAPH_CHECK(igraph_vector_int_init(&row_lengths_vector, row_count));
+    IGRAPH_FINALLY(igraph_vector_int_destroy, &row_lengths_vector);
+    IGRAPH_CHECK(igraph_vector_int_init(&row_start_vector, row_count));
+    IGRAPH_FINALLY(igraph_vector_int_destroy, &row_start_vector);
```
ntamas, line 215: Use `IGRAPH_VECTOR_INT_INIT_FINALLY(...)` here as well.

src/constructors/lattices.c
```diff
@@ -0,0 +1,308 @@
...
+            row_length--;
+        }
+    }
+
+    IGRAPH_CHECK(triangle_lattice(graph, directed, mutual, &row_lengths_vector, &row_start_vector));
+
+    igraph_vector_int_destroy(&row_lengths_vector);
+    IGRAPH_FINALLY_CLEAN(1);
+    igraph_vector_int_destroy(&row_start_vector);
+    IGRAPH_FINALLY_CLEAN(1);
```
ntamas, line 248: The first `IGRAPH_FINALLY_CLEAN(...)` is not necessary and the second one should be `IGRAPH_FINALLY_CLEAN(2)`.

ntamas (changes requested):

rfulekjames:
not sure about the reason behind the failed check which seemed to happen during the build stage.

ntamas:
Something must have gone during a recent merge as there are lots of unrelated changes now in the diff w.r.t. `master`. I'll try to revert these in the next few commits.

rfulekjames:
I was just trying to rebase this branch on `master` branch and didn't succeed. Likely due to pulling the changes from the remote branch before merging. The unrelated changes should be due to the most recent commits on master.
I can try to rebase it again.

ntamas:
No need to rebase, it's all sorted out now. It turns out that _my_ master wasn't up-to-date plus there is a Github UI glitch that prevents me from seeing the "real" diff for some reason.

szhorvat:
Thanks @rfulekjames and @ntamas! I'd like to take a closer look when I'm back from holidays in two days.

Regarding the above two points, I agree on both counts: no link to issues from docs, no `fmin()` for integers. A simple way is `a < b ? a : b`.

szhorvat:
GitHub was acting up again, and not showing the diff to master properly. Changing and reverting base seems to have fixed it.

rfulekjames:
Thanks for the feedback! 

Regarding the formatting, when running formatting on a particular source file, it does not format the file in a way other source files are formatted. I am using VS Code IDE with EditorConfig for VS Code extension. Not sure if `.editorconfig` is not taken into account or what the issue might be.


 @szhorvat  @ntamas

ntamas:
`.editorconfig` governs only the usage of tabs vs spaces and the indentation width, nothing else. So, if your IDE sets the formatting automatically to 4 spaces then it's OK. The rest of the code style is mostly governed by `.astylerc`, but we don't stick to it very strictly. I think the biggest difference between how your IDE formatted the code and what is preferred by us is that opening braces should not go on their own line separately.

szhorvat:
I tested the functionality, and it looks really great!

There are a few minor changes I'd like to ask for. Normally I'd do these myself, but at the moment I am very much pressed for time, so if you can do it @rfulekjames, that would be useful. Just let us know.

----

TODO 1: First, if any dimension is zero, the function should return a graph with zero vertices (the null graph). This is a minor but useful generalization.  Be sure to still throw errors if the `dim` length is not 1, 2 or 3. 

----

TODO 2: Let us name the function `igraph_triangular_lattice()`, as this is the usual name, and it matches "square lattice" and "hexagonal lattice" well. (To be accurate, many physicists would call this structure a hexagonal lattice, not a triangular one, but let's not get hung up on that and stick to triangular.)

----

Regarding the case of length-2 dim vectors, there are two interpretations one can use, illustrated below. Let's call them "parallelogram" and "rectangle". 

image

image

You implemented the rectangle. I think this is great, and I'm not sure how to fit _both_ into a single function. I would have suggested the rectangle as well, as the parallelogram version is much easier to generate by a simple repetition of a "unit cell". That should be a different function that supports arbitrary unit cells, as well as periodicity.

I am spelling this out so we can be conscious of what we are doing and so that @ntamas can also give his opinion on these choices. There is no TODO for this.

----

A peculiarity of the "rectangle" version is that the `(n,m)` and the `(m,n)` graphs are not generally isomorphic. See the (2,3) and (3,2) ones below, according to the current implementation:

image

image

This means that we have to make a choice. Which of the above two graphs should (2,3) generate? Currently, it generates the first. However, unfortunately, the version I have already implemented for IGraph/M uses the exact opposite interpretation:

image

So now I have a dilemma: (1) Do we change this implementation to match mine, so I can smoothly transition `IGTriangularLattice` to this new implementation? (2) Do I keep everything as-is and accept that the Mathematica interface uses a different interpretation than all the rest? (3) Do I just break API in the Mathematica interface and change how it interprets things?

I'm wondering how time-consuming it would be to make the necessary change to go with (1) while also keeping the vertex numbering reasonable. Any input on this @ntamas ?

----

The above is all closely tied up with how vertices are numbered, which should be documented, so that high-level interfaces can add on vertex coordinates if they wish. Currently, this is what we have:

`(3)`:

image


`(3,2)`:

image

`(2,3)`:

image


`(3,3,3)`:

image


`(2,3,4)`:

image

`(4,2,3)`:

image

Documenting this order is TODO 3. (I used 1-based indexing in the plots above, but of course in C we work with 0-based.)

----

The final and very minor TODO 4 is to add a small ASCII art in the docs that clarifies what sort of graphs are being generated.

src/constructors/lattices.c
```diff
@@ -0,0 +1,299 @@
...
+#include "math/safe_intop.h"
+
+#define MIN(n, m) (n < m ? n : m)
+#define MAX(n, m) (n < m ? m : n)
+
+/**
+ * \ingroup generators
+ * \function igraph_triangle_lattice
+ * \brief A triangle lattice with the given shape.
+ *
```
szhorvat, line 30: Please add `\experimental` here. It means that we do not yet commit to the API, so in case we notice that some changes should be made, e.g. while implementing hex lattices, we can still do that.

src/constructors/lattices.c
```diff
@@ -0,0 +1,299 @@
...
+
+#include "core/interruption.h"
+#include "math/safe_intop.h"
+
+#define MIN(n, m) (n < m ? n : m)
+#define MAX(n, m) (n < m ? m : n)
+
+/**
+ * \ingroup generators
+ * \function igraph_triangle_lattice
```
szhorvat, line 28: Please remove this `\function` line as it might interfere with the one before the primary function. Remove the `\ingroup` as well.

src/constructors/lattices.c
```diff
@@ -0,0 +1,299 @@
...
+    IGRAPH_FINALLY_CLEAN(2);
+
+    return IGRAPH_SUCCESS;
+}
+
+/**
+ * \ingroup generators
+ * \function igraph_triangle_lattice
+ * \brief A triangle lattice with the given shape.
+ *
```
szhorvat, line 241: Please add `\experimental` here. It means that we do not yet commit to the API, so in case we notice that some changes should be made, e.g. while implementing hex lattices, we can still do that.

src/constructors/lattices.c
```diff
@@ -0,0 +1,299 @@
...
+ * Creates a triangle lattice whose vertices have the form (i, j) for non-negative integers i and j
+ * and (i, j) is connected with (i + 1, j), (i, j + 1), and (i - 1, j + 1) provided a vertex
+ * exists. Thus, all vertices have degree at most 6.
+ *
+ * </para><para>
+ * The vertices of the resulting graph are ordered lexicographically with the 2nd coordinate being
+ * more significant, e.g., (i, j) < (i + 1, j) and (i + 1, j) < (i, j + 1).
```
szhorvat, line 37:
```suggestion
 * more significant, e.g., (i, j) &lt; (i + 1, j) and (i + 1, j) &lt; (i, j + 1).
```

src/constructors/lattices.c
```diff
@@ -0,0 +1,299 @@
...
+ * \ingroup generators
+ * \function igraph_triangle_lattice
+ * \brief A triangle lattice with the given shape.
+ *
+ * Creates a triangle lattice whose vertices have the form (i, j) for non-negative integers i and j
+ * and (i, j) is generally connected with (i + 1, j), (i, j + 1), and (i - 1, j + 1).
+ *
+ * </para><para>
+ * The vertices of the resulting graph are ordered lexicographically with the 2nd coordinate being
+ * more significant, e.g., (i, j) < (i + 1, j) and (i + 1, j) < (i, j + 1)
```
szhorvat, line 247:
```suggestion
 * more significant, e.g., (i, j) &lt; (i + 1, j) and (i + 1, j) &lt; (i, j + 1)
```

src/constructors/lattices.c
```diff
@@ -0,0 +1,299 @@
...
+ * \param graph An uninitialized graph object.
+ * \param dims Integer vector, defines the shape of the lattice. (Below the "edge length"s are in terms of graph theoretical path lengths.)
+ *        If \p dims is of length 1, the resulting lattice has a triangular shape
+ *        where each side of the triangle contains \c "dims[0]" vertices.
+ *        If \p dims is of length 2, the resulting lattice has a
+ *        "quasi rectangular" shape with the sides containing \c "dims[0]" and
+ *        \c "dims[1]" vertices, respectively.
+ *        If \p dims is of length 3, the resulting lattice has a hexagonal shape
+ *        where the sides of the hexagon contain \c "dims[0]", \c "dims[1]" and
+ *        \c "dims[2]" vertices.
```
szhorvat, line 258: Let's change `\c "dims[0]"` to `<code>dims[0]</code>`. Necessary because of the limitations of `\c`.

szhorvat (changes requested):
In addition to the below notes on the docs, please add the function to `doc/generator.xxml`, right after `igraph_square_lattice()`.

szhorvat:
I'm quite happy with how this turned out! Would you like to have a go at implementing the _dual lattice_ of this, i.e. a hexagonal one? We should think carefully about how to parametrize it and how to interpret the parameters. The simple choice would be that for any set of parameters, `igraph_hexagonal_lattice()` would generate the dual of what `igraph_triangular_lattice()` produces.

szhorvat:
Final comments:

@rfulekjames, please stop force-pushing to this PR from now on, so we can also make some minor corrections if necessary, and and avoid conflict. Just push commits normally and we'll squash them before merging.

@ntamas If you have time, can you add this to `interfaces.yaml`?

ntamas:
Function added to `interfaces.yaml` and it was renamed to `igraph_triangular_lattice()`, FYI. Also merged `master` into this one.

src/constructors/lattices.c
ntamas: Done in my commit.

rfulekjames:
> So now I have a dilemma: (1) Do we change this implementation to match mine, so I can smoothly transition IGTriangularLattice to this new implementation? (2) Do I keep everything as-is and accept that the Mathematica interface uses a different interpretation than all the rest? (3) Do I just break API in the Mathematica interface and change how it interprets things?

@szhorvat 
It makes sense to have a parallelogram implementation for 2-dim inputs since it seems to fit better with the triangle and hexagon. The current implementation just closely follows the issue's description  #1842 
Then if jagged rectangular grids are also needed, we can create a separate function.
Please, let me know what you prefer.

szhorvat:
I prefer we keep the current jagged implementation, as it's less trivial to generate.

The parallelogram should be a separate future function (not in this PR), as it is easily generalizable to other lattices. In IGraph/M I have a function which can create many different meshes (also convertible to graphs) based on a database of many different unit cells. Perhaps eventually we can add something like this to the C core as well.

However, it would be convenient for me if we can easily flip the meaning of dimensions for the 2-parameter case, i.e. (m,n) -> (n,m). If it's not easy, let's not bother.

rfulekjames:
> However, it would be convenient for me if we can easily flip the meaning of dimensions for the 2-parameter case, i.e. (m,n) -> (n,m). If it's not easy, let's not bother.

That should be done just with exchanging the roles two variables, but I am not 100% sure if it would **always** produce the same result as your function.

szhorvat:
With a simple exchange, it would produce graphs that are isomorphic to those generated by `IGTriangularLattice[]`, but they would have a different vertex ordering.

rfulekjames:
> With a simple exchange, it would produce graphs that are isomorphic to those generated by IGTriangularLattice[], but they would have a different vertex ordering.

@szhorvat I did the exchange. To preserve the vertex ordering does not appear to be easily implementable without changing most of the current logic. Another option would be to somehow relabel the vertices after the graph is constructed.

ntamas:
@szhorvat If you are happy with this PR, please merge; I don't have any further comments and the last unsettled question is whether the new implementation is conformant enough with the current version in Mathematica.

szhorvat:
Please give me a couple more days until I can get to this.

rfulekjames:
> If you are happy with this PR, please merge; I don't have any further comments and the last unsettled question is whether the new implementation is conformant enough with the current version in Mathematica.

I found a way to change the vertex labels for the rectangle shaped lattice with only a minor change, although it introduced a new parameter to `triangular_lattice(.)`.

@szhorvat

szhorvat:
I'm sorry about the slow response. It's been a rough few days. I'll try to respond about this as well as the hex lattice tomorrow morning.

Regarding the hex lattice, basically my idea was to make the triangular lattice be the dual of the hex and not the reverse. I.e. the centre of each hexagon will be a vertex in the tri lattice. But in the hex lattice, some vertices around the boundary won't correspond to a triangle. I'll make a drawing tomorrow.

rfulekjames:
> I'm sorry about the slow response. It's been a rough few days. I'll try to respond about this as well as the hex lattice tomorrow morning.

No worries!

> Regarding the hex lattice, basically my idea was to make the triangular lattice be the dual of the hex and not the reverse. I.e. the centre of each hexagon will be a vertex in the tri lattice. But in the hex lattice, some vertices around the boundary won't correspond to a triangle. I'll make a drawing tomorrow.

Ok. I think I know what you mean.

szhorvat:
This looks great! I'd like to finish properly integrating it into the Mathematica interface (including coordinate generation), but no more input is needed from others.

ntamas:
Based on discussions with @szhorvat today, we decided to switch the vertex ordering of the rectangular lattice back to lexicographical to remain consistent with the other two options; @szhorvat will make a breaking change in the Mathematica interface later to make it conform to what the C core does.

I have extended the unit tests so they now validate the exact edge set returned from the functions. I'll merge the PR when all CI tests pass.

Thanks a lot for your hard work on this!

Issue #2236, opened 2022-10-12
Set up CI run with Clang 15


szhorvat:
Clang 15 has introduced many new warnings which broke compilation with the default of `IGRAPH_WARNINGS_AS_ERROR=ON`. I fixed these, but the risk of re-introducing problems is high, at least until we pick up some new habits such as using `int main(void)` instead of `int main()`. Therefore, there should be a CI run with Clang 15.

Ideally this should have ASan enabled and should replace the existing Clang run on Azure.

One possibility is https://apt.llvm.org/ but there should be some caching to avoid continual re-downloads.

Note:

 - CRAN now does checks with Clang 15
 - We already had a user point out Clang 15 warnings.

szhorvat:
Here's the software available on runners:

https://learn.microsoft.com/en-us/azure/devops/pipelines/agents/hosted?view=azure-devops&tabs=yaml

Simplest might be to use Linuxbrew. Pay attention to the note that says that it's not in the path by default.

Can you have a look at this @ntamas, since I don't use Homebrew myself, and it's not entirely clear how to ensure that the newly installed Clang would be in the path (even without the path issue mentioned above)?

ntamas:
I don't know what the connection is between Homebrew and Linuxbrew so I can't connect on that. One thing that's for sure is that Homebrew never installs alternative versions of OS-level tools such as `clang` into the system PATH; if you install it, it is stored in a versioned folder in `/opt/homebrew/Cellar`, and you need to add it to your PATH manually if you want to use it.

Homebrew prints this when installing `llvm`, which gives you a hint about how to put it on the PATH:

```
llvm is keg-only, which means it was not symlinked into /opt/homebrew,
because macOS already provides this software and installing another version in
parallel can cause all kinds of trouble.

If you need to have llvm first in your PATH, run:
  echo 'export PATH="/opt/homebrew/opt/llvm/bin:$PATH"' >> ~/.zshrc

For compilers to find llvm you may need to set:
  export LDFLAGS="-L/opt/homebrew/opt/llvm/lib"
  export CPPFLAGS="-I/opt/homebrew/opt/llvm/include"
```

I don't know how this translates to Linuxbrew (if at all).

vtraag:
> Here's the software available on runners:
> 
> https://learn.microsoft.com/en-us/azure/devops/pipelines/agents/hosted?view=azure-devops&tabs=yaml
> 
[...]

PR #2238, opened 2022-10-13
CI: Use clang15.

vtraag:
This is an attempt at addressing issue #2236.


vtraag:
It seems that this now correctly uses `clang-15` instead of `clang-14`. Perhaps you still want to change something else @szhorvat, with regards to ASan or something else?

ntamas:
Looks fine to me!

szhorvat:
This looks good to me. As I understand, it's not cached, but I think that's too much to ask anyway.

szhorvat (approved):

vtraag:
It seems that the build also passes, so that is great. We can also test with `clang-16` by the way, but that is probably not really necessary.

szhorvat:
No, let's not test with unreleased versions. It's not necessary. Clang 15 was important because people (and CRAN) complained already.

vtraag:
> it's not cached, but I think that's too much to ask anyway.

No, it's not indeed. But the setup only takes 24s in total, so I thought it wasn't worth the trouble anyway.

szhorvat:
I don't mind if it takes 3 minutes.

I was just never sure if we were supposed to be good citizens and keep bandwidth usage low. But given how GitHub actions are typically used, they probably don't care.

PR #2240, opened 2022-10-15
add A* pathfinding algorithm

### LINKED ISSUE #403: Add A* algorithm
Should be pretty easy. The A* (A star) shortest path algorithm achieved superior performance by using an estimate of the distance to the given target node.

https://en.wikipedia.org/wiki/A*_search_algorithm

---

- date created: 2009-08-09T07:38:08Z
- owner: gabor.csardi
- the launchpad url was https://bugs.launchpad.net/bugs/410955

GroteGnoom:
Fixes #403


include/igraph_paths.h
```diff
@@ -34,6 +34,8 @@
 
 __BEGIN_DECLS
 
+typedef igraph_error_t igraph_astar_heuristic_t(igraph_real_t *result, igraph_integer_t vertex_id, void *extra); //TODD where should this go?
```
szhorvat, line 37: I think it's fine here.

GroteGnoom:
I wonder how a user would use this. I'm used to graphs embedded in a euclidean space where you use the heuristic of 'distance = euclidean distance'. But for graphs it's hard to find heuristics, especially if it's a generated graph, because you know very little about how vertices are placed. Maybe we could help the user with this? I'm now just passing on vertex id's, maybe something else is needed? Maybe it's fine, and the user just needs to do the heuristic calculations up front and pass them to `extra`?

In my test I now use a heuristic that abuses my knowledge of how the graph is generated. (It doesn't find the shortest path, because I think it's not [admissible](https://en.wikipedia.org/wiki/Admissible_heuristic) ) This is just to check if it takes fewer steps, but it depends on the specific results of igraph_square_lattice, so we can't leave it in.

szhorvat:
I think the design is fine. A typical use case would be a spatially embedded graph, as you say. Suppose there are `x` and `y` vectors for the spatial coordinates of each vertex, possibly stored in an attribute. Then you pass these vectors in `extra`, and use the vertex index to index into them within the heuristic finder function.

Nice demonstrations / tests could be the following:

 - Use `igraph_square_lattice()`. The position of the vertex within the lattice can be computed from the vertex index, see docs: https://igraph.org/c/html/latest/igraph-Generators.html#igraph_square_lattice The Euclidean distance heuristic will be admissible.
 - Use `igraph_grg_game()`. It gives you `x` and `y` coordinates directly. Compute edge weights from them. With these weights the Euclidean distance heuristic will be admissible again, but without them it won't.

In both cases, do not make the structure periodic to avoid complications.

szhorvat:
The GRG example is fussier because the graph may not be connected, and the cutoff parameter should be chosen so that the graph is not too dense but not too sparse or too likely to be disconnected. In a high-level interface I'd just keep generating graphs until they become connected.

For your reference, with 100 points the graph may look like this for various cutoffs:

image

GroteGnoom:
Thanks! I didn't really realize we offered such vertex id guarantees as in square_lattice :slightly_smiling_face:

szhorvat:
For the lattice generators, eventually I would like to have a way to also generate node coordinates. This is very useful in many applications and in high-level interfaces one can add the coordinates to the graph immediately, as an attribute. This is what I do in the Mma interface:

image

These are not force-based layouts but pre-generated coordinates.

szhorvat:
Just a heads up that `igraph_is_nan()` was deprecated. Use `isnan()` from `math.h` instead.

ntamas:
> For the lattice generators, eventually I would like to have a way to also generate node coordinates. 

Coordinates are not always unique; there are several, equally meaningful coordinate assignments for certain types of lattices. For instance, a hexagonal lattice may exist in "flat topped" and "pointy topped" variants ([source](https://www.redblobgames.com/grids/hexagons/#basics)), and I wonder whether there's any reason to choose one over the other. I wonder whether we really want to go down this road; it feels a bit out of scope for me.

szhorvat:
Of course there is no single correct coordinate assignment. However, when I work with such graph, I often want to visualize them, and the best way to do so is to assign such coordinates. Force-directed layouts do not tend to do well, except for the tiniest lattices. During practical work, I find it extremely useful to have coordinates pre-assigned when the graph is generated. I can always remove them or replace them if the defaults are not suitable.

Illustration:

image

Coordinates can be returned by the same function, or there can be a companion function for each lattice generator that produces the coordinates. I would be quite happy with the second approach, but I'm worried that this complicates things when using interface generators.

iosonofabio:
Just my two cents, FWIW: I agree with @szhorvat that if you make a lattice (triangular or otherwise) that *can* be visualized in 2 or 3D chances are you *will* plot it sooner or later, or otherwise make use of the coordinates.

Whether the C core is the right place to introduce those coordinates (as opposed to Mathematica, which is a higher-level language with integrated plotting capability) I'm not sure, but telling folks that this is a "default" layout that they can use if they want seems like a good idea to me.

szhorvat:
If the coordinates are not generated in the C core, there should still be some guarantee about which vertex is at which position in the lattice. That will make it possible to generate the coordinates elsewhere without risk of the C core pulling the rug from under you. We already give such a guarantee with `igraph_square_lattice()`:

https://igraph.org/c/html/latest/igraph-Generators.html#igraph_square_lattice

(I noticed the slight mistake in the formula in the docs, it's already fixed.)

GroteGnoom:
It seems to work, but there's a lot of copied code. Maybe I should split off the functions that handle the parents/inbound edges/vertices/edges ? They're also used by `get_shortest_paths_dijkstra`.

GroteGnoom:
Interesting, 25 passed checks, but travis fails with
```
210/524 Test #211: test::igraph_get_shortest_paths_astar ..............***Failed    0.47 sec

Test exited abnormally with error: Floating-point exception

=========================================

CMake Error at /home/travis/build/igraph/igraph/etc/cmake/run_legacy_test.cmake:32 (message):

  Exiting test.
```

ntamas:
Hmmm, FPEs like this typically occur because of division-by-zero, modulo-by-zero or doing-something-illegal-with-NaN issues. In one particular case I've seen FPEs because Intel and ARM CPUs handle out-of-bounds conversions from floats to integers differently, and one particular conversion turned something into a zero on ARM but not on Intel. So try to think along these lines.

The only case where division or modulo operations happen in your code in non-trivial code paths is the lattice heuristic, so maybe that's the cause? Try disabling any test that involves this heuristic and check whether it still fails in Travis.

szhorvat:
Floating point division by zero is fine, and we rely on it in many scenarios. Same for Inf and NaN operations. So this is quite puzzling.

My suggestion is to enable UBSan on your local machine and run tests with it. By default, UBSan does not abort, only prints findings. Check here how to make it more useful:

https://clang.llvm.org/docs/UndefinedBehaviorSanitizer.html

As I remember, you need:

```
-fsanitize=undefined -fno-sanitize-recover=undefined
```

and set the envvar `UBSAN_OPTIONS=print_stacktrace=1:color=always`.

Note that older Clang (or possible Apple Clang) will complain about division by zero, but this is not a real issue on machines that support Inf/NaN. Newer Clang does not complain about this by default.

If this reveals nothing, we should try enabling ASan on Travis ...

szhorvat:
FYI I restarted the job, let's see if a second try succeeds.

szhorvat:
I'm sorry, I'm too tired, and I did not remember that I already enabled UBSan on CI ... so the above suggestion won't help ...

GroteGnoom:
The easiest way to get a floating point exception is to do integer division or modulo by zero.

szhorvat:
You're right. And that's a very confusing name!

GroteGnoom:
I suspect
```
        x[i] = vertex_id % *d;
        vertex_id /= *d;
```
can be the only issue, because that's where I divide by an integer. But the code to set `d` is very basic, it should just always be 10. I'm using `#define` to rule those out as a possible division by zero.

Else it might be an actual floating point issue.

GroteGnoom:
Travis now gives a
'Automatic restarts limited: Please try restarting this job later or contact support@travis-ci.com.'
after 367m, so 6 hours.

szhorvat:
Did you try an `IGRAPH_ASSERT(*d != 0)`?

szhorvat:
Not sure what to do with Travis. It's still giving me the same error right now. :-(

szhorvat:
If you have the patience, you might try requesting access to an s390x machine from IBM: https://www.ibm.com/community/z/open-source/virtual-machines-request/

ntamas:
All checks passed now -- please mark it as ready for review if I should take a look at it.

src/paths/astar.c
```diff
@@ -0,0 +1,331 @@
...
+    igraph_integer_t i, to_reach;
+
+    if (weights) { //If there are no weights, they are treated as 1.
+        if (igraph_vector_size(weights) != no_of_edges) {
+            IGRAPH_ERROR("Weight vector length does not match", IGRAPH_EINVAL);
+        }
+        if (no_of_edges > 0) {
+            igraph_real_t min = igraph_vector_min(weights);
+            if (min < 0) {
+                IGRAPH_ERROR("Weight vector must be non-negative", IGRAPH_EINVAL);
```
szhorvat, line 139:
```suggestion
                IGRAPH_ERROR("Weight vector must be non-negative.", IGRAPH_EINVAL);
```

src/paths/astar.c
```diff
@@ -0,0 +1,331 @@
...
+            }
+            else if (isnan(min)) {
+                IGRAPH_ERROR("Weight vector must not contain NaN values", IGRAPH_EINVAL);
```
szhorvat, line 142:
```suggestion
                IGRAPH_ERROR("Weight vector must not contain NaN values.", IGRAPH_EINVAL);
```

src/paths/astar.c
```diff
@@ -0,0 +1,331 @@
...
+    igraph_2wheap_push_with_index(&Q, from, -heur_res);
+
+    while (!igraph_2wheap_empty(&Q) && to_reach > 0) {
+        igraph_integer_t nlen, minnei;
+        /*The sum of the distance and the heuristic is the estimate.
+         *The estimates should be on the heap,
+         *because the minimum estimate should always be handled next.
+         *The value taken off the heap is ignored, we just want the index.
+         */
+        igraph_2wheap_delete_max_index(&Q, &minnei);
```
szhorvat, line 198: missing `IGRAPH_CHECK`

src/paths/astar.c
```diff
@@ -0,0 +1,331 @@
...
+        if (!is_target[ IGRAPH_VIT_GET(vit) ]) {
+            is_target[ IGRAPH_VIT_GET(vit) ] = 1;
+        } else {
+            to_reach--;       /* this node was given multiple times */
+        }
+    }
+
+    VECTOR(dists)[from] = 0.0;  /* zero distance */
+    IGRAPH_CHECK(heuristic(&heur_res, from, extra));
+    igraph_2wheap_push_with_index(&Q, from, -heur_res);
```
szhorvat, line 189: missing `IGRAPH_CHECK`

szhorvat:
Can you add a version for a single target, i.e. `igraph_get_shortest_path_astar()` (not `path`, not `paths`)? This is probably the most common use case and other algorithms have this variant.

src/paths/astar.c
```diff
@@ -0,0 +1,418 @@
...
+#include "igraph_memory.h"
+
+#include "core/indheap.h"
+#include "core/interruption.h"
+
+/**
+ * \ingroup structural
+ * \function igraph_get_shortest_paths_astar
+ * \brief Shortest paths from a vertex with heuristic.
+ *
```
szhorvat, line 35: I suggest to make it a policy to mark all non-trivial functions added in non-major versions (i.e. mid-cycle between 0.10.0 and 0.11.0) as `\experimental`. Then most these `\experimental` labels can be removed in 0.11.0.

include/igraph_paths.h
```diff
@@ -34,6 +34,8 @@
 
 __BEGIN_DECLS
 
+typedef igraph_error_t igraph_astar_heuristic_t(igraph_real_t *result, igraph_integer_t vertex_id, void *extra);
```
szhorvat, line 37: Would be nice to use a name for this type that makes it immediately obvious that it's a function.

szhorvat:
Anything I should pay particular attention to in the review @GroteGnoom, or any other helpful comments for the review?

GroteGnoom:
It's been a while since I looked at it thoroughly, so I'm not sure. I think the biggest problem is the copied code. I think the shortest path reconstruction can at least be split off as a function, which would then also be used by Dijkstra, and maybe also Bellman-Ford.

szhorvat (changes requested):
I am confused about the meaning of the parameters of the heuristic (distance estimate) function.

The algorithm needs to be able to estimate the distance from any vertex to the target vertex. So I assume that the `vertex_id` argument is the vertex we calculate the distance _from_ to a single fixed target, which is not passed to the function. The function is just expected to know it.

 - However, the `igraph_get_shortest_paths_astar()` function takes _multiple_ target vertices. Which one is the heuristic supposed to calculate the distance to?
 - Even independently of the previous point, wouldn't it be much more convenient if the heuristic function took two arguments, both a source and a target vertex?  I understand that the target vertex can in general be passed in the `extra` argument, but that gets inconvenient in practice. Suppose I want to find the path to multiple targets, maybe even in parallel. Suppose that vertices existing space and we use an Euclidean estimate. If  the distance estimate function takes both a source and a target argument, `extra` only needs to contain the vertex coordinates, and thus it becomes re-usable between multiple `shortest_paths_astar()` calls, even parallel ones. How things currently stand, `extra` would need to be updated with information about the target before each call.

----

Additional early comments:

 - Can you change the brief description to include "A*"? "A*" needs to appear in the function index so people can easily find it.
 - Can you document the `igraph_astar_heuristic_t` type, explaining the meaning of arguments precisely?
 - Can you improve the docs a bit, so that people can use the function even if they have minimal knowledge about the algorithm? The admissibility requirements on the distance estimate function should be explained (i.e. that it should not overestimate distances, only underestimate them).

szhorvat:
I think I understand what is being done here. If I got it right, paths are extended based on the distance estimate, but the search terminates based on the value of `to`. This is a conceptually complicated, and IMO confusing approach.  In the end, the path is guaranteed to be shortest only to one vertex (the one corresponding to the heuristic function), and it seems that there is little point in ever adding more than this vertex to the `to` vertex set.

A*, as usually described, works for a single source and a single target. Why don't we then just implement that, i.e. just `igraph_get_shortest_path_astar()` (singular)?  Is there any use case you see for the much harder to understand `igraph_get_shortest_paths_astar()`? I am concerned that people will believe that it will find the shortest path to all vertices given in `to`, but it might not.

So I suggest:

 - Implement a search towards only a single target, and return only a single path.
 - The distance estimate function should take both a source and target node
 - Should there be a `parents` vector? It would be a niche usage but perhaps this is still useful for some, as it provides information about how the algorithm explored the graph. If someone wants to get actual paths, they can reconstruct them based on `parents`.

When I looked at the NetworkX A* function, another point came up:

 - Will it speed up the algorithm if the results from the heuristic function are cached? Should we implement this caching (it's quite simple)? Consider that evaluating heuristic functions written in a high-level language may be expensive.
 - Does it make sense to use a heuristic function that changes / improves its estimate across subsequent calls, which would make caching be a problem? The NetworkX docs mention that this is not possible with their implementation due to caching, so I assume there's a use case for this approach.

GroteGnoom:
> The distance estimate function should take both a source and target node

Yes, you're right.

> Implement a search towards only a single target, and return only a single path.
 This is a conceptually complicated, and IMO confusing approach. In the end, the path is guaranteed to be shortest only to one vertex (the one corresponding to the heuristic function), and it seems that there is little point in ever adding more than this vertex to the to vertex set.

This is fixed by putting the target node in the heuristic, right? Then the vertex corresponding to the heuristic function doesn't exist anymore.

> Should there be a parents vector? It would be a niche usage but perhaps this is still useful for some, as it provides information about how the algorithm explored the graph. If someone wants to get actual paths, they can reconstruct them based on parents.

It's already being generated in the multiple-endpoints version, so it's easy to pass it on. I'm not sure why it isn't passed in Dijkstra.

szhorvat:
> This is fixed by putting the target node in the heuristic, right? Then the vertex corresponding to the heuristic function doesn't exist anymore.

But the algorithm will still only work for a single target. When the path is extended, we try to extend towards one specific target vertex. At least the plain version of the algorithm works with only one target, and I don't immediately see how to generalize to multiple targets. That is not a trivial problem.

So to fix the issue, we also need to stick to the single-target version.

ntamas:
> But the algorithm will still only work for a single target.

It could actually work for multiple targets, all that you need is a heuristic that receives _all_ targets and returns an admissible estimate of the distance to the _closest_, yet-unvisited target. Note that in the typical A* usage with multiple targets, you are interested in a _single_ path that leads to _any_ of the target nodes. The classic example is an agent foraging for food in a 2D or 3D grid-like world; it knows which cells contain food and uses the Manhattan distance (or any other admissible distance function) to the nearest cell as a heuristic to drive the search.

So, I wouldn't rule out the possibility of using multiple targets, but the semantics would be different from all the other shortest-path functions.

szhorvat:
Do you have a provably correct solution for handling multiple targets and obtaining _the shortest_ path to each?  I.e. not just something that may be sufficient for routing in typical practical scenarios, but an algorithm that makes use of the distance estimate for a speedup and is provably yielding the shortest paths?

ntamas:
> Do you have a provably correct solution for handling multiple targets and obtaining the shortest path to each?

I don't think you have any better option than running A* multiple times, each time with a different target node. The output of the heuristic function inherently depends on the target being considered, so any solution that attempts to handle multiple targets in a single A* run must necessarily use a heuristic function that knows about _all_ targets and returns an estimate based on that.

szhorvat:
In that case, isn't it cleaner (and less confusing to the user) to implement the single-target version only? 

If in the future we figure out some smart solution for multi-target scenarios, we can always add that. That would be research level stuff though (which I'm in favour of, but it takes time).

Refactoring this to support only the single-target situation won't take much work, it's mainly just a question of reorganization. No work will be lost.

ntamas:
I'm okay with refactoring this to a single-target version.

GroteGnoom:
> Should there be a parents vector? It would be a niche usage but perhaps this is still useful for some, as it provides information about how the algorithm explored the graph. If someone wants to get actual paths, they can reconstruct them based on parents.

Do we need a parent and inbound edge? If I understand correctly, for the `paths` version of Dijkstra, this is useful because it's less work to reconstruct them. But this information is already in the `vertices` and `edges` data, right?
Do we want this for the single `path` version in a*?

szhorvat:
> Do we need a parent and inbound edge?

I think it's nice to keep it if it's easy, especially if it's computed internally. It provides some educational insight into what A* actually did, and how it explored the graph.  But it's not a must.

szhorvat:
The nice thing about keeping the parent vectors is that using that information one could create animations like the one here: https://en.wikipedia.org/wiki/A*_search_algorithm#/media/File:Weighted_A_star_with_eps_5.gif These are very nice for educational purposes.

GroteGnoom:
> The nice thing about keeping the parent vectors is that using that information one could create animations like the one here: https://en.wikipedia.org/wiki/A*_search_algorithm#/media/File:Weighted_A_star_with_eps_5.gif These are very nice for educational purposes.

I don't think we're talking about the same thing. If I look at the parents as defined in `igraph_get_shortest_paths_dijkstra()`, I don't think it gives any extra information about the search path. This is in the docs:

> *        If not null, a vector containing the parent of each vertex in
 *        the single source shortest path tree is returned here.
 
So it only give information about the shortest path, right? Not other things that have been tried.

szhorvat:
Hm, my understanding was that `parents` represents the search tree, which includes more nodes than those on the shortest path from source to target. At first I thought highlighting successively deeper levels of the search tree would replay the search, but now I see that this is not really the case. If the search in one direction reached a dead end, then it will be paths that were tried earlier that will be extended.

So perhaps this information won't be useful even for educational purposes. Let's skip it then, i.e. _not_ use the `parents` output argument.

ntamas:
@GroteGnoom Is this ready for review? I'm about to do a 0.10.3 soon and we can get this into the next release if I can merge it in the next few days. Otherwise it will have to wait until 0.10.4 or 0.11.0.

GroteGnoom:
> @GroteGnoom Is this ready for review? I'm about to do a 0.10.3 soon and we can get this into the next release if I can merge it in the next few days. Otherwise it will have to wait until 0.10.4 or 0.11.0.

There's still a lot of little things to clean up. Maybe it's ok to merge it, because the interface could be fine, and it also says 'experimental'. I'm not sure what's nicest for our users.

szhorvat:
We would remove the `parents` and `inbound_edges` output parameters, as you suggested, right?

GroteGnoom:
> We would remove the `parents` and `inbound_edges` output parameters, as you suggested, right?

Yes, definitely.

szhorvat:
I already made some edits to this locally, and I will finalize this PR—please do not push commits for the moment.

But first I'd like to discuss what to do when there is no path from the source to the target. Right now, it seems to me that a partial path will still be returned. A warning is given. My suggestion is to keep the warning, but make sure that empty vectors are returned instead of partial paths. IMO there isn't really any justification for outputting a partial path. Other `get_shortest_path()` functions also return empty vectors.

szhorvat:
Thinking aloud here:

What would be the complexity of this if the distance estimator is perfect, always returning the accurate distance to the target? Then the algorithm needs to take only as many steps as the length of the path. Of course, it still needs to consider all neighbours of each vertex along the path, but it does not need to consider _all_ vertices.

However, it still checks that weights are non-negative, giving an $O(E)$ component, and it still initializes the `dist` vector giving an $O(V)$ component. I guess this is not worth worrying about in practice, as this initialization should be very fast ...

GroteGnoom:
> But first I'd like to discuss what to do when there is no path from the source to the target. Right now, it seems to me that a partial path will still be returned. A warning is given. My suggestion is to keep the warning, but make sure that empty vectors are returned instead of partial paths. IMO there isn't really any justification for outputting a partial path. Other get_shortest_path() functions also return empty vectors.

Seems fine to me. Else we'd just be returning some nonsense.

> Thinking aloud here:
> 
> What would be the complexity of this if the distance estimator is perfect, always returning the accurate distance to the target? Then the algorithm needs to take only as many steps as the length of the path. Of course, it still needs to consider all neighbours of each vertex along the path, but it does not need to consider _all_ vertices.
> 
> However, it still checks that weights are non-negative, giving an O(E) component, and it still initializes the `dist` vector giving an O(V) component. I guess this is not worth worrying about in practice, as this initialization should be very fast ...

That could even also happen with Dijkstra, right?
Just to illustrate it with an example, you are using this for a game, you're in this huge open world, and the player wants to take a few steps to the right. Is it a problem that every edge weight is checked and every vertex dist is initialized? Could be! And it's not some weird edge case. Good point :)

szhorvat (approved):

szhorvat:
> That could even also happen with Dijkstra, right?

Yes, but Dijkstra is meant to compute paths from one vertex to all others anyway. Normally it processes all vertices and all edges anyway (though we do have a feature to cut the search short if all targets have been reached).

Regarding initializing a value for all vertices, it's not just the `dist` vector, but also the `2wheap`. We need the indexed heap for efficient updates of distances.

As for the edge weight checks, they could be done at the same time as accessing each edge weight during the search

ntamas:
Will review this tomorrow - it's been a busy day today. Thanks for bringing this to my attention.

I'm starting an update to the `master` branch first, that should be harmless.

szhorvat:
A short technical review will be sufficient. I think that everything is fine here, but it never hurts to have another pair of eyes.

szhorvat:
Good idea with making the heuristic optional, @ntamas !

ntamas:
Thanks a lot everyone!

ValZapod:
So how much faster it is, considering A* is proved to be optimal possible?

ntamas:
A*'s optimality depends on the heuristic that you feed to it. If the heuristic is admissible and consistent, then it is guaranteed that it won't expand nodes that are not necessary to reach the destination. But, if you have such a heuristic, then you could run Dijkstra with `w'(x,y) = w(x,y) + h(y) - h(x)` (where `h(x)` is the heuristic, and `w(x,y)` are the original weights), and then Dijkstra would find the shortest path without expanding unnecessary nodes either.

As for practical benchmarks, we haven't done any yet, but one would need to be careful with a direct comparison to Dijkstra as Dijkstra's algorithm gives you one shortest path from the source node to _any_ target node that it has happened to reach, while A* would give you the shortest path from the source to the target only as the value of the heuristic depends on the target node in all practical use-cases.

ValZapod:
>A*'s optimality depends on the heuristic that you feed to it.

I am just talking about it being optimal in this sense: https://stackoverflow.com/a/29470434/11173412

Issue #2243, opened 2022-10-22
 UMAP attractive force and positive sampling scheme


iosonofabio:
**Describe the bug**
I found a few bugs in our current, experimental implementation of UMAP.

Basically, we currently attract a vertex each epoch with force mu * force, where mu is the weight of the edge. Instead, what we should be doing is to attract it with force (without the mu), but only every mu epochs. It makes a difference because the learning rate is decreasing, and it also makes it faster.

I have a working fork on my laptop, so let me make a proper PR in the next days against this repo.

note to self: each time we attract we also repel, so rarer and stronger attractions also means less repulsion.

**Version information**
`master`

Note: xferred from the `python-igraph` repo.

GroteGnoom:
> we currently attract a vertex each epoch with force mu * force, where mu is the weight of the edge. Instead, what we should be doing is to attract it with force (without the mu), but only every mu epochs.

Just to check: So now a bigger mu means _more_ force is applied, but it should mean _less_ force (same force less often) is applied?

iosonofabio:
Sorry, sloppy explanation.

Mu is the same, it's a probability btw 0 and 1. Instead of moving a vertex by force * mu each epoch, like tic tic tic, we can touch it every 1/mu epochs and move it by a larger amount, like wait wait wait bang! That ends up making much tighter clouds.

Also, I want to improve the writing of a few internal functions and the negative sampling needs to happen way less often and I think more freely.

I'll start a PR with the various elements and describe them in some detail.

GroteGnoom:
Ah, so it should be 'every 1/mu epochs'.

iosonofabio:
Oops, thankfully GitHub allows me to edit the past to fit the present...

szhorvat:
Can this be closed @iosonofabio ?

Issue #2229, opened 2022-10-05
Proper weight support for igraph_community_edge_betweenness()


szhorvat:
`igraph_community_edge_betweenness()` implements the Girvan-Newman algorithms which builds a dendrogram by iteratively removing the highest betweenness edges and checking how the graph falls to pieces. Then the dendrogram is cut so as to maximize modularity.

When weights are passed to this function, it uses these weights as _edge lengths_ during the betweenness calculation. Therefore, high-weight edges are interpreted as representing weaker connections. During the modularity calculation, high-weight edges are interpreted as stronger connections, which is contradictory. We have discussed this before, and the stopgap measure was that the function issues a warning when it is given some weights, and a membership vector is requested. In fact, it issues two warnings, which looks a bit sloppy.

https://github.com/igraph/igraph/blob/develop/src/community/edge_betweenness.c#L506-L516

----

This is a proposal for dealing with this problem. There are actually weighted generalizations of the method in the literature.

Section 5.1 of https://doi.org/10.1016/j.physrep.2009.11.002 says:

> The method can be simply extended to the case of weighted graphs, by suitably generalizing the edge betweenness. The betweenness of a weighted edge equals the betweenness of the edge in the corresponding unweighted graph, divided by the weight of the edge [159].

Ref [159] is https://doi.org/10.1103/PhysRevE.70.056131 and says:

> The resulting algorithm is identical to that for the integer case: betweennesses are simply divided by the weight of the edge and the edge with the highest resulting score removed from the network.

----

In order to preserve the existing functionality of igraph while also interpreting weights reasonably, I propose to change `igraph_community_edge_betweenness()` to take _two_ weight-type arguments:

```c
igraph_error_t igraph_community_edge_betweenness(const igraph_t *graph,
                                      igraph_vector_int_t *result,
                                      igraph_vector_t *edge_betweenness,
                                      igraph_matrix_int_t *merges,
                                      igraph_vector_int_t *bridges,
                                      igraph_vector_t *modularity,
                                      igraph_vector_int_t *membership,
                                      igraph_bool_t directed,
                                      const igraph_vector_t *weights,
                                      const igraph_vector_t *lengths);
```

[...]

PR #2244, opened 2022-10-24
Sampling every 1/mu epochs, force functions cannot fail

iosonofabio:
There are a few outstanding bugs in our UMAP implementation, as outlined in #2243. This PR aims at fixing them.

It'd be great to not break API, however the `sampling_prob` is now not used anymore so I think it'd be better to just make a clear break here rather than drag relics around. The function is "experimental" anyway, so API breaks are expected outside of major versions.

This PR also improves readability of the force functions and uses less temporary vectors, which is nice.

A few outstanding things:
- [x] writing a proper test function. Turns out this should have at least a few hundred vertices for the algo to really kick in given the approximations. That should be ok for CI since runtime is around 5ms/epoch on a graph with 1k vertices and 20k edges and reasonable convergence can be reached usually << 100 epochs.
- [x] Documentation overall in the C file can be improved as @szhorvat suggested before. This is tricky because UMAP's paper has a very different flavour from what we usually do. They barely ackowledge that 1-simplices are edges of a graph and that UMAP is by and large a graph layout algo. We probably have to strike a balance between using igraph-style lingo (edges, vertices, edge weights) and the original lingo (1-simplices, 0-simplices, connectivities). Let's see what we can do.

Also, if we break API we might as well add a few additional arguments if you guys like, e.g.:
- [ ] number of negative samples for each positive one, currently 5

@GroteGnoom I'd be grateful for an opinion if you have time one of these days


iosonofabio:
I've written a test for my python extension fork, I'll try to convert it to a C test:

```python
    def testLarge(self):
        # Loosely connected clouds of highly connected vertices
        cloud_sizes = [50, 100, 30, 20]
        n_vertices = 0
        edges = []
        distances = []
        for cloud_size in cloud_sizes:
            for v1 in range(cloud_size):
                for v2 in range(1, 11):
                    edges.append(
                        (v1 + n_vertices, v2 + n_vertices),
                    )
                    distances.append(2)
            n_vertices += cloud_size

        g = Graph(edges)
        lo = treasuremap_igraph(
            g,
            dist=distances,
            epochs=50,
        )
        self.assertTrue(isinstance(lo, Layout))

        n_vertices = 0
        for cloud_size in cloud_sizes:
             err = np.std(lo[n_vertices: n_vertices + cloud_size], axis=0)
             self.assertTrue((err < 1.0).all())
             n_vertices += cloud_size
```

szhorvat:
There is a breaking change here. Be sure to document that in the changelog, and point out that the change is made to an experimental function (just a reminder).

We will surely get some flak for making such breaking changes in "patch" releases, even if we made it clear that we may do so for experimental functions. The problem is that UMAP is already exposed in the Python interface. Because of this breaking change, py-igraph 0.10.3 won't be compatible with igraph 0.10.2 and vice versa: the current py-igraph 0.10.2 won't work with the future igraph 0.10.3. From now on, it will be difficult to determine which py-igraph releases are compatible with which C core releases, especially if version numbers also get out of sync in the future.  This makes me a bit uncomfortable.

To be clear, I do support having "experimental" functions and reserving the right to make such breaking changes, at least while we're in the 0.x stage. I plan to make another such breaking change in the minimum cycle basis functions for 0.10.3. But I wanted to point out the difficulty that stems from py-igraph already making use of these experimental functions (both UMAP and cycle bases). When I originally proposed the idea of experimental functions, I did not consider py-igraph.

ntamas:
> py-igraph 0.10.3 won't be compatible with igraph 0.10.2 and vice versa: the current py-igraph 0.10.2 won't work with the future igraph 0.10.3. From now on, it will be difficult to determine which py-igraph releases are compatible with which C core releases

In the Python interface, there's only one guarantee ever: the version of igraph that is vendored with the Python interface _is_ compatible; otherwise all bets are off. The very reason why we've introduced experimental functions is to be able to play around with their APIs while the functions are being tested in real scenarios, and it's really hard to do that without exposing them in higher level interfaces, at least temporarily.

If we want to be somewhat nicer to maintainers who unbundle the vendored igraph from the Python interface and link to an external one instead, we could keep the current signature until 0.11, mentioning that the `sampling_prob` argument is now ignored, and add all the rest of the knobs in 0.11 only.

szhorvat:
> In the Python interface, there's only one guarantee ever: the version of igraph that is vendored with the Python interface _is_ compatible; otherwise all bets are off.

Can you document this explicitly somewhere so that when distro maintainers come screaming bloody murder, we have something to point to?

> If we want to be somewhat nicer to maintainers who unbundle the vendored igraph from the Python interface and link to an external one instead, we could keep the current signature until 0.11, mentioning that the `sampling_prob` argument is now ignored, and add all the rest of the knobs in 0.11 only.

To be clear, I do strongly support having experimental functions, and I think we should go ahead with this PR for 0.10.3, exactly for the reasons you stated:

> The very reason why we've introduced experimental functions is to be able to play around with their APIs while the functions are being tested in real scenarios, and it's really hard to do that without exposing them in higher level interfaces, at least temporarily.

----

I was just concerned about the confusion all this may cause. I suggest to document what you said above, i.e. that the only version guaranteed to be compatible is the vendored one, and make sure that the compatible C core version is always states in the release notes (so far it has been).

ntamas:
Added a section in the README of the Python interface [here](https://github.com/igraph/python-igraph/commit/8bc5d0aa07a85de4c27c72080e7c288cac850233). Feel free to change or improve the wording.

iosonofabio:
Great discussion guys, this type of problem will come up again and again.

I'm happy to take the blame from our users if they mind the break.

iosonofabio:
Ok, it now works pretty well and all tests pass.

If you have requests we can try out a few more things here, otherwise the "classical" UMAP is working fine here and I can move on to making a PR for the Python repo ;-)

iosonofabio:
oh actually there is one thing. The official UMAP implementation splits what we do here in two steps:

1. compute the weighted "connectivity" graph from the distance graph
2. layout the connectivity graph

In the standard software used by many single cell people, they also split these two steps quite clearly. So it would be normal for those users to expect us to do the same, i.e. have two public functions instead of one.

I think it's a bit of a gimmick but not terrible and can be done easily. Any other opinions?

ntamas:
No preferences from me either way; I'd probably keep the current function that does the two things at the same time, _and_ provide two additional functions that just do one step only. The current function can then call these two, one after the other.

ntamas:
Quick code review: seems correct to me, without knowing much about the UMAP internals. That being said, I think it's safe to merge once you are happy with it.

src/layout/umap.c
```diff
@@ -94,31 +94,91 @@ static igraph_error_t igraph_i_umap_find_sigma(const igraph_vector_t *distances,
...
+
+    if (distances == NULL) {
+        return IGRAPH_SUCCESS;
+    }
+
+    if (igraph_vector_size(distances) != no_of_edges) {
+        IGRAPH_ERROR("Distances must be the same number as the edges in the graph.", IGRAPH_EINVAL);
+    }
+
+    for (igraph_integer_t eid = 0; eid != no_of_edges; eid++) {
```
ntamas, line 108: We have `igraph_vector_all_ge()` so you can use that instead of rolling your own loop. (It is also NaN-aware).

iosonofabio: Thanks. Hmm, that would require allocating a vector of the same length full of zeros though, wouldn't that be slower than this one?

szhorvat: @iosonofabio Use `igraph_vector_min()`, see how it's done in `igraph_closeness()` for reference.

src/layout/umap.c
```diff
@@ -91,28 +149,92 @@ static igraph_error_t igraph_i_umap_find_sigma(const igraph_vector_t *distances,
...
+
+/**
+ * \function igraph_layout_umap_compute_connectivities
+ * \brief Compute connectivities (edge weights) for a UMAP layout starting from distances.
+ *
+ * \experimental
+ *
+ * UMAP is used to embed high-dimensional vectors in a low-dimensional space
+ * (most commonly 2D). It uses a distance graph as an intermediate data structure,
+ * making it also a useful graph layout algorithm. See igraph_layout_umap for more information.
```
szhorvat, line 183:
```suggestion
 * making it also a useful graph layout algorithm. See \ref igraph_layout_umap() for more information.
```

Use `\ref` to refer to functions. Add `()` after function names to indicate functions clearly.

src/layout/umap.c
iosonofabio: I don't want the min though, I want to quit as soon as a negative number is found, which is faster.

src/layout/umap.c
```diff
@@ -91,28 +149,92 @@ static igraph_error_t igraph_i_umap_find_sigma(const igraph_vector_t *distances,
...
+    return IGRAPH_SUCCESS;
+}
 
-/* Convert the graph with distances into a probability graph with exponential decay */
-/* NOTE: this is funny because the distance was a correlation to start with... ?? */
-static igraph_error_t igraph_i_umap_find_prob_graph(const igraph_t *graph,
-        const igraph_vector_t *distances, igraph_vector_t *umap_weights) {
+
+/**
+ * \function igraph_layout_umap_compute_connectivities
```
szhorvat, line 176: I read this doc section, but I cannot tell what "connectivity" means here, or what this function does at all. Can you explain concisely what is being calculated?

After that we should think about whether the word "connectivity" can be avoided. This word is abused too much and used for entirely unrelated things by different papers. I guess the UMAP paper uses this term?

iosonofabio: Given a graph with distances between vertices associated with the edges, UMAP wants to move those distances into a probability space. So it wants to know what's the probability that vertices a and b are *connected* (topologically) given there is an edge with distance d between them.

Now if a has only one neighbor and that is b, then either a will end up a singleton (which UMAP tries to avoid) or it's connected to b, i.e. *the connectivity associated with that edge (which has distance d) is 1.0". If a has other neighbors at shorter distance, then the probability that b is connected becomes lower, since a is already connected to something else with higher priority.

That's the idea in a nutshell. Connectivity is how they call it in the paper and in the original UMAP implementation, so I moved away from our previous lingo "umap weights" to "connectivity" to make it easier for our average user who comes from that perspective.

src/layout/umap.c
szhorvat: Is it necessary to optimize error conditions?  Error handling is already slow. This is true not only for igraph, but for almost any system, including the exception handling of most languages.

The pattern I referenced is used everywhere else in the library.

iosonofabio: Well, I just don't see why this needs to change. What if there's a NaN? What will the min function do? And will that be clear to the next person reading the code or do we have to add a comment explaining what happens in that case?

Whereas like this it's crystal clear...

szhorvat: > Well, I just don't see why this needs to change.

Leave it as it is then.



> What if there's a NaN?

Please check the function I referenced (`igraph_closeness()`), and it will be clear to you.



> And will that be clear to the next person reading the code

That is precisely the point. People read code in patterns. This check is done identically in all other igraph functions, _except_ this one. When I see your code, I have to take a moment to check if it's correct. If I see the same pattern that I've seen many times before, I know it's fine, and I don't have to stop to think about it.

src/layout/umap.c
szhorvat: Can you document how this transformation from "distance" to "connectivity" is done exactly?

ntamas: Unless it's too long, in which case we should probably just point the reader to the UMAP paper. I don't know whether the concept can be summarized in a few sentences.

src/layout/umap.c
iosonofabio: I now changed it to match other functions and moved it to the body of the caller, for simplicity. Thanks for the suggestion.

src/layout/umap.c
iosonofabio: yeah, we can't explain all the nitty gritty details here. People should read the paper if they want to understand the ins and outs of UMAP, with formulae, examples, and all.

szhorvat: I looked at the paper and searched for "connectivity". I was unable to find the concept you implement here.

Where is "connectivity" defined under this name in https://arxiv.org/abs/1802.03426, and how might one find out what it is?

iosonofabio: First you renormalize the distances via rho and sigma, see algorithms 2 and 3 in the paper. Then you compute the connectivity as:

image

szhorvat: Note to self: that formula is on page 21 in the linked preprint.

iosonofabio: see e.g. https://github.com/lmcinnes/umap/blob/a7118a004b403b8cc9d9876854fd9d7f95f7ea94/umap/umap_.py#L151

szhorvat: Yes, this is too complicated to describe in the docs, but there should still be enough information there to allow one to find the part of the paper which explains what this function does.

At this moment my biggest concern is the unusual usage of the term "connectivity". You use this to describe a quantity associated with edges, but the above paper does not use this term this way (am I missing something?). Did you borrow this term from another paper perhaps? Can this term be replaced with something that relates to the paper better?

szhorvat: > if you actually want to go through the math

I don't have time to do that now, but I want to make sure that if I wanted to do it, or even just use this function, it would be sufficiently easy for me to determine what it does after reading the docs. It should at least have non-confusing naming and enough pointers to the paper.

iosonofabio: What name do you propose? I think connectivity is just fine but am open to concrete suggestions.

szhorvat: I have to log off now. Does the paper not use any name for this quantity? Is this what it means by the "membership strength" of a 1-simplex?  Otherwise why can't we just use "weight"?

I am not aware of any comparable use the term "connectivity" anywhere, but maybe I'm missing something.

szhorvat: I should note that it is a very common situation in network science that we have a quantity associated with edges so that a large value represents a weaker connection (commonly a length / idstance), but for some algorithm we need a quantity where a large value represents a stronger connection. In this case the transform to this second form can be called "weight" or to be more specific, "(connection) strength".

Do I understand it that we have a similar situation here?

We have a similar case in https://github.com/igraph/igraph/issues/2229 as well as in a community detection algorithm I'm implementing now (no PR yet). In these two cases I went with "length" and "weight".

iosonofabio: "umap_weight" is what we have in master ATM. I can easily revert to that. It's some kind of strength of the type you are mentioning, yes.

ntamas: I think that we are diverging from the original topic of the PR here, which would have been to fix an error in the sampling process. IMHO it is totally fine to merge this PR even if the terminology is not 100% correct as long as we eventually fix it in the future (and if it is part of the public API, we should of course merge it before a release is made).

iosonofabio: fair enough. What's happening is that @szhorvat had requested better docs for a while and I agreed to improve them a little. I think we are there now. I'll go back to the "umap_weight" terminology and then we can merge.

szhorvat: This function _is_ part of the public API, and is added by this PR.

If it's merged as is, it'll likely stay as is, and any update will be forgotten or dismissed.

iosonofabio: btw I finally remembered where I borrowed the term "connectivity" from:

https://github.com/scverse/scanpy/blob/e5cdbbc02702d779352cbbdfa8e54020d93301ea/scanpy/tools/_umap.py#L107

src/layout/umap.c
```diff
@@ -92,27 +150,82 @@ static igraph_error_t igraph_i_umap_find_sigma(const igraph_vector_t *distances,
...
+ * (most commonly 2D). It uses a distance graph as an intermediate data structure,
+ * making it also a useful graph layout algorithm. See \ref igraph_layout_umap() for more information.
+ *
+ * </para><para>
+ *
+ * An early step in UMAP is to compute exponentially decaying "weights" from the
+ * distance graph. Connectivities can also be viewed as edge weights that quantify
+ * similarity between two vertices. This function computes weights from the
+ * distance graph. To compute the layout from precomputed weights, call
+ * igraph_layout_umap with the "distances_are_weights" argument set to true.
```
szhorvat, line 169:
```suggestion
 * \ref igraph_layout_umap() with the \p distances_are_weights argument set to \c true.
```

src/layout/umap.c
```diff
@@ -161,26 +274,26 @@ static igraph_error_t igraph_i_umap_find_prob_graph(const igraph_t *graph,
...
-         * If no, set the P_{->}. If yes, it contains P_{->} and now we can substitute
+         * If not, set the P_{->}. If yes, it contains P_{->} and now we can substitute
          * it with the final result.
          *
          * */
-        for (igraph_integer_t j = 1; j < no_of_neis; j++) {
+        for (igraph_integer_t j = 0; j < no_of_neis; j++) {
             eid = VECTOR(eids)[j];
             /* Basically, nodes closer than rho have probability 1, but nothing disappears */
             weight = sigma < 0 ? 1: exp(-(VECTOR(*distances)[eid] - rho) / sigma);
```
szhorvat, line 284:
```suggestion
            weight = sigma < 0 ? 1 : exp(-(VECTOR(*distances)[eid] - rho) / sigma);
```

(cosmetic nitpick, sorry)

src/layout/umap.c
```diff
@@ -92,27 +150,82 @@ static igraph_error_t igraph_i_umap_find_sigma(const igraph_vector_t *distances,
...
+ *
+ * </para><para>
+ * References:
+ *
+ * </para><para>
+ * Leland McInnes, John Healy, and James Melville. https://arxiv.org/abs/1802.03426
+ *
+ * \param graph Pointer to the distance graph.
+ * \param distances Pointer to the vector with the vertex-to-vertex distance associated with
+ *   each edge.
```
szhorvat, line 184: Can you please mention that this is allowed to be `\c NULL`?

src/layout/umap.c
```diff
@@ -92,27 +150,82 @@ static igraph_error_t igraph_i_umap_find_sigma(const igraph_vector_t *distances,
...
         }
         return IGRAPH_SUCCESS;
     }
-    /* alright, the graph is weighted */
+    /* alright, the graph has actual distances */
 
-    /* Initialize vectors and matrices */
+    /* Initialize auxiliary vectors */
     IGRAPH_VECTOR_INT_INIT_FINALLY(&eids, 0);
     IGRAPH_VECTOR_INT_INIT_FINALLY(&weight_seen, no_of_edges);
```
szhorvat, line 230: Minor question: This is an integer vector, but it is used as a Boolean in practice, i.e. the counts are recorded but it's only ever checked whether they're zero or not.  Are you planning to use the counts later, or can it be changed to an actual Boolean vector?

iosonofabio: I should make it a bool

src/layout/umap.c
iosonofabio: done

iosonofabio:
@ntamas you beat me to it - had already fixed it but forgot to push ;-)

I'll let this one sit over the weekend to see if anything of note comes to mind that needs to improve further, if not we can merge next week. Thank you

szhorvat:
Let's not forget the changelog entry for this (can be after the PR was squash-merged).

iosonofabio:
I just found another bug in the computation of the weights, so I'll wait until that's fixed before pushing this one in. Converting back to draft for now.

iosonofabio:
I fixed the bug in the computation of the UMAP weights. It turned out to be a more interesting story than anticipated.

Just FYI. The graph with distances is usually directed, e.g. a knn graph in which each edge is between a *focal* vertex and a *neighbor* vertex. When UMAP computes weights from distances, it symmetrizes the graph into an undirected one. Because weights are probabilities, the idea is that either edge counts as a good shot:
```
W = P(edge) := P(-> | <-) = P(->) + P(<-) - P(->) * P(<-) = W1 + W2 - W1 * W2
```
They call it "fuzzy union"... anyway elementary probability math. Now because symmetrization is a lossy operation, you might need fewer weights than you had distances. To avoid having to construct a separate graph, I set the weight for the redundant edges to zero and then skip them in the stochastic gradient descent for the layout itself.

I tested this against the original weight computation in UMAP and the results are 99.983% correlated. There are some boundary effects that I think have to do with numerical approximation (our own gradient descent for the a/b parameters tends to get stuck in a narrow valley and bounce around), but nothing we should worry about.

I also explained that function a bit better in the docstring.

If this passes I'll flag ready for review and we can merge. All tests pass locally.

iosonofabio:
I think this is done, let's merge?

ntamas:
Let me go through it once again with the recent commits, and then I'll merge.

ntamas:
Thanks!

szhorvat:
> Let's not forget the changelog entry for this (can be after the PR was squash-merged).

This is a breaking change. Please update the changelog.

PR #2245, opened 2022-10-24
Fix unused-but-set-parameter warning

folded:
PE is otherwise unused (and is marked as such) so modifying its value serves no purpose, and also triggers -Wunused-but-set-parameter. Remote the unnecessary postincrement.

### DIFF
--- a/src/cliques/maximal_cliques.c
+++ b/src/cliques/maximal_cliques.c
@@ -308,7 +308,7 @@ static igraph_error_t igraph_i_maximal_cliques_up(
         VECTOR(*PX)[vvpos - 1] = tmp;
         VECTOR(*pos)[vv] = XS + 1;
         VECTOR(*pos)[tmp] = vvpos;
-        PE++; XS++;
+        XS++;
     }
 
     return IGRAPH_SUCCESS;


szhorvat:
Indeed the `PE` parameter does nothing at the moment, but the warning should not appear because of the `IGRAPH_UNUSED(PE);` above. Are you using the latest version of igraph (0.10.2) when you see this warning?

If you see the issue in the latest version, I will comment out that bit, but I am reluctant to remove it. At some point someone should re-read the paper describing the algorithm, and properly comment and clean up the code. At that point it will be helpful to see these unused bits.

ntamas:
@szhorvat It could be the case that compilers have gotten smarter and they complain because the _new_ value of `PE` is indeed unused, and the earlier `IGRAPH_UNUSED(...)` macro tricks the compiler into believing that the _old_ value of `PE` is used but not the one that is valid after the loop has been executed. Maybe placing another `IGRAPH_UNUSED(...)` warning _after_ the loop would silence the warning.

Still, I'm curious to know the exact compiler version as well to understand whether this is a problem that we should anticipate popping up with newer versions.

szhorvat:
I've tested with all recent GCC and Clang, and none of them warn about this with the current igraph version, _assuming that default options are used_. However, Clang 15 (very recently released) did issue this exact warning before I added the `IGRAPH_UNUSED(PE)`. Clang 14 did not complain. @folded, I'm wondering if you were using the latest igraph, as previously you mentioned that you were using 0.9.3 (if I remember correctly).

Since WARNINGS_AS_ERRORS is enabled by default, it'd be good to understand why the warning appeared.

Of course, with extra warning options there may be extra warnings. I just fixed a bunch of warnings I saw in the Fedora build logs at https://kojipkgs.fedoraproject.org//packages/igraph/0.10.2/1.fc38/data/logs/x86_64/build.log

szhorvat:
If you need to stay on 0.9, you can use 0.9.10, or for an extra few bugfixes, this branch: https://github.com/igraph/igraph/tree/release/0.9

Otherwise, I strongly recommend to go with the latest release, 0.10.2. The 0.10 series has many bugfixes that cannot be backported to 0.9.

Do keep the reports coming, even if it's just minor things such as compiler warnings.

PR #2248, opened 2022-10-30
docs: add induced_subgraph_map

### LINKED ISSUE #1866: igraph_induced_subgraph_map is undocumented
**Describe the bug**
There is no documentation provided for `igraph_induced_subgraph_map`. This is used internally by `igraph_induced_subgraph` but offers greater flexibility and is explicitly exported.

**Version information**
Latest `develop` branch.

GroteGnoom:
Fixes #1866

I'm surprised about the interface. Maybe using -1 for non-mapped vertices would have been nicer. Or a mapping from the vids vector. Like if vids = {15, 16} then the mapping tells you where 15 and 16 map to, instead of saying {0,0,0,0,0,0,0,0,0,0, ..etc}. But maybe this works better with other functions, i_induced_subgraph_map is used in /src/connectivity/components.c

### DIFF
--- a/src/operators/subgraph.c
+++ b/src/operators/subgraph.c
@@ -364,19 +364,19 @@ igraph_error_t igraph_i_induced_subgraph_map(const igraph_t *graph, igraph_t *re
  * \function igraph_induced_subgraph_map
  * \brief Creates an induced subraph and returns the mapping from the original.
  *
- * </para><para>
  * This function collects the specified vertices and all edges between
  * them to a new graph.
  * As the vertex IDs in a graph always start with zero, this function
  * very likely needs to reassign IDs to the vertices.
+ *
  * \param graph The graph object.
  * \param res The subgraph, another graph object will be stored here,
  *        do \em not initialize this object before calling this
  *        function, and call \ref igraph_destroy() on it if you don't need
  *        it any more.
  * \param vids A vertex selector describing which vertices to keep.
- * \param impl This parameter selects which implementation should we
- *        use when constructing the new graph. Basically there are two
+ * \param impl This parameter selects which implementation should be
+ *        used when constructing the new graph. Basically there are two
  *        possibilities: \c IGRAPH_SUBGRAPH_COPY_AND_DELETE copies the
  *        existing graph and deletes the vertices that are not needed
  *        in the new graph, while \c IGRAPH_SUBGRAPH_CREATE_FROM_SCRATCH
--- a/src/operators/subgraph.c
+++ b/src/operators/subgraph.c
@@ -359,6 +359,57 @@ igraph_error_t igraph_i_induced_subgraph_map(const igraph_t *graph, igraph_t *re
     }
 }
 
+/**
+ * \ingroup structural
+ * \function igraph_induced_subgraph_map
+ * \brief Creates an induced subraph and returns the mapping from the original.
+ *
+ * </para><para>
+ * This function collects the specified vertices and all edges between
+ * them to a new graph.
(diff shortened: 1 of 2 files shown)


ntamas:
Otherwise the docs in this PR are okay and I'm happy to merge it once we decide what to do with the direct mapping.

GroteGnoom:
> igraph_delete_vertices_idx(), which also uses this weird mapping vector

That's not in the docsof  igraph_delete_vertices_idx(). This should be added to the docs, right?

ntamas:
Yes, I think it should be.

src/operators/subgraph.c
```diff
@@ -359,6 +359,57 @@ igraph_error_t igraph_i_induced_subgraph_map(const igraph_t *graph, igraph_t *re
...
+ * \ingroup structural
+ * \function igraph_induced_subgraph_map
+ * \brief Creates an induced subraph and returns the mapping from the original.
+ *
+ * </para><para>
+ * This function collects the specified vertices and all edges between
+ * them to a new graph.
+ * As the vertex IDs in a graph always start with zero, this function
+ * very likely needs to reassign IDs to the vertices.
+ * \param graph The graph object.
```
szhorvat, line 372:
```suggestion
 * This function collects the specified vertices and all edges between
 * them to a new graph.
 * As the vertex IDs in a graph always start with zero, this function
 * very likely needs to reassign IDs to the vertices.
 *
 * \param graph The graph object.
```

The `<para>` doesn't hurt, it's just unnecessary. The empty line before the `\param` is for readability in code.

src/operators/subgraph.c
```diff
@@ -359,6 +359,57 @@ igraph_error_t igraph_i_induced_subgraph_map(const igraph_t *graph, igraph_t *re
...
+ * \param res The subgraph, another graph object will be stored here,
```
szhorvat, line 373: Suggestion: rename this param to `subgraph`, both here and in the function signature.

src/operators/subgraph.c
```diff
@@ -359,6 +359,57 @@ igraph_error_t igraph_i_induced_subgraph_map(const igraph_t *graph, igraph_t *re
...
+ *        do \em not initialize this object before calling this
+ *        function, and call \ref igraph_destroy() on it if you don't need
+ *        it any more.
+ * \param vids A vertex selector describing which vertices to keep.
+ * \param impl This parameter selects which implementation should we
+ *        use when constructing the new graph. Basically there are two
```
szhorvat, line 379:
```suggestion
 * \param impl This parameter selects which implementation should be
 *        used when constructing the new graph. Basically there are two
```

szhorvat (commented):
Very minor comments on docs.

szhorvat:
I am not very comfortable about this convention, and I am in favour of changing it.

Regarding saving a `vector_fill()`, doesn't it already do a `vector_null()`, and isn't that just as slow anyway?  Am I missing something? Is there any  evidence that the change would worsen performance?

ntamas:
Added an issue about this (see #2255). Let's merge this PR as it is outside the scope of this PR to fix a more general problem with mapping vectors; this will be done in #2255

ntamas:
FWIW I don't see much benefit in renaming `res` to `subgraph`; `res` clearly communicates that this is the result and it's pretty commonly used in many other places in the library. If we change it here, then we should probably also change `igraph_subgraph_edges()` et al as well, and it's hard to draw a clear boundary between places where we should make a similar rename and places where we should not. (I started the renaming by hand, that's when I realized this).

Issue #1843, opened 2021-11-14
Create a hexagonal grid graph


szhorvat:
_This is a good first issue to work on for a new contributor interested in math._

**What is the feature or improvement you would like to see?**

Functionality to create hexagonal grid graphs such as the following:

<img width="360" alt="image" src="https://user-images.githubusercontent.com/1212871/141675624-c47654d9-cf58-4704-83e4-3ae4f21e1d1e.png">

As described in #1842, we should think about a simple but flexible interface that allows creating several kinds of hexagonal grids. This interface should probably be designed concurrently with the one for triangle grids (#1842). See that issue for more details. Note that hexagonal grids are the planar duals of triangular grids.

**Vertex coordinates**

The function should also be able to compute vertex coordinates for producing nice graph layouts, as in the illustrations above.

**Use cases for the feature**

Hexagonal grids are commonly used and are available in several other graph libraries.

PR #2262, opened 2022-11-25
Hexagonal grid graph

rfulekjames:
Implements a hexagonal grid #1843 dual to the [triangular grid](https://github.com/igraph/igraph/pull/2235)
Examples for some inputs from the unit test:
 for `dimvector=(5)`, `directed=true`, `mutual=false` 

output

 for `dimvector=(4, 5)`, `directed=true`, `mutual=true`
output

 for `dimvector=(3, 4, 5)`, `directed=false`
output


szhorvat:
Thanks for working on this, this will be great!

I have not looked at either the code or its output yet, just wanted to make sure we're on the same page about the idea.

The input should be the same as for the triangular lattice, and the output should be a hex lattice whose dual is the triangular lattice obtained with the same parameters.  Here are some illustrations:

image

image

image

Notice that every red tri lattice node is in the centre of a blue hexagon, but the reverse is not true. Not every blue node is in the middle of a triangle.  My reasoning for this choice: if we put a node in the centre of every triangle, we will not always get complete hexagons. There will be some degree-1 nodes at the corners.

I do not have hex lattice graphs implemented in Mathematica, so this implementation doesn't have to match anything. If you are wondering how I generated these: I have an implementation of various geometric _meshes_: not quite the same thing as a graph, but one can derive a graph from it.

It would be great to index the vertices in a way that it wouldn't be too hard to generate coordinates for them to get nice plots like above. The way you implemented the triangular lattice makes this easy, so that's great.

If you feel that in order to allow the tri and hex lattices to match, you should change something in the tri lattice implementation, feel free to do that.

Let us know if you have any questions!

rfulekjames:
> I do not have hex lattice graphs implemented in Mathematica, so this implementation doesn't have to match anything. If you are wondering how I generated these: I have [an implementation of various geometric meshes](http://szhorvat.net/mathematica/IGDocumentation/#iglatticemesh): not quite the same thing as a graph, but one can derive a graph from it, and in case of the hexagonal mesh it happens to give nice results.

@szhorvat Thanks for the link! For me it is also easy to draw it with `python-igraph` using `kamada_kawai` layout. Just added the figures in my comment from the above. 

The labelling at this point is _row based_ similarly as in the case of the triangular grid if interpretting the lattice as a brick wall, which should allow for constructing a layout but I expect it to be a bit more intricate than in the case of the triangular lattice.

szhorvat:
It seems I accidentally marked this as "ready for review". I reverted to draft, just let us know when you consider it ready.

The plots look great!

rfulekjames:
I marked it ready for review :-) ! @szhorvat

src/constructors/lattices.c
```diff
@@ -246,6 +250,10 @@ static igraph_error_t triangular_lattice_hex_shape(
  *
  * Creates a triangular lattice whose vertices have the form (i, j) for non-negative integers i and j
  * and (i, j) is generally connected with (i + 1, j), (i, j + 1), and (i - 1, j + 1).
+ * The function constructs a planar dual of the graph constructed by <code>igraph_hexagonal_lattice</code>.
+ * In particular, there a one-to-one correspondence between the vertices in the constructed graph
+ * and the cycles of length 6 in the graph constructed by <code>igraph_hexagonal_lattice</code>
+ * with the same \p dims parameter.
```
szhorvat, line 256:
```suggestion
 * The function constructs a planar dual of the graph constructed by \ref igraph_hexagonal_lattice().
 * In particular, there a one-to-one correspondence between the vertices in the constructed graph
 * and the cycles of length 6 in the graph constructed by \ref igraph_hexagonal_lattice()
 * with the same \p dims parameter.
```

src/constructors/lattices.c
```diff
@@ -288,20 +296,303 @@ igraph_error_t igraph_triangular_lattice(
...
+ * \function igraph_hexagonal_lattice
+ * \brief A hexagonal lattice with the given shape.
+ *
+ * \experimental
+ *
+ * Creates a hexagonal lattice whose vertices have the form (i, j) for non-negative integers i and j
+ * and (i, j) is generally connected with (i + 1, j), and if i is odd also with (i - 1, j + 1).
+ * The function constructs a planar dual of the graph constructed by <code>igraph_triangular_lattice</code>.
+ * In particular, there a one-to-one correspondence between the cycles of length 6 in the constructed graph
+ * and the vertices of the graph constructed by <code>igraph_triangular_lattice</code> function
```
szhorvat, line 537:
```suggestion
 * and the vertices of the graph constructed by \ref igraph_triangular_lattice() function
```

src/constructors/lattices.c
```diff
@@ -288,20 +296,303 @@ igraph_error_t igraph_triangular_lattice(
...
+
+/**
+ * \function igraph_hexagonal_lattice
+ * \brief A hexagonal lattice with the given shape.
+ *
+ * \experimental
+ *
+ * Creates a hexagonal lattice whose vertices have the form (i, j) for non-negative integers i and j
+ * and (i, j) is generally connected with (i + 1, j), and if i is odd also with (i - 1, j + 1).
+ * The function constructs a planar dual of the graph constructed by <code>igraph_triangular_lattice</code>.
```
szhorvat, line 535:
```suggestion
 * The function constructs a planar dual of the graph constructed by \ref igraph_triangular_lattice().
```

src/constructors/lattices.c
```diff
@@ -288,20 +296,303 @@ igraph_error_t igraph_triangular_lattice(
...
+ * \param directed Boolean, whether to create a directed graph.
+ *        If the \c mutual argument is not set to true,
+ *        edges will be directed from lower-index vertices towards
+ *        higher-index ones.
+ * \param mutual Boolean, if the graph is directed this gives whether
+ *        to create all connections as mutual.
+ * \return Error code:
+ *         \c IGRAPH_EINVAL: The size of \p dims must be either 1, 2, or 3 with all the components
+ *         at least 1.
+ *
```
szhorvat, line 564:
```suggestion
 * \sa \ref igraph_triangular_lattice() for creating a triangular lattice.
 *
```

A similar `\sa` reference should be added in the docs of `igraph_triangular_lattice()` as well.

szhorvat (commented):
Just some doc formatting for now.

In the `switch` statements, can you un-indent the body by one level so that the `case` lines up with the `switch`?

The screenshots you shared show that this works well, so I don't expect issues. I'll do partial integration into the Mma interface and test it a bit, then I think we can merge.

src/constructors/lattices.c
```diff
@@ -305,3 +314,287 @@ igraph_error_t igraph_triangular_lattice(
...
+            IGRAPH_ERRORF(
+                "row_lengths_vector vector must have non-negative coordinates, "
+                "was (%" IGRAPH_PRId ") for the (%" IGRAPH_PRId ")-th row.",
+                IGRAPH_EINVAL, VECTOR(*row_lengths_vector)[i], i);
+        }
+    }
+
+    IGRAPH_VECTOR_INT_INIT_FINALLY(&edges, 0);
+
+    COMPUTE_NUMBER_OF_VERTICES
```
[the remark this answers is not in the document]
szhorvat: Instead of just using `{ ... }`, please use `do { .... } while (0)` when defining the macro. This will shut up some linters that would complain about an empty statement when the semicolon is added.

Actually, I'm wondering if this is better converted a real function ... but that's not really important.

szhorvat:
I won't have time to integrate this into the Mma interface for a while. Feel free to merge, @ntamas, once you're happy.

src/constructors/lattices.c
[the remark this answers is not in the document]
rfulekjames: Let me know if it is Ok now.

ntamas (approved):

ntamas:
Thanks a lot!

PR #2265, opened 2022-11-30
Multi-edges support for igraph_erdos_renyi_game_gnm()

### LINKED ISSUE #2161: igraph_erdos_renyi_game_gnm() should support multi-edges

`igraph_erdos_renyi_game_gnm()` should have a `multi` boolean argument which controls whether multi-edges can be generated.

**Use cases for the feature**

Generating such graphs with multi-edges is trivial to do in high-level languages. However, if the graph is very large (billions of edges), the generation takes significantly more resources than it would in C.

At the moment, we do not have a good way to generate huge graphs for testing purposes. `igraph_erdos_renyi_game_gnp()` only works up to about a hundred million vertices, as above that overflow issues make the generation inaccurate. In 0.10 this function errors for too many vertices. `igraph_erdos_renyi_game_gnm()` does not currently refuse to create large graphs, but it is dubious whether it works correctly at all for similar overflow reasons. In practice it is extremely slow.

For all these reasons, it would be useful to have a built-in function that can create random graphs with multi-edges.

**Notes**

The same feature does not make sense for `gnp()`. We have a common interface for `gnm()` and `gnp()` through `igraph_erdos_renyi_game()`. The Python interfaces exposes this. The R interface still exposes this for backwards compatibility, but it also has the separate and more modern `sample_gnm()` and `sample_gnp()`. I suggest that the Python interface should also separate these two cases. The Mathematica interface already does.

VRajesh7649:
New feature: Multi-edges support for `igraph_erdos_renyi_game_gnm()`

Resolves #2161 

Currently, I have implemented this as a separate function. Please let me know if I should refactor `igraph_erdos_renyi_game_gnm()` instead.

I will continue to add tests for this.


src/games/erdos_renyi.c
```diff
@@ -244,6 +244,120 @@ igraph_error_t igraph_erdos_renyi_game_gnm(
...
+    if(multi) {
+
+        RNG_BEGIN();
+        for(i = 0; i < no_of_edges; i++) {
+            igraph_integer_t from, to;
+            from = RNG_INTEGER(0, no_of_nodes - 1);
+            to = RNG_INTEGER(0, no_of_nodes - 1);
+            if(!loops && from == to) {
+                to = (to + 1) % (no_of_nodes - 1);
+            }
```
szhorvat, line 303: The handling of loops, as done here, is incorrect, and does not sample graphs uniformly.

Have an entirely separate conditional branch for the case with loops, and use the following algorithm:

 - Sample `from` from $[0, n-1]$, where $n$ is the vertex count.
 - Sample `to` from $[0,n-2]$, and if `to == from` then set `to = n-1`.

Basically, you need to sample `to` uniformly from not $n$, but $n-1$ distinct values. When you do the modulo operation, as implemented now, there is a chance that it wraps around to `0`, placing more weight on `0` than on other values.

src/games/erdos_renyi.c
```diff
@@ -244,6 +244,120 @@ igraph_error_t igraph_erdos_renyi_game_gnm(
...
+            if(!directed && to < from) {
```
szhorvat, line 304: Remove this block, as it is not necessary to order values passed to `igraph_create()`.

src/games/erdos_renyi.c
```diff
@@ -244,6 +244,120 @@ igraph_error_t igraph_erdos_renyi_game_gnm(
...
+                igraph_integer_t temp = from;
+                from = to;
+                to = temp;
+            }
+            igraph_vector_int_push_back(&edges, from);
+            igraph_vector_int_push_back(&edges, to);
+        }
+        RNG_END();
+
+    } else {
```
szhorvat, line 314: This branch duplicates the existing implementation of the non-multigraph case. Do not include that in this function. Also remove the `multi` argument.

src/games/erdos_renyi.c
```diff
@@ -244,6 +244,120 @@ igraph_error_t igraph_erdos_renyi_game_gnm(
...
+            if(!loops && from == to) {
+                to = (to + 1) % (no_of_nodes - 1);
+            }
+            if(!directed && to < from) {
+                igraph_integer_t temp = from;
+                from = to;
+                to = temp;
+            }
+            igraph_vector_int_push_back(&edges, from);
+            igraph_vector_int_push_back(&edges, to);
```
szhorvat, line 310:
```suggestion
            igraph_vector_int_push_back(&edges, from); /* reserved */
            igraph_vector_int_push_back(&edges, to); /* reserved */
```

Indicate for readers that `IGRAPH_CHECK` is missing on purpose.

src/games/erdos_renyi.c
```diff
@@ -244,6 +244,120 @@ igraph_error_t igraph_erdos_renyi_game_gnm(
...
+    igraph_vector_int_t edges = IGRAPH_VECTOR_NULL;
+
+    if (n < 0) {
+        IGRAPH_ERROR("Invalid number of vertices.", IGRAPH_EINVAL);
+    }
+    if (m < 0 || m > IGRAPH_ECOUNT_MAX) {
+        IGRAPH_ERROR("Invalid number of edges.", IGRAPH_EINVAL);
+    }
+
+    if (m == 0.0 || no_of_nodes == 0) {
```
szhorvat, line 264: `m` is an integer, do not compare with `0.0` as it coerces the value to a `double`.  Also, do we need this special case if only handling `multi == true`? Shorter code is usually better.

src/games/erdos_renyi.c
```diff
@@ -244,6 +244,120 @@ igraph_error_t igraph_erdos_renyi_game_gnm(
...
+        IGRAPH_FINALLY_CLEAN(1);
+        return IGRAPH_SUCCESS;
+    }
+
+    IGRAPH_VECTOR_INT_INIT_FINALLY(&edges, 0);
+    IGRAPH_CHECK(igraph_vector_int_reserve(&edges, no_of_edges * 2));
+
+    igraph_integer_t i;
+    igraph_real_t maxedges;
+    if(multi) {
```
szhorvat, line 276: Setting and checking `maxedges` is not necessary if we only care about `multi == true`.

src/games/erdos_renyi.c
```diff
@@ -244,6 +244,120 @@ igraph_error_t igraph_erdos_renyi_game_gnm(
...
 
+igraph_error_t igraph_erdos_renyi_game_gnm_multi(
+    igraph_t *graph, igraph_integer_t n, igraph_integer_t m,
+    igraph_bool_t directed, igraph_bool_t loops, igraph_bool_t multi
+) {
+
+    igraph_integer_t no_of_nodes = n;
+    igraph_integer_t no_of_edges = m;
+    igraph_real_t no_of_nodes_real = (igraph_real_t) no_of_nodes;   /* for divisions below */
+    igraph_vector_int_t edges = IGRAPH_VECTOR_NULL;
```
szhorvat, line 255: Please do not initialize data structures to values that will never be used. I know that sometimes it's taught to always initialize everything, but doing so tends to inhibit static analysers:

 - Either they complain about setting a value that is never used.
 - Or they fail to detect an effectively uninitialized access later (if it is present).

Overall, unnecessary initialization tends to hide errors in the code.

szhorvat (changes requested):
I left a few comments, but please wait for feedback from @ntamas before proceeding, to avoid doing unnecessary work. He will ultimately decide what the final interface will be (separate function or integrate into `erdos_renyi_gnm()`).

Minot stylistic comment is to always leave a space between `if`, `for`, etc. and the following `(`.

VRajesh7649:
Thank you for the feedback. I have incorporated the changes you have suggested.

src/games/erdos_renyi.c
```diff
@@ -244,6 +244,49 @@ igraph_error_t igraph_erdos_renyi_game_gnm(
...
+        igraph_vector_int_push_back(&edges, from); /* reserved */
+        igraph_vector_int_push_back(&edges, to); /* reserved */
+    }
+    RNG_END();
+
+    IGRAPH_CHECK(igraph_create(graph, &edges, n, directed));
+    igraph_vector_int_destroy(&edges);
+    IGRAPH_FINALLY_CLEAN(1);
+    return IGRAPH_SUCCESS;
+
```
szhorvat, line 287:
```suggestion

    return IGRAPH_SUCCESS;
```

src/games/erdos_renyi.c
```diff
@@ -244,6 +244,49 @@ igraph_error_t igraph_erdos_renyi_game_gnm(
...
+    }
+    if (m < 0 || m > IGRAPH_ECOUNT_MAX) {
+        IGRAPH_ERROR("Invalid number of edges.", IGRAPH_EINVAL);
+    }
+
+    IGRAPH_VECTOR_INT_INIT_FINALLY(&edges, 0);
+    IGRAPH_CHECK(igraph_vector_int_reserve(&edges, no_of_edges * 2));
+
+    RNG_BEGIN();
+    for (igraph_integer_t i = 0; i < no_of_edges; i++) {
```
szhorvat, line 267: It's great that you localized the `i` variable to the `for` loop. Please keep doing this in the future even though this is not what you can currently see in a lot of old igraph code.

src/games/erdos_renyi.c
```diff
@@ -244,6 +244,49 @@ igraph_error_t igraph_erdos_renyi_game_gnm(
     return IGRAPH_SUCCESS;
 }
 
+igraph_error_t igraph_erdos_renyi_game_gnm_multi(
+    igraph_t *graph, igraph_integer_t n, igraph_integer_t m,
+    igraph_bool_t directed, igraph_bool_t loops
+) {
+
+    igraph_integer_t no_of_nodes = n;
+    igraph_integer_t no_of_edges = m;
```
szhorvat, line 253: This is a simple renaming of existing variables, so let's just get rid of the `no_of_nodes` and `no_of_edges` and keep using `n` and `m` throughout.  Currently both are used, somewhat inconsistently.

szhorvat (changes requested):
That was a quick response! This is very good for a first contribution, only a few nitpick comments left from my side.

Next steps are:

 1. wait for feedback from @ntamas regarding the interface
 2. finalize interface
 3. update docs

ntamas:
Looks good to me!

Things to be done:

- [ ] Update changelog
- [ ] Add new function to `functions.yaml` (just copy the definition of `igraph_erdos_renyi_gnm`)
- [ ] Add documentation and link it to `doc/generators.xxml`

Let me know when you're done with these, mark the PR as ready for review and then I'll merge it.

VRajesh7649:
Thanks for the response. I've done the next steps and have marked this PR as ready to review.

szhorvat:
@ntamas IMO it would be much better to integrate this into `igraph_erdos_renyi_gnm()` instead of adding a separate function. Is there any good reason to make it a separate function?

ntamas:
> Is there any good reason to make it a separate function?

I'm not very fond of breaking existing APIs all the time just because we add some new functionality. Whether that counts as "good reason" is up for debate, but my personal preference in cases like this is to get around without breaking the API unless it is necessary to break it.

> The most reasonable way to expose this in a high-level interface is through a single `gnm` function. 

Agreed. But note that even if the functions are separate in the C layer, nothing prevents you from creating a single function in the higher-level interface; that's what I would probably do in a newly-generated Python interface as well. New arguments are much easier to add in higher-level interfaces where you can simply provide default values for them without worrying about API breakage.

ntamas:
@szhorvat I know that `igraph_erdos_renyi_game()` with its common `m_or_p` parameter is/was probably a mistake, but do you think we should extend the `igraph_erdos_renyi_t` enum with a new option for the multi-edge variant? We would then at least have an easy unified entry point that also includes the multi-edge generator, and this can be done without breaking the API/ABI.

Also note that `igraph_erdos_renyi_game_gnp()` and `igraph_erdos_renyi_game_gnm()` are not even included in the documentation, only `igraph_erdos_renyi_game()` is documented.

szhorvat:
My preferred solution is that starting with 0.11:

 -  `igraph_erdos_renyi_game()` would be deprecated in favour of the `gnm` and `gnp` variants (just as `erdos.renyi.game()` got deprecated in favour of `sample_gnm` and `sample_gnp` in R).
 - `igraph_erdos_renyi_game_gnm()` would gain an additional `multi` option, which `igraph_erdos_renyi_game_gnp()` will not have.
 - The deprecated `igraph_erdos_renyi_game()` function will not gain the ability to generate multi-edges.

Additionally, the `gnm` and `gnp` variants would get properly documented in 0.10.3, and there would be a mention in the `igraph_erdos_renyi_game()` docs that the specialized variants are preferred (but no deprecation that causes annoying compiler warnings when building the Python interface).

ntamas:
> My preferred solution is that starting with 0.11:

I think that your preferred solution could even be implemented for 0.10, right? After all, `igraph_erdos_renyi_game()` would not be changed, it would only be deprecated. `igraph_erdos_renyi_game_gnm()` was not part of the public API yet so we are free to add a new argument to it even in 0.10.

szhorvat:
Somehow I only saw your response now.

> `igraph_erdos_renyi_game_gnm()` was not part of the public API

I always assumed that it's public API. It's in the public headers and I use it in IGraph/M. I would recommend leaving it for 0.11, but it's up to you.

ntamas:
Retargeting for `develop` so we can add the extra `multi` argument that @szhorvat proposed to `igraph_erdos_renyi_game_gnm()`.

ntamas:
Finally merged, thanks a lot!

PR #2266, opened 2022-12-02
feat: igraph_tree_from_parent_vector()

szhorvat:


ntamas:
I think that `igraph_tree_from_parent_vector()` is perfect; no need to change or shorten it.

szhorvat:
Feel free to merge if CI checks pass. I won't make more modifications. I added `\experimental` out of precaution, but I don't really expect changes. Let's just make all non-trivial functions experimental if they are added during the 0.10.x cycle. When 0.11 is released, we can mark many of those as finalized.

szhorvat:
Oops, actually there's a little bug ...

ntamas (approved):

ntamas:
Thanks!

Issue #2197, opened 2022-09-11
"Tree" and "Hourglass" algorithms for all-pairs shortest paths


szhorvat:
**What is the feature or improvement you would like to see?**

Implement the "Tree" or "Hourglass" algorithms for all-pairs shortest paths from the reference below. The paper appears to be well-written, and contains not only a theoretical analysis of the algorithm, but also an extensive empirical comparison. I think that there is a good chance that it works well in practice.

**Use cases for the feature**

I am looking for an algorithm that works well on dense graphs, i.e. the running time does not increase noticeably with the edge count when keeping the vertex count constant. The Floyd-Warshall algorithm ($O(V^3)$) has this property, but in practice it is slower than Dijkstra except for _very_ dense graphs. The reference claims an expected running time of $O(V^2 \log^2 V)$.

**References**

 - Andrej Brodnik, Marko Grgurovič, Rok Požar: Modifications of the Floyd-Warshall algorithm with nearly quadratic expected-time, https://doi.org/10.26493/1855-3974.2467.497

Nivig1310:
Hello, I am interested in working on this but I am confused about what you would like implemented. Should I apply the Floyd-Warshall algorithm based on the tips given in the reference to optimize it?

szhorvat:
Yes, the task is to implement at least one of the two algorithms described in the paper, ideally the faster one.  However, this is not a trivial task, so unless you are already comfortable with igraph, and experienced with graph algorithms, I suggest you start with a simpler issues first. I suggest #1842/#1843, or #1840 (which has stalled and is available). These would allow you to learn how to program with igraph without needing to also deal with a complex algorithm at the same time. It is up to you whether you start with something simple.

For this issue, you would need to:

 - Read the paper, understand the algorithms, summarize them briefly here.
 - Understand what is needed to create a production quality implementation that works on non-simple graphs
 - Implement one algorithm (or both). If you hourglass one, you would need to look into how to make it non-recursive, as recursion [risks stack-overflow](https://github.com/igraph/python-igraph/issues/563). A recursive version is fine for a starting point though.
 - Benchmark the implementation against Dijkstra and a plain Floyd-Warshall (#2194).

Here's what you need to know to get started: https://github.com/igraph/igraph/wiki/Quickstart-for-new-contributors

szhorvat:
[...]

PR #2267, opened 2022-12-04
All pairs shortest paths a la brodnik et al

rfulekjames:
Towards resolving issue  #2197.
- The "Tree" algorithm from https://doi.org/10.26493/1855-3974.2467.497 
- Added unit tests which are copied from `tests/unit/igraph_distances_floyd_warshall `
- Added [Benchmarks](https://github.com/igraph/igraph/issues/2197#issuecomment-1336185456)
- Optimized by following "Practical improvement" section of https://doi.org/10.26493/1855-3974.2467.497 
- Added a test for a random graph with 100 vertices


src/paths/floyd_warshall.c
```diff
@@ -140,3 +144,131 @@ igraph_error_t igraph_distances_floyd_warshall(
...
+        if (out && MATRIX(*res, from, to) > w) MATRIX(*res, from, to) = w;
+        if (in  && MATRIX(*res, to, from) > w) MATRIX(*res, to, from) = w;
+    }
+
+    igraph_t out_k;
+    igraph_vector_int_t parents;
+    igraph_matrix_int_t predecessors;
+    igraph_stack_int_t stack;
+    IGRAPH_CHECK(igraph_stack_int_init(&stack, no_of_nodes));
+    IGRAPH_CHECK(igraph_matrix_int_init(&predecessors, no_of_nodes, no_of_nodes));
```
rfulekjames, line 215: Failed tests are be due to a memory leak caused by this line and happen when there is a negative cycle in the graph which throws an error on line 258.
I suspect that adding `igraph_matrix_int_destroy(&predecessors);` before 258 would resolve this issue but perhaps that's not the best way to handle this memory leak.

szhorvat, line 215:
```suggestion
    IGRAPH_CHECK(igraph_stack_int_init(&stack, no_of_nodes));
    IGRAPH_FINALLY(igraph_stack_int_destroy, &stack);
    IGRAPH_CHECK(igraph_matrix_int_init(&predecessors, no_of_nodes, no_of_nodes));
    IGRAPH_FINALLY(igraph_matrix_int_destroy, &predecessors);
```

This should fix the memory leak (untested).

I am posting this just for illustration. `IGRAPH_FINALLY` registers a data structure for destruction in case of error. `IGRAPH_FINALLY_CLEAN(n)` discards the last `n` such registrations.

There are often macros that perform both initialization and adding the to the "finally stack". These should be preferred instead of an explicit `IGRAPH_FINALLY` when they are available, just to keep the code shorter. An example of such a macro is `IGRAPH_VECTOR_INT_INIT_FINALLY`. I do not recall immediately if these are available for matrices, but they are certainly available for stacks.

src/paths/floyd_warshall.c
```diff
@@ -140,3 +144,131 @@ igraph_error_t igraph_distances_floyd_warshall(
...
+                igraph_vector_int_destroy(&children);
+                IGRAPH_FINALLY_CLEAN(1);
+            }
+        }
+        igraph_destroy(&out_k);
+    }
+    igraph_vector_int_destroy(&parents);
+    IGRAPH_FINALLY_CLEAN(1);
+    igraph_matrix_int_destroy(&predecessors);
+    igraph_stack_int_destroy(&stack);
```
szhorvat, line 271:
```suggestion
    igraph_vector_int_destroy(&parents);
    igraph_matrix_int_destroy(&predecessors);
    igraph_stack_int_destroy(&stack);
    IGRAPH_FINALLY_CLEAN(3);
```

src/paths/floyd_warshall.c
```diff
@@ -140,3 +144,131 @@ igraph_error_t igraph_distances_floyd_warshall(
...
+        for (igraph_integer_t u=0; u < no_of_nodes; u++) {
+            /* the penultimate vertex on the shortest path from u to v */
+            MATRIX(predecessors, u, v) = u;
+        }
+    }
+    IGRAPH_VECTOR_INT_INIT_FINALLY(&parents, no_of_nodes);
+
+    for (igraph_integer_t k=0; k < no_of_nodes; k++) {
+        /* constructing the shortest path tree rooted at vertex k as out_k */
+        IGRAPH_CHECK(igraph_empty(&out_k, 0, IGRAPH_DIRECTED));
```
szhorvat, line 226:
```suggestion
```

because `igraph_tree_from_parent_vector` is a constructor and will overwrite this

src/paths/floyd_warshall.c
```diff
@@ -140,3 +144,131 @@ igraph_error_t igraph_distances_floyd_warshall(
...
+        for (igraph_integer_t v=0; v < no_of_nodes; v++) {
+            VECTOR(parents)[v] = v == k ? -1 : MATRIX(predecessors, k, v);
+        }
+        //print_vector_int2(&parents);
+        igraph_tree_from_parent_vector(&out_k, &parents, IGRAPH_TREE_OUT);
```
szhorvat, line 231:
```suggestion
        IGRAPH_CHECK(igraph_tree_from_parent_vector(&out_k, &parents, IGRAPH_TREE_OUT));
        IGRAPH_FINALLY(igraph_destroy, &out_k);
```

src/paths/floyd_warshall.c
```diff
@@ -140,3 +144,131 @@ igraph_error_t igraph_distances_floyd_warshall(
...
+                    if (i == j && MATRIX(*res, i, i) < 0) {
+                        IGRAPH_ERROR("Negative cycle found while calculating distances with Floyd-Warshall.",
+                                    IGRAPH_ENEGLOOP);
+                    }
+                }
+                igraph_vector_int_destroy(&children);
+                IGRAPH_FINALLY_CLEAN(1);
+            }
+        }
+        igraph_destroy(&out_k);
```
szhorvat, line 266:
```suggestion
        igraph_destroy(&out_k);
        IGRAPH_FINALLY_CLEAN(1);
```

szhorvat:
Some more comments:

`igraph_vector_int_t children` should be initialized once outside the main loop, and re-used repeatedly. This is for better performance, as memory allocations are expensive. `igraph_neighbors()` can write its result into any initialized vector, not only an empty one. It will resize that vector as needed.

Similarly, repeatedly creating the graph `out_k` is a heavyweight operation that almost certainly has a noticeable impact on performance. Unlike a plain vector, a graph cannot be re-used effectively between different iterations of the loop. If we can find a way to compute `children` without involving the construction of a new graph, that might help significantly. This would be something to think carefully about.

szhorvat:
Just moving the initialization of `children` out of the loop gave me a significant speedup, now the "Tree" version is about 2x faster than the plain Floyd-Warshall even in the second block of benchmarks. Before it was slightly slower—it only became faster in the third block.

szhorvat:
OK, now we're faster than the original in all benchmark. Hope I didn't step on your toes with this push.

szhorvat:
The benchmark looks like this on my machine now (note that timings are somewhat unstable on this laptop):

```
|> Benchmark file: /Users/szhorvat/Repos/igraph-main/igraph/tests/benchmarks/igraph_distances.c
|  1 vcount=100, p=0.5, Unweighted, 200x                                            0.13s  0.129s      0s
|  2 vcount=100, p=0.5, Dijkstra, 200x                                             0.781s  0.779s  0.001s
|  3 vcount=100, p=0.5, Unweighted Floyd-Warshall, 200x                            0.249s  0.248s      0s
|  3 vcount=100, p=0.5, Unweighted Floyd-Warshall-tree-speedup, 200x                0.18s   0.18s      0s
|  4 vcount=100, p=0.5, Floyd-Warshall, 200x                                       0.341s   0.34s      0s
|  4 vcount=100, p=0.5, Floyd-Warshall-tree-speedup, 200x                          0.214s  0.214s      0s
|  5 vcount=100, p=0.5, Bellman-Ford (negative), 200x                              0.869s  0.865s  0.002s
|  6 vcount=100, p=0.5, Johnson (negative), 200x                                   0.922s  0.901s  0.004s
|  7 vcount=100, p=0.5, Floyd-Warshall (negative), 200x                            0.357s  0.356s      0s
|  7 vcount=100, p=0.5, Floyd-Warshall-tree-speedup (negative), 200x               0.221s  0.219s  0.001s

|  1 vcount=500, p=0.1, Unweighted, 10x                                            0.156s  0.155s  0.001s
|  2 vcount=500, p=0.1, Dijkstra, 10x                                               1.24s   1.24s  0.001s
|  3 vcount=500, p=0.1, Unweighted Floyd-Warshall, 10x                              1.39s   1.39s  0.002s
|  3 vcount=500, p=0.1, Unweighted Floyd-Warshall-tree-speedup, 10x                0.422s  0.421s  0.001s
|  4 vcount=500, p=0.1, Floyd-Warshall, 10x                                         1.54s   1.54s  0.001s
|  4 vcount=500, p=0.1, Floyd-Warshall-tree-speedup, 10x                           0.437s  0.436s      0s
|  5 vcount=500, p=0.1, Bellman-Ford (negative), 10x                                1.69s   1.68s  0.002s
|  6 vcount=500, p=0.1, Johnson (negative), 10x                                     1.29s   1.28s  0.002s
|  7 vcount=500, p=0.1, Floyd-Warshall (negative), 10x                              1.56s   1.56s  0.002s
|  7 vcount=500, p=0.1, Floyd-Warshall-tree-speedup (negative), 10x                0.439s  0.438s  0.001s

|  1 vcount=1500, p=0.02, Unweighted, 1x                                           0.127s  0.123s  0.004s
|  2 vcount=1500, p=0.02, Dijkstra, 1x                                             0.876s  0.874s  0.001s
|  3 vcount=1500, p=0.02, Unweighted Floyd-Warshall, 1x                             4.45s   4.43s  0.007s
|  3 vcount=1500, p=0.02, Unweighted Floyd-Warshall-tree-speedup, 1x               0.986s  0.978s  0.007s
|  4 vcount=1500, p=0.02, Floyd-Warshall, 1x                                        4.81s   4.76s   0.02s
|  4 vcount=1500, p=0.02, Floyd-Warshall-tree-speedup, 1x                           1.25s   1.23s  0.008s
|  5 vcount=1500, p=0.02, Bellman-Ford (negative), 1x                               1.93s   1.92s  0.003s
|  6 vcount=1500, p=0.02, Johnson (negative), 1x                                   0.921s  0.917s  0.002s
|  7 vcount=1500, p=0.02, Floyd-Warshall (negative), 1x                             4.93s    4.9s  0.013s
|  7 vcount=1500, p=0.02, Floyd-Warshall-tree-speedup (negative), 1x                1.08s   1.08s  0.003s
```

Note that I did not try to understand the algorithm yet, I simply refactored your code.  I'll stop pushing to this PR now.  I'm quite happy about the speedup compared to the plain Floyd-Warshall, thanks for taking up this project @rfulekjames !

src/paths/floyd_warshall.c
```diff
@@ -140,3 +142,132 @@ igraph_error_t igraph_distances_floyd_warshall(
...
+        for (igraph_integer_t j=0; j < no_of_nodes; j++) {
+            igraph_real_t dkj = MATRIX(*res, k, j);
+            if (dkj == IGRAPH_INFINITY) continue;
+            IGRAPH_CHECK(igraph_stack_int_push(&stack, k));
+            while (!igraph_stack_int_empty(&stack)) {
+                igraph_integer_t v = igraph_stack_int_pop(&stack);
+                igraph_neighbors(&out_k, &children, v, IGRAPH_OUT);
+                igraph_integer_t no_of_children = igraph_vector_int_size(&children);
+                for (igraph_integer_t l=0; l < no_of_children; l++) {
+                    igraph_integer_t i = VECTOR(children)[l];
```
rfulekjames, line 247: There was a mistake in my previous version.  In fact, we need the `children`, not only their count `no_of_children `. 
It is strange that in spite of that mistake all the tests were passing. 
Also looking into this closer, the algorithm shouldn't work for undirected graphs at this point.

rfulekjames:
I am sorry for the mistake in my previous commit, but at least it led to the discovery of possible problems with the "tree" algorithm itself.

szhorvat:
Sorry, I should not have pushed the changes here before you requested a review. I was excited to find out if performance can be made better even for small graph, as the plots in the paper seem to indicate. Perhaps it would have been better to show the changes I tried on a separate branch without modifying this PR.

I see that you requested a review now. Just to clarify, now the algorithm is good, and matches what's in the paper, but only for the undirected case?  I will need some time to look at the paper more carefully.

rfulekjames:
@szhorvat 

> I see that you requested a review now. Just to clarify, now the algorithm is good, and matches what's in the paper, but only for the undirected case? I will need some time to look at the paper more carefully

Now the algorithm works for both directed and undirected but it is slower than the original floyd warshall in all current benchmarks. Perhaps, for larger dense graphs it might give something better.

It remains to see if "hourglass" would be better.

> Sorry, I should not have pushed the changes here before you requested a review. I was excited to find out if performance can be made better even for small graph, as the plots in the paper seem to indicate. Perhaps it would have been better to show the changes I tried on a separate branch without modifying this PR.

No problem. I fixed the algorithm. Tested it also thoroughly on random graphs and it works also on directed graphs, except that it is much slower.

rfulekjames:
Do you think that the creation of the `in_tree_k` is the bottleneck here?

src/paths/floyd_warshall.c
```diff
@@ -140,3 +142,132 @@ igraph_error_t igraph_distances_floyd_warshall(
...
+    IGRAPH_VECTOR_INT_INIT_FINALLY(&parents, no_of_nodes);
+
+    for (igraph_integer_t k=0; k < no_of_nodes; k++) {
+        for (igraph_integer_t v=0; v < no_of_nodes; v++) {
+            VECTOR(parents)[v] = v == k ? -1 : MATRIX(successors, v, k);
+        }
+        /* tree of so far discovered shortest paths ending at k */
+        igraph_t tree_in_k;
+        IGRAPH_FINALLY(igraph_destroy, &tree_in_k);
+        IGRAPH_CHECK(igraph_tree_from_parent_vector(&tree_in_k, &parents, IGRAPH_TREE_OUT));
```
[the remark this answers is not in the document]
szhorvat: Yes. But don't worry. The first step is to get the algorithm right. Once that's done we can look at practical constant factor speedups.

src/paths/floyd_warshall.c
```diff
@@ -140,3 +142,132 @@ igraph_error_t igraph_distances_floyd_warshall(
...
+
+    IGRAPH_STACK_INT_INIT_FINALLY(&stack, no_of_nodes);
+
+    IGRAPH_CHECK(igraph_matrix_int_init(&successors, no_of_nodes, no_of_nodes));
+    IGRAPH_FINALLY(igraph_matrix_int_destroy, &successors);
+
+    for (igraph_integer_t v=0; v < no_of_nodes; v++) {
+        for (igraph_integer_t u=0; u < no_of_nodes; u++) {
+            /* the second vertex on the shortest path from u to v */
+            MATRIX(successors, u, v) = v;
```
szhorvat, line 219: `succcessors[i,j]` corresponds to $\pi_{ij}$ in the paper in Algorithm 2, correct? Line 1 reads $\pi_{ij} := i$. However, here you are setting $\pi_{ij} = j$. Is there a problem here?

szhorvat: It's a bit confusing to follow because the notation here is different than in the paper. I can see that this is in part because you started with the existing plain Floyd–Warshall code in igraph, and kept the iterator names from there. But it would be much easier to follow whether the implementation matches the paper if we used the same notation as the paper.

It seems to me that with the assumption that `successors[j,i]` = $\pi_{ij}$ (i.e. transpose of what I wrote above), the line below saying `MATRIX(successors, i, j) = MATRIX(successors, k, j);` still does not match the paper.

Could you please refactor the notation to match the paper, so it's easier to check the algorithm?

src/paths/floyd_warshall.c
```diff
@@ -140,3 +142,132 @@ igraph_error_t igraph_distances_floyd_warshall(
...
+                igraph_integer_t v = igraph_stack_int_pop(&stack);
+                igraph_neighbors(&tree_in_k, &children, v, IGRAPH_OUT);
+                igraph_integer_t no_of_children = igraph_vector_int_size(&children);
+                for (igraph_integer_t l=0; l < no_of_children; l++) {
+                    igraph_integer_t i = VECTOR(children)[l];
+                    igraph_real_t di = MATRIX(*res, i, k) + dkj;
+                    igraph_real_t dd = MATRIX(*res, i, j);
+                    if (di < dd) {
+                        MATRIX(*res, i, j) = di;
+                        MATRIX(successors, i, j) = MATRIX(successors, k, j);
```
szhorvat, line 252: _With the current notation_, I believe this line should read

```suggestion
                        MATRIX(successors, i, j) = MATRIX(successors, i, k);
```

The role of $i$ and $j$ is reversed here compared to the paper and `successors` is transposed compared to $\pi$.

rfulekjames: Correct! Fixed.

src/paths/floyd_warshall.c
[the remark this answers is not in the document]
rfulekjames: I removed the creation of the tree and it led to much faster running times. Working just with arrays instead.

src/paths/floyd_warshall.c
rfulekjames: > Could you please refactor the notation to match the paper, so it's easier to check the algorithm?

I can refactor it to match the paper if you think the following wouldn't cause some considerable performance drawbacks.
The reason I didn't follow the original algorithm is this comment:
 //  /* Iteration order matters for performance!
        //  First j, then i, because matrices are stored as column-major. */

Exact reasons why the order matters is not clear to me, though.

rfulekjames:
New benchmarks: 

1 vcount=100, p=0.5, Unweighted, 200x                                           0.203s  0.203s      0s
|  2 vcount=100, p=0.5, Dijkstra, 200x                                              1.58s   1.58s      0s
|  3 vcount=100, p=0.5, Unweighted Floyd-Warshall, 200x                            0.577s  0.577s      0s
|  3 vcount=100, p=0.5, Unweighted Floyd-Warshall-tree-speedup, 200x               0.515s  0.515s      0s
|  4 vcount=100, p=0.5, Floyd-Warshall, 200x                                       0.713s  0.713s      0s
|  4 vcount=100, p=0.5, Floyd-Warshall-tree-speedup, 200x                          0.596s  0.596s      0s
|  5 vcount=100, p=0.5, Bellman-Ford (negative), 200x                               1.46s   1.46s      0s
|  6 vcount=100, p=0.5, Johnson (negative), 200x                                    1.71s   1.71s      0s
|  7 vcount=100, p=0.5, Floyd-Warshall (negative), 200x                            0.718s  0.718s      0s
|  7 vcount=100, p=0.5, Floyd-Warshall-tree-speedup (negative), 200x               0.612s  0.612s      0s

|  1 vcount=500, p=0.1, Unweighted, 10x                                            0.259s  0.259s      0s
|  2 vcount=500, p=0.1, Dijkstra, 10x                                               2.49s   2.49s      0s
|  3 vcount=500, p=0.1, Unweighted Floyd-Warshall, 10x                              3.44s   3.44s      0s
|  3 vcount=500, p=0.1, Unweighted Floyd-Warshall-tree-speedup, 10x                 1.07s   1.06s   0.01s
|  4 vcount=500, p=0.1, Floyd-Warshall, 10x                                         3.61s   3.61s      0s
|  4 vcount=500, p=0.1, Floyd-Warshall-tree-speedup, 10x                            1.47s   1.46s   0.01s
|  5 vcount=500, p=0.1, Bellman-Ford (negative), 10x                                2.69s   2.69s      0s
|  6 vcount=500, p=0.1, Johnson (negative), 10x                                     2.55s   2.55s      0s
|  7 vcount=500, p=0.1, Floyd-Warshall (negative), 10x                              3.72s   3.72s      0s
|  7 vcount=500, p=0.1, Floyd-Warshall-tree-speedup (negative), 10x                 1.84s   1.84s      0s

|  1 vcount=1500, p=0.02, Unweighted, 1x                                           0.189s  0.179s   0.01s
|  2 vcount=1500, p=0.02, Dijkstra, 1x                                              1.79s   1.79s      0s
|  3 vcount=1500, p=0.02, Unweighted Floyd-Warshall, 1x                             8.79s   8.79s      0s
|  3 vcount=1500, p=0.02, Unweighted Floyd-Warshall-tree-speedup, 1x                1.99s   1.99s      0s
|  4 vcount=1500, p=0.02, Floyd-Warshall, 1x                                        9.07s   9.07s      0s
|  4 vcount=1500, p=0.02, Floyd-Warshall-tree-speedup, 1x                           3.58s   3.57s   0.01s
|  5 vcount=1500, p=0.02, Bellman-Ford (negative), 1x                               3.07s   3.06s      0s
|  6 vcount=1500, p=0.02, Johnson (negative), 1x                                    1.85s   1.85s      0s
|  7 vcount=1500, p=0.02, Floyd-Warshall (negative), 1x                             9.06s   9.06s      0s
|  7 vcount=1500, p=0.02, Floyd-Warshall-tree-speedup (negative), 1x                3.82s   3.81s   0.01s

szhorvat:
Once the notation refactoring is done, I'll do a careful comparison with the algorithm in the paper. If I did this now, I would still change the notation so I can see things better. If you could do this that would be really helpful.

As the next step after that, it would be great to expand the test suite, adding using some larger unweighted graph, and comparing the result with Dijkstra (or the plain F-W), so we have more confidence that everything is correct. It's not necessary to print the result, just compare with another implementation. Why unweighted? So that numerical roundoff errors wouldn't cause a difference.

It seems that the current tests didn't catch some of the issues in the implementation, and it would give me more peace of mind to know that a more extensive test suite passes as well.

My plan for the step after that was to try to optimize the code to get a constant-factor speedup, but you have already done that (i.e. switch from an `igraph_t` to an array representation). Great!

Next, it would be nice to try to implement the further optimization suggested in the paper under the "Practical improvement" heading.

The final step would be to finalize the interface.  What I have in mind right now is:

 - We add an extra `method` parameter to `igraph_distances_floyd_warshall()`. This will select the specific algorithm to use: original, tree, or hourglass (if it's implemented later). Thus there will be a single function to access all methods. If I got this right, Figure 4 of the paper tells us that there are in fact cases where the plain F-W will perform better. Later we might even get ambitious and add an "auto" method which tries to select the best approach based on the graph's characteristics (if we can figure out a good heuristic for that).
 - The main function should handle the error checking and the initialization of the values in the `res` matrix based on `weights1, to avoid code duplication. Then it will call internal functions for the main part of the algorithm.

Does this sound good?

rfulekjames:
@szhorvat 
> Could you change around the notation with this in mind? I.e. the  index in the paper should be the i variable in the code. However, when the paper shows  we write res[j,i], reversing the indices everywhere. Then also switch around the setting of the in and out variables.

I changed the indices accordingly. 
If there are not other issues I can proceed to expanding the unit tests. 

I have already tested the latest version on large random graphs. Therefore there shouldn't be issues with correctness at this point.

rfulekjames:
> Next, it would be nice to try to implement the further optimization suggested in the paper under the "Practical improvement" heading.

This optimization gives the following which seems to be a substantial improvement over the previous version.

|  1 vcount=100, p=0.5, Unweighted, 200x                                             0.2s    0.2s      0s
|  2 vcount=100, p=0.5, Dijkstra, 200x                                              1.55s   1.55s      0s
|  3 vcount=100, p=0.5, Unweighted Floyd-Warshall, 200x                            0.558s  0.558s      0s
|  3 vcount=100, p=0.5, Unweighted Floyd-Warshall-tree-speedup, 200x               0.474s  0.464s   0.01s
|  4 vcount=100, p=0.5, Floyd-Warshall, 200x                                         0.7s    0.7s      0s
|  4 vcount=100, p=0.5, Floyd-Warshall-tree-speedup, 200x                          0.395s  0.395s      0s
|  5 vcount=100, p=0.5, Bellman-Ford (negative), 200x                               1.44s   1.44s      0s
|  6 vcount=100, p=0.5, Johnson (negative), 200x                                    1.73s   1.73s      0s
|  7 vcount=100, p=0.5, Floyd-Warshall (negative), 200x                            0.719s  0.719s      0s
|  7 vcount=100, p=0.5, Floyd-Warshall-tree-speedup (negative), 200x               0.403s  0.403s      0s

|  1 vcount=500, p=0.1, Unweighted, 10x                                            0.254s  0.254s      0s
|  2 vcount=500, p=0.1, Dijkstra, 10x                                               2.46s   2.46s      0s
|  3 vcount=500, p=0.1, Unweighted Floyd-Warshall, 10x                               3.3s    3.3s      0s
|  3 vcount=500, p=0.1, Unweighted Floyd-Warshall-tree-speedup, 10x                0.865s  0.845s   0.02s
|  4 vcount=500, p=0.1, Floyd-Warshall, 10x                                         3.55s   3.55s      0s
|  4 vcount=500, p=0.1, Floyd-Warshall-tree-speedup, 10x                           0.972s  0.962s   0.01s
|  5 vcount=500, p=0.1, Bellman-Ford (negative), 10x                                2.57s   2.57s      0s
|  6 vcount=500, p=0.1, Johnson (negative), 10x                                     2.45s   2.45s      0s
|  7 vcount=500, p=0.1, Floyd-Warshall (negative), 10x                              3.56s   3.56s      0s
|  7 vcount=500, p=0.1, Floyd-Warshall-tree-speedup (negative), 10x                0.961s  0.961s      0s

|  1 vcount=1500, p=0.02, Unweighted, 1x                                           0.177s  0.167s   0.01s
|  2 vcount=1500, p=0.02, Dijkstra, 1x                                              1.76s   1.76s      0s
|  3 vcount=1500, p=0.02, Unweighted Floyd-Warshall, 1x                             8.46s   8.46s      0s
|  3 vcount=1500, p=0.02, Unweighted Floyd-Warshall-tree-speedup, 1x                2.32s   2.31s   0.01s
|  4 vcount=1500, p=0.02, Floyd-Warshall, 1x                                        8.75s   8.75s      0s
|  4 vcount=1500, p=0.02, Floyd-Warshall-tree-speedup, 1x                           3.03s   3.02s   0.01s
|  5 vcount=1500, p=0.02, Bellman-Ford (negative), 1x                               3.03s   3.02s      0s
|  6 vcount=1500, p=0.02, Johnson (negative), 1x                                    1.86s   1.86s      0s
|  7 vcount=1500, p=0.02, Floyd-Warshall (negative), 1x                             8.83s   8.83s      0s
|  7 vcount=1500, p=0.02, Floyd-Warshall-tree-speedup (negative), 1x                3.06s   3.05s   0.01s

src/paths/floyd_warshall.c
```diff
@@ -140,3 +142,161 @@ igraph_error_t igraph_distances_floyd_warshall(
...
+            MATRIX(predecessors, u, v) = u;
+        }
+    }
+
+    for (igraph_integer_t k=0; k < no_of_nodes; k++) {
+        /* resetting no_of_children vector */
+        for (igraph_integer_t parent=0; parent < no_of_nodes; parent++)
+            VECTOR(no_of_children)[parent] = 0;
+        /* constructing the tree out_k (as in the paper) but
+           representing it as the children matrix */
```
rfulekjames, line 244: Representing the tree by a matrix gives a substantial improvement in comparison with a linear (space) tree:

|  3 vcount=100, p=0.5, Unweighted Floyd-Warshall-tree-speedup-linear-tree, 200x   0.746s  0.746s      0s
|  3 vcount=100, p=0.5, Unweighted Floyd-Warshall-tree-speedup, 200x               0.468s  0.468s      0s
|  4 vcount=100, p=0.5, Floyd-Warshall-tree-speedup-linear-tree, 200x              0.647s  0.647s      0s
|  4 vcount=100, p=0.5, Floyd-Warshall-tree-speedup, 200x                          0.394s  0.394s      0s
|  7 vcount=100, p=0.5, Floyd-Warshall-tree-speedup-linear-tree (negative), 200x   0.657s  0.657s      0s
|  7 vcount=100, p=0.5, Floyd-Warshall-tree-speedup (negative), 200x               0.413s  0.413s      0s

|  3 vcount=500, p=0.1, Unweighted Floyd-Warshall-tree-speedup-linear-tree, 10x     1.24s   1.24s      0s
|  3 vcount=500, p=0.1, Unweighted Floyd-Warshall-tree-speedup, 10x                0.798s  0.798s      0s
|  4 vcount=500, p=0.1, Floyd-Warshall-tree-speedup-linear-tree, 10x                1.27s   1.27s      0s
|  4 vcount=500, p=0.1, Floyd-Warshall-tree-speedup, 10x                           0.933s  0.933s      0s
|  7 vcount=500, p=0.1, Floyd-Warshall-tree-speedup-linear-tree  (negative), 10x    1.26s   1.26s      0s
|  7 vcount=500, p=0.1, Floyd-Warshall-tree-speedup (negative), 10x                0.934s  0.924s   0.01s

|  3 vcount=1500, p=0.02, Unweighted Floyd-Warshall-tree-speedup-linear-tree, 1x    2.69s   2.68s   0.01s
|  3 vcount=1500, p=0.02, Unweighted Floyd-Warshall-tree-speedup, 1x                2.22s   2.21s   0.01s
|  4 vcount=1500, p=0.02, Floyd-Warshall-tree-speedup-linear-tree, 1x                3.4s    3.4s      0s
|  4 vcount=1500, p=0.02, Floyd-Warshall-tree-speedup, 1x                           2.96s   2.96s      0s
|  7 vcount=1500, p=0.02, Floyd-Warshall-tree-speedup-linear-tree (negative), 1x    3.34s   3.34s      0s
|  7 vcount=1500, p=0.02, Floyd-Warshall-tree-speedup (negative), 1x                3.02s   3.02s      0s

szhorvat: The distance matrix itself already takes up $O(|V|^2)$ storage, so I think we don't need to worry about the memory use of this approach.

szhorvat:
@rfulekjames Can you please remind me where we left off? Is the algorithm ready in principle, with the current data structures?  I know that the interface is not ready, and that there are a few low-level optimizations that one might look at, such as matrix traversal orders. But those can come later.

src/paths/floyd_warshall.c
```diff
@@ -140,3 +141,160 @@ igraph_error_t igraph_distances_floyd_warshall(
...
+            for the out_k tree
+        */
+        IGRAPH_CHECK(igraph_stack_int_push(&stack, k));
+        igraph_integer_t counter = 0;
+        while (!igraph_stack_int_empty(&stack)) {
+            igraph_integer_t parent = igraph_stack_int_pop(&stack);
+            if (parent > -1) {
+                VECTOR(dfs_traversal)[counter] = parent;
+                counter++;
+                IGRAPH_CHECK(igraph_stack_int_push(&stack, -parent - 1));
```
szhorvat, line 259: Why is the pushed value negated here?

rfulekjames: We use `stack` to construct both `dfs_traversal` and `dfs_skip` array. The non-negative values are used to construct `dfs_traversal` array, and the negative values `dfs_skip` array. In particular, values `-parent - 1` can be thought of as markers that are popped from the stack right after all the descendants of `parent` were processed. At that point `counter` stores the number of vertices seen up to that point. Therefore `VECTOR(dfs_skip)[-(parent + 1)] = counter` sets the index of the first vertex in `dfs_traversal` following `parent` that is not a descendant of the `parent` as desired.

rfulekjames: I can put there a comment explaining that.

rfulekjames:
@szhorvat 

> Can you please remind me where we left off? 

Currently, we have implemented the *tree algorithm* with **practical improvements** (page 8 of the paper) using *dfs* and *skip* arrays named *dfs_traversal* and *dfs_skip* in the current implementation + the new test on a random graph with 100 vertices.

> Is the algorithm ready in principle, with the current data structures?

It should be ready for review. I don't see how to further optimize the running time of the current implementation of the tree algorithm except the low-level stuff you mentioned.

Next, I plan to look at the *hour glass algorithm*, but this algorithm seems much trickier to implement, therefore alternatively I can implement the interface with what we have now and do the hour glass algorithm in a different PR. 
What do you think?

As for hour glass algorithm a recursive implementation would not be very hard. Otherwise, there is a need for a stack that stores a pointer to a tree in every element + some other info (possibly in a different stack, not sure what would be better/more desirable, or if there is any pattern to follow here. Is there an implementation I can use as an inspiration you can point me to?). It [looks like](https://github.com/igraph/igraph/blob/b5d5b1c9a5194a3b9b96b84a29548f874d0980ef/CHANGELOG.md?plain=1#L139)  that `igraph_vector_ptr_t`  is the go-to data structure, right?

src/paths/floyd_warshall.c
szhorvat: Can you please add a short comment in the code that explains this?

szhorvat:
Yes, the hourglass algorithm should definitely be a followup PR. I can't really comment on the details until I also read how the algorithm works, but regarding the stack of trees: yes, currently that would be the `vector_ptr_t`.

ntamas:
> there is a need for a stack that stores a pointer to a tree in every element + some other info (possibly in a different stack, not sure what would be better/more desirable, or if there is any pattern to follow here. Is there an implementation I can use as an inspiration you can point me to?)

Basically, we have two things that might be of help: we have `igraph_vector_list_t` and related data structures, these always store lists of vectors and take care of memory management; this would be satisfactory to your use-case if you want to represent trees as "parent vectors" (i.e. a vector where the i-th item points to the parent of vertex i in the tree) but not if you want to store anything more complex than that. Then we have `igraph_vector_ptr_t`, which is basically a standard vector of pointers, but then _you_ are responsible for some aspects of the memory management as the vector stores only pointers without knowing anything about the semantics of those pointers, and the ownership of the objects that you put in the vector via the pointers always stays with you. A rudimentary form of automatic memory management may be used with `igraph_vector_ptr_set_item_destructor()`, which lets you define a destructor function for the items in the vector, and in this case the `igraph_vector_ptr_t` will take ownership of the pointers that you put in it and destroy the items when they leave the vector. But this has proven to be inflexible and confusing in practice so we mostly abandoned the idea; if you use `igraph_vector_ptr_t`, it's better to be explicit about the memory management and not use the item destructor functionality at all.

szhorvat:
This is the next one on my list, but I need a bit of time before I can get to it.

szhorvat:
This looks good to me but I'd like to ask @ntamas to take a look as well.

szhorvat:
The ugliness here is the `children` matrix, which is essentially a memory-inefficient way to store an adjacency list for the $OUT_k$ tree. But in the end it's only as large as `res` itself, so it only doubles memory usage while keeping things simple. It think this is worth it, but I do wonder how the original authors implemented this.

I'm happy to merge this as-is (changelog can be updated later).

szhorvat:
Note to self:

 - We should add interruption support, which was missing from the previous version of this function.
 - Perhaps it makes sense to add a `from`/`to` parameter. While using these makes no sense from a performance viewpoint, it might be useful to present a uniform interface to high-level interfaces. E.g. in R, this will probably be integrated into the `distances` function anyway, which does have `from`/`to`. See also https://github.com/igraph/igraph/pull/2293#issuecomment-1381602277

szhorvat:
I forgot to note that I did experiment with changing iterations orders (i.e. make most things column-major) to try to speed things up, but without success.

ntamas:
Thanks a lot!

rfulekjames:
@ntamas @szhorvat 

Regarding the Hourglass algorithm from https://doi.org/10.26493/1855-3974.2467.497 

> Basically, we have two things that might be of help: we have igraph_vector_list_t and related data structures, these always store lists of vectors and take care of memory management; this would be satisfactory to your use-case if you want to represent trees as "parent vectors" (i.e. a vector where the i-th item points to the parent of vertex i in the tree) 

Ok. My current idea is to use `igraph_vector_list_t` as a stack for keeping track of changes to the $OUT_k$ tree, which at each step would be just a list of edges removed from $OUT_k$, and can be represented as an integer vector twice as long as the number of edges removed. Therefore only one graph $OUT_k$ would be stored and there would be no need for  `igraph_vector_ptr_t`.

However, for this to be efficient we need to be able to add and delete an arbitrary edge from $OUT_k$ in $O(1)$ time (or in a time very close to it) while still being able to list neighbors of an arbitrary vertex of degree $d$ of the modified graph in $O(d)$ time. Is there a way to do this with some already built-in functionality for `igraph_t`?

szhorvat:
Before moving on to the hourglass algorithm, I'd like to clarify something here.

It seems that instead of constructing out-trees for paths starting at `k`, the current version of the code actually constructs in-trees that _end_ at `k`. I can see that this allows a column-major traversal of `res`. Was this done on purpose @rfulekjames ?

Notice that for each `i`, the code considers a `j -> k -> i` path, not an `i -> k -> j` path.

I do not yet see 100% clearly that this reversal is implemented correctly everywhere, though. We should also update the comments.

szhorvat:
@rfulekjames 

The modifications you refer to are lines 12 and 15 of Algorithm 5, right?

If we store $OUT_k$ as an adjacency list, i.e. a list of children for each vertex, then removing the subtree $v_j$ means emptying the neighbour list of $v_j$. How about instead of explicitly removing items, we just mark $v_j$ as "inactive"?

One way to do this would be to assign a boolean `active[v]` to each vertex. When `v` gets inactivated on line 12, it also gets pushed onto a stack. This way, in order to reactivate vertices on line 15, we just pop them off the stack. We can use a negative marker on the stack (just like you did before) to know when to stop popping off elements.

Does this sound feasible?

rfulekjames:
@szhorvat 

> It seems that instead of constructing out-trees for paths starting at k, the current version of the code actually constructs in-trees that end at k. I can see that this allows a column-major traversal of res. Was this done on purpose @rfulekjames ?

Yes. See this [comment](https://github.com/igraph/igraph/pull/2267#discussion_r1042695355) from the above.

rfulekjames:
> The modifications you refer to are lines 12 and 15 of Algorithm 5, right?

Yes. (page 9)

> If we store 
>  as an adjacency list, i.e. a list of children for each vertex, then removing the subtree 
>  means emptying the neighbour list of 
> . How about instead of explicitly removing items, we just mark 
>  as "inactive"?

> 
> One way to do this would be to assign a boolean active[v] to each vertex. When v gets inactivated on line 12, it also gets pushed onto a stack. This way, in order to reactivate vertices on line 15, we just pop them off the stack. We can use a negative marker on the stack (just like you did before) to know when to stop popping off elements.
> 
> Does this sound feasible?

It might be worth trying this and see whether it gives any performance improvement, since as you describe it, it shouldn't be hard to implement. However, I think that one of the main benefits of the hour-glass algorithm is that we shouldn't even touch nodes that were removed (until we put them back to the tree).

The descendants of a removed node wouldn't be touched, at any rate, as in the tree algorithm. 
Therefore with marking instead of removing the only improvement over the tree algorithm would be that we sometimes perform a Boolean check instead of summing up two numbers and comparing it with a third which would come at the cost of constructing the second tree.

PR #2270, opened 2022-12-10
move community_edge_betweenness example to tests

GroteGnoom:
Output can be different on different machines:

```
393/528 Test #393: example::igraph_community_edge_betweenness .........***Failed    0.14 sec
-- Test case output differs from the expected output
-- See diff below:
-- -------------------------------------------------------
--- D:/a/igraph/igraph/examples/simple/igraph_community_edge_betweenness.out	2022-12-10 10:21:50.556396600 +0000
+++ D:/a/igraph/igraph/build-i686/tests/example_igraph_community_edge_betweenness.out	2022-12-10 10:24:38.908349400 +0000
@@ -1,11 +1,11 @@
 edges:
-2 3 0 1 4 7 5 6
+2 3 6 5 0 1 4 7
 edge betweenness:
-4 5 3.33333 4 2.5 4 1 1
+4 5 3.33333 6 1.5 2 1 1
 merges:
-2 4
-1 3
-6 5
-0 7
+3 4
+1 2
+0 6
+7 5
 bridges:
 7 6 5 3
```

So this needs some way to get consistent results. We can also remove the results printing or results checking.


szhorvat:
> Output can be different on different machines

Do you know why that is and if yes, can you explain?

szhorvat:
I added comparisons with tolerances. Can you please try again if results still differ between platforms?

szhorvat:
Why was this merged into `develop` only? Shouldn't it also go on `master`?

PR #2272, opened 2022-12-11
move bipartite_projection examples to tests

GroteGnoom:
Is there a canonical way to print graphs in examples? I am used to print_graph_canon in tests.


szhorvat:
You can use `igraph_write_graph_edgelist()` to `stdout`. It won't canonicalize the edge ordering though.

szhorvat:
Why is this targeting `develop` instead of `master`?

ntamas:
Could go to `master` IMHO.

szhorvat:
I cherry picked the two recent examples-to-tests commits (edge_betweenness and fastgreedy) back to master. I hope that's okay.

GroteGnoom:
I have an issue here which also relates to #2174: we use examples as tests, but we don't have the same functions available in examples as in tests. In this case it's the canonical graph printing.

If we just copy the functions over to the separate examples, the examples get bigger and more complex than they need to be. If we publicly expose the missing functions it's not ok because our API grows. Is there some better system for this?

szhorvat:
> we use examples as tests

I would say that we _run_ examples to make sure that they work. But proper examples are not intended for testing, and don't need to be written in a way so that the output can be checked. Of course, right now many examples are _effectively_ tests, which is why they should be moved to the test. But once an example is rewritten to be a true example, we don't need to include careful checks in them.

GroteGnoom:
All right, so just remove the `.out` file if I think it is hard to get a proper output?

ntamas:
Well, ideally, the example should be converted into a test first, with the `.out` file, and with canonical graph printing. Then a similar, but stripped-down version should be provided as an example. Examples get included in the built HTML documentation so they should be concise and educational, even at the expense of not being able to check the output (this is the responsibility of the test instead).

Issue #495, opened 2013-06-12
InternalError: Error at .\src\foreign.c:580: Parse error in Pajek file, line 1 (syntax error, unexpected ALNUM, expecting $end), Parse error


jankatins:
I try to load a *.net file, which I just saved via igraph back into a Graph object:

```
g.save("graph.net")
g = igraph.load("graph.net")
```

I get the following error:

```
---------------------------------------------------------------------------
InternalError                             Traceback (most recent call last)
<ipython-input-12-066efe7d4a1e> in <module>()
      1 progress_init("load graph")
----> 2 g = igraph.load("graph.net")
      3 progress_destroy()

C:\portabel\Python27\lib\site-packages\igraph\__init__.pyc in read(filename, *args, **kwds)
   3900     @param filename: the name of the file to be loaded
   3901     """
-> 3902     return Graph.Read(filename, *args, **kwds)
   3903 load=read
   3904 

C:\portabel\Python27\lib\site-packages\igraph\__init__.pyc in Read(klass, f, format, *args, **kwds)
   2128             raise IOError("no reader method for file format: %s" % str(format))
   2129         reader = getattr(klass, reader)
-> 2130         return reader(f, *args, **kwds)
   2131     Load = Read
   2132 

InternalError: Error at .\src\foreign.c:580: Parse error in Pajek file, line 1 (syntax error, unexpected ALNUM, expecting $end), Parse error
```

The file itself is quite big (1.1GB, IGRAPH UN-- 2143148 118606193 -- \n + attr: name (v)). The head of the file is:

```
C:\data\courses\data-science\kaggle>head graph.net
*Vertices 2143148
1"0"
2"1"
3"2"
4"3"
5"4"
6"5"
7"6"
8"7"
9"8"
```

PR #2276, opened 2022-12-20
Only use "name" (no "id") attribute in Pajek reader/writer

szhorvat:
This PR completely replaces the use of the "id" vertex attribute by the "name" attribute in the Pajek reader and writer.

Previously,

 - When reading Pajek files, vertex names were placed into both an "id" and a "name" attribute
 - When writing Pajek files, the "id" attribute was used for vertex names instead of the "name" attribute. High-level interfaces generally use "name" for identification.

See discussion starting here: https://github.com/igraph/rigraph/issues/511#issuecomment-1358981605

This is targeting `develop` because it can be considered a breaking change.

Before this is merged, we should check very carefully if there were any special reasons to not use "name". Perhaps there was an issue related to uniqueness considerations, but it was not documented. Note that the Pajek format does not use the words "name" and "id"—these are choices on the igraph side.

No hurry in merging this.

@ntamas I leave it to you whether any of this should be ported back to `master` (perhaps with better compatibility, such as retaining "id" in some cases. IMO it's not necessary.


ntamas:
Notes for ourselves: it should be checked whether Pajek (i.e. the _desktop_ software, not the file format) attaches any special meaning to the `id` attribute or not.

szhorvat:
Related: https://github.com/igraph/igraph/issues/495#issuecomment-19449397

The same change was requested, but rejected with the reasoning that it is breaking.

ntamas:
Well, it wasn't _rejected_ per se, note that the issue was closed as fixed earlier than the comment suggesting the change. I just added a note to state that it's a breaking change (which it is).

szhorvat:
Yes, didn't mean it in absolute terms, just trying to summarize. I was about to edit the comment when you responded.

Pajek seems to use the term "label" for what we are discussing as "vertex name" here in this issue: http://vladowiki.fmf.uni-lj.si/doku.php?id=pajek:faq:format

It's good to mention that the format does not seem to support arbitrary attribute names. There are some predefined attributes. So _I assume_ (but did not check) that the software also won't. The vertex number is referred to as "index" and is 1-based.

szhorvat:
Rebased on latest `develop`.

ntamas:
@szhorvat What shall we do with this? Shall we merge it as-is? Or shall we ask the Pajek authors at the workshop in June? (I believe someone is going to represent them).

szhorvat:
@ntamas I think we should merge this for `develop`.

PR #2279, opened 2022-12-20
feat: igraph_induced_subgraph_edges()

### LINKED ISSUE #2277: A function to find edges contained within an induced subgraph

A function that returns the edge IDs contained within a subgraph induced by some vertices.

```
igraph_error_t igraph_induced_subgraph_edges(const igraph_t *graph, igraph_vs_t vids, igraph_vector_int_t *edges);
```

The `edges` output parameter will contain the IDs of edges whose both endpoints are within `vids`. In other words, it returns the edges in the subgraph induced by `vids`.

**Use cases for the feature**

 - Faster than `igraph_induced_subgraph()`, and sometimes we don't need a new graph, only its edges.
 - Existing functions don't give easy access to the edges in a subgraph. If we construct a new graph using `igraph_induced_subgraph(_map)`, the edges IDs in that graph do not relate to the edge IDs in the original graph.
 - See discussion at https://github.com/igraph/python-igraph/issues/507#issuecomment-1339407988
 - Suppose we need to compute statistics on edge attributes of a subgraph. E.g., what is the total edge weight contained within a community?

**Notes**

 - The name `igraph_induced_subgraph_edges()` is very close to `igraph_subgraph_edges()`, which is not ideal. Personally I do not like the name `igraph_subgraph_edges()` as it sounds as if it produced edges as _output_. In fact it takes edges as input, so I'd find the name `igraph_edge_subgraph()` more intuitive.
 - It might make sense to provide multiple implementations with different performance characteristics, just the `impl` parameter of `igraph_induces_subgraph()`. The function could work like this:

        for v in vids:
            for e in incident_edges(v):
                if other(e, v) in vids:
                    result.push_back(e)

    Now the question is how to test for membership in `vids`. One way is to allocate a boolean array of `vcount` size and mark each vertex from `vids` in it. Then we have an O(1) time, but O(vcount) memory solution.  Another way is to use the existing set data structure, which has worse time performance but much better memory performance.  We should support taking tiny subgraphs out of very large graphs.

    My preference is to _not_ have an `impl` argument, and instead use the set implementation. We can add a marker-vector implementation for the case of large subgraphs later, if benchmarking indicates a need.

szhorvat:
Not yet tested proof of concept, fixes #2277. I need feedback on this before continuing to add tests.


ntamas:
Looks good, but wouldn't this add each edge twice for undirected graphs?

szhorvat:
Thanks, fixed.

@ntamas I think `igraph_i_incident()` is buggy when using `IGRAPH_LOOPS_ONCE`. Can you have a look? In the test I added here, edge 0 should appear, but `igraph_i_incident()` does not return it.

ntamas:
Workng on it.

szhorvat:
Thanks! I am not in a hurry with this, so it's fine after Christmas. Please see #2281 and #2280 for more.

ntamas:
The bug in `igraph_i_incident()` is now resolved and the problematic `multiple` parameter is also gone. Backport is coming soon.

szhorvat:
Other than deciding on the naming, I think this is good to go.

However, I have to say that I am not happy with the implementation. Different techniques are used to prevent duplicating self-loops (`IGRAPH_LOOPS_ONCE`) vs duplicating non-loop edges (`u-v` is included only when `u <= v`). It would be much cleaner if we could somehow view the graph as directed, even if it's undirected.

Does anyone have better ideas on how to implement this well?

ntamas:
No I don't, so if you are happy with the PR apart from this small bit, please mark it as ready for review.

szhorvat:
There's the question of the naming. Perhaps others have input there as well?

ntamas:
IMHO `igraph_induced_subgraph_edges()` is okay; the problematic bit is `igraph_subgraph_edges()`, which should be renamed to `igraph_subgraph_from_edges()`.

ntamas:
Can you please extend the changelog before I merge this?

szhorvat:
Done.

Issue #2249, opened 2022-10-31
An Ant-colony optimisation metaheuristic graph colouring function.


professorcode1:
A function that takes undirected graphs and performs graph colouring using the Ant-colony optimisation meta-heuristic. It will take as argument 
1) The undirected graph
2) Optionally it can take a list of list of nodes where for each list: the nodes inside the list must have consecutive colours i.e. 
[
    [1,2,6],
    [7,8]
] the 1,2 and 6 must have consecutive colours and 7 and 8 must have consecutive colours.
3) Optionally it can take a map that maps a node to a list of colours, such that the node cannot take any value inside the list as a colour. i.e 
{
    1 => [5,7]
    2 =>  [1,2,3]
}
Then for all attempted colouring's, 5,7 will not be assigned to node 1 as colours.  
4) Optional argument  K, the number of colours to attempt to colour the graph with. If not provided the algorithm will start with K = K_start and attempt to reduce K to a smaller value till valid colouring's can be found for that K.
5) Optional argument K_start, if not provided will be equal to the number of nodes in the graph. 

Uses 
This function would be useful as without any of the optional arguments it can be used to estimate the chromatic number of the graph, and with the optional arguments it can be used to perform scheduling. 

References 
I have already implemented this function (all be it poorly) for a college project 
https://github.com/professorcode1/College-Time-Table-Scheduler
using the following paper
https://www.researchgate.net/publication/249906915_A_New_Ant_Algorithm_for_Graph_Coloring
If you assign this issue to me I would like to add it to the library with proper code.

szhorvat:
It's great that you'd like to add a new feature to igraph. You are welcome to work on it, and will find quick start instructions here: https://github.com/igraph/igraph/wiki/Quickstart-for-new-contributors

However, I must note that at this moment we have limited capacity to review new contributions. There are several outstanding pull requests that we must first deal with. Therefore, we may be slow to respond in the upcoming weeks, and it will likely take longer than usual until we can review your code.

To make things as quick and as smooth as possible, please first write up a concise but complete summary of the algorithm here in this issue. This will aid the review. Also, as noted in the quickstart guide, please open a PR as soon as you have a barebones setup—don't wait until everything is complete and working!

professorcode1:
#2250 I have written the setup

professorcode1:
[...]

PR #2284, opened 2022-12-26
Feat/d satur graph coloring

professorcode1:
Dsatur implemented as a heuristic to greedy graph coloring. Commiting to run tests.


professorcode1:
The error are cause by documentation generation but I didn't cause them. I did not alter any files that have anything to do with documentation.

ntamas:
Just a quick review; I'll have more time for this after the holiday season. I'd probably leave the example as is because it is meant for educational purposes and it's also included in the manual so it should be kept short. The code that tests the DSatur algorithm should be placed in `tests/unit` instead; I'd probably take the already-modified example, copy it to `tests/unit`, and then revert the copy in `examples/simple` to show the greedy coloring only. After that, I would refactor the part that tests the validity of the coloring in the unit test to a separate function.

As for the algorithm itself, it seems like a naive O(n<sup>2</sup>) implementation as the vertex selection takes O(n) time and it has to be repeated n times. Using a binary heap to store saturation values could bring it down to something like O((n+m) log n), which could make a real difference for large and sparse graphs. This is probably outside the context of this PR, but it has to be done as igraph aims to provide not only _some_ implementation to graph algorithms but an efficient one.

One thing that _is_ in the context of this PR, though, is to avoid having to repeatedly query the degrees of vertices with `igraph_degree_1()`. You can fetch all the degrees at once with `igraph_degree()` into a vector and then use that vector instead. The advantage of this is that you could do it in a context where you could wrap the `igraph_degree()` call in an `IGRAPH_CHECK()`; right now it's called in a place where error handling is neglected.

src/misc/coloring.c
```diff
@@ -129,6 +131,105 @@ static igraph_error_t igraph_i_vertex_coloring_greedy_cn(const igraph_t *graph,
...
+    IGRAPH_FINALLY(igraph_vector_int_destroy, &saturation_degree);
+
+    igraph_vector_int_t edge_degree;
+    IGRAPH_CHECK(igraph_vector_int_init(&edge_degree, igraph_vcount(graph) ) );
+    IGRAPH_CHECK(igraph_degree(graph, &edge_degree, igraph_vss_all(), IGRAPH_ALL, IGRAPH_LOOPS ) );
+    IGRAPH_FINALLY(igraph_vector_int_destroy, &edge_degree);
+
+    igraph_vector_int_fill(colors, -1);   // -1 as a color means uncolored
+
+    igraph_integer_t vc = igraph_vcount(graph);
```
szhorvat, line 209: Please move this to the beginning and use`vc` wherever `igraph_vcount()` is currently used.

src/misc/coloring.c
```diff
@@ -129,6 +131,105 @@ static igraph_error_t igraph_i_vertex_coloring_greedy_cn(const igraph_t *graph,
...
+static igraph_error_t igraph_i_vertex_coloring_dsatur_cn(
+    const igraph_t *graph, igraph_vector_int_t *colors
+) {
+    igraph_adjlist_t adjlist;
+    IGRAPH_CHECK(igraph_adjlist_init(graph, &adjlist, IGRAPH_ALL, IGRAPH_LOOPS_TWICE, IGRAPH_MULTIPLE));
+    IGRAPH_FINALLY(igraph_adjlist_destroy, &adjlist);
+
+    igraph_vector_int_t saturation_degree;
+    IGRAPH_CHECK(igraph_vector_int_init(&saturation_degree, igraph_vcount(graph) ) );
+    IGRAPH_FINALLY(igraph_vector_int_destroy, &saturation_degree);
```
szhorvat, line 200: Instead of a pair of `igraph_vector_int_init()` and `IGRAPH_FINALLY()`, use the convenience macro `IGRAPH_VECTOR_INT_INIT_FINALLY`. The arguments are the same as for `igraph_vector_int_init()`.

src/misc/coloring.c
```diff
@@ -129,6 +131,105 @@ static igraph_error_t igraph_i_vertex_coloring_greedy_cn(const igraph_t *graph,
...
+
+    igraph_vector_int_t edge_degree;
+    IGRAPH_CHECK(igraph_vector_int_init(&edge_degree, igraph_vcount(graph) ) );
+    IGRAPH_CHECK(igraph_degree(graph, &edge_degree, igraph_vss_all(), IGRAPH_ALL, IGRAPH_LOOPS ) );
+    IGRAPH_FINALLY(igraph_vector_int_destroy, &edge_degree);
```
szhorvat, line 205: If the `igraph_degree()` call fails, then we have a memory leak. To avoid this, `IGRAPH_FINALLY` should follow `igraph_vector_int_init()` _directly_. However, in this case, just use `IGRAPH_VECTOR_INT_INIT_FINALLY()`.

src/misc/coloring.c
```diff
@@ -129,6 +131,105 @@ static igraph_error_t igraph_i_vertex_coloring_greedy_cn(const igraph_t *graph,
...
+
+    igraph_vector_int_fill(colors, -1);   // -1 as a color means uncolored
+
+    igraph_integer_t vc = igraph_vcount(graph);
+    igraph_integer_t vertices_colored = 0;
+    while (vertices_colored < vc) {
+        igraph_integer_t node_to_color = igraph_i_dsatur_select_node(graph, colors, &saturation_degree, &edge_degree, vc);
```
szhorvat, line 212: In `igraph_i_dsatur_select_node()` you iterate through all vertices, which leads to an overall $O(|V|^2)$ complexity, as @ntamas mentioned. This colouring approach can be implemented more efficiently by using a priority queue, similar to how it's done for `IGRAPH_COLORING_GREEDY_COLORED_NEIGHBORS`. You can use `igraph_2wheap_t` data structure, just like in `igraph_i_vertex_coloring_greedy_cn()`. See that function for reference and let us know if you have questions.

src/misc/coloring.c
```diff
@@ -129,6 +131,105 @@ static igraph_error_t igraph_i_vertex_coloring_greedy_cn(const igraph_t *graph,
...
+    igraph_integer_t nbr_cnt = igraph_vector_int_size(neighbours);
+    for (igraph_integer_t nbr_indx = 0; nbr_indx < nbr_cnt; nbr_indx++){
+        igraph_integer_t nbr = VECTOR(*neighbours)[nbr_indx];
+        if (!is_color_used_by_neighbour(colors, color, igraph_adjlist_get(adjlist, nbr))) {
+            VECTOR(*saturation_degree)[nbr]++;
+        }
+    }
+}
+
+static igraph_error_t igraph_i_vertex_coloring_dsatur_cn(
```
szhorvat, line 191: Please drop the `_cn` suffix from the name of this function.

szhorvat (changes requested):
I took a look, and the approach looks correct, but the code needs to be refactored to have reasonable complexity before the PR is merged.  This is not difficult to do if you use `igraph_i_vertex_coloring_greedy_cn()` as a reference for your implementation. Please take a look and let us know if you have questions.

szhorvat:
The problem is that we want to consider both the degree and the "saturation degree", but igraph's heaps can only consider one integer value. I'd suggest addressing the other review comments first, and later (probably in the new year) we can figure out how to proceed.

There's no way to _not_ consider the degrees as at least at the beginning of the colouring procedure, as otherwise there will be many ties.

One option is to just use C++, as its standard library has general heaps, allowing comparison based on both degree and saturation degree.

Another is to extend igraph with a generic heap data structure, but that is definitely beyond the scope of this PR.

szhorvat:
There is one more issue I just noticed: the algorithm, as described in Brélaz's paper, calls for resolving ties not using the plain degree, but the degree in the subgraph of yet uncoloured vertices. Thus, when a new vertex is coloured, the degrees of its uncoloured neighbours should be decremented by 1.

Another note is that we will need to pay attention to make sure that self-loops are effectively ignored.

src/misc/coloring.c
```diff
@@ -129,6 +131,105 @@ static igraph_error_t igraph_i_vertex_coloring_greedy_cn(const igraph_t *graph,
...
+        igraph_integer_t node_to_color = igraph_i_dsatur_select_node(graph, colors, &saturation_degree, &edge_degree, vc);
+        igraph_vector_int_t *neighbours = igraph_adjlist_get(&adjlist, node_to_color);
+        for (igraph_integer_t color = 0; color < vc; color++) {
+            igraph_bool_t viable_color = !is_color_used_by_neighbour(colors, color, neighbours);
+            if (viable_color) {
+                igraph_i_dsatur_update_saturation_degree(&adjlist, &saturation_degree, neighbours, colors, color);
+                VECTOR(*colors)[node_to_color] = color;
+                break;
+            }
+        }
```
szhorvat, line 221: Here you choose a new colour by iterating through all possible colours, and for each colour, iterating through neighbours of the current vertex. I believe it would be more efficient to retrieve neighbour colours once, then sort them. Iterate through the sorted list, and choose the lowest unused colour you find. This avoids several iterations through the neighbour list.

professorcode1:
I have made all possible changes. As you pointed out, binary heap to store (saturation degree, degree) would require too much code. Rest I have made all the changes.  

> Another note is that we will need to pay attention to make sure that self-loops are effectively ignored.
The edges are queried twice, one when making the adjacency list on line 236
` 
    IGRAPH_CHECK(igraph_adjlist_init(graph, &adjlist, IGRAPH_ALL, IGRAPH_LOOPS_TWICE, IGRAPH_MULTIPLE));
`
and once on line 245
`
    IGRAPH_CHECK(igraph_degree(graph, &edge_degree, igraph_vss_all(), IGRAPH_ALL, IGRAPH_NO_LOOPS ) );
`
245 ignores self edges with the IGRAPH_NO_LOOPS. I just copied the line 236 from the greedy implementation and I cannot find any documentation on the function igraph_adjlist_init that takes 5 parameters since in the docs it takes 3 parameter 
[docs version](https://igraph.org/c/doc/igraph-Data-structures.html#igraph_adjlist_init).

When moving code around and making final changes before merging please keep the function `igraph_i_dsatur_main_loop` with the same signature. You can move the functions it calls inside it but please do not remove it. I am also implementing an ant-colony metaheuristic graph coloring algorithm as well and that one requires this function being present.

ntamas:
> The problem is that we want to consider both the degree and the "saturation degree", but igraph's heaps can only consider one integer value.

A cheap trick is to limit ourselves to graph with less than 2<sup>31</sup> vertices and then use bits 31-62 to store the saturation degree and bits 0-30 to store the degree itself, in a single 64-bit integer. Not sure if I would really want to go down that road, though; C++ heaps are probably a better option.

professorcode1:
I can quickly re write this in C++ and I would also prefer that since the Ant Colony algorithm is also coming out  needlessly complex in C and is just in general difficult to write in C.

professorcode1:
Please let me know if you would like me to shift it to C++, and also let me know if its okay if I write the ant colony algorithm in C++ as well.

szhorvat:
Please wait with the C++ rewrite for a bit.

What features, other than the max heap, do you need from C++ for the ant colony graph colouring?

----

In the meantime, can you please complete this part? This does not depend on the decision about C++ in any way:

> Just a quick review; I'll have more time for this after the holiday season. I'd probably leave the example as is because it is meant for educational purposes and it's also included in the manual so it should be kept short. The code that tests the DSatur algorithm should be placed in `tests/unit` instead; I'd probably take the already-modified example, copy it to `tests/unit`, and then revert the copy in `examples/simple` to show the greedy coloring only. After that, I would refactor the part that tests the validity of the coloring in the unit test to a separate function.

professorcode1:
I have to use a boolean array with length equal to number of vertices to represent subset of vertices. This can be replaced with sets. (There is a igraph_set_t in the code but no doc so I don't know how to use it). There are 5 array with the property that if one changes all 5 should change together. This is a textbook use case of OOP. I needed a vector of 4 tuple so I made a struct with 4 vectors in C. The code is 60% done. I will finish it soon and it looks kind of like a mess. C++ would just make it a lot cleaner is all. I will write the ant colony code and open a draft pull request soon(hopefully today). Once that is done I'll write the unit tests.

professorcode1:
The file src/core/set.h has function to create a set, push to it, clear it and check if an element belongs. But I also need to be able to take intersection and remove elements.

szhorvat:
We are well aware of the advantages of C++ when it comes to complex projects like igraph. However, given that igraph is a C library, there are good reasons to avoid C++ as much as possible.

I would like to keep the DSatur implementation as pure C. To make this possible I have (partially) implemented a generic heap data structure which you will be able to use. Let's pause this PR until that feature is merged. Right now things are moving slowly as people are on holidays. You can follow the progress in #2285.

Regarding the use of C++, some of the issues which come up here:

 1. igraph is aiming to behave predictably on all platforms. This means that the priority queue should return elements in the same order on all platforms, regardless of the specific C++ standard library implementation. It is unclear to me that this will be the case.
 2. One needs to take special care when using C++ exceptions and igraph's own error handling together. This is not an unsurmountable problem, but it does make mistakes more likely, and makes the maintenance of the code harder.
 3. It is useful to keep using a uniform approach throughout the codebase. This means that once someone has learned how igraph works, they can keep working with any part of it. IMO this is a major reason why so many open source projects stick to C instead of C++: it lowers the barrier to contributions, especially when those contributions come from students or academics (not software engineers). igraph is a scientific project after all.

All this does not mean that C++ can never be used in igraph. I'm sure you noticed that it _is_ used in several functions already. It also doesn't mean that igraph will never move to C++ in the future, but that won't happen for the 1.0 release. Thus let us discuss how to proceed with the ant colony graph coloring in #2249, and figure out if there are any deal breakers such as point 1 above.

szhorvat:
The generic heap feature is now merged into `master`. You can go ahead and use it.

professorcode1:
Ok. I will do that.  Once I do that I will write the unit tests.

I understand the reasons the repo uses C. Untill rust becomes the De facto standard and is taught in universities we all use C I guess. 

Giving you an update on the ant colony code. It has hit a hurdle. The set implementation in igraph is bugged(kind of). I will open an issue and ask to be assigned to rewrite the set using AVL trees to allow all POD's instead of just integers. untill that is done I will have to halt the work on ant colony, since it would be a nightmare to code it without sets (in terms of both code complexity which is already above 800 lines and the performance)

szhorvat:
Let's please proceed step by step, as follows:

 1. Finish this PR first before moving on to the ant colony graph colouring
 2. Discuss what missing  library features and operations are needed for the ant colony algorithm (sets, etc.)
 3. Decide whether the ant colony algorithm will use C++ or whether we need an improved set data structure

professorcode1:
Ok. Here I have written the for DSatur using gen2wheap, and am gonna write the unit tests now but there is a problem. calling `igraph_gen2wheap_destroy` causes the program to seg fault. And if i comment it then `igraph_adjlist_destroy` causes the following message to print.
```
corrupted size vs. prev_size
Aborted (core dumped)
```
I am committing so please take a look.

szhorvat:
Just a few more small notes:

 - The deactivation feature is not necessary for this algorithm, so do not use it. Simply remove the top element from the heap instead of deactivating it.
 - In `igraph_i_dsatur_update_heap()`, you cast away the `const` from the return value of `igraph_gen2wheap_get()` and modify it. The `const` is there for a reason. You _must_ not modify elements this way. Modification should only be done using `igraph_gen2wheap_modify()`.

professorcode1:
I have fixed the 2 mistakes you pointed out, but it is still throwing the heap-buffer-overflow on linux_vendor_32. I am sure that I am not accessing outside bounds anywhere in my code. You will have to take a look whenever you can. I will write the unit tests and finish the work on my part.

src/misc/coloring.c
```diff
@@ -129,6 +133,135 @@ static igraph_error_t igraph_i_vertex_coloring_greedy_cn(const igraph_t *graph,
...
+    const igraph_adjlist_t *adjlist, igraph_gen2wheap_t *node_degrees_heap, 
+    igraph_vector_int_t *neighbours, igraph_vector_int_t *colors, 
+    igraph_integer_t color
+) {
+    igraph_gen2wheap_delete_max(node_degrees_heap);
+    igraph_color_metadata_t color_metadata; 
+    igraph_integer_t nbr_cnt = igraph_vector_int_size(neighbours);
+    for (igraph_integer_t nbr_indx = 0; nbr_indx < nbr_cnt; nbr_indx++){
+        igraph_integer_t nbr = VECTOR(*neighbours)[nbr_indx];
+        color_metadata = *((igraph_color_metadata_t*)igraph_gen2wheap_get(node_degrees_heap, nbr));
```
szhorvat, line 180: This crashes because you are trying to access an element without checking whether it is actually contained within the heap. I suggest you take a look at how the `IGRAPH_COLORING_GREEDY_COLORED_NEIGHBORS` is implemented and mirror that. Most of the greedy colouring approaches, _including DSatur_, will be very similar to each other and can be implemented in similar ways.

professorcode1:
Unit tests are written and all tests passed.. I'd say this feature is complete.

src/misc/coloring.c
```diff
@@ -21,10 +21,14 @@
...
 
 #include "igraph_interface.h"
+#include "igraph_memory.h"
 #include "igraph_adjlist.h"
 
 #include "core/indheap.h"
 #include "core/interruption.h"
+#include "core/genheap.h"
+
+/* Greedy coloring heuristics */
```
szhorvat, line 31:
```suggestion
/* COLORED_NEIGHBORS: Choose vertices based on the number of already coloured neighbours. */
```

I noticed that you keep referring to the heuristic based on the number of already coloured neighbour as "greedy" and contrasting it with DSatur. However, all methods in this file are various kinds of greedy colourings. DSatur is also a greedy colouring.

src/misc/coloring.c
```diff
@@ -129,6 +133,136 @@ static igraph_error_t igraph_i_vertex_coloring_greedy_cn(const igraph_t *graph,
...
+            if(VECTOR(*colors)[neighbour] != -1){
+                VECTOR(viable_colors)[VECTOR(*colors)[neighbour]] = false;
+            }
+        }
+        igraph_integer_t color = igraph_i_vector_bool_first_true(&viable_colors);
+        igraph_i_dsatur_update_heap(adjlist, node_degrees_heap, neighbours, colors, color);
+        VECTOR(*colors)[node_to_color] = color;
+
+        vertices_colored++;
+    }
```
szhorvat, line 225: Due to the use of `viable_colors`, we have an $O(|V|^2)$ complexity. Notice that there are several iterations over `viable_colors`, including even the `igraph_vector_bool_fill()` call.

See my earlier comment on how to select new colours efficiently, https://github.com/igraph/igraph/pull/2284#discussion_r1057946570  You can use `igraph_i_vertex_coloring_greedy_cn()` for reference.

src/misc/coloring.c
```diff
@@ -129,6 +133,136 @@ static igraph_error_t igraph_i_vertex_coloring_greedy_cn(const igraph_t *graph,
...
+    igraph_integer_t vc = igraph_vcount(graph);
+
+    IGRAPH_CHECK(igraph_vector_int_resize(colors, vc));
+    igraph_vector_int_fill(colors, -1);   // -1 as a color means uncolored
+
+
+    igraph_adjlist_t adjlist;
+    IGRAPH_CHECK(igraph_adjlist_init(graph, &adjlist, IGRAPH_ALL, IGRAPH_LOOPS_TWICE, IGRAPH_MULTIPLE));
+    IGRAPH_FINALLY(igraph_adjlist_destroy, &adjlist);
+    igraph_adjlist_simplify(&adjlist);
```
szhorvat, line 243: Do we need this, and if yes, why? Where exactly does the algorithm go wrong when using the unsimplified adjlist? Can we handle that without brute-force simplification (which is expensive)?

tests/unit/igraph_coloring.c
```diff
@@ -0,0 +1,56 @@
...
+        igraph_integer_t color = VECTOR(*colors)[i];
+        if(color<0 || color>=vertex_count){
+            printf("The vertex %" IGRAPH_PRId " has invalid color %" IGRAPH_PRId "\n", i, color);
+            return IGRAPH_FAILURE;
+        }
+    }
+    return IGRAPH_SUCCESS;
+}
+
+int main(void) {
```
szhorvat, line 29: So far there's a test done on a very large random graph. That's great. We need a few more test cases for confidence though:

 - Add a few small test graphs, something small enough that it should be possible to verify the result manually. Print out the colouring, and create an `.out` file to verify it.  You'll find more information in https://github.com/igraph/igraph/wiki/How-to-write-unit-tests%3F
 - Test the function on the zero-vertex and one-vertex graphs.
 - Test the function on non-simple graphs (containing multi-edges and self-loops).

tests/unit/igraph_coloring.c
```diff
@@ -0,0 +1,56 @@
...
+
+    /* Set default seed to get reproducible results. */
+    igraph_rng_seed(igraph_rng_default(), 42);
+
+    /* Simple undirected graph. */
+    igraph_erdos_renyi_game(&graph, IGRAPH_ERDOS_RENYI_GNM, 1000, 10000, IGRAPH_UNDIRECTED, IGRAPH_NO_LOOPS);
+
+    /* Initialize the vector into which the function will write its results. */
+    igraph_vector_int_init(&colors, 0);
+    IGRAPH_CHECK(igraph_vertex_coloring_greedy(&graph, &colors, IGRAPH_COLORING_GREEDY_COLORED_NEIGHBORS));
```
szhorvat, line 41: Error handling macros should not be used in tests. Note that this construct would try to return an igraph error code, `igraph_error_t`, while `main()` must return an exit code (of type `int`) which is semantically different. The two must not be mixed.

I'll add another comment with a suggestion on how to handle failures.

tests/unit/igraph_coloring.c
```diff
@@ -0,0 +1,56 @@
...
+igraph_error_t verify_coloring(igraph_t *graph, igraph_vector_int_t *colors){
+    /* Verify that the colouring is valid, i.e. no two adjacent vertices have the same colour. */
+    igraph_integer_t i;
+    /* Store the edge count to avoid the overhead from igraph_ecount in the for loop. */
+    igraph_integer_t no_of_edges = igraph_ecount(graph);
+    for (i = 0; i < no_of_edges; ++i) {
+        if ( VECTOR(*colors)[ IGRAPH_FROM(graph, i) ] == VECTOR(*colors)[ IGRAPH_TO(graph, i) ]  ) {
+            printf("Inconsistent coloring! Vertices %" IGRAPH_PRId " and %" IGRAPH_PRId " are adjacent but have the same color.\n",
+                   IGRAPH_FROM(graph, i), IGRAPH_TO(graph, i));
+            return IGRAPH_FAILURE;
```
szhorvat, line 14: Instead of returning an error code, make this function `void` and replace the `printf` with the `IGRAPH_FATALF` macro. The argument are exactly the same as that of `printf`, so they don't need to be modified.

Normally, we use `IGRAPH_ASSERT` in tests, which calls `igraph_fatal()` internally. Here `IGRAPH_FATALF` si a better fit as it can print a more informative failure message.

These functions also ensure that a stack trace will be printed on failure, so one can tell which exact `verify_coloring()` call failed, if there are multiple.

examples/simple/igraph_coloring.c
```diff
@@ -15,21 +15,8 @@ int main(void) {
...
-        /* Store the edge count to avoid the overhead from igraph_ecount in the for loop. */
-        igraph_integer_t no_of_edges = igraph_ecount(&graph);
-        for (i = 0; i < no_of_edges; ++i) {
-            if ( VECTOR(colors)[ IGRAPH_FROM(&graph, i) ] == VECTOR(colors)[ IGRAPH_TO(&graph, i) ]  ) {
-                printf("Inconsistent coloring! Vertices %" IGRAPH_PRId " and %" IGRAPH_PRId " are adjacent but have the same color.\n",
-                       IGRAPH_FROM(&graph, i), IGRAPH_TO(&graph, i));
-                abort();
-            }
-        }
-    }
```
szhorvat, line 32: Please re-add these removed lines to the example, as they still serve an educational purpose.

szhorvat (changes requested):
I did an initial review round. The function seems to work well, but as currently written, it still has quadratic complexity. This would be the first thing to correct.

src/misc/coloring.c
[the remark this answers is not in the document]
professorcode1: I didn't write that comment. That was written by @ntamas.

szhorvat:
Sorry about that, didn't realize that you didn't add the comment.

> I will do all of these things.

Thanks!

> Also I can remove all the internal functions and move everything into 1 functions like igraph_i_vertex_coloring_greedy_cn if you like

Please use whatever structuring you find the most readable. There's nothing wrong with breaking it down to small functions.

src/misc/coloring.c
professorcode1: By account for these problems , I mean problems cause by assumption of the graph bring simple broken. A bunch of if statements will have to run in every loop there is.

professorcode1: Which will also have a performance hit.

tests/unit/igraph_coloring.c
professorcode1: Writing a .out file for colorings does not seem like a feasible idea. Since coloring is an NP Complete problem there is no correct solutions and different algorithms will come to different solutions. A coloring is a set partition is labelling is assigning a number to said paritions so the exact same coloring can have different labellings as well, so I just don't know how i would go about making a .out file.

szhorvat: Here the purpose is not to verify that the colouring is valid. We have a function for that. The aim is to detect if the algorithm gets broken in the future. Consider e.g. how the first version in this PR considered the degree in the entire graph rather than just thenon-coloured part of the graph.

 - Let the program print the colour assigned to each vertex
 - Record the output into an `.out` file

This will help detect if the code is inadvertently broken in the future.

src/misc/coloring.c
szhorvat: I am not asking for changes here. I am asking you why the simplification is necessary. Is this just a pre-emptive measure, just in case the current implementation doesn't work for multigraphs? Or do you know _how exactly_ it fails for multigraphs?

src/misc/coloring.c
professorcode1: I have copied the method from igraph_i_vertex_coloring_greedy_cn (in essence, in practice that function created a vector then sorted it, I just created a set instead). So this is done.

src/misc/coloring.c
professorcode1: It is just a pre-emptive measure when i wrote that comment. Looking at the code thoroughly, only one line will cause problems which is 270
```
       igraph_color_metadata.edge_degree = igraph_vector_int_size(igraph_adjlist_get(&adjlist, vertex));

```
When the heap to store (saturation degree, degree) is being created by inserting each vertex the degree will be calculated incorrectly. So to calculate it correctly, the simple graph  edge degree for each vertex will have to calculated anyways. And if loop and multiple edges are left in adjlist they will only slow the algorithm loop down. So it makes no sense not to do this.

professorcode1: by not to do this, I am referring to storing the adjlist as a simple graph

tests/unit/igraph_coloring.c
professorcode1: Ok. I will created a .out file

szhorvat:
Just click the "Resolve conversation" button on all review commits where there is clearly nothing more to be done. If you have questions, or are not sure that everything was addressed, just ask. Let us know when this PR is ready for the next review round.

src/misc/coloring.c
```diff
@@ -129,6 +134,150 @@ static igraph_error_t igraph_i_vertex_coloring_greedy_cn(const igraph_t *graph,
...
+    const igraph_t *graph, igraph_vector_int_t *colors
+) {
+    igraph_integer_t vc = igraph_vcount(graph);
+
+    if (vc == 0) {
+        return IGRAPH_SUCCESS ;
+    }
+
+    IGRAPH_CHECK(igraph_vector_int_resize(colors, vc));
+
```
szhorvat, line 250:
```suggestion
    IGRAPH_CHECK(igraph_vector_int_resize(colors, vc));

    if (vc == 0) {
        return IGRAPH_SUCCESS ;
    }
```

Otherwise the size of the `colors` vector won't be set correctly for the null graph.

src/misc/coloring.c
```diff
@@ -129,6 +134,150 @@ static igraph_error_t igraph_i_vertex_coloring_greedy_cn(const igraph_t *graph,
...
+    if (vc == 1) {
+        VECTOR(*colors)[0] = 0; //igraph vectors always have size 1
+        return IGRAPH_SUCCESS ;
+    }
+
+    igraph_vector_int_fill(colors, -1);   // -1 as a color means uncolored
+
+
+    igraph_adjlist_t adjlist;
+    IGRAPH_CHECK(igraph_adjlist_init(graph, &adjlist, IGRAPH_ALL, IGRAPH_NO_LOOPS, IGRAPH_NO_MULTIPLE));
```
szhorvat, line 260: Note to self: This is not yet correct. Reciprocal edges in directed graphs are effective multi-edges, so we still need to investigate whether multi-edges can be left in, or what changes the implementation needs to handle these appropriately.

src/misc/coloring.c
```diff
@@ -129,6 +134,150 @@ static igraph_error_t igraph_i_vertex_coloring_greedy_cn(const igraph_t *graph,
...
+    }
+    igraph_integer_t state = 0, prev_element, next_element;
+    igraph_set_iterate(viable_colors, &state, &prev_element);
+    while (igraph_set_iterate(viable_colors, &state, &next_element)) {
+        if (next_element != prev_element + 1) {
+            return prev_element + 1;
+        }
+        prev_element = next_element;
+    }
+    return prev_element + 1;
```
szhorvat, line 203: Here you are assuming that set elements are iterated over in increasing order. This is not guaranteed and cannot be relied on.

Why don't you use the sorted vector approach that I now suggested several times? Simply retrieve and sort the neighbour colours, then iterate through the list and find the first gap, just as here. It will be more efficient than building a set and it does not rely on internal implementation details. See the other heuristic for details.

professorcode1: Sorry I'll do that now

professorcode1:
The PR is ready for the next round of review

ntamas:
Fixed typos, formatting issues, unnecessary whitespace changes and unneeded includes. I haven't inspected `src/misc/coloring.c` thoroughly yet.

ntamas:
Tests are failing. Note that earlier tests did not use the expected output file because the expected output file had a typo in its name. Now it does, and the expected output does not match what the test case actually prints.

I'm going to commit a few improvements to the formatting of `src/misc/coloring.c`, then I'll hand it back to you to fix the code. For me it seems like the coloring returned from the algorithm is correct, but the color _index_ is one larger than expected (i.e. the indexing starts from 1 and not 0).

The `igraph_i_get_first_viable_color()` function (this is the new name that I'm about to commit soon) would benefit from some comments to explain how it works and what it's supposed to do.

szhorvat:
When I asked for printing the output, I meant including some manually specified small graphs where it is feasible to examine the output visually and reason about it. What is included now is still too large for this. Here are a few graphs that I suggest including:

```
igraph_small(&graph, 0, IGRAPH_UNDIRECTED,
    0, 3, 0, 4, 1, 4, 2, 4, 1, 5, 2, 5, 3, 5, 0, 6, 1, 6, 2, 6, 3, 6, 0, 7, 1, 7, 2, 7, 4, 7, 5, 7,
    -1);
```

image

This one has a chromatic number of 3, but the existing greedy colouring heuristic cannot seem to produce fewer than 4 colours. I wonder if DSatur can.

```
igraph_lcf(&graph, 8, 2, 8, 0);
```

image

This one has the same property: chromatic number is 4, but existing greedy colouring can only produce 5.

DSatur is proven to be exact on a few graph classes, so these should be tested as well. Wheel graph (`igraph_wheel()`) on odd/even vertices has a chromatic number of 3/4.  Bipartite graph (try a large one using `igraph_bipartite_game_gnm()` but do not print the colouring) will of course have a chromatic number of 2.

src/misc/coloring.c
szhorvat: Please revert this to preserve loops and multi-edges. The current setup is not sufficient for directed graphs anyway, and we want to see what goes wrong with non-simple graphs so we can fix the issue if necessary.

src/misc/coloring.c
```diff
@@ -129,6 +132,160 @@ static igraph_error_t igraph_i_vertex_coloring_greedy_cn(const igraph_t *graph,
...
+static igraph_error_t igraph_i_dsatur_main_loop(
+    const igraph_t *graph, igraph_vector_int_t *colors,
+    igraph_gen2wheap_t *node_degrees_heap, igraph_adjlist_t *adjlist,
+    igraph_integer_t vc
+) {
+    igraph_integer_t vertices_colored = 0;
+    igraph_vector_int_t viable_colors;
+    igraph_integer_t color;
+
+    IGRAPH_VECTOR_INT_INIT_FINALLY(&viable_colors, 0);
```
szhorvat, line 219: This vector is called "viable colors", but as far as I can see, it holds colours which are NOT viable, correct?

src/misc/coloring.c
```diff
@@ -129,6 +132,160 @@ static igraph_error_t igraph_i_vertex_coloring_greedy_cn(const igraph_t *graph,
...
+}
+
+static igraph_error_t igraph_i_get_first_viable_color(
+    igraph_vector_int_t *viable_colors, igraph_integer_t color_count,
+    igraph_integer_t *result
+) {
+    IGRAPH_CHECK(igraph_vector_int_resize(viable_colors, color_count));
+    igraph_vector_int_sort(viable_colors);
+    igraph_integer_t i = 0;
+    igraph_integer_t col = 0;
```
szhorvat, line 196: It seems to me that this should start at -1, not 0, to ensure that isolated vertices will be assigned colour 0.

The reason why I am asking for detailed tests that are easy enough to verify manually is so that these issues can be noticed.

src/misc/coloring.c
```diff
@@ -129,6 +132,160 @@ static igraph_error_t igraph_i_vertex_coloring_greedy_cn(const igraph_t *graph,
...
+        color_metadata.edge_degree--;
+        igraph_gen2wheap_modify(node_degrees_heap, nbr, &color_metadata);
+    }
+}
+
+static igraph_error_t igraph_i_get_first_viable_color(
+    igraph_vector_int_t *viable_colors, igraph_integer_t color_count,
+    igraph_integer_t *result
+) {
+    IGRAPH_CHECK(igraph_vector_int_resize(viable_colors, color_count));
```
szhorvat, line 193: What is the purpose of this line?

src/misc/coloring.c
```diff
@@ -129,6 +132,160 @@ static igraph_error_t igraph_i_vertex_coloring_greedy_cn(const igraph_t *graph,
...
+
+    while (vertices_colored < vc) {
+        igraph_integer_t node_to_color = igraph_gen2wheap_max_index(node_degrees_heap);
+        igraph_vector_int_t *neighbours = igraph_adjlist_get(adjlist, node_to_color);
+        igraph_integer_t neighbours_count = igraph_vector_int_size(neighbours);
+        igraph_integer_t color_count = 0;
+        IGRAPH_CHECK(igraph_vector_int_resize(&viable_colors, neighbours_count));
+        for (igraph_integer_t neighbor_index = 0; neighbor_index < neighbours_count; neighbor_index++) {
+            igraph_integer_t neighbour = VECTOR(*neighbours)[neighbor_index];
+            if (VECTOR(*colors)[neighbour] != -1) {
```
szhorvat, line 229: If you are going to skip `-1` value, use `igraph_vector_int_push_back()` to fill the vector. Then you only need to `igraph_vector_int_clear()` the vector at the beginning, but don't need to bother with resizing.

src/misc/coloring.c
professorcode1: yes, that is correct. I will fix that

szhorvat:
@ntamas Let us know if you're done pushing to this branch, so we don't step on each others' toes.

In the meantime I added some comments to the implementation of the other heuristic to make it clearer how it works.

src/misc/coloring.c
professorcode1: I copied it straight from igraph_i_vertex_coloring_greedy_cn but I realise now it was probably written in a different context. I will re-write it.

ntamas:
I won't push to this branch now so feel free to go ahead

professorcode1:
The tests are failing because when I wrote the tests the output I used was produces by the faulty logic that sets are iterated over in a sorted order. Now that that is fixed, the next commit will (hopefully) be the true correct result. 
@ntamas please tell me what command I can use to run only a specific test, I know I can use 
`cmake --build . --target check -j(number of cores)` to run them all.

professorcode1:
@szhorvat I have written the tests you have asked and it is working as intended. 
For the small graph it was able to produce the chromatic number 3
For the lcf graph it was not able to produce chromatic number 4, it also produced 5. However the colorings are not bijective so they are distinct.  
For the wheel graph the result is correct 
For the bipartite graph the result is correct 
For the isolated vertices graph the result is correct

ntamas:
> please tell me what command I can use to run only a specific test

`cmake --build . --target build_tests && ctest -R coloring`

The `-R` switch of `ctest` allows you to run tests matching a certain regex only. Note that the code snippet above still compiles all tests; it's not a problem if you use `ccache` or a similar compiler cache.

professorcode1:
@szhorvat its ready for review

szhorvat:
FYI I'll be pushing some changes to this branch today.

> please tell me what command I can use to run only a specific test

To add to what @ntamas said, you can also execute a test manually. In this case, the executable is `tests/test_igraph_coloring`. This won't compare the output of course.

To build just this one test, you can use `cmake --build . --target test_igraph_coloring`. Or more simply `make test_igraph_coloring`, assuming that you are using `make`.

szhorvat (approved):

szhorvat:
Thanks for your work on this! It looks good to me.

I'll leave merging to @ntamas (please use squash-and-merge!)

src/misc/coloring.c
```diff
@@ -136,6 +137,141 @@ static igraph_error_t igraph_i_vertex_coloring_greedy_cn(const igraph_t *graph,
...
+
+static igraph_error_t dsatur_get_first_viable_color(const igraph_vector_int_t *used_colors_sorted) {
+    igraph_integer_t color_count = igraph_vector_int_size(used_colors_sorted);
+    igraph_integer_t i = 0;
+    igraph_integer_t col = 0;
+    while (i < color_count && VECTOR(*used_colors_sorted)[i] == col) {
+        while (i < color_count && VECTOR(*used_colors_sorted)[i] == col) {
+            i++;
+            if (i == color_count) {
+                break; /* loop second condition could read outside bounds without this */
```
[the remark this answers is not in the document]
professorcode1: I have always been told that it can be compiler dependent and that it should not be risked.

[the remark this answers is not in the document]
szhorvat: `&&` and `||` are short-circuiting in C. This is not compiler dependent. https://en.cppreference.com/w/c/language/operator_logical

If you overload these operators in C++, there is no short-circuiting though (as they might be performing any function).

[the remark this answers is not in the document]
professorcode1: I didn't know that. Thanks!

ntamas (approved):

ntamas:
Thanks a lot! :tada:

professorcode1:
This was my first contribution to open source, thanks for being helpful and patient throughout!

PR #1939, opened 2022-01-22
Steiner Tree Feature

Nachiket18:
The functionality that is being developed is calculating the Steiner tree in a graph. Steiner Tree is created on basis of subset of terminals in the graphs with the condition that length of the tree should be minimum.

We are following the approach suggested in this paper - https://arxiv.org/abs/1710.00668

Also, we are taking reference of an FPT algorithm known as Dreyfus- Wagner algorithm - https://onlinelibrary.wiley.com/doi/10.1002/net.3230010302

----

Edit by @szhorvat:

TODO list to be completed only _after_ the first review:

 - [x] interruptability
 - [ ] functions.yaml
 - [ ] documentation
 - [x] checks on input parameters
 - [ ] overflow checks
 - [x] decide whether `weights=NULL` should be accepted (meaning identical weights) and how this is best handled

PR #2288, opened 2023-01-10
Feat/Steiner Tree

Nachiket18:
As per instructions from the previous PR [Link](https://github.com/igraph/igraph/pull/1939). We fetch upstream the master branch  , created a new branch add added the 4 files that we worked on in the previous PR to update the files.

The functionality that is being developed is calculating the Steiner tree in a graph. Steiner Tree is created on basis of subset of terminals in the graphs with the condition that length of the tree should be minimum.

We are following the approach suggested in this paper - https://arxiv.org/abs/1710.00668

Also, we are taking reference of an FPT algorithm known as Dreyfus- Wagner algorithm - https://onlinelibrary.wiley.com/doi/10.1002/net.3230010302


szhorvat:
Thanks, this is great, and will make it much easier to continue!

Can you please remove the accidentally committed `.DS_Store` files?

GroteGnoom:
Could you add the function to src/CMakeLists.txt and the test to tests/CMakelists.txt? Then the tests will run.

alekkk94:
Is the Steiner Tree algorithm available now in March 2023?

ntamas:
The PR is still unmerged; you can try checking out this branch and compiling it on your own if you want to use the code in this PR, but it's not in an official release yet.

src/paths/steiner.cpp
```diff
@@ -0,0 +1,586 @@
...
+/*
+ *  Generates Subsets of length > 1 for list of integer values.
+ *  As algorithm computes the minimum distance between set of variables ranging from
+ *  length 2 to steiner_terminals - 1, we need these subsets.
+ *  for e.g [2,3,4] --> [ [2,3], [2,4], [3,4], [2,3,4] ]
+ *  Populates the subsetMap data structure as well.
+ *
+ */
+static igraph_error_t generateSubsets(const igraph_vector_int_t *steinerTerminals, igraph_integer_t n, igraph_integer_t graphsize, dictionary& subsetMap, std::set<int_set> & allSubsets) {
+    if (n > sizeof(igraph_integer_t)){
```
GroteGnoom, line 48: is `n > sizeof(igraph_integer_t)` really the check you want? For example, if igraph_integer_t is 32-bits, sizeof(igraph_integer_t) will give 4. The maximum n for an unsigned int would be 31, and for a signed int it would be 30. So I think you want sizeof(int) *8 - 2. Right?

src/paths/steiner.cpp
```diff
@@ -0,0 +1,586 @@
...
+        IGRAPH_ERROR("igraph_integer_overflow detected. The given number of terminals is more than what the computer can handle.",IGRAPH_EINVAL);
+    }
+    igraph_integer_t count = ((igraph_integer_t)1 << n);
+    igraph_integer_t subsetIndex = graphsize;
+
+    // The outer for loop will run 2^n times to get all subset .
+    // Here variable i will act as a binary counter
+
+    for (igraph_integer_t i = 2; i < count; i++) {
```
GroteGnoom, line 57: It's probably nice to add a comment on why this can start at 2.

src/paths/steiner.cpp
```diff
@@ -0,0 +1,586 @@
...
+
+                    if (E != j) {
+                        igraph_real_t distanceEJ = MATRIX(distance, E, j);
+
+                        int_set DMinusE = D;
+
+                        /*
+                         *  A set with Singleton value E removed from subset D
+                         */
+                        for (auto item :  DMinusE) {
```
GroteGnoom, line 504: Why don't you just do
```                       
DMinusE.erase(E);
```
instead of this loop?

src/paths/steiner.cpp
```diff
@@ -0,0 +1,586 @@
...
+            }
+        }
+    }
+
+    igraph_real_t distance2 = IGRAPH_INFINITY;
+
+    for (igraph_integer_t j = 0; j < no_of_nodes; j++) {
+        igraph_real_t distance1 = IGRAPH_INFINITY;
+        IGRAPH_ALLOW_INTERRUPTION();
+        for (igraph_integer_t subset_C_iterator = 0; subset_C_iterator < no_of_terminals; subset_C_iterator++) {
```
GroteGnoom, line 536: I'm a little surprised that `subset_C_iterator` is not actually an iterator :smile:

src/paths/steiner.cpp
```diff
@@ -0,0 +1,586 @@
...
+            igraph_integer_t F = VECTOR(steiner_terminals_copy)[subset_C_iterator];
+            igraph_real_t distanceFJ = MATRIX(distance, F, j);
+
+            int_set CMinusF;
+
+            for (igraph_integer_t k = 0; k < no_of_terminals; k++) {
```
GroteGnoom, line 542: Isn't this all way easier if you make C a set too, instead of a vector? Although I'm not sure about the performance.

src/paths/steiner.cpp
```diff
@@ -0,0 +1,586 @@
...
+
+                if (VECTOR(steiner_terminals_copy)[k] != F) {
+                    CMinusF.insert(VECTOR(steiner_terminals_copy)[k]);
+                }
+            }
+
+            igraph_integer_t indexOfSubsetCMinusF = fetchIndexofMapofSets(CMinusF, subsetMap);
```
GroteGnoom, line 549: I'm not sure if this can be done better, but I just wonder: Are you sure this is the best way of implementing this loop? Because you're creating a set (CMinusF), but the only thing you're doing with it, is to get its index. Isn't there an easier way?

GroteGnoom:
I was just trying things, and ran this test:
```
    {
        igraph_t g;
        igraph_vector_int_t terminals;
        igraph_real_t value;
        igraph_vector_int_t tree_edges;
        igraph_full(&g, 10, IGRAPH_UNDIRECTED, 0);
        igraph_vector_int_init_int(&terminals, 9, 0, 1, 2, 3, 4, 5, 6, 7, 8);
        igraph_vector_int_init(&tree_edges, 0);
        igraph_steiner_dreyfus_wagner(&g, &terminals, NULL, &value, &tree_edges);
        igraph_vector_int_print(&tree_edges);
        printf("value: %f", value);
        igraph_destroy(&g);
        igraph_vector_int_destroy(&tree_edges);
        igraph_vector_int_destroy(&terminals);
    }
```
This causes a heap-buffer-overflow at
```
                MATRIX(dp_cache, indexOfSubsetD, j) = IGRAPH_INFINITY;
```
currently at line 488.
It's because indexOfSubsetD is 256, but the matrix has only `(2 ^ 7 + 10) = 138` rows.

src/paths/steiner.cpp
RonakGSahu: Yes. sorry for the silly mistake. Haven't used that function for a while. Will Fix it and push it

src/paths/steiner.cpp
Nachiket18: Thanks. Yes it makes sense instead of looping. I have made the change

src/paths/steiner.cpp
Nachiket18: Yes. I added a comment saying that algorithms needs subsets of at least size 2 as distance for singleton can be easily calculated using adjacency matrix

src/paths/steiner.cpp
Nachiket18: `typedef std::set<igraph_integer_t> int_set;` C and CMinusF are both of the type int_set. I maybe have misunderstood but it is already a set

GroteGnoom: I was calling `steiner_terminals_copy` `C` because you use `subset_C_iterator` to iterate over it, and copying all of its elements except `F` results in CMinusF.

`steiner_terminals_copy` is a vector, and therefore you can't just take it and do `.erase(F)` like you could with `DMinusE`.

GroteGnoom:
@Nachiket18 @RonakGSahu 

I think the best way to go forward is to only fix the essential stuff, test thoroughly and then merge this.
So for now, ignore all my comments about performance issues, or refactorings, and let's fix the issues in this last comment:
https://github.com/igraph/igraph/pull/2288#issuecomment-1474829820
and add some more tests, because this issue was not caught by the current tests.

Also it would be nice for us to know if you think it's ok for us to step in a bit more and fix stuff ourselves. Right now, we only review and leave the changes up to you, but things would be a bit more streamlined if we would also make commits. I'm a little worried that this would otherwise take many months to merge. Let us know what you think, and what your expectations are :slightly_smiling_face:

Nachiket18:
@GroteGnoom - It would be okay if you want to make any changes yourselves. I and @RonakGSahu would try to work on https://github.com/igraph/igraph/pull/2288#issuecomment-1474829820 this weekend.

Nachiket18:
> @Nachiket18 @RonakGSahu
> 
> I think the best way to go forward is to only fix the essential stuff, test thoroughly and then merge this. So for now, ignore all my comments about performance issues, or refactorings, and let's fix the issues in this last comment: [#2288 (comment)](https://github.com/igraph/igraph/pull/2288#issuecomment-1474829820) and add some more tests, because this issue was not caught by the current tests.
> 
> Also it would be nice for us to know if you think it's ok for us to step in a bit more and fix stuff ourselves. Right now, we only review and leave the changes up to you, but things would be a bit more streamlined if we would also make commits. I'm a little worried that this would otherwise take many more months to merge. Let us know what you think, and what your expectations are 🙂

We figured out what the issue was. Our understanding about number of subsets was incorrect. Previously we thought the value should be 2 ^ (n-1)  where n stands for number of Steiner terminals - 1, but the power set contains 2 ^ n elements. Now we don't want singleton sets and a null set so the value is 2 ^ n - (n+1). We initialized the DP table appropriately. We executed the same the test case that you have mentioned and reproduced the bug. Thanks for pointing out the issue and sending the test case it was really helpful.  As per our understanding the result should have 4 edges since the graph is complete (Each edge connects two terminals).

GroteGnoom:
> As per our understanding the result should have 4 edges since the graph is complete (Each edge connects two terminals).

But it's a single connected tree, right? And a tree with n edges only connects n + 1 nodes.

RonakGSahu:
From the documentation of igraph_full, we have that in a full graph every possible edge is present, every vertex is connected to every other vertex.

GroteGnoom:
I'm not arguing that any nodes are not connected. A Steiner tree should be a tree. A tree with n edges cannot connect more than n + 1 nodes.

Also, when I try the test with 6 terminals, it still gives me 4 edges. I expect 5, and it seems you would expect 3.
And when I try the test with 2 terminals it gives me an error.

So something is not correct. Maybe I'm completely misinterpreting things and the test is wrong, but right I now I think it's the algorithm.

Nachiket18:
> I'm not arguing that any nodes are not connected. A Steiner tree should be a tree. A tree with n edges cannot connect more than n + 1 nodes.
> 
> Also, when I try the test with 6 terminals, it still gives me 4 edges. I expect 5, and it seems you would expect 3. (And when I try the test with 2 terminals it gives me an error.) (Also, of course even if you would connect 2 terminals with every edge, for 9 terminals you would still need more than 4 edges.)
> 
> So something is not correct. Maybe I'm completely misinterpreting things and the test is wrong, but right I now I think it's the algorithm.

You are right tree with 9 terminals cannot be connected with 4 edges. It's not necessary that every edge in complete graph connect two Steiner terminals. Sorry I was mistaken. Since we changed the size of the Dynamic Programming table the changes had to be trickled down for generation of Steiner Tree. I made a change to code which is much simpler and doesn't use Combination function. I ran the test that you have prepared and it gave me 8 edges as you had expected.

GroteGnoom:
Great!
There's now also this issue:
>  (And when I try the test with 2 terminals it gives me an error.)

```
Fatal error at src/paths/steiner.cpp:89 : The Subset's index that you tried to find doesn't exist. Hence the code won't run.

Stack trace:
    #0 0x7f21882ccb9a in __sanitizer_print_stack_trace /usr/src/debug/gcc/gcc/libsanitizer/asan/asan_stack.cpp:87
    #1 0x5650d35881ee in igraph_abort /home/daniel/Documents/igraph/src/core/error.c:69
    #2 0x5650d358993e in igraph_fatal_handler_abort /home/daniel/Documents/igraph/src/core/error.c:421
    #3 0x5650d3589994 in igraph_fatal /home/daniel/Documents/igraph/src/core/error.c:430
    #4 0x5650d36c1b75 in fetchIndexofMapofSets /home/daniel/Documents/igraph/src/paths/steiner.cpp:89
    #5 0x5650d36c7754 in igraph_steiner_dreyfus_wagner /home/daniel/Documents/igraph/src/paths/steiner.cpp:537
    #6 0x5650d3587f48 in main /home/daniel/Documents/igraph/tests/unit/igraph_steiner_tree_fpt.c:282
    #7 0x7f218003c78f  (/usr/lib/libc.so.6+0x2378f)
    #8 0x7f218003c849 in __libc_start_main (/usr/lib/libc.so.6+0x23849)
    #9 0x5650d357e6f4 in _start (/home/daniel/Documents/igraph/build/tests/test_igraph_steiner_tree_fpt+0x2dc6f4)

fish: Job 1, 'tests/test_igraph_steiner_tree_…' terminated by signal SIGABRT (Abort)
```
This happens for 1, 2, and 3 steiner terminals. 0 terminals gives another error.
Could you fix that too?

Nachiket18:
@GroteGnoom - For the terminals 0,1,2,3 which graph was used? Was it same as the one with 9 terminals? I can reproduce the test and see where the code is failing.

GroteGnoom:
> @GroteGnoom - For the terminals 0,1,2,3 which graph was used? Was it same as the one with 9 terminals? I can reproduce the test and see where the code is failing.

Yes, the full graph with 10 vertices.

Nachiket18:
There's a condition to handle 0 terminals here - https://github.com/Nachiket18/igraph/blob/feat/steiner/src/paths/steiner.cpp#L374

For one terminal what should be the value that needs to be returned? It's not really a Steiner Tree since it's just one terminal. We can return 0.0 if that's okay.

For two terminals - I think it's just finding the shortest path using Dijkstra's algorithm and we can add a piece of code for that if it's necessary. We would check why it's failing for three terminals

ntamas:
> For one terminal what should be the value that needs to be returned? 

I think 0.0 is okay. 

> For two terminals - I think it's just finding the shortest path using Dijkstra's algorithm

Yes, exactly, but it should be handled anyway - we need to cover all cases.

Nachiket18:
@ntamas - Okay. I am working on 0 terminal and 1 terminal but the test case seems to be failing with the error 

"
Test exited abnormally with error: Subprocess aborted
"
I'm unable to understand why the test is failing. Is there a way to know more about the error?

szhorvat:
What you need to do when something like this happens is to run the test executable directly:

```
tests/test_igraph_steiner_tree_fpt
```

The last line of the output is informative:

```
tests/unit/igraph_steiner_tree_fpt.c:290 : Finally stack is not empty (stack size is 1). Check that the number in IGRAPH_FINALLY_CLEAN matches the IGRAPH_FINALLY count.
fish: Job 1, 'tests/test_igraph_steiner_tree_…' terminated by signal SIGABRT (Abort)
```

This is coming from `VERIFY_FINALLY_STACK();`, which contains `abort()`. I will fix up this macro so errors would be a bit more informative in the future. Please do merge `master` periodically into this branch, to keep it up to date with new development in igraph, and to pick up the improvement I'm about to make for `VERIFY_FINALLY_STACK();`

Nachiket18:
@szhorvat - Thanks. I was able to fix the error.

@GroteGnoom, @ntamas   - I was able to write a test case for 1 terminal and it was successful. Will work further on the test case for two and three terminals.

Nachiket18:
@szhorvat - I was working on test case for two terminals and it passes on my local machine however some pipelines are failing. The error is -

```
FAILED: tests/example_igraph_grg_game 
: && /usr/bin/cc -O3 -DNDEBUG -Wl,--disable-new-dtags tests/CMakeFiles/example_igraph_grg_game.dir/__/examples/simple/igraph_grg_game.c.o -o tests/example_igraph_grg_game  -Wl,-rpath,/home/vsts/work/1/s/build/src  src/libigraph.so.3.0.0 && :
/usr/bin/ld: tests/CMakeFiles/example_igraph_grg_game.dir/__/examples/simple/igraph_grg_game.c.o: undefined reference to symbol 'hypot@@GLIBC_2.35'
/usr/bin/ld: /lib/x86_64-linux-gnu/libm.so.6: error adding symbols: DSO missing from command line
```
I'm unable to understand why this issue is occurring and how to fix it.

szhorvat:
This is a result of fix @ntamas is currently working on. The failure is also present on master and is not related to your work. You can ignore it for now.

Nachiket18:
@szhorvat Thanks. I am working on test case for three terminals and I'm facing some issues with respect to memory management. I was trying to see if I could get Address Sanitizers to work . I found this - https://github.com/igraph/igraph/wiki/Using-sanitizers-to-find-bugs but not sure how to compile the code so that stack trace is generated along with memory related errors like they do for azure pipelines.

szhorvat:
The stack trace is produced by AddressSanitizer when it detects an error. Additionally, igraph is set up that when AddressSanitizer is available, any `igraph_error()` or `igraph_fatal()` will also produce a stack trace.

All you need to do is enable AddressSanitizer. Use `-DUSE_SANITIZERS=Address` at the configuration step.

Nachiket18:
> -DUSE_SANITIZERS=Address

Thanks. I tried to configure using this command cmake /home/nachiket/igraph/build -DUSE_SANITIZERS=Address and then build the code but while testing the stack trace is not produced.

GroteGnoom:
> > -DUSE_SANITIZERS=Address
> 
> Thanks. I tried to configure using this command cmake /home/nachiket/igraph/build -DUSE_SANITIZERS=Address and then build the code but while testing the stack trace is not produced even when there's a memory error.

what do you see then?

If I introduce an error myself, and then
```
cmake .. -GNinja -DUSE_SANITIZER='Address;Undefined' -DCMAKE_BUILD_TYPE=Release
```
(I just tried 'Release' to see if it would give less output, usually you do 'Debug')
and then
```
cmake --build . --target check
```
I get
```
[545/546] Executing unit tests...
Test project /home/daniel/Documents/igraph/build/tests
361/543 Test #361: test::igraph_connect_neighborhood...................***Failed    0.05 sec
Warning at src/operators/connect_neighborhood.c:78 : Order smaller than two, graph will be unchanged
/home/daniel/Documents/igraph/src/operators/connect_neighborhood.c:95:24: runtime error: store to address 0x604000000330 with insufficient space for an object of type 'igraph_integer_t'
0x604000000330: note: pointer points here
 00 00 00 00  00 00 00 00 00 00 00 00  00 00 00 00 00 00 00 00  00 00 00 00 00 00 00 00  00 00 00 00
              ^ 
=================================================================
==34843==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x604000000330 at pc 0x556d84b3095c bp 0x7ffff8fdc420 sp 0x7ffff8fdc410
WRITE of size 8 at 0x604000000330 thread T0
    #0 0x556d84b3095b in igraph_connect_neighborhood (/home/daniel/Documents/igraph/build/tests/test_igraph_connect_neighborhood+0x3fa95b) (BuildId: 78ff861a02b59f9a88aba009616d5249638fc5f0)
    #1 0x556d849d78d9 in main (/home/daniel/Documents/igraph/build/tests/test_igraph_connect_neighborhood+0x2a18d9) (BuildId: 78ff861a02b59f9a88aba009616d5249638fc5f0)
    #2 0x7effba83984f  (/usr/lib/libc.so.6+0x2384f) (BuildId: 2f005a79cd1a8e385972f5a102f16adba414d75e)
    #3 0x7effba839909 in __libc_start_main (/usr/lib/libc.so.6+0x23909) (BuildId: 2f005a79cd1a8e385972f5a102f16adba414d75e)
    #4 0x556d849d8654 in _start (/home/daniel/Documents/igraph/build/tests/test_igraph_connect_neighborhood+0x2a2654) (BuildId: 78ff861a02b59f9a88aba009616d5249638fc5f0)

0x604000000330 is located 752 bytes after 48-byte region [0x604000000010,0x604000000040)
allocated by thread T0 here:
    #0 0x7effc22e0cd1 in __interceptor_calloc /usr/src/debug/gcc/gcc/libsanitizer/asan/asan_malloc_linux.cpp:77
    #1 0x556d84b2f0fc in igraph_connect_neighborhood (/home/daniel/Documents/igraph/build/tests/test_igraph_connect_neighborhood+0x3f90fc) (BuildId: 78ff861a02b59f9a88aba009616d5249638fc5f0)
    #2 0x556d849d78d9 in main (/home/daniel/Documents/igraph/build/tests/test_igraph_connect_neighborhood+0x2a18d9) (BuildId: 78ff861a02b59f9a88aba009616d5249638fc5f0)
    #3 0x7effba83984f  (/usr/lib/libc.so.6+0x2384f) (BuildId: 2f005a79cd1a8e385972f5a102f16adba414d75e)

SUMMARY: AddressSanitizer: heap-buffer-overflow (/home/daniel/Documents/igraph/build/tests/test_igraph_connect_neighborhood+0x3fa95b) (BuildId: 78ff861a02b59f9a88aba009616d5249638fc5f0) in igraph_connect_neighborhood
Shadow bytes around the buggy address:
  0x604000000080: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x604000000100: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x604000000180: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x604000000200: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x604000000280: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
=>0x604000000300: fa fa fa fa fa fa[fa]fa fa fa fa fa fa fa fa fa
  0x604000000380: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x604000000400: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x604000000480: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x604000000500: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x604000000580: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
Shadow byte legend (one shadow byte represents 8 application bytes):
  Addressable:           00
  Partially addressable: 01 02 03 04 05 06 07 
  Heap left redzone:       fa
  Freed heap region:       fd
  Stack left redzone:      f1
  Stack mid redzone:       f2
  Stack right redzone:     f3
  Stack after return:      f5
  Stack use after scope:   f8
  Global redzone:          f9
  Global init order:       f6
  Poisoned by user:        f7
  Container overflow:      fc
  Array cookie:            ac
  Intra object redzone:    bb
  ASan internal:           fe
  Left alloca redzone:     ca
  Right alloca redzone:    cb
==34843==ABORTING
Test exited abnormally with error: 1
=========================================
CMake Error at /home/daniel/Documents/igraph/etc/cmake/run_legacy_test.cmake:32 (message):
  Exiting test.



543/543 Test #543: test::igraph_hexagonal_lattice
99% tests passed, 1 tests failed out of 543

Total Test time (real) =  33.52 sec

The following tests FAILED:
	361 - test::igraph_connect_neighborhood (Failed)
Errors while running CTest
FAILED: tests/CMakeFiles/check /home/daniel/Documents/igraph/build/tests/CMakeFiles/check 
cd /home/daniel/Documents/igraph/build/tests && /home/daniel/.local/lib/python3.11/site-packages/cmake/data/bin/ctest --progress --output-on-failure -C Release
ninja: build stopped: subcommand failed.
```

Nachiket18:
> -DCMAKE_BUILD_TYPE=Release

With the command

`cmake .. /home/nachiket/igraph/build -DUSE_SANITIZERS='Address;undefined;Leak;Memory' -DCMAKE_BUILD_TYPE=Release`

I'm getting the following

```
Test project /home/nachiket/igraph/build/tests
220/544 Test #220: test::igraph_steiner_tree_fpt.......................***Failed    0.11 sec
double free or corruption (fasttop)
Test exited abnormally with error: Subprocess aborted
=========================================
CMake Error at /home/nachiket/igraph/etc/cmake/run_legacy_test.cmake:32 (message):
  Exiting test.

```

GroteGnoom:
strange :thinking: I just get the sanitizer output.

Do you see something like
```
-- -----------[ Testing ]----------
-- Diff tool:                  diff
-- Sanitizers:    Address;Undefined
-- Code coverage:                no
-- Verify 'finally' stack:       no
```
In the cmake info after the first command?
Which compiler are you using? I think the sanitizers get silently ignored if they're not available.

GroteGnoom:
Oh there's also a spelling error in there, USE_SANITIZER, not USE_SANITIZERS. 
maybe just try this, after you've emptied your build folder?
```cmake .. -DUSE_SANITIZER='Address;Undefined' -DCMAKE_BUILD_TYPE=Debug```
If somethings wrong it should just give you a warning.

**Comment by Nachiket18** [comment]
@GroteGnoom - Thanks for the helping out. 
I was able to write code to handle test case for 1,2,3 terminals. 
There's a small thing that I noticed. We initialized the iweights when weights are not passed as input. The problem is that when we encounter special case of 1,2,3 terminals and weights is not passed we need to destroy iweights before returning. Right now we have a code that I've put in every case. 

```
if (!weights) {
            igraph_vector_destroy(&iweights);
            IGRAPH_FINALLY_CLEAN(1);   
 }
```
I was wondering what's a way to write it more neatly and avoid code duplication.

**Comment by ntamas** [comment]
One trick that you can use is to move all cleanup (deallocation) code to the end of the function, label it with `cleanup` and use `goto cleanup;` whenever you are going to exit the function in an irregular manner. Something like:

```
igraph_error_t result = IGRAPH_SUCCESS;
igraph_vector_t iweights;

/* initialize iweights if needed, and make 'weights' point to it */
if (weights == NULL) {
    igraph_vector_init(&iweights, 0);
    weights = &iweights;
}

/* do stuff */

/* whenever something bad happens, do this: */
if (something_bad_happened) {
    result = IGRAPH_FAILURE;   /* or some other error code */
    goto cleanup;
}

cleanup:

if (weights == iweights) {
    /* we were using our own weight vector, so destroy it */
    igraph_vector_destroy(weights);
    IGRAPH_FINALLY_CLEAN(1);
}

/* other destructor calls go here */

return result;
```

Of course you should be careful to destroy things in exact reverse order compared to how they are initialized.

**Comment by Nachiket18** [comment]
@ntamas - I think we were not handling cleanup when something fails. I was fixing the case where code works fine but due to corner cases and when weights are not passed then it was causing the leak. I was able to write a simple if statement with flag variable that helps to reduce code duplication.  

```
if (flag_terminals == 1){
        if (!weights) {
            igraph_vector_destroy(&iweights);
            IGRAPH_FINALLY_CLEAN(1);
        }

        return IGRAPH_SUCCESS;
    }
```
What should be the further steps? @szhorvat @GroteGnoom @ntamas

**Comment by GroteGnoom** [comment]
I've tried this test:

```
    printf("\nA graph:\n");
    igraph_vector_int_t terminals;
    igraph_vector_int_t tree_edges;
    igraph_t g;
    igraph_real_t value;
    igraph_small(&g, 10, IGRAPH_UNDIRECTED,
                 0,1, 0,2, 0,3, 0,4, 0,5, 0,6,
                 2,7, 2,8, 2,9,
                 -1);
    igraph_vector_int_init_int(&terminals, 9, 0,1,2,3,4,5,6,7,8);
    igraph_vector_int_init(&tree_edges, 0);
    igraph_steiner_dreyfus_wagner(&g, &terminals, NULL, &value, &tree_edges);
    printf("tree edges:\n");
    igraph_vector_int_print(&tree_edges);
    printf("value: %f\n", value);

    igraph_vector_int_destroy(&tree_edges);
    igraph_vector_int_destroy(&terminals);

    igraph_destroy(&g);
```

And this is the result:
```
A graph:
tree edges:
0 0 1 1 6 1 2 1 7 1 3 4 4 5
value: 10.000000
```
So if I got this correctly then
 this result has 14 edges, while I expect 9,
some of which are repeated, which shouldn't happen
the value does not correspond with either 14 or 9 edges. 

So either I'm causing some issues myself, or the result is incorrect in multiple ways. If the result is incorrect then I'm really getting worried about this algorithm because this was the first thing I tested.

**Comment by Nachiket18** [comment]
@GroteGnoom - Thanks for pointing this out. 

The first issue where the Steiner tree distance (values) is incorrect. [Gave 10 when it should be 8]. We looked the dp cache to see if the values that are generating are correct. It seems that algorithm correctly captures all the values but due to the condition in the if statement [q != j] (See below) the value is incorrect.

```
if (q != j && (MATRIX(distance, q, j) + distance1) < distance2) {
            distance2 = MATRIX(distance, q, j) + distance1;
}
```
We fixed it by removing the condition and all the other test cases pass as well. [Not pushed the code yet]

For the second issue about the edges we are still figuring out what went wrong and will update you on it.

**Comment by Nachiket18** [comment]
I and @RonakGSahu are working on fixing the issue with generation of Steiner Tree. @GroteGnoom - Do you have any other test cases (right now we are triaging the last test case that you mentioned) that need to be tested so that once we make changes we would test it extensively before pushing the code.

**Comment by Nachiket18** [comment]
@RonakGSahu and I  were able to fix the issue. Due to the structure of the graph where one node was connected to 7 others  and the algorithm was removing that critical node so Steiner Tree edges had several nodes in common so  duplicates were getting generated. We fixed the issue using `std::set` which now additionally allows to have the edge id sorted and also fixed the issue mentioned in https://github.com/igraph/igraph/pull/2288#issuecomment-1577456563.

**Comment by GroteGnoom** [comment]
All right, I pulled your latest changes, and changed the last test. I removed a vertex from the terminals, so it looks like:
>     igraph_vector_int_init_int(&terminals_new, 8, 0, 2, 3, 4, 5, 6, 7, 8);

and this is the result:

```
Tree edges:
0 1 2 3 4 5 6 7
value: 7.000000
```

there are 8 edges, but the value is 7. But it has no weights so the value should be equal to the number of edges. Right?

**Comment by Nachiket18** [comment]
@GroteGnoom - We were able to fix the test case that you had provided. The edge 0 was extra and the value 7 was correct.  We also added few test cases from our end which led us to find out some bugs in the implementation and a slight problem in our understanding of the algorithm. The new test cases are

1.  Node 1,2 removed from the terminal list 

```
  printf("\nA` graph with 7 terminals and 2 removed to test if it forces it to go through:\n");
  igraph_vector_int_init_int(&terminals_new, 7, 0, 3, 4, 5, 6, 7, 8);
  igraph_vector_int_init(&tree_edges_new, 0);
  igraph_steiner_dreyfus_wagner(&g_new, &terminals_new, NULL, &value_new, &tree_edges_new);
  printf("Tree edges:\n");
  igraph_vector_int_print(&tree_edges_new);
  printf("value: %f\n", value_new);
```

2. This test case where the structure of the graph is little different was giving value which was incorrect so we brainstormed as why this is happening. It happens that as per algorithm when split subset D into subsets E and D-E. E is not necessarily singleton. We fixed the code with all possible combinations of E except when E = D.

```
printf("\nA graph with different structure that previous\n");
igraph_t g_new_1;
igraph_real_t value_new_1;
igraph_small(&g_new_1, 9, IGRAPH_UNDIRECTED,
             0, 1, 0, 2, 0, 3, 0, 6,
             2, 7, 2, 9,
             3, 4,
             4, 5,
             7, 8,
             -1
            );
igraph_vector_int_init_int(&terminals_new, 7, 0, 1, 2, 3, 4, 5, 7);
igraph_vector_int_init(&tree_edges_new, 0);
igraph_steiner_dreyfus_wagner(&g_new_1, &terminals_new, NULL, &value_new_1, &tree_edges_new);
printf("Tree edges:\n");
igraph_vector_int_print(&tree_edges_new);
printf("value: %f\n", value_new_1);
```

**Comment by GroteGnoom** [comment]
I tried this
```
    igraph_vector_init(&weights, 12);
    igraph_vector_fill(&weights, 100);
    igraph_small(&g_new, 9, IGRAPH_UNDIRECTED,
                0,1, 1,2,
                0,3, 1,4, 2,5,
                3,4, 4,5,
                3,6, 4,7, 5,8,
                6,7, 7,8,
                 -1);
    igraph_vector_int_init_int_end(&terminals_new, -1, 0, 4, 2, 8, -1);
    igraph_vector_int_init(&tree_edges_new, 0);
    igraph_steiner_dreyfus_wagner(&g_new, &terminals_new, &weights, &value_new, &tree_edges_new);
    printf("Tree edges:\n");
    igraph_vector_int_print(&tree_edges_new);
    printf("value: %f\n", value_new);
```
(so a square grid of 9 points)
and this was the result:
```
Tree edges:
0 1 2 5 7 10 11
value: 500.000000
```
The value is the value of the correct solution. The edges form a tree which connect all the terminals. 
But the edges are not a minimal solution, and the value does not correspond to the edges.

**Comment by RonakGSahu** [comment]
After going back to the algorithm and checking every line, we figured out that there was a major bug we had missed before which was fixed a couple of days back. After that when we were trying to generate the tree for which the previous backtracking algorithm wouldn't have worked because in the new code, we were not assuming a singleton set for which Dijkstra would have failed and the algorithm required to go back and check all possible combinations of the subsets. So, I implemented a front pass-through code to generate the edges that would have been needed for the final solution and it appears to be working for all and the correct values were generated after manual checking. And leaks were fixed.

**Comment by RonakGSahu** [comment]
Hey @GroteGnoom, are all the test cases you are running on the updated code working correctly?  We feel that we have corrected the implementation of the algorithm by refining the understanding of the paper.

**Comment by GroteGnoom** [comment]
I'm randomly generating graphs, and checking with the brute forcer.
```
    igraph_vector_init_real(&weights, 8, 0.218959, 0.383502, 0.5297, 0.417486, 0.526929, 0.910321, 0.328234, 0.247039);
    igraph_small(&g, 6, IGRAPH_UNDIRECTED,
            5,2,
            5,2,
            2,0,
            4,1,
            5,3,
            2,0,
            2,1,
            3,1,
                 -1);
    igraph_vector_int_init_int_end(&terminals, -1, 2, 0, 1, -1);
```
Results in:

```
Tree edges:
2 6 2
```
So it has edge 2 twice.

Nachiket18:
> I'm randomly generating graphs, and checking with the brute forcer.
> 
> ```
>     igraph_vector_init_real(&weights, 8, 0.218959, 0.383502, 0.5297, 0.417486, 0.526929, 0.910321, 0.328234, 0.247039);
>     igraph_small(&g, 6, IGRAPH_UNDIRECTED,
>             5,2,
>             5,2,
>             2,0,
>             4,1,
>             5,3,
>             2,0,
>             2,1,
>             3,1,
>                  -1);
>     igraph_vector_int_init_int_end(&terminals, -1, 2, 0, 1, -1);
> ```
> 
> Results in:
> 
> ```
> Tree edges:
> 2 6 2
> ```
> 
> So it has edge 2 twice.



> I'm randomly generating graphs, and checking with the brute forcer.
> 
> ```
>     igraph_vector_init_real(&weights, 8, 0.218959, 0.383502, 0.5297, 0.417486, 0.526929, 0.910321, 0.328234, 0.247039);
>     igraph_small(&g, 6, IGRAPH_UNDIRECTED,
>             5,2,
>             5,2,
>             2,0,
>             4,1,
>             5,3,
>             2,0,
>             2,1,
>             3,1,
>                  -1);
>     igraph_vector_int_init_int_end(&terminals, -1, 2, 0, 1, -1);
> ```
> 
> Results in:
> 
> ```
> Tree edges:
> 2 6 2
> ```
> 
> So it has edge 2 twice.

@GroteGnoom - Thanks for this test case. This test case is different as it has parallel edges with different weights. The code causing issues as it was accessing the if condition where we were checking if Steiner terminals are 3 then do a simple pairwise calculation. This was done before we did the recent major  changes and it seems that after those major changes this condition is not needed. Our question is that what should we do about parallel edges with different weights. As of right now it uses it for computation but we don't think that the larger weight edge will ever be used and at the start we can trim it for faster computation

RonakGSahu:
Any update on this.

GroteGnoom:
Sorry, I've been away for a long time, and am also quite busy now. Last time I checked everything seemed to work correctly, but I still need double check if we haven't missed anything.

RonakGSahu:
Sounds good. Take your time.

Nachiket18:
I was wondering whether there are more steps that have to be taken from our end for this PR in terms of testing.

GroteGnoom:
Maybe I have time this weekend or the weekend after to take another look.

GroteGnoom:
I checked again, added some tests, and I can't seem to find any inputs where your algorithm goes wrong (if I remember correctly some errors could be handled better). I haven't checked the code itself, so I'll have a talk with my colleagues about merging it like this.

GroteGnoom:
All right 🥳 we will look at the code and the docs to see if it still needs work, but the functionality seems there :) So thank you for all your work, and we'll take it from here to get it merged!

ntamas:
@GroteGnoom ask me for a review when you are done with this and I'll have a final look before it is merged.

RonakGSahu:
hi, rechecking on this. if we have to do any work

PR #2293, opened 2023-01-13
Refactor widest path implementation

szhorvat:
- [x] Clarify docs
 - [x] Simplify and speed up Floyd-Warshall: (1) makes no sense to use _lazy_ inclist (2) no need to use any inclist to construct the adj. mat., just iterate through edges (3) iterate through matrices in column-major order for better performance
 - [x] Avoid using an extra matrix when computing all-pairs results in Floyd-Warshall
 - [x] Do not copy infinities into separate variables


szhorvat:
`igraph_widest_path_widths_floyd_warshall()` has `from` and `to` parameters. However, this algorithm always computes the entire matrix for all vertex pairs. It is almost pointless to have these parameters. If only a small subset is required, then Dijkstra will be faster.

PR #2294, opened 2023-01-13
perf: improve memory efficiency of the Floyd-Warshall "Tree" algorithm

szhorvat:
This change stores the adjacency list of OUT_k contiguously, in a vector if length $V-1$, instead of in a $V\times V-1$ matrix. This requires two passes over the `predecessors` matrix as we need to count children before we can construct the adjacency list. There does not seem to be a noticeable performance impact, at least not on my machine.

@rfulekjames, can you take a look at this before I merge it? Two pairs of eyes notice more than just one, and it's late here.

I saw your question about an appropriate data structure for the hourglass algorithm, but I don't have an immediate answer, unfortunately. I'll take a close look at the paper and will think about it.

### DIFF
--- a/src/paths/floyd_warshall.c
+++ b/src/paths/floyd_warshall.c
@@ -67,13 +67,21 @@ static igraph_error_t igraph_distances_floyd_warshall_tree(
     igraph_matrix_int_t predecessors;
     IGRAPH_MATRIX_INT_INIT_FINALLY(&predecessors, no_of_nodes, no_of_nodes);
 
-    /* children[u][i] is the i-th child of u in a tree of shortest distances
-       rooted at k in the main loop below */
-    igraph_matrix_int_t children;
-    IGRAPH_MATRIX_INT_INIT_FINALLY(&children, no_of_nodes, no_of_nodes > 0 ? no_of_nodes-1 : 0);
+    /* children[children_start[u] + i] is the i-th child of u in a tree of shortest paths
+       rooted at k in the main loop below (OUT_k). There are no_of_nodes-1 child vertices
+       in total, as the root vertex is excluded. This is essentially a contiguously stored
+       adjacency list representation of OUT_k. */
+    igraph_vector_int_t children;
+    IGRAPH_VECTOR_INT_INIT_FINALLY(&children, no_of_nodes-1);
 
-    /* no_of_children[u] is the number of children of u
-       rooted at k in the main loop below */
+    /* children_start[u] indicates where the children of u are stored in children[].
+       These are effectively the cumulative sums of no_of_children[], with the first
+       element being 0. The last element, children_start[no_of_nodes], is equal to the
+       total number of children in the tree, i.e. no_of_nodes-1. */
+    igraph_vector_int_t children_start;
+    IGRAPH_VECTOR_INT_INIT_FINALLY(&children_start, no_of_nodes+1);
+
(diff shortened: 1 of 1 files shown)


src/paths/floyd_warshall.c
```diff
@@ -96,20 +104,34 @@ static igraph_error_t igraph_distances_floyd_warshall_tree(
...
             VECTOR(no_of_children)[parent]++;
         }
-        /* constructing dfs-traversal and dfs-skip arrays for the out_k tree */
+
+        igraph_integer_t cumsum = 0;
+        for (igraph_integer_t v = 0; v < no_of_nodes; v++) {
+            VECTOR(children_start)[v] = cumsum;
+            cumsum += VECTOR(no_of_children)[v];
+        }
+        VECTOR(children_start)[no_of_nodes] = cumsum;
```
ntamas, line 120: We have an `igraph_vector_int_cumsum()` function - could we use it here?

szhorvat: No. I actually added a comment explaining why. I will re-add it.

Given a length-`n` vector `v`, `vector_cumsum()` computes another length-n vector `c` where `c[0] == v[0]` and the last element of `c` is `sum(v)`.

Here we want to start at 0, i.e. `c[0] == 0`, and the length of `c` is `n+1`.

szhorvat:
Regarding [this comment](https://github.com/igraph/igraph/pull/2267#issuecomment-1382731017), I updated the code comments to match what the implementation does. It appears correct to me, but I'd appreciate a quick look before I merge this.

ntamas:
Looks good to me! If you want to, let's wait for a nod from @rfulekjames , otherwise feel free to merge.

src/paths/floyd_warshall.c
```diff
@@ -87,29 +102,46 @@ static igraph_error_t igraph_distances_floyd_warshall_tree(
...
+           one longer, with a 0 being prepended to what vector_cumsum() would produce. */
+        igraph_integer_t cumsum = 0;
+        for (igraph_integer_t v = 0; v < no_of_nodes; v++) {
+            VECTOR(children_start)[v] = cumsum;
+            cumsum += VECTOR(no_of_children)[v];
+        }
+        VECTOR(children_start)[no_of_nodes] = cumsum;
+
+        /* Constructing the tree IN_k (as in the paper) and representing it
+           as a contiguously stored adjacency list. The entires of the no_of_children
```
rfulekjames, line 133: entRIEs?

src/paths/floyd_warshall.c
```diff
@@ -61,19 +61,34 @@ static igraph_error_t igraph_distances_floyd_warshall_tree(
...
+    /* This is the "Tree" algorithm of Brodnik et al.
+     * A difference from the paper is that instead of using the OUT_k tree of shortest
+     * paths _starting_ in k, we use the IN_k tree of shortest paths _ending_ in k.
+     * This makes it easier to iterate through matrices in column-major order,
+     * i.e. storage order, thus increasing performance. */
+
     igraph_integer_t no_of_nodes = igraph_vcount(graph);
 
-    /* predecessors[u][v] is the second but last vertex on the shortest path from u to v */
+    /* predecessors[v][u] is the second vertex on the shortest path from v to u,
```
rfulekjames, line 72: to me calling it `succcessors` sounds more appropriate after this change

rfulekjames:
@szhorvat 

A nice optimization. I haven't spotted any problems with the implementation.

PR #2295, opened 2023-01-20
feat: igraph_matrix_subset_vertices()

szhorvat:
@ntamas Have a look before I continue.

I find it a bit unpleasant that now `igraph_matrix.h` needs to include all headers that defined `igraph_t` and `igraph_vs_t`.

Once we're happy with the general approach, I'll add another function for vectors, and will start using this in Floyd-Warshall.

### DIFF
--- a/src/internal/utils.c
+++ b/src/internal/utils.c
@@ -1,0 +1,95 @@
+/*
+   IGraph library.
+   Copyright (C) 2023  The igraph development team <igraph@igraph.org>
+
+   This program is free software; you can redistribute it and/or modify
+   it under the terms of the GNU General Public License as published by
+   the Free Software Foundation; either version 2 of the License, or
+   (at your option) any later version.
+
+   This program is distributed in the hope that it will be useful,
+   but WITHOUT ANY WARRANTY; without even the implied warranty of
+   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+   GNU General Public License for more details.
+
+   You should have received a copy of the GNU General Public License
+   along with this program.  If not, see <https://www.gnu.org/licenses/>.
+*/
+
+#include "igraph_interface.h"
+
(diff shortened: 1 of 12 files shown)


ntamas:
Well, I don't like the new includes either; could we find another place for this function?

We could make it internal and then we'd only need an internal header, nicely tucked away in `src/paths`. We could also put the implementation of the function there in a `utils.c` file or something like that, and then we could leave `igraph_matrix.h` alone. OTOH it might be more complicated in the end due to the mess with the `.pmt` files.

Ultimately, I think I'd leave it up to you. I'm happy with the PR otherwise, so feel free to merge if you decide that it's not worth the fuss.

ntamas:
Hmmm, actually, don't we need to reset the inner iterator at the end of each iteration of the outer for loop?

szhorvat:
We can make it internal and implement it only for `igraph_matrix_t`, without making it generic. Does that sound good? I grepped the headers for `igraph_matrix_int_t`, and it seems there are no functions that produce an integer matrix where rows/columns correspond to vertices.

ntamas:
Yes, that sounds good to me!

szhorvat:
@ntamas This is now ready, but I did it in a rush, so please take a careful look.

You can use "Rebase and merge"—please do not squash the commits.

src/internal/utils.c
```diff
@@ -53,7 +53,7 @@ igraph_error_t igraph_i_matrix_subset_vertices(
 
     igraph_integer_t no_of_nodes = igraph_vcount(graph);
     igraph_integer_t ncol = igraph_matrix_ncol(m);
-    igraph_integer_t nrow = igraph_matrix_ncol(m);
+    igraph_integer_t nrow = igraph_matrix_nrow(m);
```
szhorvat, line 56: 🤦 sorry

ntamas:
Thanks a lot!

Issue #2306, opened 2023-02-11
Weighted clique results are sometimes wrong when using identical weights


szhorvat:
**Describe the bug**

Weighted clique results are sometimes wrong when using identical weights.

**To reproduce**

I added tests that demonstrate this in b99ef43eefcec7a37fc2fac31e80a28fa6bd935b

**Version information**

Tested in 0.10.4.

Bug is not present in 0.9, this is a regression in 0.10.

szhorvat:
@ntamas Would you have time to take a look at this?

ntamas:
I think we aren't doing much in our Cliquer wrapper apart from converting the vertex weight vector to Cliquer's format, right? So if it's a bug, it's either a bug in the stock Cliquer, or something went wrong during the integration of Cliquer's source into igraph.

szhorvat:
The issue is not present in 0.9, only in 0.10. I tested this using IGraph/M.

ntamas:
Inside `clique_max_weight()`, `clique_find_single()` returns `NULL` for `&s` when invoked with the P_2 graph with (5, 5) as the weights. I need to dig further into `clique_find_single()` to see what's happening there.

ntamas:
Okay, so, the allocation / deallocation structure of Cliquer is weird and I don't really understand the purpose of the `ENTRANCE_SAVE()` and `ENTRANCE_RESTORE()` calls, but here's what I've found:

* `clique_max_weight()` first calls `clique_find_single()`
* `clique_find_single()` eventually calls `clique_unweighted_find_single()` when it realizes that all the weights are the same
* `clique_unweighted_find_single()` correctly finds the largest clique in `current_clique`, but then, upon exit, it calls `ENTRANCE_RESTORE()`, which restores `current_clique` to an empty set, and that's why we are seeing that outside.

PR #2308, opened 2023-02-12
sets benchmark 

professorcode1:
Hey. 
I need your help. Please review my code and let me know what I am doing wrong. Its causing a segfault and I cannot figure out why.


professorcode1:
@ntamas  please take a look

szhorvat:
Can you please run it with AddressSanitizer enabled, and if you are still not sure what's wrong, post the AddressSanitizer output here?

Run `ccmake .` in the build directory to get the CMake GUI, then make these adjustments:

`CMAKE_BUILD_TYPE` should be `Debug`

`USE_SANITIZER` should be set to `Address`

Hit `c`, then `g`, then build and run the benchmark.

professorcode1:
ok thanks. I will try that

professorcode1:
@szhorvat please let me know how to add rule for a benchmark. Cause ` cmake --build . --target benchmark_igraph_set` is not working.

GroteGnoom:
Did you add it to tests/CMakeLists.txt? If you search for `add_benchmarks` there, you'll see a list of benchmarks.

ntamas:
@professorcode1 See the comment from @GroteGnoom ; you need to add the benchmark explicitly to `tests/CMakeLists.txt`, otherwise it won't compile and won't run. That's one thing. The other thing is that there's a failure now in the `igraph_weighted_cliques()` test case, but that's okay; the bug is tracked in https://github.com/igraph/igraph/issues/2306 and the test case is there to reproduce the bug. So it will keep on failing until we fix #2306, which I can start doing next week -- but it's completely unrelated to your code.

tests/benchmarks/igraph_set.c
```diff
@@ -0,0 +1,64 @@
...
+    igraph_integer_t induced_subgraph_vertice_count, igraph_integer_t benchmark_count,
+    const char* bench_message
+) {
+    igraph_vector_int_t edges;
+    printf("bench called\n");
+
+    ;
+    BENCH(bench_message,
+          REPEAT(
+                igraph_induced_subgraph_edges(graph, igraph_vss_range(0, induced_subgraph_vertice_count),&edges),
```
szhorvat, line 19: You are using `edges` without initializing it first.

professorcode1: Can't believe I still made that mistake. Thank you

professorcode1:
@ntamas @GroteGnoom Thank you its working now.

ntamas:
Thanks, looks good now! I'll merge this now and we can add other graph types to the benchmark later if needed.

Issue #2287, opened 2022-12-30
Sets should be re written using Red Black Trees. 


professorcode1:
Sets are currently implemented as an array of integers. When a new element is inserted binary search is used to find its position and the entire array is shifted. This a sub-optimal performance vice.  Also the state is just an index in the array, this means if an element is added in the set while use are iterating before it then the element will be read twice. This is a bug, to replicate it
```
    igraph_set_t set;
    IGRAPH_CHECK(igraph_set_init(&set, 0));
    IGRAPH_FINALLY(igraph_set_destroy, &set);

    igraph_set_add(&set,1);
    igraph_set_add(&set,2);
    igraph_set_add(&set,3);
    igraph_set_add(&set,4);
    igraph_set_add(&set,5);

    igraph_integer_t state, element;
    for(state=0;igraph_set_iterate(&set, &state,&element) ;){
        if(element==3){
            igraph_set_add(&set, 0);
        }
        printf("%ld\n", element);
    }
``` 
3 will get printed twice. 
There is also no functionality to remove elements from the set.  And if there was one added now it would also become a bug because of how the iterate function works. Hence the set class needs to be re written properly. 

To allow for all POD I intend to use the methods and API described here [here](https://github.com/igraph/igraph/issues/2285). 

**Use Case**
Complex graph algorithms like ant-colony coloring need sets to be implemented, and are a good tool to have in general. To keep the current code that use the sets working, a wrapper can be make around the RBT to make a set that only store integers.

**References**
CLSR has a chapter on how to implement red and black tress. 

Please assign the issue to me.

ntamas:
I was wondering whether we should simply make use of many of the existing generic header-only C implementations of sets instead of rolling our own. E.g., here is this one:

https://github.com/rurban/ctl/blob/master/docs/set.md

professorcode1:
Please let me know once it is merged into master on the ant colony thread.

ntamas:
oh sorry, I meant to assign this to you; it looks like I was quite tired last night when I did this. Feel free to decide how to go ahead with the set implementation, I was only suggesting that maybe it's not worth reinventing the wheel if there are suitable solutions out there.

szhorvat:
> To allow for all POD I intend to use the methods and API described [here](https://github.com/igraph/igraph/issues/2285).
> EDIT: by that I mean I intend to create an API similar to this one and use the same void pointer method to emulate comparison overloading.

[...]

PR #2313, opened 2023-02-17
RBTree for set

professorcode1:
I have written the set api to use Red and Black tree, but the degree sequence game internally relies on the logic that a set iterator will be an int. It an easy fix. Another 2 integers to keep track of number of elements iterated over. I'll update it soon


szhorvat:
Unfortunately I won't have time to look at this for a while.

Whoever does the review, please check that there is no performance degradation in the `IGRAPH_DEGSEQ_CONFIGURATION_SIMPLE` method of `igraph_degree_sequence_game()`, which relies heavily on sets.

ntamas:
I'll take a look at it early next week.

ntamas:
FYI, this is not forgotten, I've started writing the benchmark, hopefully it will be ready today.

ntamas:
Benchmark results are in; it looks like the proposal in this PR did not improve performance, although there's probably still room for optimization.

*Old version*:

igraph_set_benchmark_old

*Proposal in this PR*:

image

Compiled in release mode, with LTO. I don't have time to debug this further, but it's probably worth looking at the code with a profiler.

ntamas:
One caveat with the benchmark above is that the implementation of `igraph_degree_sequence_game()` seems to be _very_ sensitive on whether the algorithm gets "lucky" and finds an appropriate wiring of the graph with no multiple edges. The timings on the same graph, with the same algorithm, vary wildly depending on the random seed being used. There's a slim chance that changing the set implementation to RB-trees also changes the generated random numbers in the algorithm and the differences that we see are not due to the performance of the RB-trees but due to the fact that the algorithm spends more time trying to find an adequate wiring of the graph. We should probably count how many times `igraph_set_contains()` (or some other crucial set function) is called in the benchmarks; if there's a huge difference, this could mean that we are not testing the same thing in the two benchmarks even if the random seed and everything else is the same. Does the RB tree uses randomness anywhere?

src/core/set.c
```diff
@@ -198,58 +315,212 @@ igraph_integer_t igraph_set_size(const igraph_set_t* set) {
...
+    if (igraph_set_contains(set, e)) {
+        return IGRAPH_SUCCESS;
     }
 
-    if (right >= 0 && SET(*set)[left] != e && SET(*set)[right] == e) {
-        left = right;
+    struct Node* newNode = IGRAPH_CALLOC(1, struct Node);
+    if(newNode == NULL){
+        IGRAPH_CHECK_OOM(newNode, "Cannot reserve space for the new set element.");
     }
```
szhorvat, line 325: `IGRAPH_CHECK_OOM` already does the `NULL`-check, so the `if` is unnecessary.

src/core/set.c
```diff
@@ -198,58 +315,212 @@ igraph_integer_t igraph_set_size(const igraph_set_t* set) {
...
-            left = middle;
-            break;
-        }
+    if (igraph_set_contains(set, e)) {
+        return IGRAPH_SUCCESS;
     }
 
-    if (right >= 0 && SET(*set)[left] != e && SET(*set)[right] == e) {
-        left = right;
+    struct Node* newNode = IGRAPH_CALLOC(1, struct Node);
```
szhorvat, line 322: If every `set_add()` involves a memory allocation, that is going be problematic for performance, as `malloc()` is expensive.

professorcode1:
@ntamas  I mentioned 3 alternatives to dropping the reserve functionality [here](https://github.com/igraph/igraph/issues/2287). I think there are must be 2 things causing the slowdown. First and foremost would be the constant cash misses. Second as @szhorvat  mentioned must be the expensive calloc call. I'll try the  alternatives and see how it goes.

ntamas:
Extended the `igraph_degree_sequence_game()` benchmark a bit so we have more data points to work with.

ntamas:
I would probably try to go along the line of having a large pre-allocated pool of tree nodes such that only a part of the array is actually used by the tree. When the pool is full and all nodes are used up, you can realloc a pool twice that large and go on. The tree structure itself would only contain indexes into the node pool. Does that sound reasonable?

Anyhow, before changing anything, try to profile the current benchmarks and see which set-related function the code spends the most time in. If it is `igraph_set_contains()`, there's no point in optimizing the allocations yet.

ntamas:
Another, generic comment to the entire PR: please keep the return value of `igraph_set_init()` as an `igraph_error_t` and keep the `IGRAPH_CHECK(...)` calls around `igraph_set_init()`. Even if the function cannot fail _now_, it is worth keeping the return value as `igraph_error_t` because then we can change the implementation in a way that can potentially fail in the future, without breaking the API. If we commit ourselves to a `void` return code now, it means that any change in the internals of `igraph_set_init()` that can potentially fail will be an API-breaking change as we would need to change the return type of `igraph_set_init()`.

professorcode1:
Just letting everyone know that I haven't given up on this. I got a job recently and I need to write a report to participate in a hackathon by this Wednesday. I will profile the code and upload the results here probably by Sunday, latest by next Wednesday. Then continue on improving the timing.

professorcode1:
Hi @szhorvat  @ntamas 
I am really sorry it took so long to do this. I am still adjusting to the new job life. 


I ran the analysis and [here](https://github.com/igraph/igraph/pull/2313/commits/963439c4d71debd270fc8e727722e06da4b8cf84) are the results.
There are screenshots of the times and flame-graphs for the profiling. 

Here is the time taken by the igraph set implementation 

>1 Degree sequence of undirected k-regular graph, N=1000, k=7, CONFIGURATION_SIMPLE 
> 1.3s 1.2s 0s
>2 Degree sequence of undirected BA graph, N=1000, m=1, CONFIGURATION_SIMPLE 
> 0.789s 0.728s 0s
>3 Degree sequence of undirected Erdos-Renyi graph, N=150, m=u450, CONFIGURATION_SIMPLE 
> 0.804s 0.742s 0s
>4 Degree sequence of GRG graph, N=10000, r=0.013, CONFIGURATION_SIMPLE 
> 0.156s 0.144s 0s
>5 Degree sequence of directed k-regular graph, N=1000, k=5, CONFIGURATION_SIMPLE 
> 0.865s 0.798s 0s
>6 Degree sequence of directed BA graph, N=500, m=2, CONFIGURATION_SIMPLE 
>2.22s  2.05s 0s
>7 Degree sequence of directed Erdos—Renyi graph, N=15000, m=u45000, CONFIGURATION_SIMPLE 
>0.653s 0.603s 0s


Here is the time taken by the simple rb tree set implementation. The commit "set constructor error t"

>1.  2.400s 2.400s 0s
>2.  1.680s 1.680s 0s
>3.  1.460s 1.460s 0s
>4.  0.294s 0.294s 0s
>5.  1.610s 1.610s 0s
>6.  4.610s 4.610s 0s
>7.  1.290s 1.290s 0s

Next is the pool implementation [here](https://github.com/igraph/igraph/pull/2313/commits/5703961747921539badc1db3c0984fad9b39ad74).
It is roughly 3 times slower than the simple rb tree implementation.

Lastly there is the multi pool implementation which is 5 times slower than the pool implementation. The multi pool implementation works by making a new pool rather than reallocating the old pool and updating the addresses. I don't understand why it is spending 60% of the time in reserve function, so its possible there's some bug there that I am not seeing(all the tests are passing though).

Here's something that instantly jumped out to me. In the simple implementation most time is wasted just deallocating the tree(44.21%). So if it is somehow possible to start a task asynchronously it will exactly be only slightly slower than the current implementation. Does igraph have a standard for threading? If yes, by making the contains function iterative instead of recursive and making the deletion asynchronous, I feel it should be as fast as the current implementation. 

If not can we please just implement it alongside the current set implementation as igraph_rbtree_t? Developers can opt in to use it in case they need the iterator to behave like c++ stl iterators.

ntamas:
> Does igraph have a standard for threading? 

No; igraph is inherently single-threaded at the moment and there are no immediate plans to add support for threading in general. Also note that many _host_ languages that igraph embeds itself in (Python, Mathematica and R being the most prominent) are also single-threaded so it's not clear yet how much we could gain by embracing the additional complexity that comes with threads. So far we've been using OpenMP in places in the code where we could immediately benefit from parallelizing a for loop, but your use-case does not seem to fit the bill.

> If not can we please just implement it alongside the current set implementation as igraph_rbtree_t?

I don't have a strong opinion here but I'm slightly leaning towards saying no. igraph's goal is not to provide a full general-purposes data structure library for C. If there is a place in igraph's current code where we could benefit from using a red-black tree instead of sets, I'm okay with adding such a data type, but adding a red-black tree just for the sake of having it means more maintenance burden on our end for little gain. I'm pretty sure that if someone badly needs a red-black tree in C, there are plenty of open-source libraries that already provide that and it's unlikely we can do any better than those without putting significant effort in. (Opinions @szhorvat ?).

professorcode1:
@ntamas  I have implemented the indexed pool igraph as you suggested. It is exactly as slow as the regular pool implementation with an identical flame graph. I guess each calloc call takes less time when you ask for very few bytes since the OS doesn't have to move memory a bunch of memory.

> If there is a place in igraph's current code where we could benefit from using a red-black tree instead of sets, I'm okay with adding such a data type, but adding a red-black tree just for the sake of having it means more maintenance burden on our end for little gain. 

But I do need the RB tree based sets to implement my ant colony graph coloring algorithm. The reason I cannot use the igraph set is because its iterator doesn't satisfy my use case: if an element is remove from the set the iterator will point to the wrong element, and I need to delete elements from the set while iterating over it.

It seems we have 2 options, either implementing RB Tree as an igraph export or to use it inside the ant colony graph coloring algorithm file only. Please let me know which one you would like me to proceed with.

ntamas:
> But I do need the RB tree based sets to implement my ant colony graph coloring algorithm.

I see three possible ways forward:

1. Just start using `std::set` in your own implementation of the ant colony graph coloring algorithm, and we can accept that this will be a small part of the igraph library that uses C++. As long as you expose a C API, I think it's okay to use C++ internally. There are already other parts in the igraph library that use C++, but they expose a C API so they can be used from pure C programs.

2. If you can make your `igraph_rbtree_t` ipmlementation more performant than an `std::set`, you can implement your own in the context of the ant colony optimization PR and we can merge that as is. `igraph_rbtree_t` might or might not be exposed as a public API, we need to decide that together with the dev team.

3. If you do not want to use `std::set` but you are not confident that you can make your `igraph_rbtree_t` at least as performant as an `std::set`, you can probably make use of the [uthash](https://troydhanson.github.io/uthash/userguide.html) library. This is a hash table so it's suitable to be used as a set as long as the items are hashable, it's pure C, it's header-only, and it supports deletion-safe iteration, so it probably fits your use-case.

professorcode1:
I have tried 3 other method to make the set faster(5 total now). So far the simplest implementation is still the fastest. 

I can't make much sense of the flame graphs. I used them, hypothesized what might be causing the problem, tried to resolve it with another method and failed. 

Right now my current hypothesis on why its slow is that the degree sequence game test make a lot (at least 100 if not thousand) individual sets of size 1 and 2 and then discards them, which is why the more complex tree implementation is slower.  Since we are basically comparing how long it takes to allocate the 24 bytes for the current implementation vs the 32 bytes of the tree implementation times 1000. 

I will implement the std::set wrapper and bench that. Just writing this comment to update you on the progress.

professorcode1:
Hey

I haven't been able to figure out how to integrate C++ into the project. Can you please point me in the right direction on how I can go about doing it. 

Thanks.

PR #2322, opened 2023-03-10
Improved spinglass documentation

vtraag:
This PR improves some of the documentation for the spinglass function. It now also includes a check for negative weights when selecting the original implementation.

### DIFF
--- a/src/community/spinglass/clustertool.cpp
+++ b/src/community/spinglass/clustertool.cpp
@@ -108,7 +108,7 @@ static igraph_error_t igraph_i_community_spinglass_negative(
  * \param weights The vector giving the edge weights, it may be \c NULL,
  *     in which case all edges are weighted equally. The edge weights
  *     must be positive unless using the \c IGRAPH_SPINCOMM_IMP_NEG
- *     implementation. This condition is not verified by the function.
+ *     implementation.
  * \param modularity Pointer to a real number, if not \c NULL then the
  *     modularity score of the solution will be stored here. This is the
  *     gereralized modularity that simplifies to the one defined in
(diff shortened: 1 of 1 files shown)


src/community/spinglass/clustertool.cpp
```diff
@@ -149,29 +146,20 @@ static igraph_error_t igraph_i_community_spinglass_negative(
...
- *     algorithm. This defines the weight of the missing and existing
- *     links in the quality function for the clustering. The default
- *     value in the original code was 1.0, which is equal weight to
- *     missing and existing edges. Smaller values make the existing
- *     links contibute more to the energy function which is minimized
- *     in the algorithm. Bigger values make the missing links more
- *     important. (If my understanding is correct.)
+ * \param gamma Real number. The gamma parameter of the algorithm,
+ *     acting as a resolution parameter. Smaller values typically lead to
+ *     larger clusters, larger values typically lead to smaller clusters.
```
szhorvat, line 151: Is this equivalent to `resolution` in Leiden, and if yes, would it make sense to rename it accordingly? Parameter renamings are not API-breaking in C.

src/community/spinglass/clustertool.cpp
```diff
@@ -256,6 +244,9 @@ static igraph_error_t igraph_i_community_spinglass_orig(
             IGRAPH_ERROR("Invalid weight vector length", IGRAPH_EINVAL);
         }
         use_weights = 1;
+        if (igraph_vector_min(weights) < 0) {
+            IGRAPH_ERROR("Cannot work with negative weights in this implementation.", IGRAPH_EINVAL);
```
szhorvat, line 248: Let's be a bit more informative. Users of high-level interface may not realize what "this implementation" refers to, as they are likely not selecting the implementation explicitly.

```suggestion
            IGRAPH_ERROR("Weights must not be negative when using the original implementation of spinglass communities. Select the implementation meant for negative weights.", IGRAPH_EINVAL);
```

src/community/spinglass/clustertool.cpp
ntamas: Let's do this even if it's not exactly equivalent. `resolution` is better in explaining what the parameter does than `gamma`.

ntamas: Hmm, actually, looking at this again: there's a `gamma_minus` parameter in the version of the function that also allows negative weights. Having `resolution_minus` would be awkward and it's not that clear any more once you get to that function that it controls the resolution directly. I propose keeping the name as is.

PR #2333, opened 2023-04-08
Voronoi communities

szhorvat:
This PR adds a new community detection method based on Voronoi partitionings of vertices. The function is marked as experimental.


ntamas:
@szhorvat is this still a draft or can we start reviewing it?

szhorvat:
Ready for review.

src/community/voronoi.c
```diff
@@ -0,0 +1,638 @@
...
+    igraph_real_t denom = f3 * (x1 - x2)   + f1 * (x2 - x3)   + f2 * (x3 - x1);
+    return 0.5 * num / denom;
+}
+
+/**
+ * Simple Brent's method optimizer. It must be called with x2 > x1.
+ * The optimal argument is the last one for which f() is invoked.
+ * f() is expected to record this in 'extra'.
+ */
+static igraph_error_t brent_opt(optfun_t *f, igraph_real_t x1, igraph_real_t x2, void *extra) {
```
ntamas, line 316: I wonder whether we should refactor this to a file of its own (not exposed as part of the public API of course). Can it happen in the future that we find other functions for which this optimizer would be useful, or is this highly specific to the Voronoi communities algorithm?

src/community/voronoi.c
```diff
@@ -0,0 +1,638 @@
...
+            x1 = x2; f1 = f2;
+            x3 = 0.5 * (x1 + x2);
+            IGRAPH_CHECK(f(x3, &f3, extra));
+
+            if (f3 >= f2) break;
+        }
+    }
+
+    /* Limit iterations to 20 */
+    for (int i=0; i < 20; ++i) {
```
ntamas, line 358: Feels like a magic constant to me. Should we make this a parameter maybe?

src/community/voronoi.c
```diff
@@ -0,0 +1,638 @@
...
+
+
+/* Objective function used with brent_opt(), it computes the modularity for a given radius. */
+static igraph_error_t get_modularity(igraph_real_t r, igraph_real_t *modularity, void *extra) {
+    get_modularity_work_t *gm = extra;
+
+    IGRAPH_CHECK(choose_generators(gm->graph, gm->generators, NULL,
+                                   gm->local_dens, gm->lengths, gm->mode,
+                                   r));
+    igraph_voronoi(gm->graph, gm->membership, NULL,
```
ntamas, line 440: Why don't we need an `IGRAPH_CHECK(...)` around this? (and later in the call to `igraph_modularity()`)?

ntamas (commented):
Some general questions in comments; no explicit approval yet but it looks good to me otherwise.

src/community/voronoi.c
```diff
@@ -0,0 +1,638 @@
...
+#include "igraph_iterators.h"
+#include "igraph_nongraph.h"
+#include "igraph_paths.h"
+#include "igraph_structural.h"
+#include "igraph_transitivity.h"
+
+#include "core/indheap.h"
+
+/**
+ * \function igraph_local_relative_density
```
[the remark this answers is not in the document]
szhorvat: It was meant to be exposed, but I changed my mind. It's a very specific thing, and I would rather not expose it. I'll remove the `\function` for now.

src/community/voronoi.c
szhorvat: Unfortunately I needed to add things that are specific to this application. The function is not smooth, and may not have an optimum within the interval. There are imperfect mechanisms that attempt to deal with this.

Thus I wouldn't factor it out at this moment.

src/community/voronoi.c
szhorvat: Oops ... will fix it.

ntamas (approved):
LGTM now.

ntamas:
Feel free to merge if you are ready (I don't know whether you still want to modify this branch or not).

szhorvat:
I'll merge at a later time, travelling now. You're right about that hard coded iteration count being ugly, I want to have another look.

PR #2337, opened 2023-04-23
remove local_scan_1_sumweights

### LINKED ISSUE #2163: weighted, undirected local_scan_1_ecount is handled in 2 different functions, depending on where you start
for the weighted, undirected case, when `igraph_local_scan_k_ecount` is called it with k=1 it will handle it itself, but when `igraph_local_scan_1_ecount` is called, it's passed on to `igraph_i_local_scan_1_sumweights`. They are now both tested, since 8d84c9f3a9f7526adc46520215b67a219a008f1e, and give the same results for a basic testcase. 

Either `igraph_i_local_scan_1_sumweights` is unnecessary and can be removed, or it is somehow better (faster?) than calling `igraph_local_scan_k_ecount`, and `igraph_local_scan_k_ecount` should redirect the weighted, undirected, k=1 case to `igraph_i_local_scan_1_sumweights`.

GroteGnoom:
igraph_i_local_scan_sumweights was faster than using scan_k_ecount, but incorrect. It didn't count loops on neighbors, it counted loops on the starting vertex twice, and it counted only one of the multiple edges between two neighbors. When counting edge weights this is almost always incorrect because you need some way of combining weights, and not just pick a random one. I didn't really find a way to fix this.

Fixes #2163


szhorvat:
Can you update the changelog, and mention the user-visible aspect of this fix? I.e., mention under what circumstances the result was wrong.

ntamas:
Thanks!

PR #2338, opened 2023-04-26
Add failing test: deleting edge with attribute

GenieTim:
This is rather a bug report, or a question of what I am doing wrong I guess: if the graph has edge attributes, the `igraph_delete_edges` results in a segfault.

I might update this PR with a fix in case I can figure out what the issue is.
The current understanding is this:
In `type_indexededgelist.c`, the `igraph_delete_edges` function calls "only" `igraph_i_attribute_permute_edges`, which in turn has the comment `If they are _not_ the same, it is assumed that the new graph has no edge attributes yet`, which is not applicable here.

### DIFF
--- a/tests/unit/igraph_delete_edges.c
+++ b/tests/unit/igraph_delete_edges.c
@@ -47,6 +47,14 @@ int main(void) {
         return 7;
     }
 
+    /* no error test, deleting edges with attributes */
+    igraph_set_attribute_table(&igraph_cattribute_table);
+    igraph_cattribute_EAN_set(&g, "test", 2, 9);
+    igraph_delete_edges(&g, igraph_ess_1(1)); // CHECK_NO_ERROR
+    if (igraph_ecount(&g) != 2) {
+        return 9;
+    }
+
     igraph_es_destroy(&es);
     igraph_destroy(&g);
 


GenieTim:
In the original version of this PR, I absolutely agree, this was on me.
However, in the code that this example was based on, my mistake was a different one.
Namely, the failing test should have looked like this:

```c++
igraph_set_attribute_table(&igraph_cattribute_table);
igraph_t g;
igraph_small(&g, 4, IGRAPH_UNDIRECTED, 0,1, 1,2, 2,2, -1);
igraph_cattribute_EAN_set(&g, "test", 2, 9); // no segfault below if we don't do this here
igraph_attribute_combination_t comb;
igraph_attribute_combination_init(&comb);
igraph_delete_edges(&g, igraph_ess_1(0)); // no segfault now
igraph_simplify(&g2, true, true, &comb); // empty comb -> expect empty edge attributes afterwards
igraph_delete_edges(&g, igraph_ess_1(0)); // segfault now
```

which I think is a bit less clear/less expected. 
While I do expect that the edge attributes are empty afterwards (as is documented), I would not expect a segfault then.
I feel like there is some internal state that is not reset appropriately here, as the segfault only happens if there have been edge attributes. Despite spending time looking through the code today, I am still too far from seeing the big picture here.

PR #2341, opened 2023-04-29
add graph_power

### LINKED ISSUE #2063: Replace `connect_neighborhood` by `graph_power`
The function `igraph_connect_neighborhood()` is somewhat naively named, and behaves naively as well.

This operation is better known as "graph power", and is well-defined for a power of 1. Currently, `connect_neighborhood` issues a warning when `order` is zero or one, and returns the unchanged graph in both cases. This warning message is not productive, and has already caused inconvenience to me.

I suggest adding a function named `igraph_graph_power()`:

 - It shows no warning for an input of one or zero.
 - For an input of one it returns the same graph.
 - For an input of zero is returns a graph with no edges. This is a reasonable generalization of the concept. In the graph raised to power `k`, vertices are connected if they are at a distance `<= k` in the original graph.
 - It produces a new graph instead of modifying the old one. **Discuss!**
 - Do we then deprecate `igraph_connect_neighborhood()` in favour of `igraph_graph_power()` or just keep both with slightly different behaviour? **Discuss!**
 - If we retain `igraph_connect_neighborhood()`, there would be room for weighted generalizations, such as connecting neighbours within a given weighted distance, which is a useful operation, but it is not necessarily appropriate to keep calling it "graph power".

I'd like to have some input about the last two points.

GroteGnoom:
fixes #2063

### DIFF
--- a/interfaces/functions.yaml
+++ b/interfaces/functions.yaml
@@ -165,7 +165,7 @@ igraph_connect_neighborhood:
     PARAMS: INOUT GRAPH graph, INTEGER order=2, NEIMODE mode=ALL
 
 igraph_graph_power:
-    PARAMS: IN GRAPH graph, OUT GRAPH res, INTEGER order=2, NEIMODE mode=ALL
+    PARAMS: IN GRAPH graph, OUT GRAPH res, INTEGER order=2
 
 igraph_linegraph:
     PARAMS: GRAPH graph, OUT GRAPH linegraph
--- a/interfaces/functions.yaml
+++ b/interfaces/functions.yaml
@@ -165,7 +165,7 @@ igraph_connect_neighborhood:
     PARAMS: INOUT GRAPH graph, INTEGER order=2, NEIMODE mode=ALL
 
 igraph_graph_power:
-    PARAMS: IN GRAPH graph, OUT GRAPH res, INTEGER order=2
+    PARAMS: IN GRAPH graph, OUT GRAPH res, INTEGER order
 
 igraph_linegraph:
     PARAMS: GRAPH graph, OUT GRAPH linegraph
--- a/interfaces/functions.yaml
+++ b/interfaces/functions.yaml
@@ -164,6 +164,9 @@ igraph_extended_chordal_ring:
 igraph_connect_neighborhood:
     PARAMS: INOUT GRAPH graph, INTEGER order=2, NEIMODE mode=ALL
 
+igraph_graph_power:
+    PARAMS: IN GRAPH graph, OUT GRAPH res, INTEGER order=2, NEIMODE mode=ALL
+
 igraph_linegraph:
     PARAMS: GRAPH graph, OUT GRAPH linegraph
 
--- a/src/operators/connect_neighborhood.c
+++ b/src/operators/connect_neighborhood.c
@@ -196,7 +196,7 @@ igraph_error_t igraph_graph_power(const igraph_t *graph, igraph_t *res,
     igraph_vector_int_t neis;
 
     if (order < 0) {
-        IGRAPH_ERRORF("Order can not be negative, found %" IGRAPH_PRId ".",
+        IGRAPH_ERRORF("Order cannot be negative, found %" IGRAPH_PRId ".",
                 IGRAPH_EINVAL, order);
     }
 
--- a/src/operators/connect_neighborhood.c
+++ b/src/operators/connect_neighborhood.c
@@ -175,15 +175,11 @@ igraph_error_t igraph_connect_neighborhood(igraph_t *graph, igraph_integer_t ord
  * symmetric: if vertex A can be reached from vertex B in at
  * most \p order steps, then the opposite is also true. Only one
  * undirected (A,B) edge will be added in this case.
- * \param graph The input graph.
+  *
+  * \param graph The input graph.
  * \param res The resulting graph.
  * \param order Integer constant, it gives the distance within which
  *    the vertices will be connected to the source vertex.
- * \param mode Constant, it specifies how the neighborhood search is
- *    performed for directed graphs. If \c IGRAPH_OUT then vertices
- *    reachable from the source vertex will be connected, \c IGRAPH_IN
- *    is the opposite. If \c IGRAPH_ALL then the directed graph is
- *    considered as an undirected one.
  * \return Error code.
  *
  * Time complexity: O(|V|*d^k), |V| is the number of vertices in the
--- a/src/operators/connect_neighborhood.c
+++ b/src/operators/connect_neighborhood.c
@@ -189,8 +189,8 @@ igraph_error_t igraph_connect_neighborhood(igraph_t *graph, igraph_integer_t ord
  * Time complexity: O(|V|*d^k), |V| is the number of vertices in the
  * graph, d is the average degree and k is the \p order argument.
  */
-igraph_error_t igraph_graph_power(const igraph_t *graph, igraph_t *res, igraph_integer_t order,
-                                igraph_neimode_t mode) {
+igraph_error_t igraph_graph_power(const igraph_t *graph, igraph_t *res,
+                                  igraph_integer_t order) {
 
     igraph_integer_t no_of_nodes = igraph_vcount(graph);
     igraph_dqueue_int_t q;
@@ -215,12 +215,9 @@ igraph_error_t igraph_graph_power(const igraph_t *graph, igraph_t *res, igraph_i
     }
 
     IGRAPH_CHECK(igraph_copy(res, graph));
+    IGRAPH_CHECK(igraph_simplify(res, true, true, NULL));
     if (order == 1) {
         return IGRAPH_SUCCESS;
-    }
-
-    if (!igraph_is_directed(graph)) {
-        mode = IGRAPH_ALL;
     }
 
     IGRAPH_VECTOR_INT_INIT_FINALLY(&edges, 0);
@@ -233,7 +230,7 @@ igraph_error_t igraph_graph_power(const igraph_t *graph, igraph_t *res, igraph_i
 
     for (i = 0; i < no_of_nodes; i++) {
         added[i] = i + 1;
-        IGRAPH_CHECK(igraph_neighbors(res, &neis, i, mode));
+        IGRAPH_CHECK(igraph_neighbors(res, &neis, i, IGRAPH_OUT));
         in = igraph_vector_int_size(&neis);
         if (order > 1) {
             for (j = 0; j < in; j++) {
@@ -248,7 +245,7 @@ igraph_error_t igraph_graph_power(const igraph_t *graph, igraph_t *res, igraph_i
             igraph_integer_t actnode = igraph_dqueue_int_pop(&q);
             igraph_integer_t actdist = igraph_dqueue_int_pop(&q);
             igraph_integer_t n;
-            IGRAPH_CHECK(igraph_neighbors(res, &neis, actnode, mode));
+            IGRAPH_CHECK(igraph_neighbors(res, &neis, actnode, IGRAPH_OUT));
             n = igraph_vector_int_size(&neis);
 
             if (actdist < order - 1) {
@@ -258,14 +255,9 @@ igraph_error_t igraph_graph_power(const igraph_t *graph, igraph_t *res, igraph_i
                         added[nei] = i + 1;
                         IGRAPH_CHECK(igraph_dqueue_int_push(&q, nei));
                         IGRAPH_CHECK(igraph_dqueue_int_push(&q, actdist + 1));
-                        if (mode != IGRAPH_ALL || i < nei) {
-                            if (mode == IGRAPH_IN) {
-                                IGRAPH_CHECK(igraph_vector_int_push_back(&edges, nei));
-                                IGRAPH_CHECK(igraph_vector_int_push_back(&edges, i));
-                            } else {
-                                IGRAPH_CHECK(igraph_vector_int_push_back(&edges, i));
-                                IGRAPH_CHECK(igraph_vector_int_push_back(&edges, nei));
-                            }
+                        if (igraph_is_directed(res) || i < nei) {
+                            IGRAPH_CHECK(igraph_vector_int_push_back(&edges, i));
+                            IGRAPH_CHECK(igraph_vector_int_push_back(&edges, nei));
                         }
                     }
                 }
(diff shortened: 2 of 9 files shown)


src/operators/connect_neighborhood.c
```diff
@@ -162,3 +162,145 @@ igraph_error_t igraph_connect_neighborhood(igraph_t *graph, igraph_integer_t ord
...
+        return IGRAPH_SUCCESS;
+    }
+
+    if (!igraph_is_directed(graph)) {
+        mode = IGRAPH_ALL;
+    }
+
+    IGRAPH_VECTOR_INT_INIT_FINALLY(&edges, 0);
+    added = IGRAPH_CALLOC(no_of_nodes, igraph_integer_t);
+    if (added == 0) {
```
ntamas, line 229: I think we have the `IGRAPH_CHECK_OOM` to shorten this.

ntamas (approved):

ntamas:
Looks good to me and I'm okay with merging this. There's only one minor comment (see above) and we should not forget to update the changelog before or after the merge.

szhorvat:
I'd suggest some changes in behaviour.

We already have `connect_neighborhood()`, which is more oriented towards data analysis, where attributes are important. Let's make `graph_power()` more oriented towards graph theory.

We need to consider what to do with non-simple graphs. The concept of graph power is not normally defined for multigraphs, so this operation shouldn't normally be applied to multigraphs. With a focus on graph without multi-edges, I suggest the following:

 - Discard all edge attributes. We can keep graph and vertex attributes because it's very straightforward. Thus, start by creating a new graph with the same number of vertices, and use `IGRAPH_I_ATTRIBUTE_COPY()` to copy over graph and vertex attributes, but not edge attributes.
 - The result should never have any multi-edges, not even for `order == 1`. The assumption is that this function wouldn't normally be used with multigraphs. This can be implemented by starting with an empty graph with as many vertices as the original, and adding new edges as needed, even for `order <= 1`.
 - What do we do with simple self-loops? They fit neatly into the definition of connecting vertices at a distance of `<= order`. A vertex is at distance 0 from itself, so self-loops are effectively preserved, including with `order == 0`.

szhorvat:
I'll take a look at the code later.

GroteGnoom:
It seems like your proposal conforms to the definition of 'The $k$ th power of a simple graph $G = (V, E)$ is the graph $G^k$ whose vertex set is $V$, two distinct vertices being adjacent in $G^k$ if and only if their distance in $G$ is at most $k$' (from Bondy and Murty), where it also sounds like it's a different graph, and we shouldn't expect $G^1=G$.

szhorvat:
Note to self for when I get home:

This should be marked as experimental.

Behaviour of mode parameter, compare directions in neighborhood calculation vs direction of added edges.

Do we have a textbook reference for the directed case?

szhorvat:
Let's talk about directed graphs. I should not that I have not tried this function yet, but from the documentation it seems that the `mode` parameter works the same way as in `connect_neighborhood()`. Correct me if I'm wrong @GroteGnoom 

Let us consider a simple 3-vertex graph, `A -> B -> C`, and `order=2`.

Notice that there is no difference between `mode=OUT` and `mode=IN`. The connection `A -> C` will be added in both cases. With `mode=OUT`, `C` is in `A`'s neighbourhood. With `mode=IN`, `A` is in `C`'s neighbourhood.

What about `mode=ALL`? With this graph, it yields the same result again. But let us now consider `A -> B <- C`, or alternatively, `A <- B -> C`.  `mode=IN` and `mode=OUT` add no connections. `mode=ALL` should add two reciprocal connections: `A <-> C`. However, with `connect_neighborhood` I get only `A -> C`. This seems to be a bug in `connect_neighborhood()`.

Conclusion: If `mode` controls how neighbourhoods are computed, then there is no difference between `IN` and `OUT`. Thus we should have a boolean `directed` parameter instead of a `mode` one.

I wasn't able to find any information on directed generalizations of the graph power concept, but we can make some connections with powers of matrices. If $A$ denotes the adjacency matrix of graph $G$, then $G^2$ is related to $A^2$ (with the diagonal zeroes out). More generally, $G^k$ is related to $\sum_{i=1}^k A^i$. This definition in terms of adjacency matrices extends naturally to directed graphs and gives precisely what we have now. Alslo, notice that $(A^T)^k = (A^k)^T$, which again shows that there is no difference between `IN` and `OUT`.

Generalization:

The current `mode` argument controls how neighbourhoods are computed. But we can also think about what kinds of edges are added for a given relation.  If `B` is in `A`'s neighbourhood, we can add `A -> B` (`OUT`), `B -> A` (`IN`) or `A <-> B` (`ALL`). This allows for creating many different patterns, but I'm not sure if it's useful in practice. If we ever consider this, I would rather add it to `connect_neighborhood()` (which we made up) rather than `graph_power()` (which has an established definition, so I wouldn't go _too_ far with generalizations).

szhorvat:
Let me know if I missed something. If not, I suggest:

 - Replace the `mode` argument of `graph_power()` with a `directed` Boolean argument.
 - Should the default value of `directed` be `true` (instead of the current effectively `false`)? **Discuss!** I am in favour of using `true` by default.
 - Make sure that the bug with `ALL` is not present in `graph_power()` (I have not checked)
 - Fix the bug with `ALL` in `connect_neighborhood()`. Before doing this, check where `connect_neighborhood()` is used internally and make sure that we don't break anything.
 - Change `connect_neighborhood()` to use a `directed` argument as well. This is breaking, so we can only do it for 0.11.
 - Should `connect_neighborhood()` get a different kind of `mode` argument controlling the direction of edges added, in 0.11? **Discuss!** No strong opinion from my side.

ntamas:
>  from the documentation it seems that the mode parameter works the same way as in connect_neighborhood().

Even if it doesn't, it would make sense to ensure that the two are equivalent. (I assumed that they are, judging from the fact that @GroteGnoom even made some changes to the connect_neighborhood() tests in this PR).

> However, with connect_neighborhood I get only A -> C. This seems to be a bug in connect_neighborhood().

Could be. IMHO in the case of `mode=ALL`, the returned graph must either be undirected (unconditionally), or if we decide to keep directedness then obviously it has to have mutual edges, right? If this is not the case, then this is a bug in `connect_neighborhood()`.

> Conclusion: If mode controls how neighbourhoods are computed, then there is no difference between IN and OUT. Thus we should have a boolean directed parameter instead of a mode one.

Maybe we could reverse the edges in the directed case if mode=IN? I mean, if the graph is A -> B -> C and we are looking for the 2nd power, with mode=IN, then we could make it so that we add not the A -> C edge but the C -> A edge. Then there would be a substantial and meaningful difference between mode=IN and mode=OUT.

szhorvat:
> Maybe we could reverse the edges in the directed case if mode=IN?

This is what I discussed when talking about controlling how the neighbourhood is computed vs the direction of the inserted edges. These are two distinct features, and they can't both be controlled with a single `mode` parameter.

----

> IMHO in the case of `mode=ALL`, the returned graph must either be undirected (unconditionally), or if we decide to keep directedness then obviously it has to have mutual edges, right?

Right. And I realize that I made a mistake above. If we add mutual edges, then `A -> B <- C` turns not into `A -> B <- C, A <-> C`, but into `A <-> B <-> C, A <-> C`.  So yes, you are right, _all_ edges would turn into mutual ones. Thus it seems to make no sense to return a directed graph, as you said.

So perhaps let's do this:

 - If `directed=false`, then edge directions are ignored and the result is undirected
 - If `directed=true`, then in the graph `A -> B -> C`, `C` is reachable from `A` within `order=2` steps.  That's a directed relationship between `A` and `C`. We can add a corresponding edge `A -> C` (`OUT`), a reversed edge `C -> A` (`IN`) or mutual edges `A <-> C` (`ALL`). I'm not sure if such detailed control of the edge directions is useful though.

szhorvat:
If we do add a `mode` parameter controlling the direction of added edges, I think it should only be done for `connect_neighbors()` but not for `graph_power()`. This idea doesn't fit well with the matrix multiplication picture.

ntamas:
>  I'm not sure if such detailed control of the edge directions is useful though.

I'm not sure either, especially because we have `igraph_reverse_edges()`. The concept is a bit hard to explain and seems like a niche application for me. I'd vote for leaving the `mode` argument out completely in case of `graph_power()`, and I'm also leaning towards leaving it out for `connect_neighbors()`.  That would leave us with only two things to do: 1) remove the `mode` argument from `graph_power()` and 2) fix the bug in `connect_neighborhood()`. The latter is a separate issue so we could close and merge this PR if we remove the `mode` argument and we are convinced that the bug in `connect_neighborhood() ` is not present here.

interfaces/functions.yaml
```diff
@@ -164,6 +164,9 @@ igraph_extended_chordal_ring:
 igraph_connect_neighborhood:
     PARAMS: INOUT GRAPH graph, INTEGER order=2, NEIMODE mode=ALL
 
+igraph_graph_power:
+    PARAMS: IN GRAPH graph, OUT GRAPH res, INTEGER order=2, NEIMODE mode=ALL
```
ntamas, line 168: There's no `mode` parameter any more, right?

ntamas (approved):
LGTM, but `functions.yaml` should be fixed before merging. (+ changelog)

GroteGnoom:
I'm also not really sure if it should have a default order of 2.

ntamas:
I don't think so either.

szhorvat:
I think the `directed` argument would actually be useful, but it should not block this PR. Let's merge this without that feature.

Let's add `\experimental`. We should do this for all new functions, at least until the next major version. There is still a possibility to add the `directed` argument later. It's not a priority. People can convert to undirected before computing the graph power to obtain the same result.

interfaces/functions.yaml
```diff
@@ -164,6 +164,9 @@ igraph_extended_chordal_ring:
 igraph_connect_neighborhood:
     PARAMS: INOUT GRAPH graph, INTEGER order=2, NEIMODE mode=ALL
 
+igraph_graph_power:
+    PARAMS: IN GRAPH graph, OUT GRAPH res, INTEGER order=2
```
szhorvat, line 168:
```suggestion
    PARAMS: IN GRAPH graph, OUT GRAPH res, INTEGER order
```

just as a reminder

src/operators/connect_neighborhood.c
```diff
@@ -162,3 +161,131 @@ igraph_error_t igraph_connect_neighborhood(igraph_t *graph, igraph_integer_t ord
...
+ * undirected (A,B) edge will be added in this case.
+ * \param graph The input graph.
+ * \param res The resulting graph.
+ * \param order Integer constant, it gives the distance within which
+ *    the vertices will be connected to the source vertex.
+ * \param mode Constant, it specifies how the neighborhood search is
+ *    performed for directed graphs. If \c IGRAPH_OUT then vertices
+ *    reachable from the source vertex will be connected, \c IGRAPH_IN
+ *    is the opposite. If \c IGRAPH_ALL then the directed graph is
+ *    considered as an undirected one.
```
szhorvat, line 186:
```suggestion
```

src/operators/connect_neighborhood.c
```diff
@@ -162,3 +161,131 @@ igraph_error_t igraph_connect_neighborhood(igraph_t *graph, igraph_integer_t ord
...
+
+    igraph_integer_t no_of_nodes = igraph_vcount(graph);
+    igraph_dqueue_int_t q;
+    igraph_vector_int_t edges;
+    igraph_integer_t i, j, in;
+    igraph_integer_t *added;
+    igraph_vector_int_t neis;
+
+    if (order < 0) {
+        IGRAPH_ERRORF("Order can not be negative, found %" IGRAPH_PRId ".",
```
szhorvat, line 203:
```suggestion
        IGRAPH_ERRORF("Order cannot be negative, found %" IGRAPH_PRId ".",
```

szhorvat:
Instead of copying the graph and deleting edges or simplifying, can we create a new empty graph, then copy over graph/vertex (but not edge) attributes? This is cleaner and more efficient. It avoids several attribute-related operations. For example, it avoids (1) edge attribute copying (2) edge attribute combination during simplification. It also avoids edge deletion.

Then we can add new edges to this empty graph.

src/operators/connect_neighborhood.c
```diff
@@ -162,3 +161,131 @@ igraph_error_t igraph_connect_neighborhood(igraph_t *graph, igraph_integer_t ord
...
+ * This function adds new edges to the input graph. Each vertex is connected
+ * to all vertices reachable by at most \p order steps from it
+ * (unless a connection already existed).  In other words, the \p order power of
+ * the graph is computed.
+ *
+ * </para><para> For undirected graphs reachability is always
+ * symmetric: if vertex A can be reached from vertex B in at
+ * most \p order steps, then the opposite is also true. Only one
+ * undirected (A,B) edge will be added in this case.
+ * \param graph The input graph.
```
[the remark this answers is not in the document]
GroteGnoom: I didn't look correctly, and I think this was a mistake, so I'll change it back :)

szhorvat:
Let me know when this is considered ready.

GroteGnoom:
It's ready.

ntamas:
This one still has my approval after going through the recent commits :)

src/operators/connect_neighborhood.c
```diff
@@ -162,3 +163,128 @@ igraph_error_t igraph_connect_neighborhood(igraph_t *graph, igraph_integer_t ord
...
+    if (order == 0) {
+        IGRAPH_CHECK(igraph_empty(res, igraph_vcount(graph), igraph_is_directed(graph)));
+        if (igraph_has_attribute_table()) {
+            IGRAPH_I_ATTRIBUTE_DESTROY(res);
+            IGRAPH_I_ATTRIBUTE_COPY(res, graph, 1, 1, 0);
+        }
+        return IGRAPH_SUCCESS;
+    }
+
+    IGRAPH_CHECK(igraph_copy(res, graph));
```
GroteGnoom, line 216: I forgot to fix this part (only copy the necessary parts, instead of copying and removing), so I'm not actually finished yet.

GroteGnoom:
What would be the proper way of copying over edges? I thought about using an inclist, but that doesn't have multi-edge control.

ntamas:
How about simply using `igraph_adjlist_t`? It allows you to filter loop and multiple edges upfront (which is what you'd do anyway). Sure, we would not be preserving the order of edges, not even for `order == 1`, but we never guaranteed that anyway.

GroteGnoom:
The copying is fixed, it's ready again.

ntamas:
Still approved, waiting for @szhorvat 's blessings.

szhorvat:
Sorry about that, I'll get back to this in a couple of days.

Issue #2300, opened 2023-01-29
Performance impact of error checking


GroteGnoom:
Because of the discussion that started here:

https://github.com/igraph/igraph/pull/2240#issuecomment-1399479647

I feel like we (or at least I) haven't been too careful about the effects of error checking on performance. In some situations, like finding small paths in a huge graph, almost all time could be spent on error checking. Should we change this?

szhorvat:
In algorithms where each weight value is accessed at most once, such as A*, the weights could be checked at access time instead of at the beginning of the function.

For A* and Dijkstra, this technically means that the result won't be valid in some cases. It is possible that there is a negative weight edge somewhere further in the graph and the actual _shortest_ path passes through it. But of course these algorithms wouldn't be able to find the shortest path with negative weights anyway—this could only throw an error. I think that for A* this is a very reasonable trade off, and I suggest you implement it.

For Dijkstra, the typical use case is one-to-all shortest paths, so it's less clear if this optimisation is worth it. Some benchmarking would be useful. However, A* reduces to Dijkstra with the null heuristic anyway, so users would still have easy access to the faster version even if we don't implement it for Dijkstra.

If you think of other algorithms than shortest paths where this might be relevant, please point them out. We're looking for situations where typically not all weights would be accessed, yet we naively check all weights at the beginning of the function.

For heap initialisation, have you done any benchmarks? I expect this is less of a problem as it's done with memset() under the hood, which is very fast, even if technically it forces an $O(V)$ complexity. We have multiple kinds of "indexed heap" data structures which store both a value and an associated index (typically vertex index). What `2wheap` provides is $O(1)$ lookup based on index, which we use when modifying values. I think the others have $O(n)$ lookup, so it's not clear how we can improve here.

szhorvat:
Thinking about which other functions may be affected:

With Dijkstra, it is possible to specify more than one source node. If we do, edges/weights would be accessed more than one time. Thus checking weights at access time is not reasonable, at least not for all inputs (not when there are multiple sources). This also applies to `closeness_cutoff()` and `betweenness_cutoff()`

Grepping the sources for `igraph_vector_min(weights)` gives:

 - Layout functions: Not relevant since all weights are accessed under all conditions.
 - Community detection: Presumably all community detection functions access the weights of all edges anyway, so not relevant.
[...]

PR #2342, opened 2023-04-30
add astar benchmark

GroteGnoom:
part of #2300

```
|> Benchmark file: /home/daniel/Documents/igraph/tests/benchmarks/error_checking.c
| singleton graph                                                                      0s      0s      0s
| graph with 50000000 nodes, no edges                                              0.061s   0.01s  0.052s
| vector init of 50000000                                                              0s      0s      0s
| vector fill of 50000000 INFINITY                                                 0.049s  0.006s  0.043s
| vector destroy                                                                   0.001s      0s      0s
| vector fill of 50000000 1s                                                       0.048s  0.006s  0.042s
| 50000000 vertices ring graph astar from 0 to 0, 1x                               0.092s  0.051s  0.041s
| 50000000 vertices ring graph astar from 0 to 50000000 -1, 1x                      4.01s   3.14s  0.865s
| 10000 vertices full graph astar from 0 to 0, 1x                                  0.031s  0.031s      0s
| vector_min on 50000000 numbers 1x                                                0.031s  0.031s      0s
| 2wheap init of 50000000, 1x                                                          0s      0s      0s
| IGRAPH_CALLOC of 50000000 igraph_integer_t, 1x                                       0s      0s      0s
| 2wheap destroy                                                                       0s      0s      0s
```

Added some benchmarks for things that almost take no time, because they did take a significant amount of time with a debug build with sanitizers on.
So for the case where you path length is negligible:
just checking weights (vector_min) takes up almost all the time for the full graph.
error checking + filling the dists vector (which scales with no_of_nodes) takes about 8/9 of the time for the ring graph. I find it difficult to find where the rest of the time goes.
I've also put in the longest path scenario, then the overhead is less than 1/40.

### DIFF
--- a/src/paths/astar.c
+++ b/src/paths/astar.c
@@ -123,7 +123,7 @@ igraph_error_t igraph_get_shortest_path_astar(const igraph_t *graph,
         heuristic = null_heuristic;
     }
 
-    if (0 && weights) { /* If there are no weights, they are treated as 1. */
+    if (weights) { /* If there are no weights, they are treated as 1. */
         if (igraph_vector_size(weights) != no_of_edges) {
             IGRAPH_ERRORF("Weight vector length (%" IGRAPH_PRId ") does not match number of edges (%" IGRAPH_PRId ").",
                           IGRAPH_EINVAL, igraph_vector_size(weights), no_of_edges);
--- a/src/paths/astar.c
+++ b/src/paths/astar.c
@@ -123,7 +123,7 @@ igraph_error_t igraph_get_shortest_path_astar(const igraph_t *graph,
         heuristic = null_heuristic;
     }
 
-    if (weights) { /* If there are no weights, they are treated as 1. */
+    if (0 && weights) { /* If there are no weights, they are treated as 1. */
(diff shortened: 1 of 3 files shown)


szhorvat:
Even though this would be difficult for us to implement (without moving to C++), I wanted to mention that some systems store metadata for vectors that indicate the vector contents. For example, I _think_ recent R versions keep track of whether a vector contains NA / NaN.  Mathematica appears to do something similar (although strictly speaking it doesn't allow NaNs in vectors).  This is much like the cache we use in `igraph_t`.

If we did have this feature, it could speed up check. We could even set this cache when transferring vectors from R or Mathematica.

But in plain C, there is no good way that I can think of to catch vector modification operations and clear this cache. So I don't think we can do it now. C++ would be different ...

szhorvat:
In some cases, the input checks can be moved into the core algorithm. This may be an advantage (when the algorithm accesses only a small fraction of edges and their weights) or a disadvantage (which the algorithm accesses the same edge / weight many times).

PR #2352, opened 2023-05-28
Direct betweenness

### LINKED ISSUE #1796: Create a direct implementation of betweenness for testing
Write a direct implementation of betweenness for testing purposes, to verify the main implementation. This will be based on direct enumeration of all shortest paths between vertex pairs.

See `igraph_get_all_shortest_paths` and `igraph_get_all_shortest_paths_dijkstra`.

GroteGnoom:
Fixes #1796

Making cutoffs work seems like a lot of work, because the only shortest path functions with cutoffs are `igraph_distances_cutoff`, `igraph_distances_dijkstra_cutoff`, which don't actually give me the paths.

subset betweenness should be doable.

### DIFF
--- a/src/paths/all_shortest_paths.c
+++ b/src/paths/all_shortest_paths.c
@@ -35,7 +35,7 @@
  * \function igraph_get_all_shortest_paths
  * \brief All shortest paths (geodesics) from a vertex.
  *
- * When there is more than one shortest path between two vertices,
+ * when there is more than one shortest path between two vertices,
  * all of them will be returned.
  *
  * \param graph The graph object.
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -284,6 +284,7 @@ add_legacy_tests(
   all_shortest_paths
   assortativity
   coreness
+  direct_betweenness
   efficiency
   global_transitivity
   hub_and_authority
--- a/tests/unit/direct_betweenness.c
+++ b/tests/unit/direct_betweenness.c
@@ -1,0 +1,18 @@
+//igraph/src/centrality/betweenness.c 500
+//
+#include "test_utilities.h"
+
+igraph_error_t igraph_direct_betweenness_cutoff(const igraph_t *graph, igraph_vector_t *res,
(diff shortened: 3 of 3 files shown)


GroteGnoom:
If I have an undirected star:
> igraph_small(&graph, 5, IGRAPH_UNDIRECTED, 0,4, 1,4, 2,4, 3,4, -1);
and I do a subset betweenness between sources 0 and 1, and target 2 and 3:
> igraph_vss_range(0, 2), igraph_vss_range(2,4)
Then I should get 4, right? Not 2.

The original implementation just divides the end result by 2 for undirected situations, but I think that's incorrect. That only works if your targets and sources are the same. If there's no overlap then there is no division needed at all, and if there is a little overlap, you can't use a basic division at the end to fix everything.

szhorvat:
> Making cutoffs work seems like a lot of work, because the only shortest path functions with cutoffs are `igraph_distances_cutoff`, `igraph_distances_dijkstra_cutoff`, which don't actually give me the paths.

Let's skip this until the necessary functions are implemented. #2358

szhorvat:
> If I have an undirected line:
> 
> > ```
> > igraph_small(&graph, 5, IGRAPH_UNDIRECTED, 0,1, 1,2, -1);
> > ```
> 
> and I do a subset betweenness between source 0, and target 2:
> 
> > igraph_vss_range(0, 1), igraph_vss_range(2,3))
> 
> Then for vertex 1 I should get a betweenness of 1, right? Not 0.5.

Correct.

> The original implementation just divides the end result by 2 for undirected situations, but I think that's incorrect. That only works if your targets and sources are the same. If there's no overlap then there is no division needed at all, and if there is a little overlap, you can't use a basic division at the end to fix everything.

I believe you are right. I'd have to look into this in more detail to see how the subset version works and how it can be fixed most effectively.

But we are making progress, we discovered a bug, so that's good. This was precisely the purpose of the direct implementation.

szhorvat:
Opened a new issue for this:

 - #2359

ntamas:
> I believe you are right.

Yes, you are correct. This seems to be a bug indeed. Great job!

Issue #2360, opened 2023-06-06
Pointer attribute type


tkram01:
Would it be possible to add a pointer (or int64) attribute type so we can easily attach pointers to more complicated objects to nodes and edges?

ntamas:
Hmmm, in some sense we have that - the R and Python interfaces use `IGRAPH_ATTRIBUTE_OBJECT` from the `igraph_attribute_type_t` enum, but the C attribute handler does not make use of this member of the enum.

Can you please describe your use-case in more detail so we could determine what exactly would be needed from our side to make this work for you?

tkram01:
Exposing IGRAPH_ATTRIBUTE_OBJECT to the C attribute handler sounds like a great solution. We have some complicated objects that we want to attach to each node that we use in isomorphism checks to determine if 2 nodes match. We currently do this by adding a bunch of float and string attributes to each node but this is cumbersome and there are some properties that have more complicated structures that this doesn't work well for. It would be much better if we can just attach a pointer to the object we already have constructed that we can then reference in the isomorphism callback.

ntamas:
I've started working on this in #2361. No unit tests yet so it might not work at all yet, but if you feel adventurous, feel free to try and give feedback in the PR :)

PR #2361, opened 2023-06-09
Pointer attributes in C

ntamas:
This PR adds support for storing pointers as attribute values in C, in response to #2360.

Still a work in progress, no unit tests yet, I just added the new bits of code by copying similar handlers for numeric attributes and changing the data types.


szhorvat:
What's the approach here? Will we have custom constructors and destructors?  Or just plain pointers with null pointers also allowed, leaving memory management to the user?

Suppose a vertex is deleted. How is the pointed-to object deallocated?  What happens when adding a new vertex?  How is the object created?

src/graph/cattributes.c
```diff
@@ -537,6 +591,11 @@ static void igraph_i_attribute_permutation_work_area_destroy(
         igraph_Free(work_area->boolean);
         work_area->boolean = NULL;
     }
+    if (work_area->pointers != NULL) {
+        igraph_vector_ptr_destroy(work_area->pointers);
+        igraph_Free(work_area->pointers);
```
szhorvat, line 596: This should be `IGRAPH_FREE`

szhorvat:
I see that the approach was not to take ownership of the pointed to object.

But this is only really useful for static graphs. For dynamically changing graphs it is very limited. For a static graph, one might as well use a plain vector of objects without resorting to attributes. The true value of attributes is that they get automatically managed when the graph changes.

szhorvat:
@ntamas What do we do with this? Are interested in pushing it further for 1.0?

ntamas:
No, not really. I have plenty of other things on my plate.

PR #2365, opened 2023-06-28
Graph power

### LINKED ISSUE #2063: Replace `connect_neighborhood` by `graph_power`

szhorvat:
Continuation of #2341

Fixes #2063

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [master]
 
+### Added
+
+ - `igraph_graph_power()` computes the kth power of a graph.
+
 ### Changes
 
  - `igraph_community_walktrap()` no longer requires `modularity` and `merges` to be non-NULL when `membership` is non-NULL.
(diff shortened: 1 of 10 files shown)


szhorvat:
I'm not sure what happened. I used `gh pr checkout 2341`, then made modifications and pushed to it. This normally shows up in the PR itself.

Can we have this in the next release @ntamas ?

szhorvat:
Please squash into a single commit when merging!

ntamas (approved):

PR #2370, opened 2023-07-11
WIP: Change IGRAPH_BOOL_TYPE to int

krlmlr:
This is a test to see if we can successfully change the definition of the `IGRAPH_BOOL_TYPE` macro. This seems necessary for the R package.

Do not merge.

### DIFF
--- a/include/igraph_config.h.in
+++ b/include/igraph_config.h.in
@@ -48,7 +48,7 @@ __BEGIN_DECLS
  * Any other use-case of overriding igraph's bool type is completely
  * unsupported.
  */
-#define IGRAPH_BOOL_TYPE bool
+#define IGRAPH_BOOL_TYPE int
 
 __END_DECLS
 


krlmlr:
Thanks for taking this on!

krlmlr:
Thanks. How can we ensure that this remains operational? The failures seem to occur on Azure Pipelines only, do we want to construct a failing test on GitHub Actions (e.g., by adding a macos entry to the build matrix)? Or can we add that test to Azure Pipelines?

szhorvat:
The ARPACK-related failures should be fixed in ARPACK, and are unrelated to this change. I have an open PR for that.

`igraph_bool_t = int` seems completely functional, I don't see any issues. I think you can go ahead on the R/igraph side.

krlmlr:
Thanks. `igraph_bool_t = int` is functional now, it wasn't this morning. How can we ensure that it remains functional in the future? Of course, we'll see it when we use a new version in the R package, that might just be good enough.

szhorvat:
IMO that's good enough for now, but if there are repeated breakages, we can set up a check directly in this repo.

The reason why I wouldn't do it now is that we already have as many checks as can run in parallel. Adding more will slow things down.

PR #2371, opened 2023-07-13
Add citation information.

vtraag:
I've added citation information now to the README.md file, and I added a `CITATION.cff` file that makes it easier to cite. The 2006 article is in the preferred citation in `CITATION.cff`, so that should be shown by GitHub. 

Additionally, `CITATION.cff` also provides the information from Zenodo. At the moment it really reflects the most recent release, including that particular DOI. Preferablly, we shouldn't have to update this manually of course, but that would be of later concern.

The citation itself is already in the documentation, so that wouldn't need any update. Let me know if this should still require some changes.

### DIFF
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -1,0 +1,57 @@
+# This CITATION.cff file was generated with cffinit.
+# Visit https://bit.ly/cffinit to generate yours today!
+
+cff-version: 1.2.0
+title: igraph
+message: >-
+  If you use igraph, please cite it using the
+  metadata from this file.
+type: software
+authors:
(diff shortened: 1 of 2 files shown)


szhorvat:
> At the moment it really reflects the most recent release, including that particular DOI.

There is also a master DOI for all versions.

vtraag:
> There is also a master DOI for all versions.

Yeah, indeed, but then it feels weird to include version information. Anyway, not sure what's the best solution there.

szhorvat:
We can skip the version for the sake of a stable citation and to save the trouble of having to edit the file on each release?

PR #2372, opened 2023-07-15
multiarch support: config headers in a per-architecture location

jgmbenoit:
Description: upstream: fix: multiarch: file-conflict
 This patch installs architecture/built specific C headers
 in a per-architecture location to avoid conflict on multiarch
 systems.
Origin: debian
Last-Update: 2023-07-15

### DIFF
--- a/etc/cmake/pkgconfig_helpers.cmake
+++ b/etc/cmake/pkgconfig_helpers.cmake
@@ -68,6 +68,11 @@ endif()
 
 join_paths(PKGCONFIG_LIBDIR "\${exec_prefix}" "${CMAKE_INSTALL_LIBDIR}")
 join_paths(PKGCONFIG_INCLUDEDIR "\${prefix}" "${CMAKE_INSTALL_INCLUDEDIR}")
+join_paths(PKGCONFIG_CFLAGS_ONLY_I "-I\${includedir}" "igraph")
+if(CMAKE_LIBRARY_ARCHITECTURE)
+  join_paths(PKGCONFIG_CFLAGS_ONLY_I_ARCH "-I\${includedir}" "${CMAKE_LIBRARY_ARCHITECTURE}" "igraph")
+  set(PKGCONFIG_CFLAGS_ONLY_I "${PKGCONFIG_CFLAGS_ONLY_I_ARCH} ${PKGCONFIG_CFLAGS_ONLY_I}")
+endif()
 configure_file(
   ${PROJECT_SOURCE_DIR}/igraph.pc.in
   ${PROJECT_BINARY_DIR}/igraph.pc
--- a/igraph.pc.in
+++ b/igraph.pc.in
@@ -10,4 +10,4 @@ URL: @PROJECT_HOMEPAGE_URL@
 Libs: -L${libdir} -ligraph
 Libs.private: @PKGCONFIG_LIBS_PRIVATE@
 Requires.private: @PKGCONFIG_REQUIRES_PRIVATE@
-Cflags: -I${includedir}/igraph
+Cflags: @PKGCONFIG_CFLAGS_ONLY_I@
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -425,6 +425,11 @@ use_all_warnings(igraph)
 # GNUInstallDirs be included before generating the pkgconfig file, as it defines
 # CMAKE_INSTALL_LIBDIR and CMAKE_INSTALL_INCLUDEDIR variables.
 include(GNUInstallDirs)
+if(CMAKE_LIBRARY_ARCHITECTURE)
+  set(CMAKE_INSTALL_INCLUDEDIR_ARCHITECTURE "${CMAKE_INSTALL_INCLUDEDIR}/${CMAKE_LIBRARY_ARCHITECTURE}")
+else()
+  set(CMAKE_INSTALL_INCLUDEDIR_ARCHITECTURE "${CMAKE_INSTALL_INCLUDEDIR}")
+endif()
 
 # Generate pkgconfig file
 include(pkgconfig_helpers)
@@ -468,7 +473,7 @@ install(
 )
 install(
   DIRECTORY ${PROJECT_BINARY_DIR}/include/
-  DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}/igraph
+  DESTINATION ${CMAKE_INSTALL_INCLUDEDIR_ARCHITECTURE}/igraph
   FILES_MATCHING PATTERN "*.h"
 )
 install(


szhorvat:
Let's spell out what the problem is.

Normally there are only separate shared libraries for different architectures, not separate headers. But in the case of igraph, headers are also architecture specific because of the definition of `IGRAPH_INTEGER_SIZE` in `igraph_config.h`. This may be 64 or 32 depending on the platform. 

What I am concerned about is whether the locations introduced by this PR are standard across different Linux distros. Do you have any knowledge about this, @jgmbenoit ? Is there no way to let GNUInstallDirs handle this?

How do you handle architecture-specific headers in Fedora, @limburgher?

jgmbenoit:
This is true that this patch very near to the limit-line. It follows [Debian support for multiarch systems](https://wiki.debian.org/Multiarch?action=show&redirect=MultiArch). So it might sound Debian oriented. On the other hand, it makes sense for systems with multi arch (eg, i386/amd64). I pondered the issue before the request. I finally shared this patch because I would be glad as superuser/root of a Un*x box (with or without distribution) to have this behaviour.
Your concerns about GNUInstallDirs handling is addressed in this [Debian wiki on Multiarch implementation](https://wiki.debian.org/Multiarch/Implementation) (section CMake) along with this old Debian bug-report [#796545](mailto:796545@bugs.debian.org).

szhorvat:
Can we get away with simply usign

```cmake
include(GNUInstallDirs)
if(CMAKE_LIBRARY_ARCHITECTURE)
  set(CMAKE_INSTALL_INCLUDEDIR "include/${CMAKE_LIBRARY_ARCHITECTURE}")
endif()
```
without any additional complications?

While we're waiting for an answer about how Fedora handles multi-arch headers ([apparently not with CMAKE_LIBRARY_ARCHITECTURE](https://bugzilla.redhat.com/show_bug.cgi?id=1531678)), do you know if there are any non-Debian-derived distros where `CMAKE_LIBRARY_ARCHITECTURE` may be set?

szhorvat:
@jgmbenoit It seems that there's a bit of an issue. GNUInstallDirs will only set `CMAKE_INSTALL_LIBDIR` to include the architecture triplet when `CMAKE_INSTALL_PREFIX` is set to `/usr`.  When `CMAKE_INSTALL_PREFIX` has other values, including the default of `/usr/local`, then the multiarch directory is not added. This is discussed at https://gitlab.kitware.com/cmake/cmake/-/issues/20565

However, the patch in this PR will add the architecture triplet to the include directory unconditionally for as long as `CMAKE_LIBRARY_ARCHITECTURE` is set. This creates an inconsistency between the library directory and the include directory.

We could add some logic to check the value of `CMAKE_INSTALL_PREFIX`, but at this point it feels like we would be going to great lengths to match complex and barely documented behaviour from GNUInstallDirs. This is fragile as the behaviour of GNUInstallDirs may change and then we need to adapt ...

yurivict:
@szhorvat 
FreeBSD normally doesn't install per-architecture files for regular ports.
When the system is installed - it has only one architecture. There is no need to create multi-architecture paths.

Please do not do this on the FreeBSD OS.

jgmbenoit:
> Can we get away with simply using
> 
> ```cmake
> include(GNUInstallDirs)
> if(CMAKE_LIBRARY_ARCHITECTURE)
>   set(CMAKE_INSTALL_INCLUDEDIR "include/${CMAKE_LIBRARY_ARCHITECTURE}")
> endif()
> ```
> 
> without any additional complications? Why did you choose to add both `-I${includedir}/igraph` and `-I${includedir}/triplet/igraph` to the pkg-config file instead of just the latter?

The idea is to share the headers that are common to all architectures.

szhorvat:
@yurivict Can you check if `CMAKE_LIBRARY_ARCHITECTURE` has a value on FreeBSD?

You can put this in a `CMakeLists.txt`:

```cmake

cmake_minimum_required(VERSION 3.18)
project(Foobar)

include(CMakePrintHelpers)
cmake_print_variables(CMAKE_LIBRARY_ARCHITECTURE)
```

Then do `mkdir build && cd build && cmake ..`

In the output, you will see a line similar to:

```
-- CMAKE_LIBRARY_ARCHITECTURE=""
```

szhorvat:
> The idea is to share the headers that are common to all architectures.

Ah, I see you set it up so that all generated files are installed in the arch-specific directory, the rest in the usual place. However, not all generated files are arch-specific. Also, for better predictability with `include`s, I'd rather keep all header files in the same directory.

jgmbenoit:
> This is fragile as the behaviour of GNUInstallDirs may change and then we need to adapt ...

I agree. Now I think that the burden must be kept at the packaging stage. Ideally the CMake machinery may attribute a specific directory name for architecture-specific headers so that distributions and superusers can apply their favourite policy (probably via GNUInstallDirs).
In short, just forget this patch.

szhorvat:
Thanks for bringing it up though, as we never thought of this issue before. I did ask a question on the CMake forum (before I saw your response): https://discourse.cmake.org/t/how-to-handle-multiarch-headers/8529

yurivict:
@szhorvat 

```CMAKE_LIBRARY_ARCHITECTURE``` is defined as an empty string on FreeBSD.

If the idea is to separate architecture-specific files just for this to be obvious - I guess this is ok.
But every time when on linux 64-bit libraries are installed into lib64 and not into lib - this needs to be patched away.
A single OS installation only runs binaries for a single architecture.

**Comment by jgmbenoit** [comment]
> Also, for better predictability with `include`s, I'd rather keep all header files in the same directory.

This is a superuser and distribution choice. I can see pros and cons.



> However, not all generated files are arch-specific.

What can be useful here, instead of the patch, is to maintained, if possible, a list of the architecture-specific headers.

**Comment by jgmbenoit** [comment]
> If the idea is to separate architecture-specific files just for this to be obvious

The idea is to support better multi-arch boxes; for examples box where a i386 system coexists along a amd64 system.

**Comment by szhorvat** [comment]
I'll close this now, but if we find a cleaner way to support parallel installations for multiple architectures, we can come back to it.

**Comment by limburgher** [comment]
> Let's spell out what the problem is.
> 
> Normally there are only separate shared libraries for different architectures, not separate headers. But in the case of igraph, headers are also architecture specific because of the definition of `IGRAPH_INTEGER_SIZE` in `igraph_config.h`. The default choice may be 64 or 32 depending on `sizeof(size_t)`.
> 
> What I am concerned about is whether the locations introduced by this PR are standard across different Linux distros. Do you have any knowledge about this, @jgmbenoit ? Is there no way to let GNUInstallDirs handle this?
> 
> How do you handle architecture-specific headers in Fedora, @limburgher?

I don't believe there's a standard method.

**Comment by tornaria** [comment]
> Ping @antonio-rojas for Arch, @yurivict for FreeBSD, @tornaria for Void Linux, @orlitzky for Gentoo.

In void-linux multilib is only to support running legacy 32-bit binaries on x86_64. That 
means we ship 32-bit libraries in `/usr/lib32` and that's it. I'm not sure we support building 32-bit binaries on 64-bit system except with cross compilation.

The way cross compiling works is by creating a parallel tree at `/usr/$TRIPLET/` in which one can install packages for the target architecture. The cross compiler will use header files and libraries from there (say `/usr/$TRIPLET/usr/include` and `/usr/$TRIPLET/usr/lib`, etc) so they should be the correct ones for the target architecture.

To build i686 packages (ie 32 bit x86) we don't cross compile, rather we run a container with a complete i686 system so also the correct ones.

**Comment by yurivict** [comment]
You should make the use of architecture-specific paths conditional on the configure-time variable, because on FreeBSD we also don't cross-compile at the ports' level. It should just install files for the current architecture.

**Comment by orlitzky** [comment]
Gentoo supports "multilib" x86/amd64 systems, but I don't think we want any special treatment from the build system. My understanding is that if the x86 and amd64 builds install different headers, we will (ourselves) install a wrapper that chooses the correct arch-specific header and put the wrapper directly in the main system include directory (e.g. `/usr/include`).

Ping @mgorny just to be sure because I've never used it myself.

**Comment by mgorny** [comment]
> Gentoo supports "multilib" x86/amd64 systems, but I don't think we want any special treatment from the build system. My understanding is that if the x86 and amd64 builds install different headers, we will (ourselves) install a wrapper that chooses the correct arch-specific header and put the wrapper directly in the main system include directory (e.g. `/usr/include`).

Yes, we do create wrappers that determine the correct target based on compiler defines and `#include` the right header. The files look like this:

```
#if defined(__x86_64__) /* amd64 */
#	if defined(__ILP32__) /* x32 ABI */
#		error "abi_x86_x32 not supported by the package."
#	else /* 64-bit ABI */
#		include <x86_64-pc-linux-gnu/gpg-error.h>
#	endif
#elif defined(__i386__) /* plain x86 */
#	include <i686-pc-linux-gnu/gpg-error.h>
[…]
```

So as long as the compiler has the correct definitions, there is no need for special support.

szhorvat:
Thanks for the feedback everyone!

PR #2377, opened 2023-07-26
fix: cleaner way to enable posix features, fixes #2376

### LINKED ISSUE #2376: _GNU_SOURCE should probably be set globally with a compiler flag
The purpose of defining [the `_GNU_SOURCE` macro](https://www.gnu.org/software/libc/manual/html_node/Feature-Test-Macros.html) is to enable some standard library extensions, including both GNU and POSIX extensions. Some examples of these extensions are `strcasecmp()` or `stpcpy()`. These are used not only by hand-written igraph code, but also by the Flex/Bison-generated sources.

The igraph headers set `_GNU_SOURCE` in two locations:

 - In `igraph.h`
 - In `igraph_types.h` which is effectively included in all igraph source files

It seems to me that this is not the right approach, because:

 - These macros should be set for CMake feature tests as well, so they can detect the availability of these extensions
 - I do not see the point of setting these in `igraph.h`. `igraph.h` is only included by user code, which does not generally need these extensions. It is not included by igraph's own sources, which is why the macro definitions are also present in `igraph_types.h`
 - Only GNU-compatible compilers understand these macros, while most of the extensions we need are standard POSIX ones. The standard way to enable them is to define `_POSIX_C_SOURCE`.

Thus I propose removing these macros from the igraph headers, and instead setting them in the build system using [`add_compile_definitions()`](https://cmake.org/cmake/help/latest/command/add_compile_definitions.html). Both `_GNU_SOURCE=1` and `_POSIX_C_SOURCE=200809L` should be set.

----

It seems that there is some issue finding these extensions on the outdated CentOS 7, see https://github.com/igraph/rigraph/issues/560 I'm not sure why. Can anyone who is more Docker-proficient figure out what's going on there? @ntamas , @krlmlr ?

szhorvat:
Fixes #2376.

This should in principle improve cross-platform compatibility, at least on platforms which try to follow standards.

It seems that with _only_ `_GNU_SOURCE` defined, `strcasecmp()` is available in `string.h` (which is always present as it's part of the standard C library). However, with `_POSIX_C_SOURCE` defined, this function is only available in `strings.h` (note plural "strings"!) which is not available on all systems. We have alreary been looking for `strcasecmp` in `strings.h` using CMake's `check_symbol_exsits()`, thus if `strcasecmp` is found, then `strings.h` should be included.

### DIFF
--- a/src/core/math.h
+++ b/src/core/math.h
@@ -19,14 +19,12 @@
 #ifndef IGRAPH_CORE_MATH_H
 #define IGRAPH_CORE_MATH_H
 
-/* Use math constants with MSVC */
-#if defined(_MSC_VER) && !defined(_USE_MATH_DEFINES)
-#define _USE_MATH_DEFINES
-#endif
-
 #include <math.h>
 
 /* Math constants are not part of standard C */
+
+/* The following definitions contain enough precision for
+ * an IEEE-754 quadruple-precision floating point format. */
 
 #ifndef M_E
 #define M_E         2.71828182845904523536028747135266250
--- a/tests/unit/igraph_complex.c
+++ b/tests/unit/igraph_complex.c
@@ -22,11 +22,9 @@
 */
 
 #include <igraph.h>
-/* This definition ensures math symbols are also available when
- * compiling with MSVC.
- */
-#define _USE_MATH_DEFINES
-#include <math.h>
+
+/* Ensure that math constants are available. */
+#include "core/math.h"
 
 #include "test_utilities.h"
 


szhorvat:
`linux_clang_17` seems to fail because of some issue with https://apt.llvm.org/, unrelated to this PR.

szhorvat:
It seems that `_POSIX_C_SOURCE` has been supported for a very long time (it's part of the POSIX standard as well), so we do not need `_GNU_SOURCE`. I would rather not define `_GNU_SOURCE` only for the theoretical possibility of some ancient compiler that does not understand `_POSIX_C_SOURCE` but can otherwise compile igraph, at least not without evidence that such a compiler exists.

PR #2379, opened 2023-07-29
Add get_adjacency_submatrix

### LINKED ISSUE #2196: Adjacency matrix for only some vertices

Currently there is `igraph_get_adjacency(_sparse)()` to construct the adjacency matrix of a graph. There is also `igraph_get_incidence()` for a bipartite incidence matrix.

I propose a generalization of both where the vertices corresponding to the matrix rows/column can be specified:

```
igraph_error_t igraph_get_adjaceny_submatrix(
        const igraph_t *graph, igraph_matrix_t *res,
        const igraph_vs_t rows, const igraph_vs_t cols,
        const igraph_vector_t *weights, igraph_loops_t loops);
```

This would generate the submatrix corresponding the specified row-vertices and column-vertices.

There should be a sparse version as well.

**Use cases for the feature**

 - Fast way to obtain the adjacency matrix of a small subgraph (when rows and cols are the same).
 - A single function handles both the adjacency matrix and bipartite incidence matrix cases.
 - Obtain connectivity between different blocks, e.g. two clusters out of many, obtained with a community detection algorithm.
 - There are matrix-based algorithms that must be applied per-connected-component, such as [Seidel's shortest path algorithm](https://en.wikipedia.org/wiki/Seidel%27s_algorithm). This is a convenient way to get a matrix for each component.

One of the purposes of this function would be to offer better performance and better memory usage than generating the entire adjacency matrix first then subsetting it.

**References**

—

GroteGnoom:
Fixes #2196

### DIFF
--- a/include/igraph_conversion.h
+++ b/include/igraph_conversion.h
@@ -43,6 +43,11 @@ IGRAPH_EXPORT igraph_error_t igraph_get_adjacency(
    const igraph_t *graph, igraph_matrix_t *res, igraph_get_adjacency_t type,
    const igraph_vector_t *weights, igraph_loops_t loops
 );
+IGRAPH_EXPORT igraph_error_t igraph_get_adjacency_submatrix(
+    const igraph_t *graph, igraph_matrix_t *res, igraph_get_adjacency_t type,
+    const igraph_vs_t rows, const igraph_vs_t cols,
+    const igraph_vector_t *weights, igraph_loops_t loops
+);
 IGRAPH_EXPORT igraph_error_t igraph_get_adjacency_sparse(
    const igraph_t *graph, igraph_sparsemat_t *res, igraph_get_adjacency_t type,
    const igraph_vector_t *weights, igraph_loops_t loops
--- a/include/igraph_conversion.h
+++ b/include/igraph_conversion.h
@@ -44,7 +44,7 @@ IGRAPH_EXPORT igraph_error_t igraph_get_adjacency(
    const igraph_vector_t *weights, igraph_loops_t loops
 );
 IGRAPH_EXPORT igraph_error_t igraph_get_adjacency_submatrix(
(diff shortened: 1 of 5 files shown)


szhorvat:
The time complexity of `igraph_get_all_eids_between()` should be investigated and included in the docs. It should be logarithmic due to the binary search, but logarithmic in what?

szhorvat:
Either way, I suspect this will be too slow. Technically, this function will always take time proportional to the number of matrix elements due to `igraph_matrix_null()`. But in practice `igraph_matrix_null()` is extremely fast, and won't dominate the timings. What matters is the other operations.

This particular solution does as many edge lookups as the number of matrix elements, i.e. $N_\text{from} \times N_\text{to}$.

Nevertheless, this implementation is so simple that we can have high confidence in it. Thus let's use it to expand the tests before moving on to implementing a faster one.

What I proposed in the issue using sorted intersections should take time $N_\text{from} \times \langle d \rangle_\text{from}$ where $\langle d \rangle$ denotes the mean degree.  I expect that in most practical cases $N_\text{to} >> \langle d \rangle_\text{from}$, but this is not necessarily the case, especially for small submatrices.

So it will be useful to think about various types of applications. Will the submatrices be smaller or larger than the typical degree?

ntamas:
@szhorvat So, am I right to assume in light of your comments that you are in favour of merging this PR even if you have concerns about the performance?

@GroteGnoom is this still a draft or is it ready to be merged?

szhorvat:
No, I think we should fix the performance.

It is not ready, loop edges are not handled. Although that would be easy with something like `if (from == to) continue;`

GroteGnoom:
> It is not ready, loop edges are not handled. 

I think they are?

But it's still a mess, so I'll at least clean up first. And I don't mind looking at performance either.

szhorvat:
> I think they are? (the TODO can be removed I think)

Sorry about that, I didn't look carefully enough.

PR #2387, opened 2023-08-28
refactor!: igraph_vector_swap() and igraph_matrix_swap() no longer return an error code

szhorvat:
`igraph_vector_swap()` and `igraph_matrix_swap()` no longer return an error code.

Reasoning: The current implementation can never fail and I do not see ever needing to account for failures. There is a pattern for adding new elements into vector lists or matrix lists that is inconvenient to implement if this function fails. See 55948f90fc783c2ee90d337c2a1a331cf6bb664a for an example.

Suppose we have `igraph_vector_t *v` and we need to insert it into an `igraph_vector_list_t *list` and never use it again. One way is:

```
IGRAPH_CHECK(igraph_vector_list_push_back_copy(list, v));
```

But this is inefficient as it copies the vector.  A faster way is the following:

```
igraph_vector_t *new_v;
IGRAPH_CHECK(igraph_vector_list_push_back_new(list, *new_v));
igraph_vector_swap(v, new_v);
```

If `igraph_vector_swap()` can fail, and we put an `IGRAPH_CHECK` on it, then a failure may cause `list` to be in an inconsistent state (containing an empty vector when it shouldn't). In some situations this is undesirable. This behaviour is conveniently avoided if `igraph_vector_swap()` can never fail, which is currently cannot. So let's make that official and not return an error code.

I expect this change may be controversial, hence the PR.


ntamas (approved):
LGTM, thanks!

PR #2390, opened 2023-09-04
ci: use windows-2022 runner and compiler with VS 2017

szhorvat:
This PR switches the Windows tests to the windows-2022 (latest) image from the previous windows-2019. This image no longer includes the VS 2015 tools, so we need to switch to VS 2017.

We do not need to make this switch right now, but this PR will stay ready to merge for when we _are_ ready.

Perhaps it'd be best to rebase this for `develop`, meaning that we make the switch for 0.11.

### DIFF
--- a/.azure/build-win.yml
+++ b/.azure/build-win.yml
@@ -73,8 +73,13 @@ steps:
 
   # Notes:
   #  - We call vcvarsall.bat to make sure the compiler (cl.exe) is in the path, and so that we can select the desired MSVC version.
-  #      https://docs.microsoft.com/en-us/cpp/build/building-on-the-command-line?view=msvc-160#vcvarsall-syntax
+  #      https://docs.microsoft.com/en-us/cpp/build/building-on-the-command-line?view=msvc-170#vcvarsall-syntax
   #      This is necessary when not using the Visual Studio CMake generator.
+  #      Location for VS2019: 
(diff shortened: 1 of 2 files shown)


ntamas:
Any disadvantages of using the new image and tools? It's only CI after all, it's not like we publish pre-compiled binaries, so the only problem that we could have is that we start adding stuff to the code that does not work with VS 2015 any more. Is there a real risk to that?

szhorvat:
Yes, the only concern is that we may accidentally add code that does not compile with VS 2015.

The risk is probably low. Here's some info on standards support in various VS versions: https://learn.microsoft.com/en-us/cpp/overview/visual-cpp-language-conformance?view=msvc-170

It was VS 2015 that brought the biggest improvements. For as long as we stay with C99 / C++11, and stay away from the more exotic language features, we're probably fine.

ntamas:
I'd say let's go ahead and merge this if you want to.

szhorvat:
Merging with this reasoning:

 - It makes sense to test with a more modern, and therefore more commonly used MSVC version. We do not currently test with any newer versions.
 - We can always create a run-on-demand (instead of run-on-push) CI test for special configurations. In fact I think we should reduce the amount of on-push CI runs and set things up so special platforms can be tested on demand, or perhaps using a cron job.

ntamas:
> In fact I think we should reduce the amount of on-push CI runs and set things up so special platforms can be tested on demand, or perhaps using a cron job.

I'm fully in favour; a full CI run is already taking a long time.

Issue #2400, opened 2023-10-03
Pajek parser is likely broken


szhorvat:
**Describe the bug**

The Pajek parser contains code that tries to control tokenization (i.e. the Flex part) by setting certain flags from Bison actions. This is likely broken, as Bison may read one token ahead in some situations. Thus this token is already parsed by Flex code before the Bison action had a chance to set the flag.

The flag is the `mode` field of the `context` object.

For example, the Flex code contains a line like this:

```flex
ic      { if (yyextra->mode==1) { return VP_IC; } else { return ALNUM; } }
```

`mode=1` is meant to indicate that we are looking for keywords indicating vertex properties. The keyword `ic` indicates that the next word(s) specify a vertex colour. `ic` should translate to the `VP_IC` token, but if the next word also happens to be `ic`, it should now translate to the generic `ALNUM` token instead, as it is a colour name.

This is attempted with the following Bison rule:

```bison
vpword:
     /* ... */
     | VP_IC { context->mode=3; } vpwordpar {
         context->mode=1;
         IGRAPH_YY_CHECK(igraph_i_pajek_add_string_vertex_attribute("color", $3.str, $3.len, context));
     }
```

Thus as soon as the parser sees a `VP_IC` token, it sets `mode=3`, indicating that the next word should be treated as an `ALNUM` (colour name). Once the colour name is read, we switch back to `mode=1` to look for keywords again.

The problem is that by the time Bison executes this code, it may have already have requested the next token from Flex (see [lookahead tokens](https://www.gnu.org/software/bison/manual/html_node/Lookahead.html)). Thus Flex read the next token before `mode` was changed. The `mode` flag is not serving its purpose.

**To reproduce**

To test whether Bison does indeed look ahead, I used `IGRAPH_ASSERT(yychar == YYEMPTY)` after setting `mode`. Yes, it does look ahead in some situations. (See [Bison docs](https://www.gnu.org/software/bison/manual/html_node/Action-Features.html).)

Alternatively, take `tests/unit/pajek5.net`, and change `ic Green` to `ic ic`. This causes an error with the test `test::pajek`:

```
Error at src/io/pajek.c:213 : Parse error in Pajek file, line 2 (syntax error, unexpected VP_IC, expecting number or ALNUM or QSTR) - Parse error.
```

Switching the mode was meant to prevent parsing as `VP_IC`, but it didn't work, as the switch couldn't happen in time.

**Practical consequences**

[...]

PR #2401, opened 2023-10-03
fix: remove the mode-switching hack from Pajek parser, refs #2400

szhorvat:
Mode switching wasn't functioning properly. Now we accept that vertex/edge property keywords are explicitly allowed and recognized everywhere in their respective sections, and do not attempt to prevent treating them as ALNUM. There is a slight mitigation using flex start conditions: now vertex property keywords are only recognized in the `*vertices` section (everywhere within that section!), and similar for edge property keywords.

See #2400 for more explanation.

This PR _may_ change how some files are handled.

### DIFF
--- a/src/io/pajek-header.h
+++ b/src/io/pajek-header.h
@@ -41,7 +41,7 @@
 
 typedef struct {
     void *scanner;
-    int eof;
+    igraph_bool_t eof;
     char errmsg[300];
     igraph_error_t igraph_errno;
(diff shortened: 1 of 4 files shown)


ntamas:
Thanks! I'm also thinking that a proper solution to the underlying problem was to re-implement the Pajek parser from scratch, without relying on flex/bison, but it seems like a daunting task given the lack of formal specification.

PR #2404, opened 2023-10-09
fix: propagate -DIGRAPH_STATIC usage requirement

szhorvat:
This should ensure that CMake-based projects that use igraph through ` target_link_libraries()` automatically have `-DIGRAPH_STATIC` if igraph was compiled into a static lib. Thus we can remove the explicit definition of this symbols from tests/benchmarks.

I don't quite remember _if_ we still need this and _why_ we need this, so asking for a review.


vtraag:
Looks good to me! Don't know whether it's necessary to address any specific issue, but it certainly seems to be the better way of doing it.

vtraag (approved):

szhorvat:
As I recall this was relevant only when using the same header both with static and shared builds. But the standard setup doesn't really make it possible to install both a shared and static build properly anyway.

PR #2406, opened 2023-10-23
Improved Pajek reader

szhorvat:


ntamas:
I have to admit that I'm not too familiar with the parser syntax and/or the internals of the Pajek parser, so if you think that this is okay, and all CI tests pass, then IMHO it's good to go.

szhorvat:
Fuzzer found a crash. I'll merge once I understand how that happens and I fix it.

szhorvat:
@ntamas There was an issue with allowing custom parameter names. They could conflict with standard names. I'm now solving this by adding a `_` at the end. My fix is a bit ugly, but I don't have patience or energy for anything better. Have a look at the next to last commit please.

szhorvat:
If no objections to the ugly fix, please merge.

szhorvat:
With this approach, we can usually read the first network successfully from Pajek project files (`.paj`, not `.net`). That does not mean proper `.paj` file support though.

szhorvat:
Merging now, further fixes can be made later.

PR #2407, opened 2023-10-24
Joint degree matrix

larah19:
Added:
- src/tests/unit/jdm.c
- src/tests/unit/jdm.out
- igraph_construct_jdm function in src/properties/degrees.c

igraph_construct_jdm constructs a joint degree matrix of a graph that either contains the number of edges (if NULL is passed as the weights parameter) or the sum of the weights (if the weights vector is passed) between vertices of degree i and degree j for every (i, j). The function populates a joint degree matrix and works on directed and undirected graphs. Self-loops are counted once.


szhorvat:
Thanks @larah19, that's a great start!

src/properties/degrees.c
```diff
@@ -517,3 +517,138 @@ igraph_error_t igraph_sort_vertex_ids_by_degree(const igraph_t *graph,
...
+ */
+
+igraph_error_t igraph_construct_jdm(const igraph_t* graph,
+                                    igraph_matrix_int_t* jdm,
+                                    igraph_integer_t dout,
+                                    igraph_integer_t din,
+                                    const igraph_vector_int_t* weights) {
+    igraph_eit_t eit;
+    igraph_es_t es;
+    igraph_integer_t no_of_edges = igraph_ecount(graph);
```
szhorvat, line 551:
```suggestion
    igraph_integer_t no_of_nodes = igraph_vcount(graph);
    igraph_integer_t no_of_edges = igraph_ecount(graph);
```

Let's add this now, as it will be useful for later comments I'll make.

src/properties/degrees.c
```diff
@@ -517,3 +517,138 @@ igraph_error_t igraph_sort_vertex_ids_by_degree(const igraph_t *graph,
...
+        }
+
+        igraph_vector_int_destroy(&out_degrees);
+        igraph_vector_int_destroy(&in_degrees);
+
+    } else {
+        igraph_vector_int_t degrees;
+        igraph_integer_t max_degree;
+        // Compute max degrees
+        IGRAPH_CHECK(igraph_vector_int_init(&degrees, 0));
```
szhorvat, line 614: You probably noticed now that some of the CI tests fail because of memory leaks. For example:

https://dev.azure.com/igraph-team/igraph/_build/results?buildId=5789&view=logs&j=1a52b443-5fb8-53f1-821f-633f5d086e98&t=16950513-9d69-54bb-da2c-179f24a13cca&l=1108

This happens not because of incorrect code, but because we did not yet add memory management. Looking at the error, 

```
Direct leak of 20 byte(s) in 1 object(s) allocated from:
    #0 0x7f03d84b4c38 in __interceptor_realloc ../../../../src/libsanitizer/asan/asan_malloc_linux.cpp:164
    #1 0x561732898b3b in igraph_vector_int_reserve /home/vsts/work/1/s/src/core/vector.pmt:474
    #2 0x56173289abc2 in igraph_vector_int_resize /home/vsts/work/1/s/src/core/vector.pmt:1270
    #3 0x5617328e0a9f in igraph_degree /home/vsts/work/1/s/src/graph/type_indexededgelist.c:1232
    #4 0x5617328e9c77 in igraph_construct_jdm /home/vsts/work/1/s/src/properties/degrees.c:573
    #5 0x56173286e866 in main /home/vsts/work/1/s/tests/unit/jdm.c:240
    #6 0x7f03d7a29d8f  (/lib/x86_64-linux-gnu/libc.so.6+0x29d8f)
```

it tells us the line number to look at, line 240 in the test file. This is where you tested incorrect input. What happens is that when the function quits due to the `IGRAPH_ERROR` below, the `degrees` vector is not de-allocated. To fix this, we can do this:

```suggestion
        IGRAPH_CHECK(igraph_vector_int_init(&degrees, 0));
        IGRAPH_FINALLY(igraph_vector_int_destroy, &degrees);
```

This registers the vector for de-allocation in case of errors. We will also need to de-register it after we destroy it manually. Look for a comment on this below.

There is a shorthand for the above two lines,

```suggestion
        IGRAPH_VECTOR_INT_INIT_FINALLY(&degrees, 0);
```

This macro is here: https://github.com/igraph/igraph/blob/master/include/igraph_vector.h#L123

There is one more small improvement we can make. We can directly allocate the required vector size, so the `igraph_degree()` function wouldn't have to resize it later.

```suggestion
        IGRAPH_VECTOR_INT_INIT_FINALLY(&degrees, no_of_nodes);
```

This is a minor optimization which probably makes very little difference here, but I wanted to show you the principle. Let's go with this last version.

Please make this change for the directed version as well.

src/properties/degrees.c
```diff
@@ -517,3 +517,138 @@ igraph_error_t igraph_sort_vertex_ids_by_degree(const igraph_t *graph,
...
+                MATRIX(*jdm, v2deg - 1, v1deg - 1)++;
+            } else {
+                MATRIX(*jdm, v1deg - 1, v2deg - 1) += VECTOR(*weights)[eid];
+                MATRIX(*jdm, v2deg - 1, v1deg - 1) += VECTOR(*weights)[eid];
+            }
+
+            IGRAPH_EIT_NEXT(eit);
+        }
+
+        igraph_vector_int_destroy(&degrees);
```
szhorvat, line 648:
```suggestion
        igraph_vector_int_destroy(&degrees);
        IGRAPH_FINALLY_CLEAN(1);
```

This de-registers the last data structure we registered for de-allocation. We need to do this since we de-allocated it ourselves here.

If we have `n` instead of `1` as the argument, it would de-register the last `n` instead of the last one.

szhorvat (changes requested):
This is a first review showing how to eliminate the memory leak. More comments on functionality will come later.

src/properties/degrees.c
```diff
@@ -517,3 +517,138 @@ igraph_error_t igraph_sort_vertex_ids_by_degree(const igraph_t *graph,
...
+        igraph_vector_int_destroy(&out_degrees);
+        igraph_vector_int_destroy(&in_degrees);
+
+    } else {
+        igraph_vector_int_t degrees;
+        igraph_integer_t max_degree;
+        // Compute max degrees
+        IGRAPH_CHECK(igraph_vector_int_init(&degrees, 0));
+        IGRAPH_CHECK(igraph_degree(graph, &degrees, igraph_vss_all(), IGRAPH_ALL, true));
+        IGRAPH_CHECK(igraph_maxdegree(graph, &max_degree, igraph_vss_all(), IGRAPH_ALL, true));
```
szhorvat, line 616: Instead of re-computing degrees again, we can do a small optimization here and take the max of the existing degree vector using `igraph_vector_int_max()`. This function must not be called for empty vectors, so let us check `no_of_nodes` and if it is zero, do not call it (just use `0` as the "maximum").

src/properties/degrees.c
```diff
@@ -517,3 +517,139 @@ igraph_error_t igraph_sort_vertex_ids_by_degree(const igraph_t *graph,
...
     return IGRAPH_SUCCESS;
 }
+
+/**
+ * \function igraph_construct_jdm
+ * \brief Constructs a joint degree matrix.
+ *
+ * The joint degree matrix of a graph contains the number of edges, or the sum of the weights, between vertices of
+ * degree i and degree j for every (i, j). The function populates a joint degree matrix and works on directed
+ * and undirected graphs. Self loops are only counted once. If either \p dout or \p din are <= 0, the JDM
```
szhorvat, line 527:
```suggestion
 * and undirected graphs. Self loops are only counted once. If either \p dout or \p din are &lt;= 0, the JDM
```

To fix the doc build. (I thought you were on strike today.)

larah19: Unfortunately, my course deadlines wouldn't strike with me 😆

src/properties/degrees.c
```diff
@@ -22,7 +22,7 @@
 */
 
 #include "igraph_structural.h"
-
+#include "igraph_mixing.h"
```
szhorvat, line 25: Let's move the function to `mixing.c` and remove this header include from here.

src/properties/degrees.c
```diff
@@ -517,3 +517,152 @@ igraph_error_t igraph_sort_vertex_ids_by_degree(const igraph_t *graph,
     }
     return IGRAPH_SUCCESS;
 }
+
+/**
+ * \function igraph_construct_jdm
+ * \brief Constructs a joint degree matrix.
+ *
```
szhorvat, line 524:
```suggestion
 *
 * \experimental
 *
```

src/properties/degrees.c
```diff
@@ -517,3 +517,152 @@ igraph_error_t igraph_sort_vertex_ids_by_degree(const igraph_t *graph,
...
+        }
+        if (igraph_vector_int_size(&in_degrees) == 0) {
+            max_in_degree = 0;
+        } else {
+            max_in_degree = igraph_vector_int_max(&in_degrees);
+        }
+
+        if (dout > 0 && din > 0) {
+            if (dout < max_out_degree || din < max_in_degree) {
+                IGRAPH_ERRORF("dout (%" IGRAPH_PRId ") or din (%" IGRAPH_PRId ") are smaller than the smallest required dimensions of the matrix (%" IGRAPH_PRId ", %" IGRAPH_PRId ")",
```
szhorvat, line 589: Let's not make this an error.

tests/unit/jdm.c
```diff
@@ -0,0 +1,334 @@
...
+
+   You should have received a copy of the GNU General Public License
+   along with this program.  If not, see <https://www.gnu.org/licenses/>.
+*/
+
+#include <igraph.h>
+#include "test_utilities.h"
+
+void print_and_destroy(igraph_t* g, igraph_matrix_int_t* jdm, igraph_vector_int_t* weights) {
+      igraph_matrix_int_print(jdm);
```
szhorvat, line 23: Please change this to `print_matrix_int()`, which is declared in `test_utilities.h`. It will produce an output that is a bit clearer in case of empty matrices.

src/properties/degrees.c
```diff
@@ -517,3 +517,154 @@ igraph_error_t igraph_sort_vertex_ids_by_degree(const igraph_t *graph,
...
+ *
+ * Time complexity: O(E),
+ * where E is the number of edges in input graph.
+ */
+
+igraph_error_t igraph_construct_jdm(const igraph_t* graph,
+                                    igraph_matrix_int_t* jdm,
+                                    igraph_integer_t dout,
+                                    igraph_integer_t din,
+                                    const igraph_vector_int_t* weights) {
```
szhorvat, line 550: I am just noticing one more issue that we need to fix.  We should operate with real-values weights, which implies that we should also use real-values matrices.

Please refactor to use `igraph_matrix_t` for the JDM and `igraph_vector_t` for `weights` instead of the current `igraph_matrix_int_t` and `igraph_vector_int_t`.

szhorvat:
Thanks @larah19!  Let me know when you consider it done, and I will take it from there, as I'd like you to start focusing on the bipartite degree realization problem.

It seems to me that functionally this is good, and the documentation is very well written. The test suite is extensive, so we feel confident to make changes.

What I'm going to change:

 - Simplify some of the code.
 - Update `interface/functions.yaml`.

I'll do this myself, but I'd like to take a look when done, as a demonstration of good practice, as well as some small technical things we did not yet discuss.

So once again, let me know when you consider it done and are ready to hand it over. Then I'll push changes to this branch in your fork.

larah19:
If there are no additional comments I'm ready to hand it over :) Thanks for the guidance in this!

szhorvat:
It seems there was a bug in that the diagonal of the matrix was doubled.

I also changed the auto-calculation of the max degree so that row and column counts of the matrix are determined separately, and auto-calculation happens only if the input is negative (not when it is zero).

src/misc/mixing.c
```diff
@@ -414,3 +414,141 @@ igraph_error_t igraph_assortativity_degree(const igraph_t *graph,
...
+
+/**
+ * \function igraph_jdm
+ * \brief The joint degree matrix of a graph.
+ *
+ * \experimental
+ *
+ * This function constructs the joint degree matrix of the graph \p jdm.
+ * <code>jdm[i-1, j-1]</code> will contain the number of edges (or sum of edge
+ * weights) between vertices of degrees <code>i-1</code> and <code>j-1</code>.
```
ntamas, line 426: Shouldn’t this be i and j here instead of i-1 and j-1?

ntamas (changes requested):
General comment: can we spell out “joint degree matrix” explicitly, at least in the function names? jdm as an abbreviation in argument names is okay; but I’d rather not have it in function names.

src/misc/mixing.c
szhorvat: No. This is a deliberate choice. We omit the row and column corresponding to zero degrees, as they are always all-zero, and not useful. If we use 1-based indexing, then $J_{ij}$ is indeed the number of connections between vertices of degree $i$ and $j$.

src/misc/mixing.c
```diff
@@ -414,3 +414,141 @@ igraph_error_t igraph_assortativity_degree(const igraph_t *graph,
...
     return IGRAPH_SUCCESS;
 }
+
+/**
+ * \function igraph_jdm
+ * \brief The joint degree matrix of a graph.
+ *
+ * \experimental
+ *
+ * This function constructs the joint degree matrix of the graph \p jdm.
```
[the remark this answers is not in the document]
szhorvat: I meant to put a comma there, but I see your point.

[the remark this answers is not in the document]
szhorvat: See if you like it with the comma, otherwise I'll try to rephrase.

szhorvat:
@ntamas Please leave merging to me, just leave a review.

One more thing I'm trying to decide is what we should put on the diagonal of the JDM: the number of connections between degree-i vertices, or twice that value? This is similar to self-loop counting, but not the same. If we put twice that value, the interpretation of the matrix is a bit harder, but some formulas become simpler (e.g. relation to assortativity).

src/misc/mixing.c
ntamas: Okay, but that's not what you wrote there. Substituting `i = 1` and `j = 1`, the sentence reads as "jdm[0, 0] will contain the number of edges (or sum of edge weights) between vertices of degrees 0 and 0".

Initially I wanted to suggest that the end of the sentence should read as "vertices of degrees i and j", but considering that even we got confused by the change in indexing, I think we should just keep everything 0-based for sake of consistency and accept that the first column and the first row of the matrix will be zero by definition. We only waste O(n) space with it, while the size of the matrix is O(n^2) so it's negligible as n goes to infinity.

src/misc/mixing.c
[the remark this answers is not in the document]
ntamas: I'll add a proposal.

ntamas:
```suggestion
 * This function constructs the joint degree matrix \p jdm of the graph \p g.
```

ntamas:
> If we put twice that value, the interpretation of the matrix is a bit harder, but some formulas become simpler (e.g. relation to assortativity).

We can add an `igraph_loops_t loops` argument and the user can decide what to do, similarly to how we do it in other functions.

src/misc/mixing.c
szhorvat: Sorry about that, I fixed is to say "vertices of degrees i and j".

The literature on the subject generally defines the JDM _without_ the zero-degree row/column. Adding that row/column for indexing reasons would give a preference to zero-based indexing languages, but igraph is also used from 1-based indexing ones. We have other functions with similar issues where the first entry corresponds to degree 1, e.g. the `knnk` output parameter of `igraph_avg_nearest_neighbor_degree()`. For these reasons, I would rather omit the first column.

ntamas: Good point on the index base being language-dependent. Also, if the JDM is defined this way in the literature, let's be consistent with the literature. Thanks for fixing the docs, this is now good to go.

ntamas (approved):

szhorvat:
Thinking more carefully about this, I'm starting to think that in order to minimize user confusion, we should have two closely related functions. Let us call them 

 - `igraph_joint_degree_matrix()`, $J_{ij}$
 - `igraph_degree_mixing_matrix()`, $M_{ij}$ (for lack of a better name)

## Joint degree matrix

You will find this concept in graph theory papers such as https://arxiv.org/abs/1302.3548

$J_{ij}$ is defined plainly as the number of connections form degree $i$ vertices to degree $j$ vertices. It counts individual connections. 

In undirected graphs: Self-loops are counted once, since they are one connection. Connections between same-degree vertices are counted once, since they are one connection.  Note that self-loops still contribute 2 to degrees.

In directed graphs: $i$ is an out-degree and $j$ is an in-degree.

Examples:

image
image

----

image
image
image
image

----

For undirected $J_{ij}$ it is true that $\sum_{i\ge j} J_{ij}$ gives the total number of edges. $\sum_i J_{ik}$ gives the number of edges incident to vertices of degree $k$.

Generally, $(i,j)$ pairs are considered unordered in undirected graphs and ordered in directed ones. $i$ and $j$ run from 1 to the max degree.

$J_{ij}$ is proportional to the probability that a randomly chosen edge connected vertices of degrees $i$ and $j$.

## Degree mixing matrix (for lack of a better name for now).

This concept comes up in the study of degree correlations, but is not always properly named. It is sometimes referred to as a joint degree distribution, but IMO that is not accurate. It is really a conditional distribution.

$M_{ij}$ is proportional to the conditional probability $P(j|i)$ that a vertex with degree $j$ connected to a vertex with degree $j$. We pick two vertices, one of degree $i$ and one of degree $j$ at random. $M_{ij}$ is proportional to the probability that they are connected.

In the directed case, $M_{ij}$ and $J_{ij}$ are the same.

We get the undirected case of $M_{ij}$ by treating each undirected edge as a pair of reciprocal directed edges. This will not be the same as $J_{ij}$. It's diagonal will be twice larger.

In the undirected case, $\sum_{i,j} M_{ij}$ gives twice the edge count. $\sum_{i} M_{ik}$ gives the degree distributions of neighbours of vertices of degree $k$.

When studying degree correlations, sometimes people are looking to relate the in-degree of $u$ to the in-degree of $v$ when there is an $u\longrightarrow v$ connection. The in-degree of $u$ may be zero. Therefore, for $M_{ij}$, $i$ and $j$ should run from 0 to the max degree. See the recent discussion about $k_{nn}(k)$ on the forum.

szhorvat:
If we do go with the two function solution, I propose to keep the JDM function ($J_{ij}$) mostly as is. The $M_{ij}$ function would take additional arguments, similar to what I did in #2419, and would include rows/columns for zero degrees.

szhorvat:
I'd like to merge this in its current state, and add a second `igraph_joint_degree_distribution()` function in a new PR, or perhaps in PR #2419. Is that okay with you @ntamas ?

ntamas:
Yes, that's okay, but we need to be extra careful about explaining the difference between `igraph_joint_degree_matrix()` and `igraph_joint_degree_distribution()` and helping users in choosing the right function for their purposes. It took me a while to wrap my head around it.

PR #2408, opened 2023-10-24
Fix docs and behaviour of min/max functions

szhorvat:
The smallest/largest element of an empty vector/matrix is not defined.

As I recall, out decision was that vector min/max functions should not be legal to call with empty structures.  This is the assumption I've been following when using these functions.

But looking at the docs, this was not explained well in all cases. I fixed this here.

Looking at the code, this is not always what the functions do. In particular, this is not what `vector_which_min` and `vector_which_max` do. They return `-1` instead. However, `vector_which_minmax` assumes that the vector is not empty, which is inconsistent with the other two.

`matrix_which_min/max/minmax` also seem to assume that the matrix is not empty, even though it calls `vector_which_min/max/minmax`. The return value is nonsensical otherwise.

Since these functions are not exposed to high-level languages, I propose that we forbid empty structures, and put in `IGRAPH_ASSERT`s where necessary.  Reasoning: when C code misuses these functions (due to an oversight), it's better to get an assertion failure which alerts the programmer to their error than to get a bad result.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
  - Resolved "ignoring duplicate libraries" warning when building tests with Xcode 15 on macOS.
  - Fixed the handling of duplicate vertex IDs in `igraph_induced_subgraph()`.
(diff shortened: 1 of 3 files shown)


szhorvat:
Tests pass, no new clang-analyzer warnings, no new Coverity warnings.

ntamas:
> I propose that we forbid empty structures
Agreed.

Issue #2413, opened 2023-10-26
Investigate NaN handling in adjacency matrix functions


szhorvat:
NaN handling should be cleaned up in adjacency matrix functions.

Proposal:

 - In both the weighted and unweighted matrix -> graph conversion functions, any value that is not `is_finite()` should trigger an error (i.e. both NaN and Inf).
 - Leave the weighted graph -> matrix conversion function as-is.

It is tempting to try to pass through these values to the weight vector in the weighted function, but this is probably not very useful, and it is difficult to implement. Currently, probably even `matrix_is_symmetric()`, used by the UNDIRECTED mode, doesn't work well with NaNs. Several of the matrix conversion modes rely on arithmetic and comparison, which does not play well with NaN. Implementing special NaN handling for all of these is a lot of work for little gain.

The same issue applied to `biadjacency` functions as well.

szhorvat:
If we want to be rigorous about this, we could use `igraph_i_safe_real_to_int()` instead of `isfinite()`. Performance impact unclear. `isfinite()` is inlined so its performance impact is very small.

ntamas:
I guess that with LTO `igraph_i_safe_real_to_int()` would be inlined as well, right? I think we can safely assume that for performance-critical applications one would compile with LTO anyway.

szhorvat:
Or we can even convert `igraph_i_safe_real_to_int()` to a macro. I checked using GodBolt that this function compiles to very few instructions, so it is probably fast once inlined.

PR #2417, opened 2023-10-28
fix: pass through NaN values when building graph from weighted adjacency matrix

szhorvat:
Ref #2413 

NaN values are now passed through when building a _weighted_ graph  from an adjacency matrix.

This still needs tests, but I'd like a review now @ntamas.

There were two issues that needed to be addressed:

 - max(NaN, anything) should give NaN with the "max" method
 - Adjacency matrices with NaN in symmetric positions should be treated as symmetric. `igraph_matrix_is_symmetric()` doesn't give `true` for these because `NaN != NaN`. A clean way to change the behaviour of `igraph_matrix_is_symmetric()` would be to [define the `EQ` macro](https://github.com/igraph/igraph/blob/master/src/core/matrix.pmt#L1579) so that it treats NaN as equal to NaN, but this would have further implications which are undesirable. In almost all contexts, we want NaN != NaN. Therefore I re-implemented the symmetry check.

It is interesting to note that some systems, like MATLAB, consider any matrix with NaNs to be non-symmetric. R, however, does accept them as symmetric.

```
> isSymmetric(m)
[1] TRUE
> m<-cbind(c(0,NaN),c(NaN,0))
> isSymmetric(m)
[1] TRUE
> NA==NA
[1] NA
> NaN==NaN
[1] NA
```

It seems safer to me to not change `igraph_matrix_is_symmetric()` in igraph and let it continue rejecting NaNs. This is actually useful in numerical contexts.

### DIFF
--- a/src/constructors/adjacency.c
+++ b/src/constructors/adjacency.c
@@ -512,7 +512,7 @@ static igraph_error_t igraph_i_weighted_adjacency_max(
         for (j = i + 1; j < no_of_nodes; j++) {
             M1 = MATRIX(*adjmatrix, i, j);
             M2 = MATRIX(*adjmatrix, j, i);
-            if (M1 < M2) {
+            if (M1 < M2 || isnan(M2)) {
                 M1 = M2;
             }
             if (M1 != 0.0) {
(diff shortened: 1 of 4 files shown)


src/constructors/adjacency.c
```diff
@@ -531,11 +531,21 @@ static igraph_error_t igraph_i_weighted_adjacency_undirected(
...
-        IGRAPH_ERROR(
-            "Adjacency matrix should be symmetric to produce an undirected graph.",
-            IGRAPH_EINVAL
-        );
+    /* We do not use igraph_matrix_is_symmetric() for this check, as we need to
+     * allow symmetric matrices with NaN values. igraph_matrix_is_symmetric()
+     * returns false for these as NaN != NaN. */
+    igraph_integer_t n = igraph_matrix_nrow(adjmatrix);
+    for (igraph_integer_t i=0; i < n; i++) {
+        for (igraph_integer_t j=0; j < n; j++) {
```
ntamas, line 539: Aren't we comparing each `(i, j)` pair twice this way? I think it is enough to start `j` from `i+1`, right?

szhorvat:
Rebased on current `master`.

szhorvat:
@ntamas I don't want to leave this PR hanging forever. What we have should fix NaN passthrough in `igraph_weighted_adjacency()`. Let's merge this and deal with the rest later. I added TODO items in #2413 and added tests to for the fixes in this PR.

ntamas (approved):

PR #2421, opened 2023-10-30
feat: connected component calculations now use the cache

szhorvat:
.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@
 ### Other
 
  - Updated vendored mini-gmp to 6.3.0.
+ - `igraph_connected_components()` makes better use of the cache, improving overall performance.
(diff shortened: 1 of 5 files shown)


szhorvat:
The reason is that `igraph_connected_components()` is buggy and returns the wrong community size vector. This PR caused the correct `csize` vector to be returned and thus changed the result of `igraph_communities_leading_eigenvector()`. A bug in `igraph_connected_components()` is very concerning.

CC @Antonov548, in case this bug may be related to [the investigation you are doing at the moment](https://github.com/igraph/rigraph/pull/840#issuecomment-1785276444).

szhorvat:
@Antonov548 I'm sorry, I was confused. The bug was in this PR, not the original code.

ntamas (approved):

PR #2422, opened 2023-10-30
feat: igraph_joint_degree_distribution

szhorvat:
Docs and implementation are complete, but no tests yet.

I'd like to have #2419 merged before adding tests and finalizing the cross referencing of docs.

The following is a nice visual illustration of why we need the joint degree distribution (and not joint degree matrix) for statistical applications dealing with degree correlations.

Join degree matrix of a large undirected Erdős-Rényi graph:

image

Joint degree distribution of the same:

image


szhorvat:
Here's a thought: Perhaps we should generalize this to work with arbitrary integer "vertex types", which can be used in place of the degrees? Then supplying in- or out-degrees would be just one choice.

szhorvat:
> Here's a thought: Perhaps we should generalize this to work with arbitrary integer "vertex types", which can be used in place of the degrees? Then supplying in- or out-degrees would be just one choice.

This is now implemented as `igraph_mixing_matrix()`. I believe this name is used in the social sciences, but perhaps it is not the best fit here, given that it is a counterpart of `igraph_joint_degree_distribution()` and not `igraph_joint_degree_matrix()`.

@mbojan, perhaps you have some input on this function and its naming?

src/misc/mixing.c
```diff
@@ -576,3 +576,322 @@ igraph_error_t igraph_joint_degree_matrix(const igraph_t* graph,
...
+ * Computes the mixing matrix M_ij, i.e. the joint distribution of vertex categories.
+ * Categories are represented by non-negative integer indices, passed in
+ * \p from_types and \p to_types. The row and column counts of \p m will be one
+ * larger than the largest source and target type, respectively. Re-index type
+ * vectors using \ref igraph_reindex_membership() if they are not contiguous integers,
+ * to avoid producing a very large matrix.
+ *
+ * </para><para>
+ * M_ij is proportional to the probability that a randomly chosen ordered vertex pair
+ * have types \c i and \c j.
```
ntamas, line 867:
```suggestion
 * has types \c i and \c j.
```

src/misc/mixing.c
```diff
@@ -576,3 +576,322 @@ igraph_error_t igraph_joint_degree_matrix(const igraph_t* graph,
...
+ *
+ * \param graph The input graph.
+ * \param m The mixing matrix will be stored here.
+ * \param from_types Vertex types for source vertices. These must be non-negative integers.
+ * \param to_types Vertex types for target vertices. These must be non-negative integers.
```
ntamas, line 872: Does it make sense to have `to_types` for undirected graphs? I would assume it does as each undirected edge is counted in "both" directions and the types of the vertices can then differ in both configurations, but I wonder whether it would really make a difference.

src/misc/mixing.c
szhorvat: This was intentional, but now it is rephrased.

src/misc/mixing.c
szhorvat: Good question. The matrix _would_ be different, since the row and column counts may not be the same anymore. Is it a useful matrix? I am not sure. I would leave it as-is.

szhorvat:
I renamed the second function to `igraph_joint_type_distribution()` for consistency with `joint_degree_distribution()`, but left the term "mixing matrix" in the short description for the sake of searchability in the index.

Test for `igraph_joint_degree_distribution()` is added. One more piece is missing, a test for `igraph_joint_type_distribution()`. But that function is almost certainly fine, since it shares an implementation with `igraph_joint_degree_distribution()`.

szhorvat:
@ntamas This is now ready for review (and hopefully merging, assuming tests pass).

szhorvat:
There's a small bug in the one feature I didn't test ...

szhorvat:
It's good now.

ntamas (approved):

ntamas:
Thanks a lot!

szhorvat:
@ntamas Thanks! AFAIC we're good to release 0.10.8. The one thing I'd do beforehand is check on why Travis is not working, and if easy, run the Travis tests once (it's exotic hardware and old compilers, so it's good to run a test before release).

PR #2423, opened 2023-11-01
fix: pass correct integer type to vararg functions

dtelsing:
When building igraph with IGRAPH_INTEGER_SIZE=64 on i686, I noticed that some tests fail with incorrect integer values reported. After a lot of debugging, this turned out to be because some variadic functions expect the arguments to be 'int' (32 bit), but igraph_integer_t (64 bit) was passed as argument.

### DIFF
--- a/src/core/matrix.pmt
+++ b/src/core/matrix.pmt
@@ -1862,9 +1862,9 @@ igraph_error_t FUNCTION(igraph_matrix, fprint)(const TYPE(igraph_matrix) *m, FIL
                 fputc(' ', file);
             }
 #ifdef FPRINTFUNC_ALIGNED
-            FPRINTFUNC_ALIGNED(file, VECTOR(column_width)[j], MATRIX(*m, i, j));
+            FPRINTFUNC_ALIGNED(file, (int) VECTOR(column_width)[j], MATRIX(*m, i, j));
 #else
-            fprintf(file, format, VECTOR(column_width)[j], MATRIX(*m, i, j));
+            fprintf(file, format, (int) VECTOR(column_width)[j], MATRIX(*m, i, j));
 #endif
         }
         fprintf(file, "\n");
--- a/examples/simple/igraph_union.c
+++ b/examples/simple/igraph_union.c
@@ -103,7 +103,7 @@ int main(void) {
     igraph_vector_ptr_init(&glist, 10);
     for (i = 0; i < igraph_vector_ptr_size(&glist); i++) {
         VECTOR(glist)[i] = calloc(1, sizeof(igraph_t));
-        igraph_vector_int_init_int_end(&v, -1, i, i + 1, 1, 0, -1);
+        igraph_vector_int_init_int_end(&v, -1, (int) i, (int) i + 1, 1, 0, -1);
         igraph_create(VECTOR(glist)[i], &v, 0, IGRAPH_DIRECTED);
         igraph_vector_int_destroy(&v);
     }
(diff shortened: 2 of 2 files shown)


ntamas:
Hmmm, I'm not sure if we actually support building a 64-bit igraph on a 32-bit platform (@szhorvat), but this is nevertheless a good catch that should probably be fixed.

@dtelsing Is this an exhaustive list of such incorrect calls in the codebase, or are these only the ones that you happened to stumble upon? I'm trying to figure out whether we should scan the codebase for more problems like this.

szhorvat:
I'll test this on the RPi when I get to it.

I'm surprised by this: why isn't it an issue on x86_64? These arguments are still expected to be of type `int`.

szhorvat:
> When building igraph with IGRAPH_INTEGER_SIZE=64 on i686

We never really considered this case, but I don't see why it wouldn't work. However, it is probably not a good idea, as address space is limited on i686, and using 64-bit integers doubles the memory use of many functions. I expect it will also affect performance.

dtelsing:
I was updating the package in GNU Guix and noticed that it failed to build on a 32-bit platform, but with this patch all tests pass. IGRAPH_INTEGER_SIZE is set to match the integer size of the (external) SuiteSparse package, which is always 64 in newer versions.
This should be a problem whenever sizeof(igraph_integer_t) != sizeof(int), right?

I checked igraph_vector_int_init_int, igraph_vector_init_int, but otherwise not exhaustively.

Mhm, the calling convention for varags is different on x86_64: https://www.uclibc.org/docs/psABI-x86_64.pdf#subsection.3.5.7. Maybe that's why it works? I didn't check in detail.

szhorvat:
> IGRAPH_INTEGER_SIZE is set to 64 to match the integer size of the (external) SuiteSparse package, which is always 64 in newer versions.

igraph no longer depends on an external SuiteSparse. Now we rely on a bundled SuiteSparse, precisely because we need to control the type of integers it uses. Please do not attempt to unbundle this dependency ... that has a risk of breaking things.

> This should be a problem whenever sizeof(igraph_integer_t) != sizeof(int), right?

I believe so. That is why I do not understand why it does _not_ fail on 64-bit systems. It seems to me that it _should_ fail.



> Mhm, the calling convention for varags is different on x86_64: https://www.uclibc.org/docs/psABI-x86_64.pdf#subsection.3.5.7. Maybe that's why it works? I didn't check in detail.

Thanks for the link, it's likely related to that. I'll dig deeper into this and will see if there are other instances that need to be fixed.

szhorvat:
@ntamas I looked for GCC features to get warning for these, but I could not find any.

It seems to me that passing variables to functions like `igraph_vector_init_int()` is a misused. These functions should really be used with literals. If we accept that, a solution like this could work in future versions, and would provide more type safety:

https://stackoverflow.com/a/10571237/695132

src/core/matrix.pmt
```diff
@@ -1862,9 +1862,9 @@ igraph_error_t FUNCTION(igraph_matrix, fprint)(const TYPE(igraph_matrix) *m, FIL
                 fputc(' ', file);
             }
 #ifdef FPRINTFUNC_ALIGNED
-            FPRINTFUNC_ALIGNED(file, VECTOR(column_width)[j], MATRIX(*m, i, j));
+            FPRINTFUNC_ALIGNED(file, (int) VECTOR(column_width)[j], MATRIX(*m, i, j));
```
szhorvat, line 1865: This one should not be needed (no action needed on your part though).

szhorvat:
Thanks @dtelsing !  I fixed these in two separate commits. The patch you proposed looks good, and I can recommend it for Guix until igraph 0.10.8 comes out with this fix.

PR #2424, opened 2023-11-01
refactor: set 'pure' attribute on some functions

szhorvat:
This PR adds GCC's "pure" attribute to certain functions. The hope is that this can improve performance. In particular, it should help with code like the following by preventing repeated calls to `vector_size()`:

```
for (igraph_integer_t i=0; i < igraph_vector_size(v); i++) {
    // read VECTOR(v)[i] and do something with it
}
```

See https://gcc.gnu.org/onlinedocs/gcc/Common-Function-Attributes.html

The attribute is added fairly conservatively. It could be applied in many more places, but doing so seems like an unnecessary risk.

MSVC does not have an equivalent attribute.

[According to an old version of the GCC docs](https://gcc.gnu.org/onlinedocs/gcc-4.1.2/gcc/Function-Attributes.html), this attribute was added in GCC 2.96, released in 2001.


szhorvat:
Strangely I see a small but consistent performance improvements _even with LTO_.

ntamas (approved):

ntamas:
Thanks!

PR #2426, opened 2023-11-04
feat: new internal macro for rate-limited interruption checks

szhorvat:
This is a suggestion to add a macro that allows limiting the number of interruption checks, and thus increasing performance.

### DIFF
--- a/src/constructors/regular.c
+++ b/src/constructors/regular.c
@@ -362,6 +362,7 @@ igraph_error_t igraph_square_lattice(
     igraph_integer_t *coords, *weights;
     igraph_integer_t i, j;
     int carry, pos;
+    int iter;
 
     if (igraph_vector_int_any_smaller(dimvector, 0)) {
         IGRAPH_ERROR("Invalid dimension vector.", IGRAPH_EINVAL);
@@ -410,7 +411,7 @@ igraph_error_t igraph_square_lattice(
 #define IS_PERIODIC(dim) ((periodic && VECTOR(*periodic)[dim]))
 
     for (i = 0; i < no_of_nodes; i++) {
-        IGRAPH_ALLOW_INTERRUPTION();
+        IGRAPH_ALLOW_INTERRUPTION_LIMITED(iter, 1 << 10);
 
         /* Connect the current node to the "next" node along each dimension */
         for (j = 0; j < dims; j++) {
--- a/src/core/interruption.h
+++ b/src/core/interruption.h
@@ -52,6 +52,30 @@ extern IGRAPH_THREAD_LOCAL igraph_interruption_handler_t *igraph_i_interruption_
         } \
     } while (0)
 
+/**
+ * \define IGRAPH_ALLOW_INTERRUPTION_LIMITED
+ *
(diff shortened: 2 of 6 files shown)


src/constructors/regular.c
```diff
@@ -410,7 +411,7 @@ igraph_error_t igraph_square_lattice(
 #define IS_PERIODIC(dim) ((periodic && VECTOR(*periodic)[dim]))
 
     for (i = 0; i < no_of_nodes; i++) {
-        IGRAPH_ALLOW_INTERRUPTION();
+        IGRAPH_ALLOW_INTERRUPTION_LIMITED(iter, 1 << 10);
```
ntamas, line 414: Can't we use `i` here directly?

src/core/interruption.h
```diff
@@ -52,6 +52,21 @@ extern IGRAPH_THREAD_LOCAL igraph_interruption_handler_t *igraph_i_interruption_
         } \
     } while (0)
 
+/**
+ * \define IGRAPH_ALLOW_INTERRUPTION_LIMITED
+ *
+ * This is a variant of IGRAPH_ALLOW_INTERRUPTION() that checks for interruption
+ * only on every 'skips' call. It uses the 'iter' variable, which must be declared
+ * separately and initialized to 0, to keep track of the number of iterations.
```
ntamas, line 60: If we enforce the iteration tracker variable to be named as `iter` then there is no need to pass it as an argument to the macro. I'd suggest calling it `__iter` in the _macro_ and keeping the first argument of the macro so the user can choose whichever variable they want in the code to be used as an iteration variable.

src/paths/simple_paths.c
```diff
@@ -152,10 +148,7 @@ igraph_error_t igraph_get_all_simple_paths(const igraph_t *graph,
             VECTOR(nptr)[up] = 0;
         }
 
-        iteration++;
-        if (iteration >= 10000) {
-            iteration = 0;
-        }
+        IGRAPH_ALLOW_INTERRUPTION_LIMITED(iter, 1 << 13);
```
ntamas, line 151: I guess `1 << 13` is okay but the original code used `10000` so I wonder whether this is a typo.

src/constructors/regular.c
szhorvat: If we do that, we need to use `i % skips == 0`. As I recall, `%` was slow, or perhaps not as amenable to compiler optimization across the loop. But I did not test this now.

src/core/interruption.h
szhorvat: Sorry, I forgot the update the comment after I made the variable name a parameter.

src/constructors/regular.c
ntamas: Good point!

src/core/interruption.h
ntamas: Okay, no worries. If you fix the comments, then we're done, feel free to merge then. Let me know when you think it's time for a new C patch release.

ntamas (approved):

src/core/interruption.h
szhorvat: @ntamas See chat about C patch release.

PR #2433, opened 2023-11-11
refactor: general cleanup and const correctness for plfit

szhorvat:
I'd like to get rid of more static analyser warnings in plfit, and this helps see what is happening. Types of changes:

 - const correctness
 - match types where appropriate (i.e. `int` vs `size_t`)
 - readability (e.g. `0` vs `NULL`)
 - remove unused headers
 - use full precision (e.g. `0.1666666666666` -> `1.0 / 6.0`)

UPDATE:

 - Also fixed some memory errors, see 3rd commit

### DIFF
--- a/vendor/plfit/lbfgs.c
+++ b/vendor/plfit/lbfgs.c
@@ -70,7 +70,6 @@ licence.
 #endif
 
 #include <stdio.h>
-#include <stdlib.h>
 #include <math.h>
 
 #include "lbfgs.h"
--- a/vendor/plfit/mt.c
+++ b/vendor/plfit/mt.c
@@ -11,8 +11,6 @@
  * This file has been placed in the public domain.
  */
 
-#include <stdlib.h>
-
 #include "igraph_random.h"
 
 #include "plfit_mt.h"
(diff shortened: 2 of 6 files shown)


szhorvat:
@ntamas Can you also update the changelog for this? I'm not sure if you want the memory-leak-on-error fix to go in the igraph changelog or plfit changelog.

ntamas:
Can you send these changes to https://github.com/ntamas/plfit instead, and then I'll copy the changes from there? I'd like to keep the two versions as close as possible.

Re the memory leak fix: there's a `cleanup` label at the end of that function if I remember correctly so you chould just bail out by setting `retval` to the appropriate error code and the cleanup part takes care of freeing everything. This way there would be no need to duplicate the `free()` calls in the middle of the function (and that's why I've had the `if (...)` guards around the `free()` calls to cater for cases when we bail out before allocating one of the pointers to be freed there).

ntamas:
Of course if you don't have time for sending these to the upstream `plfit` repo, let me know and I'll do it, I just wanted to make sure that you get credit for your work on `plfit`.

szhorvat:
> Can you send these changes to https://github.com/ntamas/plfit instead

OK



> Re the memory leak fix: there's a `cleanup` label at the end of that function if I remember correctly so you chould just bail out by setting `retval` to the appropriate error code and the cleanup part takes care of freeing everything

I saw that, but then an error message is not issued along the error code. Are you sure you want it that way? Other instances of `goto cleanup` are where catching a non-success error code from another function. Presumably, that other function does report the error.

szhorvat:
> and that's why I've had the `if (...)` guards around the `free()` calls to cater for cases when we bail out before allocating one of the pointers to be freed there

`free` is safe to call on `NULL` pointers, so it's a stylistic choice

szhorvat:
See https://github.com/ntamas/plfit/pull/43

ntamas (approved):

PR #2434, opened 2023-11-11
Some cleanup in the spinglass community detection code

szhorvat:
This PR aggressively removes unused code from spinglass community detection, and makes the code a bit friendlier to static analysers.

In the past I was against removing parts of vendored code like this, partly to make sure we an understand it in the future. But no one else maintains this code, and we won't add new features to it.

This PR eliminates some Coverity warnings (and thanks to reduced code complexity, it enables Coverity to find some new ones)


ntamas:
Still a draft or can this be merged?

szhorvat:
It can be merged. I added a test to check that the output is not changing.

But there is something that is bothering me. The output is not invariant to a linear rescaling of weigths, not even an integer rescaling. Is that normal @vtraag ?  But that was already so before this PR.

szhorvat:
I see this comment in the sources:

> //sollen doppelte Links erlaubt sein??  NEIN

Should we check for multigraphs and forbid them?

szhorvat:
This is such a mess. Do we have anyone who understands this code? @vtraag ?

Does the following mean that temperatures may be negative? What's that absolute value about? It seems to me that the temperature should be positive.

https://github.com/igraph/igraph/blob/16c36d036aeb818e9d20a2b5464f8d393b6d3cc2/src/community/spinglass/clustertool.cpp#L257-L260

There is a lot of code dealing with the "zero temperature" case, but the docs are silent on what this means. Does it means that the start and stop temperatures are auto-determined?  If I set that value, things go wrong, NaNs are generated, and the program crashes with assertion failures.

https://github.com/igraph/igraph/blob/16c36d036aeb818e9d20a2b5464f8d393b6d3cc2/src/community/spinglass/clustertool.cpp#L304-L313

szhorvat:
I think I managed to clean this up and avoid the crash. It seems to me that the intention is that either `starttemp > stoptemp > 0` or `starttemp == stoptemp == 0`.

szhorvat:
@ntamas I had enough of this and won't touch it anymore. Please merge.

According to static analysers, there are some invalid memory access bugs in the DLList data structure here, but they never get triggered, and probably can't be triggered at all. The code is too messy for me to understand what exactly is going on. This bug is not due to this PR, but this PR cleans up things enough that there's a chance of fixing them now. I won't do that at this time though.

ntamas:
Thanks a lot!

PR #2436, opened 2023-11-13
Simplify vertex_disjoint_paths() and add more tests for it

szhorvat:
UPDATE: This PR was changed to merely add tests and simplify code. It no longer adds support for using the same vertex for source and target.

----

As the title says, vertex/edge disjoint path calculations now support the source=target case. In this case we simply count the number of self-loops.

We do not need to merge this for 0.10.8! It's good to think about this carefully, and whether it makes sense to do this at all. An alternative way to think about this is to say that we have already arrived to the destination before traversing any edge, and in fact traversing the self-loop makes this a cycle and not a path. Thus we can return 1, saying that the only path is the "empty path" (similar to how we return an empty path for the shortest path between a vertex and itself).

### DIFF
--- a/src/flow/flow.c
+++ b/src/flow/flow.c
@@ -1898,7 +1898,7 @@ static igraph_error_t igraph_i_st_vertex_connectivity_undirected(const igraph_t 
  *     \c IGRAPH_VCONN_NEI_IGNORE, ignore the fact that the two vertices
  *        are connected and calculate the number of vertices needed
  *        to eliminate all paths except for the trivial (direct) paths
- *        between \p source and \p vertex. TODO: what about neighbors?
+ *        between \p source and \p vertex.
  * \return Error code.
  *
  * Time complexity: O(|V|^3), but see the discussion at \ref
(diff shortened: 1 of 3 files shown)


ntamas:
> Does anyone know what this TODO is about?

Nope. I think we can get rid of it.

> Why does igraph_vertex_disjoint_paths() delete edges between source and target if it is already using IGRAPH_VCONN_NEI_IGNORE?

Seems like this is not needed. I went back in the history and this part of the code was essentially unchanged since 2006. I think that if our current test cases cover this part of the code, then we can safely remove the part where we remove  the edges.

ntamas (approved):

ntamas:
@szhorvat take a look at my modifications and feel free to merge if it's OK

szhorvat:
I think counting self-loops is not good here. If we count self-loops, we should also count longer cycles, which then becomes complicated.

szhorvat:
This requires much more careful thought. I will remove the feature for using the same source/target from this PR, and will keep the updated tests as well as the changes you made.

PR #2439, opened 2023-11-18
Use correct `arpack_defaults()` default

krlmlr:
I wonder if the C core is really the right place for this configuration.

This would have to be forward-ported to the main branch too.

### DIFF
--- a/interfaces/functions.yaml
+++ b/interfaces/functions.yaml
@@ -647,35 +647,35 @@ igraph_eigenvector_centrality:
         GRAPH graph, OUT VERTEXINDEX vector, OUT REAL value,
         BOOLEAN directed=False, BOOLEAN scale=True,
         EDGEWEIGHTS weights=NULL,
-        INOUT ARPACKOPT options=arpack_defaults
+        INOUT ARPACKOPT options=arpack_defaults()
     DEPS: weights ON graph, vector ON graph
 
 igraph_hub_score:
     PARAMS: |-
         GRAPH graph, OUT VERTEXINDEX vector, OUT REAL value,
         BOOLEAN scale=True, EDGEWEIGHTS weights=NULL,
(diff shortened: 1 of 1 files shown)


szhorvat:
For now, let's do whatever is necessary to get the R release out. The `release/0.9` branch exists mostly for the benefit of R/igraph anyway.

In the longer term, there should be a proper solution that doesn't involve writing R-specific expressions, and can be handled from multiple languages.

krlmlr:
This should also find its way into the mainline because https://github.com/igraph/rigraph/pull/840 relies on it.

szhorvat:
Yes, and there it may interfere with the new Python interface, so let's wait for @ntamas.

In the long term we'll definitely need something more structured. We'll need to avoid writing R code or Python code in `functions.yaml`, or just relying on R and Python code looking the same. Mathematica code will definitely not look the same: underscores are not allowed in function names and function calls don't use parentheses.

But for the moment, the priority is to get the new R release out, not to have a perfect solution here.

szhorvat:
Eventually, on the R side, `arpack_defaults()` should probably rely on `igraph_arpack_options_init()`, to make sure that different interfaces use the same default parameters. That's probably a bit messy to do at the moment.

ntamas:
> In the longer term, there should be a proper solution that doesn't involve writing R-specific expressions,

`stimulus` treats the default values in an abstract manner and you can map them to real target-language-specific constructs in `types.yaml` like this:

```yaml
ARPACK_OPTIONS:
    DEFAULT:
        some_abstract_value: arpack_options()
```

So we can use a language-agnostic construct like `default_arpack_options` and then map it to whatever the target language supports in a language-specific `types.yaml` file.

szhorvat:
No longer needed, this was handled according to Tamás's suggestion.

PR #2440, opened 2023-11-19
HRG code cleanup

szhorvat:
TODO before merging:

 - [ ] Tests
 - [ ] Refactor `inters::getRandomEdge()` to use `RNG_INTEGER`. This must be the last step as this changes the output.

Things to investigate:

 - [x] `splittree::returnArrayOfKeys()` is assigning `-1` to a string. *This is present in the original consensusHRG code. Commented out for now. Nothing seems to have broken.*
 - [x] In `dendro::cullSplitHist()`, `int tot` is assigned a real value. *Also present in original code. Changed the type to `double`. Nothing seems to change in the test output.*


ntamas:
LGTM so far. Let me know when / if you need a review.

szhorvat:
Not yet. We need tests first. I'm pretty sure I broke something.

szhorvat:
@ntamas I'm not sure this is reviewable at all, but have a look. The approach was to fix a lot of little things that static analysers are unhappy with, which then improved the ability of these analysers to find more significant issues.

Some of the tests had notes that the output is not printed because it would be different on i386 due to numerical roundoff errors. I checked that this PR does not change the output if we do print it, but I did not enable printing in the tests.

I think I'd like to merge this soon.

ntamas:
Went through the PR in the last hour or so and couldn't spot any obvious mistakes or omissions. I think that if the CI runs pass then we can merge this.

szhorvat:
OK, I will merge this soon, but I am going to squash all the commits, as this is a mess. Is that fine? It would make a rather large commit.

szhorvat:
Or I could squash just part of it to reduce the number of commits, but give us a  better chance to discover what went wrong if issues are discovered in the future.

ntamas:
Squash all that you want, I'm not particularly picky about keeping the history accurate.

Issue #2309, opened 2023-02-13
Warning with recent GCC and LTO enabled


szhorvat:
When compiling with recent GCC and LTO enabled, there are some warnings mostly "exceeds maximum object size". This issue is for investigating whether these are harmful, and whether they are fixable. CRAN may eventually start requiring a fix for these.

To reproduce, compile with `-DBUILD_SHARED_LIBS=ON -DIGRAPH_ENABLE_LTO=ON` and use GCC 12. It is easiest to do this on a system where GCC is the default compiler, such as Linux. Getting LTO to work with GCC on macOS with MacPorts / Homebrew is more involved.

The following are as of 925f5243fc0c8dd58fc34cfbe01ca4480f042355

```
In member function ‘convert_from_igraph’,
    inlined from ‘igraph_community_walktrap’ at /home/szhorvat/Repos/igraph/src/community/walktrap/walktrap.cpp:191:5:
/home/szhorvat/Repos/igraph/src/community/walktrap/walktrap_graph.cpp:168:42: warning: argument 1 value ‘1844674407370955161’ exceeds maximum object size 9223372036854775807 [-Walloc-size-larger-than=]
  168 |     G.vertices = new Vertex[G.nb_vertices];
      |                                          ^
/usr/local/gcc/gcc-12.2.0/include/c++/12.2.0/new: In function ‘igraph_community_walktrap’:
/usr/local/gcc/gcc-12.2.0/include/c++/12.2.0/new:128:26: note: in a call to allocation function ‘operator new []’ declared here
  128 | _GLIBCXX_NODISCARD void* operator new[](std::size_t) _GLIBCXX_THROW (std::bad_alloc)
      |                          ^
In member function ‘__ct ’,
    inlined from ‘igraph_community_walktrap’ at /home/szhorvat/Repos/igraph/src/community/walktrap/walktrap.cpp:191:5:
/home/szhorvat/Repos/igraph/src/community/walktrap/walktrap_communities.cpp:399:51: warning: argument 1 value ‘18446744073709551615’ exceeds maximum object size 9223372036854775807 [-Walloc-size-larger-than=]
  399 |     communities = new Community[2 * G->nb_vertices];
      |                                                   ^
/usr/local/gcc/gcc-12.2.0/include/c++/12.2.0/new: In function ‘igraph_community_walktrap’:
/usr/local/gcc/gcc-12.2.0/include/c++/12.2.0/new:128:26: note: in a call to allocation function ‘operator new []’ declared here
[...]

PR #2441, opened 2023-11-22
Work around false-positive GCC warnings when using LTO

szhorvat:
This PR adds an IGRAPH_ASSUME macro and uses it, along with IGRAPH_ASSERT, to work around false-positive GCC warnings when compiling with LTO. See #2309.

### DIFF
--- a/include/igraph_decls.h
+++ b/include/igraph_decls.h
@@ -11,13 +11,6 @@
 /* This is to eliminate gcc warnings about unused parameters */
 #define IGRAPH_UNUSED(x) (void)(x)
 
-/* Include the definition of macros controlling symbol visibility */
-#include "igraph_export.h"
-
-/* Used instead of IGRAPH_EXPORT with functions that need to be tested,
- * but are not part of the public API. */
-#define IGRAPH_PRIVATE_EXPORT IGRAPH_EXPORT
-
 /* The pure function attribute of GCC-compatible compilers indicates
  * that the function does not have side-effects, i.e. it does not
  * modify global memory. This enables additional compiler optimizations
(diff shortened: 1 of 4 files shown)


krlmlr (approved):
Nice, looks good to me. What's the advantage of assume over assert -- saving a few cycles?

szhorvat:
> Nice, looks good to me. What's the advantage of assume over assert -- saving a few cycles?

You're right, assert would have been fine. I went with assert in several places where I have less confidence in the code. This is all borrowed code, not originally written for igraph, and not always as robust as our own.

But perhaps `IGRAPH_ASSUME` will be nice to use in some situations. It's not bad to have it.

szhorvat:
`__builtin_assume()` is mentioned in the Clang 3.6 docs, not sure if it was added earlier. https://releases.llvm.org/3.6.0/tools/clang/docs/LanguageExtensions.html Do we want to support older Clang? This is 2015.

`__builtin_unreachable()` was added in GCC 4.5 in 2010. This is definitely old enough. https://gcc.gnu.org/gcc-4.5/changes.html

szhorvat:
Compiler compatibility should be good now. It no longer requires Clang 3.6.

I was unable to find anything in igraph's code that could be sped up with the use of IGRAPH_ASSUME and doesn't already have IGRAPH_ASSERT. Most likely, IGRAPH_ASSERT is good enough in most cases.

I'd still like to add IGRAPH_ASSUME so that it would be easy to experiment with it in the future.

szhorvat:
`IGRAPH_ASSERT` is _always_ checked, not just in debug mode, but there's no noticeable performance degradation from this. If it were disabled, it would no longer help the compiler do optimizations. This is the case with `assert()`, which does get disabled in release mode.

`IGRAPH_ASSUME` is never checked, but it communicates the same information to the compiler as an assertion would, and allows better optimization.  It is not valid to put things with potential side effects (e.g. function calls) in `IGRAPH_ASSUME`. Clang will throw an error.

PR #2442, opened 2023-12-01
Fix compiler error with libxml2 2.12 and clang

Biswa96:
This commit fixes the following compiler error.

    graphml.c:1613:39: error: incompatible function pointer types passing
    'void (*)(void *, xmlErrorPtr)' (aka 'void (*)(void *, struct _xmlError *)') to
    parameter of type 'xmlStructuredErrorFunc' (aka 'void (*)(void *, const struct _xmlError *)')
     [-Wincompatible-function-pointer-types]
     1613 |     xmlSetStructuredErrorFunc(&state, &igraph_i_libxml_structured_error_handler);
          |                                       ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    libxml2/libxml/xmlerror.h:898:29: note: passing argument to parameter 'handler' here
      898 |                                  xmlStructuredErrorFunc handler);
          |                                                         ^

    According to the libxml2 changelog, this is due to a change in version 2.12:
    https://download.gnome.org/sources/libxml2/2.12/libxml2-2.12.0.news

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
  - More robust error handling in HRG code.
  - Fixed infinite loop in `igraph_hrg_sample_many()`.
+ - Compatibility with libxml2 version 2.12 and above.
 
 ### Other
 
--- a/src/io/graphml.c
+++ b/src/io/graphml.c
@@ -344,7 +344,11 @@ static void igraph_i_graphml_parser_state_set_error_from_varargs(
 }
 
 static void igraph_i_graphml_parser_state_set_error_from_xmlerror(
+#if LIBXML_VERSION >= 21200
+    struct igraph_i_graphml_parser_state *state, const xmlError *error
+#else
     struct igraph_i_graphml_parser_state *state, const xmlErrorPtr error
+#endif
 ) {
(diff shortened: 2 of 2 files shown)


Biswa96:
It seems that the compiler error happens with newer libxml2 versions. I shall add the libxml2 version macro in the condition.

szhorvat:
Thanks for the report!

According to the libxml2 changelog, this is due to a change in version 2.12:

https://download.gnome.org/sources/libxml2/2.12/libxml2-2.12.0.news

I'm with limited internet connection at the moment and it is difficult for me to download libxml2 to test. As you say, there should be a macro guard for the version. I believe this should be:

```
#if LIBXML_VERSION < 21200
// old version
#else
// new version
#endif
```

Could you please use this and test that it works fine? Perhaps we should have an appropriate typedef to deal with this.

Could you please also update CHANGELOG.md and add something along the lines of "Compatibility with libxml2 2.12" under "Fixed"?

Biswa96:
Thank you for the review. I have added necessary conditions to be compatible with older libxml2 versions.

ntamas (approved):

Issue #2444, opened 2023-12-05
STR() vs strvector_get()


szhorvat:
There are two ways to access elements of string vectors: `STR` and `igraph_strvector_get()`. At the moment, both are used in the library, with `igraph_strvector_get()` being more common.

This issue is to discuss whether it would be worth getting rid of `STR` entirely.

Reasons for keeping `STR`:

 - Theoretically faster access to elements, similar to how `VECTOR` is preferred to `igraph_vector_get()`.

Reasons for dropping `STR`:

 - Provides raw access to the string vector internals, for example one can do `STR(v,i) = "asd"`. A user may assume that this is valid, since it _is_ valid with `VECTOR`. But in fact with `STR` it is illegal.
 - Faster access is likely not nearly as beneficial as with `VECTOR`, since strings are heavier objects anyway
 - An opaque interface like `igraph_strvector_get()` provides optimization opportunities. For example, internally `""` could be represented as `NULL`, saving a lot of `calloc()` calls when initializing an string vector. This would speed up string vector initialization significantly, and will speed up some file format readers in cases when only a few elements have an associated non-empty string. It would be a major win for fuzzing efficiency, in particular.
 - Some complained that `STR` is too collision-prone as a macro name, #1001

szhorvat:
I propose that we deprecate `STR` in 0.10.9 and define the macro in terms of `igraph_strvector_get()`. Then we remove `STR` entirely in 0.11 / 1.0. This would not cause major inconvenience for users since they can define `STR` in terms of `igraph_strvector_get()` easily, or just do a find-and-replace.

ntamas:
Agree with the deprecation. I don't think it has been used much outside our codebase and it is indeed dangerous.
On the other hand: `STR` is a macro so I wonder whether there's an easy way to deprecate it that also triggers a compilation warning.

I'll take a look at your PR.

PR #2445, opened 2023-12-05
refactor: strvector: represent empty strings as NULL

szhorvat:
Proof of concept for #2444

### DIFF
--- a/include/igraph_strvector.h
+++ b/include/igraph_strvector.h
@@ -36,6 +36,7 @@ __BEGIN_DECLS
  */
 
 typedef struct s_igraph_strvector {
+    /* Empty strings "" are represeted using NULL. */
     char **stor_begin;
     char **stor_end;
     char **end;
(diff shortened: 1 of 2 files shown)


include/igraph_strvector.h
```diff
@@ -36,6 +36,7 @@ __BEGIN_DECLS
  */
 
 typedef struct s_igraph_strvector {
+    /* Empty strings "" are represeted using NULL. */
```
ntamas, line 39:
```suggestion
    /* Empty strings "" are represented using NULL. */
```

include/igraph_strvector.h
```diff
@@ -56,7 +57,7 @@ typedef struct s_igraph_strvector {
  *
  * Time complexity: O(1).
  */
-#define STR(sv,i) ((const char *)((sv).stor_begin[(i)]))
+#define STR(sv,i) igraph_strvector_get(&sv, i)
```
ntamas, line 60: Is this a breaking change? This can surely not be used as an lvalue any more, but maybe the earlier version was not suitable for an lvalue either due to casting?

szhorvat: You are right, this is not a breaking change.

szhorvat:
I'm going to merge this and see how it goes. If some major issues come to light, we can always revert.

PR #2451, opened 2023-12-13
Add fast label propagation

vtraag:
This PR adds the fast label propagation algorithm as proposed by @lovre and me in:

Traag, V. A., & Šubelj, L. (2023). Large network community detection by fast label propagation. Scientific Reports, 13(1) https://doi.org/10.1038/s41598-023-29610-z https://arxiv.org/abs/2209.13338

It is a cleaned up version of an earlier branch in https://github.com/vtraag/igraph/tree/feature/flpa, which contains the code we used for the publication. Several changes to the original LPA implementation required a substantial clean-up, most notably the integer transition & cleanup.

In addition to the fast variant, it also contains the implementation of an alternative strategy we labelled "retention" in our paper.

I believe it should be all correct, but I still need to check some things. Everything does compile correctly, but there are some tests that fail. Perhaps it's just a matter of updating the output to the algorithm. But I don't have the time right now to look into this, I'll try to do this asap.

### DIFF
--- a/interfaces/types.yaml
+++ b/interfaces/types.yaml
@@ -492,6 +492,11 @@ WHEEL_MODE:
     CTYPE: igraph_wheel_mode_t
     FLAGS: ENUM
 
+LPA_VARIANT:
+    # Enum that describes how a tree graph should be constructed
+    CTYPE: igraph_lpa_variant_t
+    FLAGS: ENUM
+
 ###############################################################################
 # Switches / flags / bits
 ###############################################################################
--- a/src/community/label_propagation.c
+++ b/src/community/label_propagation.c
@@ -27,7 +27,7 @@
 #include "igraph_random.h"
 
 int igraph_i_community_label_propagation(const igraph_t *graph,
-    igraph_vector_t *membership,
+    igraph_vector_int_t *membership,
     igraph_neimode_t mode,
     const igraph_vector_t *weights,
     igraph_vector_bool_t *fixed)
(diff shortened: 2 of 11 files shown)


vtraag:
I believe it is now all in order. There were some minor issues with documentation and Stimulus name arguments.

There was still one major error as a result of putting a brace somewhere incorrectly. In addition, the presence of unlabelled nodes presented some problems. Both are now also fixed.

I believe all CI now pass, only the arm64 and ppc64 build still need to pass, but they take forever.

tests/unit/community_label_propagation.c
```diff
@@ -33,19 +33,24 @@ int main(void) {
...
     igraph_vector_bool_t fixed;
     igraph_integer_t i, j;
+    igraph_lpa_variant_t variant;
 
     /* Simple triangle graph, the output should be always one community */
     igraph_small(&g, 0, IGRAPH_UNDIRECTED, 0,  1,  0,  2,  1,  2, -1);
     igraph_vector_int_init(&membership, 0);
 
-    for (j = 0; j < 100; j++) {
+    for (variant = 0; variant < 3; variant++)
```
szhorvat, line 42: Can you please refactor this to use explicit enum constants? For example, store the constants in an array and loop through that array.

Also, can you keep the opening brace after `for` on the same line, to keep with the existing style?

tests/unit/null_communities.c
```diff
@@ -85,7 +85,7 @@ int main(void) {
 
     igraph_vector_int_resize(&membership, 1);
 
-    igraph_community_label_propagation(&g, &membership, IGRAPH_ALL, NULL, NULL, NULL);
+    igraph_community_label_propagation(&g, &membership, IGRAPH_ALL, NULL, NULL, NULL, IGRAPH_LPA_DOMINANCE);
```
szhorvat, line 88: Are you sure we don't need extra lines for all other variants here? It would be simple to add them. From my experience, proper null graph handling tends to break easily when refactoring code.

src/community/label_propagation.c
```diff
@@ -103,11 +503,14 @@
...
- *   \em directed modularity if the input graph is directed, even if
- *   you set \p mode to \c IGRAPH_ALL
+ * \param lpa_variant Which variant of label propagation algorithm to run.
+ *   One of
+ *   \c IGRAPH_LPA_DOMINANCE Check for dominance of all nodes after each
+ *                           iteration.
+ *   \c IGRAPH_LPA_RETENTION Keep current label if among dominant labels,
+ *                           only check if labels changed.
+ *   \c IGRAPH_LPA_FAST      Sample from dominant labels, only check
+ *                           neighbors.
```
szhorvat, line 513: Can you elaborate a bit more on what these mean in the main text? Make sure to define "dominant label" explicitly.

src/community/label_propagation.c
```diff
@@ -213,163 +609,35 @@ igraph_error_t igraph_community_label_propagation(const igraph_t *graph,
...
-        } else {
-            /* Shuffle the node ordering vector if we are in the label updating iteration */
-            igraph_vector_int_shuffle(&node_order);
-        }
-
-        RNG_BEGIN();
-        /* In the prescribed order, loop over the vertices and reassign labels */
-        for (i = 0; i < no_of_not_fixed_nodes; i++) {
-            v1 = VECTOR(node_order)[i];
+    switch(lpa_variant)
```
szhorvat, line 619: Can you run this file through astyle and add an extra commit with the result of that formatting?

src/community/label_propagation.c
```diff
@@ -213,163 +609,35 @@ igraph_error_t igraph_community_label_propagation(const igraph_t *graph,
...
-                    }
-                }
-                else {
-                    /* Select randomly from the dominant labels */
-                    k = RNG_INTEGER(0, igraph_vector_int_size(&dominant_labels) - 1);
-                    VECTOR(*membership)[v1] = VECTOR(dominant_labels)[k];
-                }
-            }
+      case IGRAPH_LPA_RETENTION:
+        IGRAPH_CHECK(igraph_i_community_label_propagation(graph, membership, mode, weights, fixed_copy, /* retention */ 1 ));
```
szhorvat, line 626: Please use `true` and `false` instead of `1` and `0` everywhere in this file. This does make a real difference for readability. After having changed so much of the existing code to a more readable form, let's make sure new code uses readable forms that encode the type everywhere.

Same applied to `0` vs `NULL` for pointers.

src/community/label_propagation.c
```diff
@@ -26,15 +26,410 @@
...
+    igraph_vector_bool_t *fixed) {
+
+    igraph_integer_t no_of_nodes = igraph_vcount(graph), no_of_not_fixed_nodes = 0;
+    igraph_integer_t i, j, k;
+    igraph_inclist_t il;
+    igraph_adjlist_t al;
+
+    igraph_vector_t label_counters;
+    igraph_vector_int_t dominant_labels, nonzero_labels, node_order;
+    igraph_dqueue_t queue;
```
szhorvat, line 267: Shouldn't this be an `igraph_dqueue_int_t`?

src/community/label_propagation.c
```diff
@@ -26,15 +26,410 @@
...
+    IGRAPH_VECTOR_INIT_FINALLY(&label_counters, no_of_nodes);
+    IGRAPH_VECTOR_INT_INIT_FINALLY(&dominant_labels, 0);
+    IGRAPH_VECTOR_INT_INIT_FINALLY(&nonzero_labels, 0);
+    IGRAPH_CHECK(igraph_vector_int_reserve(&dominant_labels, 2));
+
+    RNG_BEGIN();
+
+    /* Initialize node ordering vector with only the not fixed nodes */
+    IGRAPH_CHECK(igraph_dqueue_init(&queue, no_of_nodes));
+    IGRAPH_FINALLY(igraph_dqueue_destroy, &queue);
```
szhorvat, line 291: Please use `IGRAPH_DQUEUE_INT_INIT_FINALLY` (after also changing the queue to store integers).

src/community/label_propagation.c
```diff
@@ -26,15 +26,410 @@
...
+    igraph_vector_t label_counters;
+    igraph_vector_int_t dominant_labels, nonzero_labels, node_order;
+    igraph_dqueue_t queue;
+    igraph_vector_bool_t in_queue;
+    igraph_neimode_t reverse_mode;
+
+    reverse_mode = IGRAPH_REVERSE_MODE(mode);
+
+    if (weights) {
+        IGRAPH_CHECK(igraph_inclist_init(graph, &il, reverse_mode, IGRAPH_LOOPS_ONCE));
```
szhorvat, line 274: I know that the `LOOPS_ONCE` choice was not made in this PR, but do we have a reasoning for using it?

Almost everywhere else, we use `LOOPS_TWICE`, unless there is a clear reason not to. In order for igraph to behave predictably, it makes sense to me to do the same here. Since your understanding of label propagation is fresh at the moment, @vtraag, do you see a reason to keep this `LOOPS_ONCE` on `develop`?

src/community/label_propagation.c
```diff
@@ -26,15 +26,410 @@
...
+    igraph_neimode_t mode,
+    const igraph_vector_t *weights,
+    igraph_vector_bool_t *fixed) {
+
+    igraph_integer_t no_of_nodes = igraph_vcount(graph), no_of_not_fixed_nodes = 0;
+    igraph_integer_t i, j, k;
+    igraph_inclist_t il;
+    igraph_adjlist_t al;
+
+    igraph_vector_t label_counters;
```
szhorvat, line 265: There was a comment on this in the original code explaining why this is of real type, and not integer (despite the confusing name): `/* real type, stores weight sums */`. Please restore this comment, since it seems to be valid for this algorithm variant as well.

szhorvat (changes requested):
GitHub is being blocked here again this morning, so I cannot check out the branch for a proper review. I can only access GitHub through the browser. I'm posting the partial review I have.

Currently the code checks for negative weights, but not zero weights. Do the new algorithm variants work properly with zero weights?

src/community/label_propagation.c
vtraag: Added.

src/community/label_propagation.c
vtraag: I think there's no strong argument for either. Self-loops in this case would matter for how strong the effect is on keeping the same label. For example, suppose we have a node with four neighbours and a self-loop, and two neighbours have label A and two have label B, while the node's own label is label C. Then with `LOOPS_ONCE` label C would not be dominant (i.e. a maximal label), because label C is counted only once with a self-loop, while with `LOOPS_TWICE` is would be. I'm not sure what the original paper suggested (if anything), but to me `LOOPS_ONCE` actually seems more reasonable than `LOOPS_TWICE` here.

src/community/label_propagation.c
vtraag: Yes, indeed, good catch, thanks!

src/community/label_propagation.c
vtraag: Yes, good point. This was indeed mostly a remnant of old code.

src/community/label_propagation.c
vtraag: Sure, I've added an explanation.

tests/unit/community_label_propagation.c
vtraag: Yes, that's indeed cleaner.

Change the style while at it. Doesn't astyle pick up on this btw?

tests/unit/null_communities.c
vtraag: Sure.

src/community/label_propagation.c
vtraag: Done, but barely made any changes.

vtraag: Done, but barely made a difference.

vtraag:
> Currently the code checks for negative weights, but not zero weights. Do the new algorithm variants work properly with zero weights?

Good point. I briefly thought it would be a problem for both old and new lagorithms, but it's actually not a problem for either.

szhorvat:
@vtraag I don't have a full understanding of what's going on here, but it does not look right to me that on line 724:

https://github.com/igraph/igraph/blob/15441b575d34f1596b6438802270e1853351480e/src/community/label_propagation.c#L724

the iteration goes to `no_of_nodes`.  The vector `node_order` has length `no_of_not_fixed_nodes`, not `no_of_nodes`.

Can you check if this is buggy, and if yes, can you add a test that would have caught it?

src/community/label_propagation.c
```diff
@@ -213,163 +639,35 @@ igraph_error_t igraph_community_label_propagation(const igraph_t *graph,
...
-            if (!VECTOR(*fixed_copy)[i]) {
-                VECTOR(node_order)[j] = i;
-                j++;
-            }
-        }
-    } else {
-        IGRAPH_CHECK(igraph_vector_int_init_range(&node_order, 0, no_of_nodes));
-        IGRAPH_FINALLY(igraph_vector_int_destroy, &node_order);
+    case IGRAPH_LPA_DOMINANCE:
+    default:
```
szhorvat, line 660: The `default` label should be separate and throw an error (incorrect LPA variant).

src/community/label_propagation.c
szhorvat: > but barely made a difference

I believe this is because you did not use our astyle settings. Use the `.astylerc` in the top level directory of the repo. This should do it:

```
astyle --project=.astylerc src/community/label_propagation.c
```

Do you want me to do it and push to this branch (I hope I can manage to do that)?

vtraag:
I'll look into it later. That code came from the existing code on the dev branch I believe. If I recall correctly you contributed it. It might be that something was incorrectly resolved when going over all conflicts when rebasing.

src/community/label_propagation.c
vtraag: Sure, we can do that.

vtraag:
> I don't have a full understanding of what's going on here, but it does not look right to me that on line 724:
> 
> https://github.com/igraph/igraph/blob/15441b575d34f1596b6438802270e1853351480e/src/community/label_propagation.c#L724-L725
> 
> the iteration goes to `no_of_nodes`. The vector `node_order` has length `no_of_not_fixed_nodes`, not `no_of_nodes`.

You were correct to see that this needed to be correct. This already was an existing bug on `develop`, and was already previously buggy. I've now corrected this in the last commit in this PR (3995409339a5a7362208c5d3cd9b3a73a41aa861), but I suggest not to add a specific test for it.

szhorvat:
> This already was an existing bug on `develop`

Can you check if it exists on `master`? I don't think so, but please have a look.

interfaces/types.yaml
```diff
@@ -492,6 +492,11 @@ WHEEL_MODE:
     CTYPE: igraph_wheel_mode_t
     FLAGS: ENUM
 
+LPA_VARIANT:
+    # Enum that describes how a tree graph should be constructed
```
szhorvat, line 496: Please fix the description

szhorvat:
After this is merged (or even within this PR) we should rename all instances `label_counters` to `label_weights`. The name is too confusing, as it suggests an integer type. In fact this variable was once accidentally changed to integer, which caused a bug.

src/community/label_propagation.c
```diff
@@ -213,163 +636,37 @@ igraph_error_t igraph_community_label_propagation(const igraph_t *graph,
...
-    if (weights) {
-        igraph_inclist_destroy(&il);
-    } else {
-        igraph_adjlist_destroy(&al);
-    }
-    IGRAPH_FINALLY_CLEAN(1);
 
-    /* Shift back the membership vector, permute labels in increasing order */
-    /* We recycle label_counters here :) and use it as an integer vector from now on */
+    /* Permute labels in increasing order */
```
szhorvat, line 664: Can you please restore the comment that explains that `label_counters` is used for an unrelated purpose here?

```
/* We recycle label_counters here :) and use it as an integer vector from now on */
```

It's quite confusing otherwise.

In fact it might be better to just eliminate any re-use and create a separate integer vector for this purpose.

vtraag: I've remove the `label_counters` vector, it wasn't used anywhere, and replaced it by an int vector.

interfaces/types.yaml
vtraag: Fixed.

vtraag:
> we should rename all instances `label_counters` to `label_weights`

Done.

vtraag:
> Check if that bug is present on master

Fixed on `master` in 69f69e97e739c58e8e01474e8b3f46416dee1f93.

vtraag:
> * Then squash-and-merge at your own convenience?

I think it would be cleaner to neatly rebase and then merge. The different commits may make it a bit easier to track and understand what happened.

vtraag:
It's now all neatly rebased, let me know if you're OK with simply merging this, instead of squash-and-merge @szhorvat .

szhorvat:
@vtraag Can you please resolve conflicts and rebase only latest develop (which should fix the doc build)?

I would suggest squash-and-merge unless you have a reason to preserve the commit history.

Please go ahead with merging when ready, I have no objections.

szhorvat:
A ping on this one. Let's merge it soon!

szhorvat:
This function is also called in the `weighted_community` fuzz target

PR #2457, opened 2023-12-27
Add target to build documentation in info format

aagon:
Hello,

First of all, thank you for this very useful library.

I spend quite a lot of time perusing the documentation in PDF format, and I
thought it would be nice to have the possibility to read it in GNU Info format
as well.

Since you build the documentation in Docbook format in the first place, it is
quite trivial.

This PR adds an extra Cmake target to build the documentation in Info format,
and reuses parts of the commands used to build the PDF docs.

This target has two new dependencies :
- docbook2x, which provides the command docbook2x-texi, to convert from Docbook
  to texi.
- texinfo, which provides the command makeinfo, to build info from texi.

I have documented this new target and its dependencies.

I have also added it to the cmake summary.

This should require little to no effort on your part to maintain, and I can be
pinged to deal with any issues that might arise.

Thank you in advance for your time.

Best,

Aymeric Agon-Rambosson, a happy user of your library

### DIFF
--- a/doc/installation.xml
+++ b/doc/installation.xml
@@ -471,8 +471,8 @@ $ cmake --build . --target html
     Building the HTML documentation requires Python 3, <literal>xmlto</literal>
     and <literal>source-highlight</literal>. Building the PDF documentation also
     requires <literal>xsltproc</literal>, <literal>xmllint</literal> and
-    <literal>fop</literal>. Building the Info documentation also requires
-    <literal>docbook2x</literal>, <literal>xmllint</literal> and
+    <literal>fop</literal>. Building the Texinfo documentation also requires
+    the docbook2X package, <literal>xmllint</literal> and
     <literal>makeinfo</literal>.
   </para>
 </section>
--- a/doc/installation.xml
+++ b/doc/installation.xml
@@ -459,18 +459,21 @@ pacman -S \
     directory.
   </para>
   <para>
-    To build the documentation for the development version, simply build
-    the <literal>html</literal> or <literal>pdf</literal> targets for the
-    HTML and PDF versions of the documentation, respectively.
+    To build the documentation for the development version, simply build the
+    <literal>html</literal>, <literal>pdf</literal> or <literal>info</literal>
+    targets for the HTML, PDF and Info versions of the documentation,
+    respectively.
   </para>
 <programlisting>
 $ cmake --build . --target html
 </programlisting>
   <para>
     Building the HTML documentation requires Python 3, <literal>xmlto</literal>
-    and <literal>source-highlight</literal>. Building the PDF documentation
-    also requires <literal>xsltproc</literal>, <literal>xmllint</literal>
-    and <literal>fop</literal>.
+    and <literal>source-highlight</literal>. Building the PDF documentation also
+    requires <literal>xsltproc</literal>, <literal>xmllint</literal> and
+    <literal>fop</literal>. Building the Info documentation also requires
+    <literal>docbook2x</literal>, <literal>xmllint</literal> and
+    <literal>makeinfo</literal>.
   </para>
 </section>
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@
 ### Other
 
  - Reduced memory usage and improved initialization performance for `igraph_strvector_t`.
+ - The documentation is now also generated in Texinfo format.
  - Documentation improvements
 
 ## [0.10.8] - 2023-11-17
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -150,7 +150,7 @@ jobs:
       - script: sudo apt-get update
         displayName: Update package registry
 
-      - script: sudo apt-get install ninja-build xmlto texinfo source-highlight libxml2-utils xsltproc fop -y
+      - script: sudo apt-get install ninja-build xmlto texinfo source-highlight libxml2-utils xsltproc fop docbook2x -y
         displayName: Install dependencies
 
       - task: CMake@1
--- a/doc/CMakeLists.txt
+++ b/doc/CMakeLists.txt
@@ -95,6 +95,11 @@ find_program(FOP_COMMAND fop)
 find_program(XMLLINT_COMMAND xmllint)
 find_program(XSLTPROC_COMMAND xsltproc)
 
+# GNU info documentation additionally requires docbook2x and makeinfo
+# (and xmllint as well)
+find_program(DOCBOOK2XTEXI_COMMAND docbook2x-texi)
+find_program(MAKEINFO_COMMAND makeinfo)
+
 if(Python3_FOUND AND SOURCE_HIGHLIGHT_COMMAND)
   set(DOC_BUILD_SUPPORTED TRUE)
 else()
@@ -111,6 +116,12 @@ if(DOC_BUILD_SUPPORTED AND XMLLINT_COMMAND AND XSLTPROC_COMMAND AND FOP_COMMAND)
   set(PDF_DOC_BUILD_SUPPORTED TRUE)
 else()
   set(PDF_DOC_BUILD_SUPPORTED FALSE)
+endif()
+
+if(DOC_BUILD_SUPPORTED AND XMLLINT_COMMAND AND DOCBOOK2XTEXI_COMMAND AND MAKEINFO_COMMAND)
+  set(INFO_DOC_BUILD_SUPPORTED TRUE)
+else()
+  set(INFO_DOC_BUILD_SUPPORTED FALSE)
 endif()
 
 if(DOC_BUILD_SUPPORTED)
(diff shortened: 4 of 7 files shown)


ntamas:
Thank you for your contribution, it looks great! Let's see how it goes in CI and I'll merge if everything passes.

ntamas:
I'm getting these warnings in CI (but otherwise the build seems to succeed):

```
igraph-docs.texi:3: warning: encoding `utf8' is not a canonical texinfo encoding
igraph-docs.texi:3: warning: unrecognized encoding name `utf8'
```

@aagon I wonder whether this is something we could fix easily - or is it safe to ignore?

szhorvat:
Should it be `utf-8` instead of `utf8`?

szhorvat:
On my system, the command seems to be called `docbook2texi` and not `docbook2x-texi`. Is the latter a Debian-specific thing by any chance? The documentation of that package also mentioned `docbook2texi`: https://docbook2x.sourceforge.net/latest/doc/docbook2X.html

szhorvat:
After changing `utf8` to `utf-8` and using the executable name `docbook2texi`, there are two remaining warnings during the build:

```
[192/193] Converting DocBook XML to GNU Texinfo format
error : xmlAddEntity: invalid redeclaration of predefined entity
docbook2texi://section[@id='igraph-Installation-vcpkg']: section is too deep
[193/193] Generating info documentation with GNU Makeinfo
```

This is with docbook2x 0.8.8 and texinfo 0.7.1 installed with MacPorts.

szhorvat:
A last missing piece would be to try to support installation of the texinfo docs with `make install`.

ntamas:
> On my system, the command seems to be called docbook2texi and not docbook2x-texi.

One of these is in `docbook2x`, the other one is in `docbook-utils`, at least on Ubuntu. The CMake file correctly mentions `docbook2x` as a requirement so I guess this is okay.

ntamas:
> This could also be done later (after this PR) as we'd need some way of controlling whether installation should happen.

With the HTML docs we chose to include it in the source tarball so maintainers only need to copy these from the source tarball to whatever place the distro needs. Can we do the same thing with the Texinfo docs?

ntamas:
One argument in favour of using `docbook2texi` instead of `docbook2x-texi` is that `docbook2texi` is also available on macOS. @aagon Do you have any concerns with using `docbook2texi`?

szhorvat:
> Can we do the same thing with the Texinfo docs?

Yes, we should just do that and let package maintainers handle the installation manually.

doc/installation.xml
```diff
@@ -459,18 +459,21 @@ pacman -S \
...
   <para>
     Building the HTML documentation requires Python 3, <literal>xmlto</literal>
-    and <literal>source-highlight</literal>. Building the PDF documentation
-    also requires <literal>xsltproc</literal>, <literal>xmllint</literal>
-    and <literal>fop</literal>.
+    and <literal>source-highlight</literal>. Building the PDF documentation also
+    requires <literal>xsltproc</literal>, <literal>xmllint</literal> and
+    <literal>fop</literal>. Building the Info documentation also requires
+    <literal>docbook2x</literal>, <literal>xmllint</literal> and
+    <literal>makeinfo</literal>.
```
szhorvat, line 476:
```suggestion
    and <literal>source-highlight</literal>. Building the PDF documentation also
    requires <literal>xsltproc</literal>, <literal>xmllint</literal> and
    <literal>fop</literal>. Building the Texinfo documentation also requires
    <literal>docbook2texi</literal>, <literal>xmllint</literal>
    and <literal>makeinfo</literal>.
```

This suggestion assumes that we go with `docbook2texi`, which I think we should. This command is provided by at least these two separate packages:

 - https://docbook2x.sourceforge.net/
 - https://sourceware.org/docbook-tools/

The unusual naming in Debian is very likely an attempts at avoiding naming conflicts between these two packages. With MacPorts these two cannot be installed at the same time.

szhorvat: It seems that I was right about why Debian renames this program: https://bugs.gentoo.org/205703

But I was wrong in assuming that the two programs called `docbook2texi` are interchangeable. The one from docbook-tools / docbook-utils does not work here. I'm getting this:

```
FAILED: doc/igraph-docs.texi /Users/szhorvat/Repos/igraph-main/igraph/build/doc/igraph-docs.texi 
cd /Users/szhorvat/Repos/igraph-main/igraph/build/doc && /opt/local/bin/docbook2texi --encoding=utf-8//TRANSLIT --string-param output-file=igraph-docs --string-param directory-category=Libraries --string-param directory-description="A\ fast\ graph\ library\ (C)" igraph-docs-with-resolved-includes.xml
-e Usage: jw [<options>] <sgml_file>
where <options> are:
...
```

This present a dilemma as hard-coding `docbook2x-texi` will prevent the build from working on some systems, but using `docbook2texi` is also not fully reliable.

CMake's `find_program` allows specifying multiple names, so we could prefer `docbook2x-texi`, then fall back to `docbook2texi` if that is not present.

`find_program` also has a feature to validate candidates, but this feature is only available in CMake 3.25 or later.

What do you suggest @ntamas?

szhorvat:
I pushed a change. I'm happy with merging like this.

ntamas:
Thank you all!

aagon:
Thank you very much for your quick answer, and your various corrections.

I saw that you finally decided to prefer docbook2x over docbook-tools. I agree
with this, the dependencies are much smaller (on Debian at least).

@szhorvat I have no idea whether those last two warnings are important :

```
[192/193] Converting DocBook XML to GNU Texinfo format
error : xmlAddEntity: invalid redeclaration of predefined entity
docbook2texi://section[@id='igraph-Installation-vcpkg']: section is too deep
[193/193] Generating info documentation with GNU Makeinfo
```

I ignored them since they did not seem to block the correct building. I've
looked at the generated file a little bit, everything seems to be in order.

> With the HTML docs we chose to include it in the source tarball so maintainers
> only need to copy these from the source tarball to whatever place the distro
> needs. Can we do the same thing with the Texinfo docs?

> Yes, we should just do that and let package maintainers handle the
> installation manually.

Just for information, the Debian Maintainers will likely want to rebuild the
info file anyway, as they already do with HTML and PDF. So, if you do not want
to add dependencies to your source tarball target, feel free to not distribute
the finalised info file.

On top of that, the changes needed in the Debian packaging to rebuild and
distribute the generated info files are trivial: I have already prepared and
tested them, I will submit them to the Debian Maintainers as soon as applicable.

Thank you again !

Best,

Aymeric

aagon:
Thank you very much for your quick answer, and your various corrections.

I saw that you finally decided to prefer docbook2x over docbook-tools. I agree
with this, the dependencies are much smaller (on Debian at least).

@szhorvat I have no idea whether those last two warnings are important :

```
[192/193] Converting DocBook XML to GNU Texinfo format
error : xmlAddEntity: invalid redeclaration of predefined entity
docbook2texi://section[@id='igraph-Installation-vcpkg']: section is too deep
[193/193] Generating info documentation with GNU Makeinfo
```

I ignored them since they did not seem to block the correct building. I've
looked at the generated file a little bit, everything seems to be in order.

> With the HTML docs we chose to include it in the source tarball so maintainers
> only need to copy these from the source tarball to whatever place the distro
> needs. Can we do the same thing with the Texinfo docs?

> Yes, we should just do that and let package maintainers handle the
> installation manually.

Just for information, the Debian Maintainers will likely want to rebuild the
info file anyway, as they already do with HTML and PDF. So, if you do not want
to add dependencies to your source tarball target, feel free to not distribute
the finalised info file.

On top of that, the changes needed in the Debian packaging to rebuild and
distribute the generated info files are trivial: I have already prepared and
tested them, I will submit them to the Debian Maintainers as soon as applicable.

Thank you again !

Best,

Aymeric

ntamas:
> Just for information, the Debian Maintainers will likely want to rebuild the
info file anyway, as they already do with HTML and PDF.

@szhorvat Would this be an argument in favour of _not_ distributing the Texinfo files? It is ~2.6MB uncompressed so it inflates the tarball a bit and it's unlikely that many people would want to use it as it's basically Linux-only (unlike the HTML files, which are useful on all the supported platforms).

szhorvat:
No strong opinions from me. I don't read docs with `info` (don't like reading docs in the terminal) so I can't judge how big the demand is for this. Usually there is a HTML alternative.

The main target audience for this are likely distro maintainers, who can probably build this without much trouble.

Maybe we should look at what other projects do?

szhorvat:
> @szhorvat Would this be an argument in favour of _not_ distributing the Texinfo files? It is ~2.6MB uncompressed so it inflates the tarball a bit and it's unlikely that many people would want to use it as it's basically Linux-only (unlike the HTML files, which are useful on all the supported platforms).

Just a reminder. Will you remove mandatory inclusion in the distributed archive?

aagon:
Since the PR discussion is still active, I use the opportunity to provide a
small aesthetic correction to it:

``` diff
diff --git a/doc/CMakeLists.txt b/doc/CMakeLists.txt
index 1291df870..3219d103a 100644
--- a/doc/CMakeLists.txt
+++ b/doc/CMakeLists.txt
@@ -349,7 +349,7 @@ if(DOC_BUILD_SUPPORTED)
       --encoding=utf-8//TRANSLIT
       --string-param output-file=igraph-docs
       --string-param directory-category=Libraries
-      --string-param directory-description="A fast graph library \(C\)"
+      --string-param directory-description='A fast graph library \(C\)'
       igraph-docs-with-resolved-includes.xml
       MAIN_DEPENDENCY igraph-docs-with-resolved-includes.xml
       COMMENT "Converting DocBook XML to GNU Texinfo format"
```

This improves the display of the short description in the top-level dir file
from :

"A\ fast\ graph\ library\ (C)"

to the cleaner :

"A fast graph library (C)"

Sorry I did not catch this earlier. I did not want to open a new PR for this,
but I can if you prefer.

ntamas:
@szhorvat Yup, the Info docs will be removed from the tarball, but I'm knee deep in the attribute handler refactoring now and don't want to switch context. I'll get back to this later.

@aagon Thanks for the correction, no need to send the PR, I'll incorporate this on my own. (Unless you want to get credit for the commit, in which case please send a PR and I'll merge).

ntamas:
@szhorvat Texinfo docs are not distributed in the tarball any more.

@aagon I've incorporated the fix in `master`, thanks a lot!

PR #2458, opened 2023-12-30
feat: Distinguish between `VERTEX_QTY` and `ALL_VERTEX_QTY`

krlmlr:
Follow-up to https://github.com/igraph/rigraph/pull/1037 which will need a fix.

Proposing to distinguish between a quantity for all vectors (which doesn't need `vids`) and a quantity for subsets (the current implementation).

I don't understand the semantics of `igraph_edge_betweenness_subset` and `igraph_betweenness_subset`, do they also need to use `ALL_VERTEX_QTY` ?

### DIFF
--- a/interfaces/functions.yaml
+++ b/interfaces/functions.yaml
@@ -652,7 +652,7 @@ igraph_betweenness_subset:
         BOOLEAN directed=True, VERTEX_SELECTOR sources=ALL, VERTEX_SELECTOR targets=ALL,
         EDGEWEIGHTS weights=NULL
     DEPS: |-
-        vids ON graph, weights ON graph, res ON graph, sources ON graph, targets ON graph
+        vids ON graph, weights ON graph, res ON graph vids, sources ON graph, targets ON graph
 
 igraph_edge_betweenness:
     PARAMS: |-
@@ -695,7 +695,7 @@ igraph_pagerank:
         REAL damping=0.85, EDGEWEIGHTS weights=NULL,
         INOUT PAGERANKOPT options=NULL
     DEPS: |-
-        vids ON graph, weights ON graph, vector ON graph,
+        vids ON graph, weights ON graph, vector ON graph vids,
         options ON algo
 
 igraph_personalized_pagerank:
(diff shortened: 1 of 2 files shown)


interfaces/functions.yaml
```diff
@@ -642,7 +642,7 @@ igraph_betweenness_subset:
         BOOLEAN directed=True, VERTEX_SELECTOR sources=ALL, VERTEX_SELECTOR targets=ALL,
         EDGEWEIGHTS weights=NULL
     DEPS: |-
-        vids ON graph, weights ON graph, res ON graph, sources ON graph, targets ON graph
+        vids ON graph, weights ON graph, res ON graph vids, sources ON graph, targets ON graph
```
krlmlr, line 645: Is this even right?

szhorvat:
`edge_betweenness_subset` gives results for edges, not vertex, do `VERTEX_QTY` is definitely a bug in `interfaces.yaml`.

I am still getting to grips with stimulus, so ping @ntamas. Do we have an `EDGE_QTY`? Why do we need a `VERTEX_QTY` at all? How does this differ from a simple `VECTOR`?

szhorvat:
OK, it's actually explained in `types.yaml`:

```
    # A vector of floating-point numbers where each entry corresponds to
    # one of the vertices in a graph and its value represents some quantity
    # associated to the vertex with the same index. Higher-level interfaces may
    # use this type to provide a "named vector" such that each entry can be
    # indexed either by the vertex index or by the vertex name.
```

Since edges don't have names, we don't need `EDGE_QTY`. I'll change this to `VECTOR` in the C core.

krlmlr:
Thanks. Please feel free to push here.

szhorvat:
I already fixed this on `master`. I did not realize that this was in the `igraph` repo, and not the `rigraph` repo.

krlmlr:
Do we need anything else here?

szhorvat:
This should be reviewed by @ntamas, as the new Py interface does use Stimulus, but the Mma interface doesn't. Let's not leave PRs hanging for too long.

ntamas (approved):

szhorvat:
@krlmlr @Antonov548 @ntamas Does this not need a corresponding update in Stimulus then? It will certainly need adaptations in R.

ntamas:
It does not need any updates in Stimulus itself. The R interface needs to provide the snippets that the code generator should use when handling an argument of type `ALL_VERTEX_QTY`.

PR #2462, opened 2024-01-19
Attribute handler refactoring

ntamas:
Attribute handling related functions now use a typed `igraph_attribute_vector_list_t` list to declare attributes instead of an untyped `void*` pointer.

Still in draft phase; I want to get some feedback on coverage before proceeding further.


szhorvat:
Started a Coverity run now.

szhorvat:
7 defects eliminated, 3 newly detected.

Of the 3, only one seems to be of real concern. In the GraphML reader, `attr_name`, allocated on line 658, is not protected with `IGRAPH_FINALLY` and will leak if `IGRAPH_CHECK` causes a `return`.

Can you have a look @ntamas ?

ntamas:
That's a false positive as well, `attr_name` is protected indirectly via `IGRAPH_FINALLY(safely_free_optional_string, &attr_name)` at the very beginning. I had to use this trick because `attr_name` can be `NULL` or a valid pointer to a string.

szhorvat:
Can you either bring down the `IGRAPH_FINALLY` to where the allocation happens, or if that's not possible or not a good idea, then just add a comment that `IGRAPH_FINALLY` is at the top?

ntamas:
I'm not expecting a thorough review from any of you guys as the PR is _huge_ and I did not really manage to modularize it (everything was just too much entangled with everything else). The increase in coverage as well as the fact that all CI tests pass gives me enough confidence that the refactoring is likely to be correct. Additional testing can start when we merge it back to `develop` and I update the higher level interfaces to use the new conventions.

@szhorvat just shout if you are okay with merging it as is or if you want to take a look at it nevertheless before merging.

ntamas:
> Can you either bring down the `IGRAPH_FINALLY` to where the allocation happens, or if that's not possible or not a good idea, then just add a comment that `IGRAPH_FINALLY` is at the top?

I'll add a comment. I can't bring it down to where the allocation happens because I need to re-allocate the string every time I encounter a new `name` attribute in XML. In theory there should only be one but you never know ;)

szhorvat:
Please go ahead and merge. I ran it through Clang Analyzer and it did not find anything new either.

PR #2464, opened 2024-01-22
Add skeleton for max matching function, pseudo code, and tests

EdgarAlenPoe:
The skeleton of the function and tests are put together, and pseudo code for algorithm is in place.

There will be two internal functions used. Pseudo code for them is included, but not function signatures.
In addition I'll be using a specific implementation of sets and a linked list for the algorithm.
I didn't see any linked list in the data structures igraph has, did I miss it?

### DIFF
--- a/tests/unit/igraph_maximum_matching.c
+++ b/tests/unit/igraph_maximum_matching.c
@@ -1,0 +1,26 @@
+/*
+   IGraph library.
+   Copyright (C) 2022  The igraph development team <igraph@igraph.org>
+
+   This program is free software; you can redistribute it and/or modify
+   it under the terms of the GNU General Public License as published by
+   the Free Software Foundation; either version 2 of the License, or
+   (at your option) any later version.
+
+   This program is distributed in the hope that it will be useful,
+   but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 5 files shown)


szhorvat:
> I didn't see any linked list in the data structures igraph has, did I miss it?

There is none at the moment. You may implement your own or just use C++ for this purpose if easier ...

ntamas:
If you decide to go for linked lists in C, I would highly recommend using [this library](https://troydhanson.github.io/uthash/utlist.html). It is header-only and BSD-licensed, and I've been using it in many other projects without problems.

That being said, feel free to use C++ for this purpose if that seems easier.

tests/unit/igraph_maximum_matching.c
```diff
@@ -17,10 +17,133 @@
...
+    if (test_bipartite_graphs()) {
+        result = 1;
+    }
+    if (test_general_graphs()) {
+        result = 1;
+    }
+    if (test_petersen_graph()) {
+        result = 1;
+    }
+    return result;
```
szhorvat, line 148: Just a heads up that we are moving away from using the return value for tests. You can use `IGRAPH_ASSERT()` or even `IGRAPH_FATAL()`. These have the advantage that when they fail, and ASan support is enabled, you will get a stack trace.

Issue #2469, opened 2024-01-28
Wishlist: Biconnectedness test


szhorvat:
**What is the feature or improvement you would like to see?**

A function to check if a graph is biconnected (i.e. 2-vertex-connected): `igraph_is_biconnected()`.

**Use cases for the feature**

Performance: While this is already possible using `igraph_biconnected_components()`, and checking the number of components, a specialized function that only verifies biconnectedness, without finding the components, can be considerably faster.

**Optimization opportunities**

 - Use the cache:
    * If the graph is not connected, it is not biconnected
    * If the graph is acyclic, it is not biconnected
 - As soon as an articulation point is found, the function can return with a negative answer

**References**
—

ntamas:
Note that currently we define biconnectedness only for undirected graphs. (Internally, `igraph_biconnected_components()` traverses the graph using an incidence list constructed with `IGRAPH_ALL`, therefore directed graphs are treated as undirected). This means that this is not true:

> If the graph is acyclic, it is not biconnected

Consider `a -> b -> c, a -> d -> c`. This is acyclic, but if you ignore the edge directions then it is biconnected.

On the other hand, we can say that is a graph is a forest, then it is not biconnected.

I've started a PR in #2470 - let's continue the discussion there.

szhorvat:
Oh wow, I didn't expect you'd do this so quickly.

PR #2470, opened 2024-01-29
Add igraph_is_biconnected() to decide whether a graph is biconnected

ntamas:
As requested in https://github.com/igraph/igraph/issues/2469

### DIFF
--- a/examples/simple/igraph_is_biconnected.c
+++ b/examples/simple/igraph_is_biconnected.c
@@ -1,0 +1,37 @@
+/* -*- mode: C -*-  */
+/*
+   IGraph library.
+   Copyright (C) 2006-2023  The igraph development team
+
+   This program is free software; you can redistribute it and/or modify
+   it under the terms of the GNU General Public License as published by
+   the Free Software Foundation; either version 2 of the License, or
+   (at your option) any later version.
+
+   This program is distributed in the hope that it will be useful,
+   but WITHOUT ANY WARRANTY; without even the implied warranty of
+   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+   GNU General Public License for more details.
+
+   You should have received a copy of the GNU General Public License
+   along with this program; if not, write to the Free Software
+   Foundation, Inc.,  51 Franklin Street, Fifth Floor, Boston, MA
+   02110-1301 USA
+
+*/
+
+#include <igraph.h>
+#include <stdio.h>
+
+int main(void) {
+
+    igraph_t g;
+    igraph_bool_t result;
+
+    igraph_small(&g, 7, 0, 0, 1, 1, 2, 2, 3, 3, 0, 2, 4, 4, 5, 2, 5, -1);
+    igraph_is_biconnected(&g, &result);
+    printf("Graph is%sbiconnected.\n", result ? " " : " not ");
+    igraph_destroy(&g);
+
+    return 0;
+}
--- a/examples/simple/igraph_is_biconnected.c
+++ b/examples/simple/igraph_is_biconnected.c
@@ -1,7 +1,6 @@
-/* -*- mode: C -*-  */
 /*
    IGraph library.
-   Copyright (C) 2006-2023  The igraph development team
+   Copyright (C) 2024  The igraph development team <igraph@igraph.org>
 
    This program is free software; you can redistribute it and/or modify
    it under the terms of the GNU General Public License as published by
@@ -14,10 +13,7 @@
    GNU General Public License for more details.
 
    You should have received a copy of the GNU General Public License
-   along with this program; if not, write to the Free Software
-   Foundation, Inc.,  51 Franklin Street, Fifth Floor, Boston, MA
-   02110-1301 USA
-
+   along with this program.  If not, see <https://www.gnu.org/licenses/>.
 */
 
 #include <igraph.h>
--- a/src/connectivity/components.c
+++ b/src/connectivity/components.c
@@ -1278,7 +1278,7 @@ igraph_error_t igraph_biconnected_components(const igraph_t *graph,
  * \example examples/simple/igraph_is_biconnected.c
  */
 
-igraph_error_t igraph_is_biconnected(const igraph_t *graph, igraph_bool_t *result) {
+igraph_error_t igraph_is_biconnected(const igraph_t *graph, igraph_bool_t *res) {
 
     igraph_integer_t no_of_nodes = igraph_vcount(graph);
     igraph_vector_int_t nextptr;
@@ -1397,8 +1397,8 @@ exit:
 
 exit2:
 
-    if (result) {
-        *result = is_biconnected;
+    if (res) {
+        *res = is_biconnected;
     }
 
     /* We do not touch the cache for graphs with less than three nodes because
--- a/src/connectivity/components.c
+++ b/src/connectivity/components.c
@@ -1290,6 +1290,16 @@ igraph_error_t igraph_is_biconnected(const igraph_t *graph, igraph_bool_t *resul
     igraph_integer_t counter, rootdfs = 0;
     igraph_bool_t is_biconnected = true;
 
+    if (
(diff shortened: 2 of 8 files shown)


szhorvat:
Let's figure out the edge cases:

 - The null graph is not connected, so also not biconnected
 - We'll probably agree that the singleton should not be biconnected
 - I'm not so sure about what to do with $K_2$ ...

ntamas:
Re $K_2$, the documentation of `igraph_biconnected_components()` already settles this:

> Note that some authors do not consider the graph consisting of two connected vertices as biconnected, however, igraph does.

I guess we should just stick to this.

szhorvat:
### Arguments for treating $K_2$ as biconnected

 - If we partition the edges of the graph into disjoint biconnected components, $K_2$ will have a _single_ component

### Survey on how other packages treat $K_2$

#### Mathematica: `False`

```
In[21]:= KVertexConnectedGraphQ[CompleteGraph[2], 2]
Out[21]= False

In[22]:= KVertexConnectedComponents[CompleteGraph[2], 2]
Out[22]= {}

In[23]:= KVertexConnectedComponents[PathGraph[{1, 2, 3}], 2]
Out[23]= {{1, 2}, {2, 3}}
```

I find it strange that some of the vertices of $K_2$ belong to no biconnected component according to Mathematica.

#### NetworkX: `True`

image

szhorvat:
@ntamas I wrote up some argument in favour / against treating $K_2$ as biconnected. This is rather messy, but I'm leaning towards treating it as biconnected because we want to stay consistent with `igraph_biconnected_components()`, and I'd like to say that every edge is part of precisely one biconnected component. (We don't say the same about vertices: isolated vertices are not part of any biconnected component.)

I would suggest treating graphs with fewer than 3 vertices separately in the implementation, and not touching the cache for such small graphs. One issue we have is that $K_2$ will be biconnected, yet it's still acyclic / a forest.

All this issues should be written up in the docs.

What do you think?

szhorvat:
> > Note that some authors do not consider the graph consisting of two connected vertices as biconnected, however, igraph does.
> 
> I guess we should just stick to this.

I'm sorry, I didn't notice this. It seems we agree then.

examples/simple/igraph_is_biconnected.c
```diff
@@ -0,0 +1,37 @@
...
+   but WITHOUT ANY WARRANTY; without even the implied warranty of
+   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+   GNU General Public License for more details.
+
+   You should have received a copy of the GNU General Public License
+   along with this program; if not, write to the Free Software
+   Foundation, Inc.,  51 Franklin Street, Fifth Floor, Boston, MA
+   02110-1301 USA
+
+*/
```
szhorvat, line 21:
```suggestion
/*
   IGraph library.
   Copyright (C) 2024  The igraph development team <igraph@igraph.org>

   This program is free software; you can redistribute it and/or modify
   it under the terms of the GNU General Public License as published by
   the Free Software Foundation; either version 2 of the License, or
   (at your option) any later version.

   This program is distributed in the hope that it will be useful,
   but WITHOUT ANY WARRANTY; without even the implied warranty of
   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
   GNU General Public License for more details.

   You should have received a copy of the GNU General Public License
   along with this program.  If not, see <https://www.gnu.org/licenses/>.
*/
```

src/connectivity/components.c
```diff
@@ -1246,6 +1251,159 @@ igraph_error_t igraph_biconnected_components(const igraph_t *graph,
...
+ * or \ref igraph_articulation_points() if the graph is likely not to be
+ * biconnected as it can bail out early without counting all the biconnected
+ * components or finding all articulation points.
+ *
+ * </para><para>
+ * Note that some authors do not consider the graph consisting of
+ * two connected vertices as biconnected, however, igraph does.
+ *
+ * \param graph The input graph. It will be treated as undirected.
+ * \param result If not a NULL pointer, the result will be returned here.
```
szhorvat, line 1272:
```suggestion
 * \param result If not a \c NULL pointer, the result will be returned here.
```

tests/unit/igraph_is_biconnected.c
```diff
@@ -0,0 +1,73 @@
...
+   but WITHOUT ANY WARRANTY; without even the implied warranty of
+   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+   GNU General Public License for more details.
+
+   You should have received a copy of the GNU General Public License
+   along with this program; if not, write to the Free Software
+   Foundation, Inc.,  51 Franklin Street, Fifth Floor, Boston, MA
+   02110-1301 USA
+
+*/
```
szhorvat, line 21:
```suggestion
/*
   IGraph library.
   Copyright (C) 2024  The igraph development team <igraph@igraph.org>

   This program is free software; you can redistribute it and/or modify
   it under the terms of the GNU General Public License as published by
   the Free Software Foundation; either version 2 of the License, or
   (at your option) any later version.

   This program is distributed in the hope that it will be useful,
   but WITHOUT ANY WARRANTY; without even the implied warranty of
   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
   GNU General Public License for more details.

   You should have received a copy of the GNU General Public License
   along with this program.  If not, see <https://www.gnu.org/licenses/>.
*/
```

szhorvat (changes requested):
A few nitpick changes and one big one:

The cache use is currently invalid for $K_2$, which is both biconnected and acyclic. Let us not touch the cache at all if the vertex count `n` is less than 3. There should be a separate code path for the `n < 3` case.

`n == 0` and `n == 1` are easy: not biconnected.

For `n == 2`, I'm not sure what's the simplest solution. We could iterate through edges until we find an edge connecting the two vertices. If we found one, then the graph is biconnected.  But this is not simple code given that it handles just one specific case.

Alternatively, the caching part could be guarded by an `if (n >= 3)` (both reading and setting the cache). Then the main code can handle `n == 2` correctly.

We should add comments in the source code about the special `n == 2` case.

szhorvat:
Let's also add this to `functions.yaml`.

src/connectivity/components.c
```diff
@@ -979,6 +979,11 @@ static igraph_error_t igraph_i_decompose_strong(const igraph_t *graph,
  * A vertex is an articulation point if its removal increases
  * the number of (weakly) connected components in the graph.
  *
+ * </para><para>
+ * Use \ref igraph_is_biconnected() if you are only interested in whether the
+ * graph has at least one articulation point: if it is biconnected, it does not
+ * have any articulation points, otherwise it has at least one.
```
szhorvat, line 985: This is not quite accurate. Once again $K_2$ has no articulation points yet it is biconnected.

Suggestion:

```suggestion
 * Use \ref igraph_is_biconnected() if you are only interested in whether the
 * graph has precisely one biconnected component.
```

ntamas:
Ready for another round here.

src/connectivity/components.c
```diff
@@ -1246,6 +1250,167 @@ igraph_error_t igraph_biconnected_components(const igraph_t *graph,
...
+
+exit2:
+
+    if (res) {
+        *res = is_biconnected;
+    }
+
+    /* We do not touch the cache for graphs with less than three nodes because
+     * of all the edge cases. */
+    if (is_biconnected && no_of_nodes > 2) {
```
szhorvat, line 1406:
```suggestion
    if (is_biconnected && no_of_nodes >= 3) {
```

Just for consistency with the lines at the top that read the cache.

szhorvat (approved):

szhorvat:
Do you mind if I backport this to `master` and add it to Python?

ntamas:
Nope, actually I started this on `master` but then I realized that there isn't an `Added` section yet in the changelog and did not want to start one. But if you need this, then sure, go ahead.

You can use the code for `igraphmodule_Graph_is_connected()` as an inspiration for `is_biconnected()`.

szhorvat:
Backport done. I marked it as experimental on `master` only (not on `develop`). Just in case ...

PR #2474, opened 2024-01-31
Fix to build with invalid pointer type enforced.

limburgher:


szhorvat:
Thanks! Can you please let me know what specific compiler warning/error this fixes with what version of libxml2? I though we were fine even with the most recent libxml2 2.12, which I tested a while ago. Do you have a link to a build log which shows the error or warning?

limburgher:
This is due to some new default flags for GCC 14 in Fedora.
Info on the changes: https://fedoraproject.org/wiki/Toolchain/PortingToModernC

Build/buildroot logs: https://koji.fedoraproject.org/koji/taskinfo?taskID=112302364

szhorvat:
Thanks! I'll merge this to make sure you get credit, but I may undo this change and come up with a better fix later. I'll have to see if the type is the same in all supported libxml2 versions (and if you already checked this, let me know).

Something I noticed in the build log is that you are passing `-DIGRAPH_USE_INTERNAL_CXSPARSE=0` to CMake. This flag is no longer used sice igraph 0.10 and can be removed.

limburgher:
Sounds good, thank you! Yeah, no smoke came out but best to double check. :) I'll drop that flag.

szhorvat:
Turns out this was already fixed in #2442 but I didn't realize that it's not yet released. It's a libxml2 compatibility fix.

szhorvat:
@limburgher We are planning a new release before the end of this week (maybe even tomorrow), so you won't need to patch this in Fedora

limburgher:
Excellent, thank you!

PR #2477, opened 2024-02-01
fix: fix cache initialization when igraph_bool_t size is not 1 bytes

szhorvat:
Looks like this could be serious when `sizeof(igraph_bool_t) != 1`, which is the case in R only. Please double-check @ntamas 

CC @krlmlr


ntamas (approved):

ntamas:
Yes, this seems serious, I don't know what I was thinking when I wrote this :-D

PR #2478, opened 2024-02-03
refactor: remove MPI code from DrL

szhorvat:
This PR removed parallel MPI code from the DrL layout.

Reasoning:

 - DrL was wastefully calling the RNG without using the result, just to deal with parallelization, which we don't use
 - igraph might use parallelization in the future, but not MPI, so we won't need this code
 - Compilation would have failed if someone accidentally defined `MUSE_MPI`. Unfortunately some packagers insert defines for no good reason (once MacPorts broke igraph by forcibly defining the `DEBUG` macro)
 - This eliminates some minor (harmless) compiler warnings about signed-unsigned comparisons

This PR may change the output for the same random seed through the removal of the RNG calls with unused results.


ntamas:
Thanks!

PR #2486, opened 2024-02-11
feat: `igraph_are_adjacent()` replaces `igraph_are_connected()`.

szhorvat:
I would like to rename `igraph_are_connected()` to `igraph_are_adjacent()` to avoid confusion and better line up with standard graph theory terminology.

"Connected" is ambiguous. $u$ and $v$ are connected can mean that there is a path between them, or that there is an edge between them. It's best to avoid this term altogether in relation to vertices and use "$u$ and $v$ are adjacent" when there is an edge between them, and "$v$ is reachable from $u$" when there is a path from $u$ to $v$.

@krlmlr, if this PR gets merged, I recommend the same change in R.


szhorvat:
For comparison, Mathematica uses `EdgeQ` for this operation, through it's flawed in multigraphs. IGraph/M uses `IGAdjacentVerticesQ`. NetworkX uses `has_edge()`.

One might argue that it's better to use `igraph_are_adjacent_vertices()` to leave room for `igraph_are_adjacent_edges()`, but I find this name too long; furthermore, checking if two edges are adjacent is trivial, so I would not add it in C.

vtraag:
Completely agree!

ntamas (approved):

PR #2496, opened 2024-02-15
refactor: `igraph_i_property_cache_set_bool()` now verifies cache consistency when setting an already known property

szhorvat:


ntamas:
I'd rather keep the functionality of the `set_bool()` function as is, without surprises (i.e. just set the specified value, no additional checks), and introduce another `set_or_check_bool()` or `set_once_bool()` function to implement the functionality that you added here. What you are doing here is not really "verifying cache consistency" yet, it's more like a functionality to prevent changing the value of a cached property once it has been set.

szhorvat:
Is there any scenario where it makes sense to _change_ a value?

 - When we first compute values, we set them
 - When we redundantly recompute values, we might still set them, and this is an opportunity for a check
 - When we modify the graph, we invalidate values, but don't change them

When would we actually want to _change_ a value (as opposed to setting it, or invalidating it)?

szhorvat:
Let me put it differently: At first I wanted to do what you suggested. Then I realized that all current usages would need to be updated to the checked version of the function, so in practice it's easier to rename the existing function to "checked", and add an unchecked version with the original name. But this unchecked version wouldn't be used at the moment. In fact, I think it mustn't ever be used without going through proper invalidation first, as the risk of putting the cache into an inconsistent state (i.e. giving the properties logically inconsistent values) is too high.

ntamas:
> Is there any scenario where it makes sense to change a value?

There could be. For instance, adding a loop edge to a graph that has no loop edges yet. At one point you might want to say that you change the value of the cached "has loop" property to `true` instead of just invalidating it and letting the system figure it out the next time it is needed.

Anyhow, what I want to avoid is _surprises_. A function named `igraph_i_property_cache_set_bool()` is expected to set a boolean value in the cache, period. There is nothing in the name of the function that would imply that the new value is compared with the old value and that changing previously-set values is forbidden. If you rename the function to something else that makes this clear, then I'm okay with the change, even if we don't add an "unchecked" version yet -- although the way I'd do it is that the checked version would call the unchecked version if the conditions pass so it won't be unused.

ntamas:
PR now rewritten to use `igraph_i_property_cache_set_bool_checked()`. @szhorvat Let me know whether it's okay this way and then we can merge.

ntamas:
Note to ourselves: when this is merged into `develop`, we will also need to replace all occurrences of `igraph_i_property_cache_set_bool()` that are in `develop` but not in `master`.

szhorvat:
Yes, this is perfectly fine. Sorry about being so slow. The reason why I want this is to have extra confidence when writing code that uses the cache.

The next step is to implement the deduction engine, but I need to finish a lot of non-igraph-related work first.

PR #2504, opened 2024-02-19
fix: `write_graph_lgl/ncol()` now refuse to write invalid files

szhorvat:
This prevents the NCOL and LGL writers from creating invalid files that cannot be read back, and protects users from data loss. It also makes it easier to implement fuzzers for file format writers. 

UTF-8 won't be affected as it uses `>= 128` bytes for encoding higher code points.


szhorvat:
I've just pushed the fuzz target I used to test this.

PR #2509, opened 2024-02-23
Fix `is_separator()` on disconnected graphs

szhorvat:
The R (but not C) documentation includes:

> In the special case of a fully connected graph with $n$ vertices, each set of $n−1$ vertices is considered to be a vertex separator.

This is probably for consistency with the usual convention that assigns vertex connectivity $n-1$ to $K_n$.

I am not in favour of maintaining this definition of separators in complete graphs. However, we will need to make very sure that if we change this, we won't break any functions that call `igraph_is_separator()`.

### DIFF
--- a/src/connectivity/separators.c
+++ b/src/connectivity/separators.c
@@ -34,97 +34,269 @@
 
 #include "core/interruption.h"
 
-static igraph_error_t igraph_i_is_separator(const igraph_t *graph,
-                                 igraph_vit_t *vit,
-                                 igraph_integer_t except,
-                                 igraph_bool_t *res,
-                                 igraph_vector_bool_t *removed,
-                                 igraph_dqueue_int_t *Q,
-                                 igraph_vector_int_t *neis,
-                                 igraph_integer_t no_of_nodes) {
 
-    igraph_integer_t start = 0;
+/**
+ * \function igraph_i_is_separator
+ *
+ * Checks if vertex set \c S is a separator. Optionally, also checks if it
+ * is a _minimal_ separator.
+ *
+ * \param graph The graph, edge directions are ignored.
+ * \param S Candidate vertex set.
+ * \param is_separator Pointer to boolean, is S a separator?
+ * \param is_minimal If not \c NULL, it is also checked whether the separator
+ *    is minimal. This takes additional time.  If S is not a separator, this is
+ *    set to false
+ */
+static igraph_error_t igraph_i_is_separator(
+    const igraph_t *graph,
+    igraph_vs_t S,
+    igraph_bool_t *is_separator,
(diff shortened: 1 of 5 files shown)


szhorvat:
@ntamas, can you please have a first read-through? What I am looking for at this stage is if the approach and comments are clear enough. Feel free to suggest more comments.

`igraph_is_minimal_separator()` needs more tests.

src/connectivity/separators.c
```diff
@@ -34,97 +34,253 @@
...
+    IGRAPH_CHECK(igraph_vit_create(graph, S, &vit));
+    IGRAPH_FINALLY(igraph_vit_destroy, &vit);
+
+    IGRAPH_VECTOR_CHAR_INIT_FINALLY(&mark, vcount);
+    IGRAPH_VECTOR_INT_INIT_FINALLY(&neis, 10);
+    IGRAPH_DQUEUE_INT_INIT_FINALLY(&Q, 100);
+
+#define VISITED(x) (VECTOR(mark)[x] & 1) /* Was x visited? */
+#define SET_VISITED(x) (VECTOR(mark)[x] |= 1) /* Mark x as visited. */
+#define INS(x) (VECTOR(mark)[x] & 2) /* Is x in S? */
```
ntamas, line 80: I'd call this `IN_S`; `INS` makes me think about insertion.

ntamas (commented):
Looks clear enough to me so far, I've posted one comment related to naming only.

src/connectivity/separators.c
szhorvat: Sure, I'll change that.

The minimal separator test is wrong, I'm working on fixing that now.

szhorvat:
@ntamas Can you see if there's an easy fix for the CodeQL failure, which is unrelated to this PR?

I believe this is now correct, but let me sleep on it once. I'll let you know when it's ready for a check. I meant to write a test that first decomposes the graph into components, and tests each one. But `igraph_decomponse()` doesn't keep track of the vertex IDs between the original graph and components, unfortunately, so this would take an inconvenient amount of work.

Regarding the handling of size $n-1$ vertex subsets of the $K_n$ complete graph: the note in the R docs was actually incorrect. I'm going to remove it.

szhorvat:
This is ready for review now.

ntamas (approved):
Looks good to me after reading the code. I trust that the tests are extensive enough to cover the corner cases that I wasn't thinking about ;)

szhorvat:
OK, I'll merge this an later I'll do #2512 and #2516 which will make testing easier. #2516 will be an internal function for now. We can add the extra features to the existing public functions in 1.0.

Issue #2490, opened 2024-02-12
Wishlist: `igraph_is_complete()` decides whether a graph is complete


aagon:
I was surprised to see that there was no function to decide whether a graph is
complete or not. Of course, one might decide to compute the largest clique of
the graph and compare it to the graph itself, but that would be very costly.

The context: I am generating a lot of random geometric graphs for my
experiments, and I simply wanted to discard the complete ones. I gave myself a
very simple implementation that compares the amount of vertices and edges and
concludes (I'm only dealing with simple graphs).

I thought I would contribute an implementation of such a function, but I would
like to see whether you're interested first, and I had a couple of questions.

The heuristic would be the following:

- If the amount of edges is low enough that the graph cannot possibly be complete,
  we return early (O(1)).

- Then, we test if the graph is simple, with `igraph_is_simple()` O(e + v) in
  the worst case, O(1) if cache hit. If the graph is simple, we can compare and
  conclude.
  
  (Side question here: why do you cache the result of `igraph_is_simple()` only
  if it is positive ?)

- If the graph is not simple, we count the loops with function
  `igraph_is_loop()` (O(e)).
  
- We test again whether the amount of edges (minus the loops) might not be low
  enough for us to conclude directly (O(1)).

- If it is not the case, we need to test whether there are multiple edges with
  `igraph_has_multiple()`. This should be O(1), since we should have cached the
  result from `igraph_is_simple()`.
  
- If there are none, we can conclude by simple comparison.

- If there are some, we use `igraph_is_multiple()` to count the multiple edges.
  (O(e*d)) with d the average degree of the graph.
  
- We can now conclude by comparing and accounting for both loops and multiple
  edges.

This heuristic is maybe a little complicated, but ensures that we conclude as
soon as we can, and tries to shove the expensive operations at the end.

Also, the memory consumption is minimal. We only need one vector of size e, to
store the result of functions `igraph_is_loop()` and `igraph_is_multiple()`.

I also had in mind another, simpler heuristic that would duplicate the graph,
and then simplify the duplicate with `igraph_simplify()`, but this would require
something like O(e + v) memory for the new graph.

I'll let you decide whether you prefer to optimise for memory or simplicity.

I wanted to ask you a couple of questions before proceeding :

- Are you interested at all by such a function ?

- In what file would you want me to put it ?

- A complete undirected graph is a graph such that there is an undirected edge
  between every couple of vertices, but a complete directed graph can
[...]

PR #2510, opened 2024-02-24
feat: igraph_is_complete

aagon:
Follow-up of #2490.

### DIFF
--- a/interfaces/functions.yaml
+++ b/interfaces/functions.yaml
@@ -2512,6 +2512,9 @@ igraph_to_prufer:
 igraph_tree_from_parent_vector:
     PARAMS: OUT GRAPH graph, INDEX_VECTOR parents, TREE_MODE type=OUT
 
+igraph_is_complete:
+    PARAMS: GRAPH graph, OUT BOOLEAN res
+
 igraph_minimum_spanning_tree:
     PARAMS: GRAPH graph, OUT EDGE_INDICES res, EDGEWEIGHTS weights=NULL
     DEPS: res ON graph, weights ON graph
--- a/src/properties/complete.c
+++ b/src/properties/complete.c
@@ -26,6 +26,8 @@
  * \function igraph_is_complete
  * \brief Decides whether the graph is complete.
  *
+ * \experimental
+ *
  * A graph is considered complete if all pairs of different vertices are
  * adjacent.
  *
@@ -46,7 +48,6 @@ igraph_error_t igraph_is_complete(const igraph_t *graph, igraph_bool_t *res)
     const igraph_integer_t ecount = igraph_ecount(graph);
     igraph_integer_t complete_ecount;
     igraph_bool_t simple, directed = igraph_is_directed(graph);
-    igraph_integer_t i;
     igraph_vector_int_t neighbours;
 
     /* If the graph is the null graph or the singleton graph, return early */
@@ -108,15 +109,15 @@ igraph_error_t igraph_is_complete(const igraph_t *graph, igraph_bool_t *res)
 #  error "Unexpected IGRAPH_INTEGER_SIZE value."
 #endif
 
-/* If the amount of edges is strictly lower than what it should be for a
-   complete graph, return early */
+    /* If the amount of edges is strictly lower than what it should be for a
+       complete graph, return early */
 
-if (ecount < complete_ecount) {
-    *res = false;
-    return IGRAPH_SUCCESS;
-}
+    if (ecount < complete_ecount) {
+        *res = false;
+        return IGRAPH_SUCCESS;
+    }
 
-/* If the graph is simple, compare and conclude */
+    /* If the graph is simple, compare and conclude */
     IGRAPH_CHECK(igraph_is_simple(graph, &simple));
 
     if (simple) {
@@ -127,7 +128,7 @@ if (ecount < complete_ecount) {
     /* Allocate memory for vector of size v */
     IGRAPH_VECTOR_INT_INIT_FINALLY(&neighbours, vcount);
 
-    for (i = 0; i < vcount; ++i) {
+    for (igraph_integer_t i = 0; i < vcount; ++i) {
 
         igraph_vector_int_clear(&neighbours);
 
--- a/src/properties/complete.c
+++ b/src/properties/complete.c
@@ -1,0 +1,154 @@
+/*
+  IGraph library.
+  Copyright (C) 2024 The igraph development team <igraph@igraph.org>
+
+  This program is free software; you can redistribute it and/or modify it under
+  the terms of the GNU General Public License as published by the Free Software
+  Foundation; either version 2 of the License, or (at your option) any later
+  version.
+
+  This program is distributed in the hope that it will be useful, but WITHOUT
+  ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+  FOR A PARTICULAR PURPOSE. See the GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License along with
+  this program. If not, see <https://www.gnu.org/licenses/>.
+*/
+
+
+#include "igraph_interface.h"
+#include "igraph_structural.h"
+
+#include "graph/internal.h"
+
+/**
+ * \ingroup structural
+ * \function igraph_is_complete
+ * \brief Decides whether the graph is complete.
+ *
+ * A graph is considered complete if all pairs of different vertices are
+ * adjacent.
+ *
+ * </para><para>
+ * The null graph and the singleton graph are considered complete.
+ *
+ * \param graph The graph object to analyze.
+ * \param res Pointer to a logical variable, the result will be stored here.
+ *
+ * \return Error code.
+ *
+ * Time complexity: O(|V| + |E|) at worst.
+ */
+
+igraph_error_t igraph_is_complete(const igraph_t *graph, igraph_bool_t *res)
+{
+    const igraph_integer_t vcount = igraph_vcount(graph);
+    const igraph_integer_t ecount = igraph_ecount(graph);
+    igraph_integer_t complete_ecount;
+    igraph_bool_t simple, directed = igraph_is_directed(graph);
+    igraph_integer_t i;
+    igraph_vector_int_t neighbours;
+
+    /* If the graph is the null graph or the singleton graph, return early */
+    if (vcount == 0 || vcount == 1) {
+        *res = true;
+        return IGRAPH_SUCCESS;
+    }
+
+    /* Compute the amount of edges a complete graph of vcount vertices would
+       have */
+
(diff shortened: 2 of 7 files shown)


szhorvat:
Let me know when this is ready for review. To fix the CodeQL CI failure, you can rebase on the latest `master` branch (but this is not critical).

interfaces/functions.yaml
```diff
@@ -2512,6 +2512,9 @@ igraph_to_prufer:
 igraph_tree_from_parent_vector:
     PARAMS: OUT GRAPH graph, INDEX_VECTOR parents, TREE_MODE type=OUT
 
+igraph_is_complete:
+    PARAMS: GRAPH graph, PRIMARY OUT BOOLEAN res
```
szhorvat, line 2516: @ntamas What does `PRIMARY` do?

ntamas: I guess it's not needed here, but most likely it doesn't hurt. According to the Stimulus docs (see `docs/index.mdx` in the Stimulus repo):

> The `PRIMARY` keyword can be used to designate one or more _output_ arguments of a function as "primary" outputs. Code generators may use this keyword to generate two versions of a function: one that returns only the primary outputs and one that returns all of them. Alternatively, code generators may also generate a _single_ function with an additional `details` argument letting the user decide whether the result should include only the primary results or all of them. Again, it is up to the code generator to support the `PRIMARY` keyword.

ntamas: Curently the R code generator supports this. If there are both primary and non-primary out-args, the R code generator adds a `details=FALSE` argument to the function and only the primary out-args will be returned if `details=FALSE`.

szhorvat: Ah, so like `is_tree()`. Then we indeed do not need this here.

aagon: I've removed it

aagon:
It can be reviewed now, or I can add the support for directed graphs, or even `igraph_is_clique`, whatever you prefer.

szhorvat:
I had a look at the code (not the test though). Generally, it is very well written.

Regarding the ordering of neighbour lists: it is indeed guaranteed that these are sorted. This is documented for `igraph_neighbors()` and some igraph code relies on it, including the code that "simplifies" the neighbour lists.

We are missing support for directed graphs. Can you please add this?

Regarding running `igraph_is_simple()`: On a simple graph, this function needs to check each edge. It basically performs the same traversal as the following code does, iterating through all neighbour lists. Thus if we call this function, and the graph is non-simple, we end up doing twice the amount of work we'd need to. However, on simple graphs it has no disadvantage, and it is a cheap and safe way to make good use of the cache.

Thus I agree with you that this is a good way to implement `is_complete()`.

One more thing that is missing is an overflow check when computing the largest possible number of edges. What if `vcount*vcount` overflows? A quick way would be to use the macros from `math/safe_intop.h`, which even has a `IGRAPH_SAFE_N_CHOOSE_2`. However, these macros all throw an error on overflow, and this is not what we want. If the vertex count is so high that `vcount*(vcount-1)/2` is not representable, then the graph obviously can't be complete. It would be better to compare `vcount` against the largest vertex count with which a complete graph is still representable with igraph. But I'm not sure how to compute this value at compile time ...

Maybe instead of computing it, we can just assume that we're working with either standard 32-bit or 64-bit signed integers, and hard-code the limit for these two cases, similar to how it's done here:

https://github.com/igraph/igraph/blob/07d794d00d7785e348d5d8fc80a8016157391fa0/src/math/safe_intop.c#L74-L82

szhorvat:
So, in the undirected case, the max possible `vcount` is $2^{32} = 4294967296$ with 64-but and $2^{16} = 65536$ with 32-bit integers.

In the directed case, this will be `3037000500` and `46341`, respectively. I hope this helps.

szhorvat:
Finally, we still need to be smart about computing `(vcount * vcount) - vcount) / 2` in case `vcount` is too close to the above mentioned values. We could do something like:

```c
complete_ecount = vcount % 2 == 0 ? (vcount/2) * (vcount - 1) : vcount * ((vcount-1)/2)
```

ntamas:
> Maybe instead of computing it, we can just assume that we're working with either standard 32-bit or 64-bit signed integers, and hard-code the limit for these two cases

I agree. It's unlikely that we'll ever want to support architectures that are not 32-bit or 64-bit anyway.

src/properties/complete.c
```diff
@@ -0,0 +1,118 @@
...
+ * The null graph and the singleton graph are considered complete.
+ *
+ * \param graph The graph object to analyse.
+ * \param res Pointer to a logical variable, the result will be stored here.
+ *
+ * Time complexity: O(1) for simple and sparse graphs, O(|E|) at worst for graphs
+ * with either loops or multiple edges.
+ *
+ * Memory complexity: O(1) for simple and sparse graphs, O(|V|) at worst for
+ * graphs with either loops or multiple edges.
```
szhorvat, line 47: The time complexity would be O(|V| + |E|), i.e. no better than `igraph_is_simple()`, which this function calls. Of course if it was already cached that the graph is simple, then this is improved. But before knowing that, we can't guarantee anything better.

aagon:
Thank you for the review.

> Regarding the ordering of neighbour lists: it is indeed guaranteed that these are sorted.

I did not mean the neighbour list, but the list of all edges in the graph, iterated on according to edge ID. Those are not always sorted according to the "lowest adjacent vertex -- highest adjacent vertex" scheme I was mentioning, unless I'm mistaken. I asked because it would have allowed to count the edges directly on this list, rather than rely on `igraph_i_neighbours`. This would have avoided to check edges twice on undirected graphs. But this was too complicated anyway.

> We are missing support for directed graphs. Can you please add this?

Done.

> In the directed case, this will be 3037000500 and 46341, respectively. I hope this helps.

This was most useful, I am glad not to have had to do this myself. I've added a comment on how those numbers are produced.

> `complete_ecount = vcount % 2 == 0 ? (vcount/2) * (vcount - 1) : vcount * ((vcount-1)/2)`

This was very smart indeed, it is guaranteed to be very cheap.

I think that the result of this function should be cached in a cache entry "is_complete". If you agree, I will add a placeholder comment to remind myself to add caching once you're finished refactoring it.

tests/unit/igraph_is_complete.c
```diff
@@ -0,0 +1,183 @@
...
+  You should have received a copy of the GNU General Public License along with
+  this program. If not, see <https://www.gnu.org/licenses/>.
+
+*/
+
+#include <igraph.h>
+
+#include "test_utilities.h"
+
+static int check(const igraph_bool_t directed, const igraph_integer_t vertices)
```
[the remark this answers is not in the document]
szhorvat: Can you please add a _very brief_ comment about what this function does, especially starting at _ /* If directed, we must add the mutuals manually */_? I'm a bit confused about this part.

szhorvat (changes requested):
I added a few comments, but haven't had time to examine the test in detail yet.

Can you please include this function in the docs? I suggest adding it in `docs/cliques.xxml`. Feel free to propose a different location if you find one.

I'll get back to this later, including answering your questions.

szhorvat:
> I think that the result of this function should be cached in a cache entry "is_complete". If you agree, I will add a placeholder comment to remind myself to add caching once you're finished refactoring it.

I thought about this a bit, and it seems to me that this property would not be needed very often. If the graph is simple, caching that already achieved O(1). So caching whether the graph is complete is relevant only for multigraphs.

The original motivation for caching was to avoid running the most common checks multiple times. There are several functions which need to check if the graph is simple, or connected, before running their algorithm. Thus these checks were frequently run redundantly.

Do you have a use case in mind where it is important to not repeat complete graph checks on a multigraph?

src/properties/complete.c
```diff
@@ -0,0 +1,157 @@
...
+                                        IGRAPH_NO_LOOPS, IGRAPH_NO_MULTIPLE));
+
+        if ((igraph_vector_int_size(&neighbours) < vcount - 1)) {
+            *res = false;
+            goto cleanup;
+        }
+    }
+
+    /* If we arrive here, we have found no neighbour vector of size stricly less
+       than vcount - 1. The graph is therefore complete */
```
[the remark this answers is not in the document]
aagon: Corrected, thanks.

src/properties/complete.c
```diff
@@ -0,0 +1,157 @@
...
+ *
+ * A graph is considered complete if all pairs of different vertices are
+ * adjacent.
+ *
+ * The null graph and the singleton graph are considered complete.
+ *
+ * \param graph The graph object to analyse.
+ * \param res Pointer to a logical variable, the result will be stored here.
+ *
+ * \return Error code.
```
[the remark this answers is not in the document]
aagon: Done, thanks.

src/properties/complete.c
```diff
@@ -0,0 +1,157 @@
+/* -*- mode: C -*-  */
+/* vim:set ts=4 sw=4 sts=4 et: */
+
```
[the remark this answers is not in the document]
aagon: I don't, I merely found them in other source files and copied them.

aagon:
> Can you please include this function in the docs? I suggest adding it in `docs/cliques.xxml`.

Done. I'm fine with adding it there, but then maybe I should change the header file in which it is declared ? I'll let you decide.

> Can you please add a very brief comment about what this function does, especially starting at _ /* If directed, we must add the mutuals manually */_? I'm a bit confused about this part.

I refactored the unit test, added comments, and used the occasion to remove all the unneeded memory management and calls to `IGRAPH_CHECK`.

> If the graph is simple, caching that already achieved O(1).

Agreed. But in that case, we should try and cache the simplicity of a graph wherever we can. For instance, a graph generated with `igraph_grg_game` is necessarily simple, but this is not cached (I thought of this because this is how I generate my graphs, but there may be a lot of other places).

> Do you have a use case in mind where it is important to not repeat complete graph checks on a multigraph?

No. I do not use multigraphs. I agree with your decision to not cache the completeness. But then, see my other point about caching simplicity much more aggressively (I can help track the places where this can be done, if you want).

szhorvat:
@ntamas What do you think about adding an `igraph_bool_t` directed parameter here, and making it possible to treat directed graphs as undirected?  Clique functions ignore edge directions (although they do issue a warning). Perhaps there is value in making this possible?

ntamas:
> What do you think about adding an igraph_bool_t directed parameter here, and making it possible to treat directed graphs as undirected?

Seems OK to me, we do this in many other places. But I th9nk we can merge the current PR as is and then we can add this later, right?

ntamas (approved):
Okay, this looks good to me then.

szhorvat:
@aagon Would you be interested in implementing `igraph_is_clique()` as well?

PR #2523, opened 2024-03-01
feat: igraph_is_simple() uses the cache in a more granular way

szhorvat:
Asking for a review, since this touches code that sets the cache.

This PR makes more granular use of the cache in `igraph_is_simple()`. Previously, it would only use the cache when the status of _both_ HAS_LOOP and HAS_MULTI was known.

However, if either of these is true, we already know that the graph is not simple without having to check the other.

Similarly, we can set set the cache in a more granular way when either was found.

The current plan for property deduction is to NOT include a new flag for IS_SIMPLE, therefore we can handle this directly in the `igraph_is_simple()` function instead.

### DIFF
--- a/src/properties/multiplicity.c
+++ b/src/properties/multiplicity.c
@@ -50,47 +50,76 @@ igraph_error_t igraph_is_simple(const igraph_t *graph, igraph_bool_t *res) {
     igraph_integer_t vc = igraph_vcount(graph);
     igraph_integer_t ec = igraph_ecount(graph);
 
-    if (
-        igraph_i_property_cache_has(graph, IGRAPH_PROP_HAS_LOOP) &&
-        igraph_i_property_cache_has(graph, IGRAPH_PROP_HAS_MULTI)
-    ) {
-        /* use the cached result */
-        *res = (
-            !igraph_i_property_cache_get_bool(graph, IGRAPH_PROP_HAS_LOOP) &&
-            !igraph_i_property_cache_get_bool(graph, IGRAPH_PROP_HAS_MULTI)
-        );
-        return IGRAPH_SUCCESS;
+    /* Is it already known whether the graph has loops or multi-edges? */
+    igraph_bool_t known_loop, known_multi;
+
+    /* If it is known, does the graph have them? */
+    igraph_bool_t has_loop, has_multi;
+
+    known_loop  = igraph_i_property_cache_has(graph, IGRAPH_PROP_HAS_LOOP);
+    if (known_loop) {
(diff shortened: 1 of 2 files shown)


src/properties/multiplicity.c
```diff
@@ -50,46 +50,74 @@ igraph_error_t igraph_is_simple(const igraph_t *graph, igraph_bool_t *res) {
...
+        }
         igraph_vector_int_destroy(&neis);
         IGRAPH_FINALLY_CLEAN(1);
     }
 
-    /* If the graph turned out to be simple, we can cache that it has no loop
-     * and no multiple edges */
-    if (*res) {
+    if (known_loop) {
         igraph_i_property_cache_set_bool(graph, IGRAPH_PROP_HAS_LOOP, false);
```
ntamas, line 117: Should `false` be `has_loop` here instead?

src/properties/multiplicity.c
```diff
@@ -50,46 +50,74 @@ igraph_error_t igraph_is_simple(const igraph_t *graph, igraph_bool_t *res) {
...
+    }
+
+    if (known_multi) {
         igraph_i_property_cache_set_bool(graph, IGRAPH_PROP_HAS_MULTI, false);
```
ntamas, line 121: Should `false` be `known_multi` here instead?

src/properties/multiplicity.c
szhorvat: Yes, this is why I asked for a review.

src/properties/multiplicity.c
```diff
@@ -50,47 +50,75 @@ igraph_error_t igraph_is_simple(const igraph_t *graph, igraph_bool_t *res) {
...
         IGRAPH_FINALLY_CLEAN(1);
     }
 
-    /* If the graph turned out to be simple, we can cache that it has no loop
-     * and no multiple edges */
-    if (*res) {
-        igraph_i_property_cache_set_bool(graph, IGRAPH_PROP_HAS_LOOP, false);
-        igraph_i_property_cache_set_bool(graph, IGRAPH_PROP_HAS_MULTI, false);
+    if (known_loop) {
+        igraph_i_property_cache_set_bool(graph, IGRAPH_PROP_HAS_LOOP, has_loop);
```
ntamas, line 117: This branch will never be hit if `vc == 0 || ec == 0` and neither the other. Is this intentional? If it is, then I think the current commit is fine.

szhorvat:
@ntamas This is good to go now.

ntamas:
LGTM. One last CI check with an updated `master` and if it passes, I'll merge. Thanks!

PR #2527, opened 2024-03-02
fix: Replace check of constant with a static assertion

krlmlr:
Should the new `IGRAPH_STATIC_ASSERT` find a better home?

This is needed to avoid `R CMD check` warnings in R.

### DIFF
--- a/include/igraph_error.h
+++ b/include/igraph_error.h
@@ -794,9 +794,8 @@ IGRAPH_EXPORT int IGRAPH_FINALLY_STACK_SIZE(void);
 #define IGRAPH_CHECK_CALLBACK(expr, code) \
     do { \
         igraph_error_t igraph_i_ret = (expr); \
-        if (code) { \
-            *(code) = igraph_i_ret; \
-        } \
+        IGRAPH_STATIC_ASSERT(code); \
+        *(code) = igraph_i_ret; \
         if (IGRAPH_UNLIKELY(igraph_i_ret != IGRAPH_SUCCESS && igraph_i_ret != IGRAPH_STOP)) { \
             IGRAPH_ERROR("", igraph_i_ret); \
         } \
--- a/include/igraph_error.h
+++ b/include/igraph_error.h
@@ -794,7 +794,6 @@ IGRAPH_EXPORT int IGRAPH_FINALLY_STACK_SIZE(void);
 #define IGRAPH_CHECK_CALLBACK(expr, code) \
     do { \
         igraph_error_t igraph_i_ret = (expr); \
-        IGRAPH_STATIC_ASSERT(code); \
         *(code) = igraph_i_ret; \
         if (IGRAPH_UNLIKELY(igraph_i_ret != IGRAPH_SUCCESS && igraph_i_ret != IGRAPH_STOP)) { \
             IGRAPH_ERROR("", igraph_i_ret); \


krlmlr:
@GroteGnoom: `IGRAPH_STATIC_ASSERT()` currently lives in `hacks.h` . How to promote it so that we can use it internally but it's not exposed to the outside world?

ntamas:
Should we use it at all? It's a horrible hack whose only aim is to simulate `static_assert(...)` in pure C. If we can safely assume that the R interface is always compiled with a C++11 compiler, then we can simply use `static_assert(...)`.

How about this? We wrap `IGRAPH_STATIC_ASSERT` with an `#ifndef` block in `hacks.h`, thus allowing the user to override it. Then, in the R interface, we simply `#define IGRAPH_STATIC_ASSERT static_assert`.

krlmlr:
Can we add a `cmake` rule? The R client could pick it up from there: https://github.com/igraph/rigraph/issues/1189#issuecomment-1975602925

krlmlr:
Also, I feel these are two separate points:

1. How to use `IGRAPH_STATIC_ASSERT()` in this macro
2. How to implement `IGRAPH_STATIC_ASSERT()`

ntamas:
CMake already has a variable for checking whether the compiler has static assert support; see the `cxx_static_assert` flag [here](https://cmake.org/cmake/help/latest/prop_gbl/CMAKE_CXX_KNOWN_FEATURES.html). The problem is that we cannot simply replace `IGRAPH_STATIC_ASSERT` with `static_assert` in the C core because you cannot use ` static_assert` in C files.

Why does `R CMD check` complain about `IGRAPH_STATIC_ASSERT` at all?

krlmlr:
The CI/CD here complains about not finding `IGRAPH_STATIC_ASSERT()` . How to fix idiomatically?

krlmlr:
AFAIU, `IGRAPH_STATIC_ASSERT()` is how you implement static assertions in C. We need that in C code. I'd not count it as a hack.

`R CMD check` complains about checking a condition that is always true.

szhorvat:
This modification makes it invalid to use `NULL` for `code`. That is not unreasonable, since we'd always want to store the return code and check for `IGRAPH_STOP`.

I would skip the `IGRAPH_STATIC_ASSERT`, it's not strictly necessary. It's implementation is hack (with no better substitute) so I wouldn't want to expose it in a public header.

ntamas:
I don't mind exposing `IGRAPH_STATIC_ASSERT` in public headers. It seems portable enough. We can add it in a public header that is not included from `igraph.h` but has to be included explicitly.

krlmlr:
I'd need help with that.

ntamas:
Would that help if I moved the macro to a public header in the `master` branch?

krlmlr:
Yes, please. Thanks!

ntamas:
https://github.com/igraph/igraph/commit/a4f6a43c162bd68ac3b8db00b2b0a5d1c7855415

ntamas:
I'm not really worried about it. C11 has `_Static_assert()` anyway so it's only a concern for compilers that support C99 but not C11; we can switch to `_Static_assert()` for C11 if we want to.

szhorvat:
It seems to me that nothing more needs to be done here. Please re-open if I'm mistaken.

szhorvat:
I'm sorry, this is not actually done. @krlmlr @ntamas can you please complete it so we can merge and close?

ntamas:
I decided to remove the `IGRAPH_STATIC_ASSERT` block entirely from the expansion of `IGRAPH_CHECK_CALLBACK`, the reason being that having it there would require the user to `#include "internal/hacks.h"`. Essentially we were providing a macro in a public header that does not work without the additional inclusion of a private header.

Let's see if the CI passes and then I'll merge.

szhorvat:
> Essentially we were providing a macro in a public header that does not work without the additional inclusion of a private header.

You did add `IGRAPH_STATIC_ASSERT` to a public header though. Do we keep it there then?

(No strong opinion from me, but if you do ask me, I'd not make it public because it's essentially a hack.)

ntamas:
> You did add IGRAPH_STATIC_ASSERT to a public header though. Do we keep it there then?

No, that's exactly the point of my last commit: `IGRAPH_STATIC_ASSERT` is not referred to from any public headers any more.

szhorvat:
I think we have a misunderstanding.

You recently moved `IGRAPH_STATIC_ASSERT` to a public header, `igraph_misc.h`. It is still in that header now, and NOT in `internal/hacks.h`. Do we really want to keep it there, or do we move it back to an internal header?

This was your commit: https://github.com/igraph/igraph/commit/a4f6a43c162bd68ac3b8db00b2b0a5d1c7855415#diff-b884378d5b5db1523e2a568f53c474a82e0e257c9cee570ff2beed88dd8cd706

I am in favour of moving it back to the internal header. I don't want to commit to supporting the use of this macro for general users.

ntamas:
Hmmm yeah, strange, I was looking at the source tree in this PR and here the macro was definitely in `src/internal/hacks.h`. Okay, let's revert `a4f6a43`

szhorvat:
Can you please do that, and before you push, make sure that headers are updated for any more recently added usages of `IGRAPH_STATIC_ASSERT` (such as the one I added yesterday)?

PR #2528, opened 2024-03-05
Fix `-Wstrict-aliasing`

SoapGentoo:
* Casting a `uint64_t*` to `double*` invokes undefined behavior, since it violates the strict aliasing rules of ISO C. Instead of casting pointers, let's read through a union which is supported by C and yields the same performant assembly code.

Closes: https://bugs.gentoo.org/924864

### DIFF
--- a/src/random/random.c
+++ b/src/random/random.c
@@ -681,8 +681,13 @@ igraph_real_t igraph_rng_get_unif01(igraph_rng_t *rng) {
          * Then we subtract 1 to arrive at the [0; 1) interval. This is fast
          * but we lose one bit of precision as there are 2^53 possible doubles
          * between 0 and 1. */
-        uint64_t r = (igraph_i_rng_get_random_bits_uint64(rng, 52) & 0xFFFFFFFFFFFFFull) | 0x3FF0000000000000ull;
-        return *(double *)(&r) - 1.0;
+        union {
+            uint64_t as_uint64_t;
+            double as_double;
+        } value;
+        value.as_uint64_t =
+            (igraph_i_rng_get_random_bits_uint64(rng, 52) & 0xFFFFFFFFFFFFFull) | 0x3FF0000000000000ull;
+        return value.as_double - 1.0;
     }
 }
 


src/random/random.c
```diff
@@ -681,8 +681,12 @@ igraph_real_t igraph_rng_get_unif01(igraph_rng_t *rng) {
          * Then we subtract 1 to arrive at the [0; 1) interval. This is fast
          * but we lose one bit of precision as there are 2^53 possible doubles
          * between 0 and 1. */
-        uint64_t r = (igraph_i_rng_get_random_bits_uint64(rng, 52) & 0xFFFFFFFFFFFFFull) | 0x3FF0000000000000ull;
-        return *(double *)(&r) - 1.0;
+        union {
+            uint64_t as_uint64_t;
+            double as_double;
+        } value = {.as_uint64_t =
+            (igraph_i_rng_get_random_bits_uint64(rng, 52) & 0xFFFFFFFFFFFFFull) | 0x3FF0000000000000ull};
```
szhorvat, line 688: Please do not use designated initializers as they don't work in some versions of MSVC (as I recall they don't work in VS2015).

SoapGentoo: amended

szhorvat:
I'll note that recent versions of MSVC generate different code for this version on x64 than the plan pointer cast. There's one extra instruction, and as far as I can tell it's writing the value to the main memory and reading it back instead of using only registers. https://godbolt.org/z/WEWPsc3cK

A similar difference persists with the actual igraph use case as well.

SoapGentoo:
The current code invokes undefined behavior, and just because it hasn't failed catastrophically yet, doesn't mean you should continue using it. Correct, MSVC seems to be storing it to the stack and loading it back from the stack, mainly because its codegen is terrible. I **strongly** doubt this will have an impact on runtime performance, since modern CPUs partially virtualize the stack anyways.

If you want a simple example how violating strict aliasing breaks code, just look at https://godbolt.org/z/rE4sW5K53. `bar()` return `0.0` even though the last store to the pointer wrote `(uint64_t)1` to the memory location, and the compiler never read it back from memory, since casting `double*` to `uint64_t*` and dereferencing the pointer invokes UB. People think UB is some kind of esoteric concept, and code that works is "fine" since it has been tested. Invoking UB is never fine, and this code will break at some point in the future.

eli-schwartz:
> Correct, MSVC seems to be storing it to the stack and loading it back from the stack, mainly because its codegen is terrible. I **strongly** doubt this will have an impact on runtime performance, since modern CPUs partially virtualize the stack anyways.

Also mingw GCC and mingw clang both exist, and both don't have this problem...

genbtc (approved):
union good

Issue #2521, opened 2024-02-28
`minimum_size_separators()` gives invalid result


szhorvat:
**Describe the bug**

`igraph_minimum_size_separators()` sometimes gives invalid results.

Note that the bug fix #2509 affects this function, but does not appear to fix this particular issue. This is expected, since this function executes its main algorithm only on connected graphs, and #2509 addresses disconnected graphs.

**To reproduce**

In Python,

```python
g=igraph.Graph([[0,8],[0,9],[1,2],[1,4],[2,4],[2,5],[3,4],[3,8],[5,6],[5,7],[6,9],[7,8]])
g.minimum_size_separators()
```

The result is:

```
[[2, 4],
 [8, 9],
 [6, 8],
 [5, 8],
 [3, 9],
 [3, 6],
 [3, 5],
 [4, 8, 9],
 [4, 6, 8],
 [4, 5],
 [4, 8],
 [5, 9],
 [0, 5],
 [0, 6],
 [2, 8],
 [2, 3]]
```

First, these obviously cannot all be minimum size separators, since some have a larger size than others. They can't even be minimal, since some are subsets of others (compare `[4,8,9]` and `[4,8]`).

Second, some of these are not separators at all. The graph looks like this:

<img width="360" alt="image" src="https://github.com/igraph/igraph/assets/1212871/d1e34498-daa4-4191-b261-c2751dfa280f">

The removal of `[3,9]` and `[3,6]` does not disconnect the graph.

_Note:_ This example was found by fuzzing, and the fuzzer couldn't minimize it further.

**Version information**
Current `master`.

szhorvat:
Two more examples on 8 vertices, with 1-based vertex IDs:

```
{{2, 5}, {6, 7}, {5, 7}, {1, 6}, {4, 5}, {3, 4}, {4, 8}, {6, 8}, {1, 2}, {3, 8}}
```

<img width="360" alt="image" src="https://github.com/igraph/igraph/assets/1212871/8289e090-7c34-4c80-a163-07a77fbb5927">

```
{{2, 7}, {4, 8}, {3, 6}, {5, 8}, {5, 6}, {1, 2}, {2, 6}, {1, 5}, {3, 5}, {4, 7}, {5, 7}, {3, 4}, {1, 8}}
```

<img width="360" alt="image" src="https://github.com/igraph/igraph/assets/1212871/86c909dd-8825-41ec-ad73-ad92fbe14ba0">

szhorvat:
[...]

PR #2533, opened 2024-03-15
Fix `igraph_all_st_mincuts` to only return minimal cuts

### LINKED ISSUE #2532: `igraph_all_st_mincuts` does not return only minimum cuts
**Describe the bug**
`igraph_all_st_mincuts` returns non-minimum cuts.

**To reproduce**

```C
igraph_t graph;

// create a graph
igraph_small(&graph, 20, IGRAPH_DIRECTED,
    1, 11,
    3, 13,
    4, 14,
    5, 15,
    6, 16,
    7, 17,
    8, 18,
    9, 19,
    0, 8,
    18, 0,
    0, 9,
    19, 0,
    2, 1,
    11, 2,
    14, 1,
    11, 4,
    2, 4,
    14, 2,
    2, 5,
    15, 2,
    14, 3,
    13, 4,
    18, 3,
    13, 8,
    16, 5,
    15, 6,
    17, 5,
    15, 7,
    19, 6,
    16, 9,
    18, 7,
    17, 8,
    -1
);

igraph_vector_int_list_t stcuts;

igraph_vector_int_list_init(&stcuts, 0);

igraph_all_st_mincuts(&graph, /*value=*/ NULL,
                        /*cuts=*/ &stcuts,
                        /*partition1s=*/ NULL,
                        /*source=*/ 2,
                        /*target=*/ 0,
                        /*capacity=*/ NULL);


// print out the vector
printf("st-cuts:\n");
for (int i = 0; i < igraph_vector_int_list_size(&stcuts); i++)
{
    igraph_vector_int_t* cut = igraph_vector_int_list_get_ptr(&stcuts, i);
    printf("total cut capacity (%ld): ", igraph_vector_int_size(cut));
    igraph_vector_int_print(cut);
}
printf("\n");

igraph_vector_int_list_destroy(&stcuts);
igraph_destroy(&graph);
```

In particular, this produces a number of cuts of 3 edges, while other cuts have 2 edges. Since all cuts have to be a minimal size, this obviously should not be possible.

**Version information**

Latest `develop`, 9695ba1659217da1c3ebc1d47642ae6666329cb6.

vtraag:
This fixes #2532 by correcting two issues:
- The calculation of $I(S, v)$ was incorrect.
- The identification of minimal nodes was incorrect.

Preliminary checks show that the example reported in https://github.com/igraph/igraph/issues/2532#issuecomment-1997752428 is now processed correctly, and only correct, minimal cuts are found. It would be good to double check these fixes compared to the original publication [1]. In addition, it would be good to test a few more of the reported examples.

Finally, once these fixes check out, we should also add a simple test of this mincut functionality that checks two things: (1) if cuts are indeed correct cuts; and (2) if cuts are indeed minimal, or at least, compared to the reported minimal value.

References
[1] Provan, J. S., & Shier, D. R. (1996). A paradigm for listing (s, t)-cuts in graphs. Algorithmica, 15(4), 351–372. https://doi.org/10.1007/BF01961544

### DIFF
--- a/src/flow/st-cuts.c
+++ b/src/flow/st-cuts.c
@@ -31,6 +31,7 @@
 #include "igraph_operators.h"
 #include "igraph_stack.h"
 #include "igraph_visitor.h"
+#include "igraph_topology.h"
 
 #include "core/estack.h"
 #include "core/marked_queue.h"
@@ -1141,11 +1142,10 @@ igraph_error_t igraph_all_st_cuts(const igraph_t *graph,
    element from which 'v' is reachable. (Not necessarily through
    active vertices.)
 
-   We calculate the in-degree of all vertices in Sbar first. Then we
-   look at the vertices with zero in-degree. If these are active,
-   then they are minimal. If they are are not active, then we remove
-   them from the graph, and check whether they resulted in more
-   zero-indegree vertices.
+   We do so by traversing the nodes in topological sort order. The nodes that
+   are processed first and are not yet connected to any minimal nodes are then
+   marked as minimal (if active). Any subsequent nodes that are connected to
+   minimal nodes are then not marked as minimal.
 */
 
 static igraph_error_t igraph_i_all_st_mincuts_minimal(const igraph_t *Sbar,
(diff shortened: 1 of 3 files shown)


vtraag:
Preliminary checks suggest that this PR might also fix #2521, but this should again also be double checked. If this is the case, again, a test should still be added for issue #2521.

vtraag:
Only now notice that existing `igraph_all_st_mincuts` test. My guess is that the previous output was actually incorrect, but I'd have to check the results in more detail.

szhorvat:
I'll give this a bit of a stress test over the weekend through the Mathematica interface.

szhorvat:
I would need this to be moved to `master` before I can test it, but `master` was broken by the last commit (CC @ntamas). Perhaps you can base it on the next to last commit of `master`?

vtraag:
OK, I've now rebased onto 857a125069c226f266562b3781b82833fe1d59d9 (last commit in `master` that passed all tests).

vtraag:
(First time to use `git rebase --onto ...` which worked pretty easily :))

szhorvat:
Thank you! However, it turns out the problem wasn't the last commit after all. Could you rebase onto the very latest `master`, where I think I worked around the failure?

vtraag:
Sure, rebased against latest master (ce52cdaf4aca1b51e315a48f4aad7c5370673c62).

szhorvat:
Thanks! I'm now testing using the Mathematica interface, and will figure out if the old test is indeed incorrect.

szhorvat:
The 4th graph in the test is this, and we're looking to cut between `0` and `3`:

image

The original test expected `0 -> 2, 1 -> 3` as a cut. Obviously this is not a cut.

The other results are good.

----

Graph 7 is this, cutting between 0 and 8"

image

The following previously returned cuts were all bad: `{{0 -> 7, 4 -> 5}, {0 -> 7, 5 -> 3}, {0 -> 7, 3 -> 8}}`

----

The rest are all fine.

szhorvat (approved):
Thanks so much @vtraag ! I'll leave this to you to merge in case you decide to add more comments.

Can you please also update the changelog (either before or after the merge, whichever's more convenient)?

szhorvat:
> Finally, once these fixes check out, we should also add a simple test of this mincut functionality that checks two things: (1) if cuts are indeed correct cuts; and (2) if cuts are indeed minimal, or at least, compared to the reported minimal value.

Would be good to open an issue for this so it doesn't get forgotten.

szhorvat:
> Preliminary checks suggest that this PR might also fix #2521, but this should again also be double checked.

I verified the cases mentioned in #2521 and did some extra fuzzing. This also seems to be fine now.



> If this is the case, again, a test should still be added for issue #2521.

Yes, we need to do this, and add the test cases from #2521 into a unit test.

szhorvat:
Actually, I'm going to merge now this so the fuzzer can be happy and look for further issues. You can commit additional changes on `master`.

vtraag:
> Can you please also update the changelog (either before or after the merge, whichever's more convenient)?

Update the changelog in 770225d82163ef8f40f0cc82a3e6440b3739241a.

> > Finally, once these fixes check out, we should also add a simple test of this mincut functionality that checks two things: (1) if cuts are indeed correct cuts; and (2) if cuts are indeed minimal, or at least, compared to the reported minimal value.
> 
> Would be good to open an issue for this so it doesn't get forgotten.

I'll open an issue for this.

> > Preliminary checks suggest that this PR might also fix #2521, but this should again also be double checked.
> 
> I verified the cases mentioned in #2521 and did some extra fuzzing. This also seems to be fine now.
> 
> > If this is the case, again, a test should still be added for issue #2521.
> 
> Yes, we need to do this, and add the test cases from #2521 into a unit test.

We can close #2521 then now, right?

I'll also open a (separate) issue for adding tests.

vtraag:
Opened issue #2536 for adding tests (combined an issue for cuts / separators).

PR #2535, opened 2024-03-18
Improve efficiency of all_st_mincuts

vtraag:
Following up on #2533 , that fix repeatedly constructed a graph and calculated a topological sort order. This is not necessary, as it can be done simply only once. This PR improves that and should make the function somewhat (or perhaps even substantially) faster. In addition, it also simplifies the code, since it involves fewer mappings of vertices to track and understand.


ntamas (approved):
LGTM.

src/flow/st-cuts.c
```diff
@@ -1439,6 +1410,26 @@ igraph_error_t igraph_all_st_mincuts(const igraph_t *graph, igraph_real_t *value
...
 
+    /* We relabel the residual graph so that it is in topological sort order. */
+    igraph_vector_int_t order;
+    igraph_t tmpgraph;
+    IGRAPH_VECTOR_INT_INIT_FINALLY(&order, no_of_nodes);
+    IGRAPH_CHECK(igraph_topological_sorting(&residual, &order, IGRAPH_OUT));
+
+    // Relabel mapping
+    for (i = 0; i < igraph_vcount(&residual); i++)
+        VECTOR(NtoL)[i] = VECTOR(order)[VECTOR(NtoL)[i]];
```
szhorvat, line 1421:
```suggestion
    for (i = 0; i < igraph_vcount(&residual); i++) {
        VECTOR(NtoL)[i] = VECTOR(order)[VECTOR(NtoL)[i]];
    }
```

Nitpick: We always use braces after `for`, `if`, `while`, etc. It's not my personal preference either, but since we were doing it consistently, let's keep to it.

szhorvat:
I don't want to hold this up. If Tamás already checked it, and there's no special reason for me to take a careful look, please just go ahead and merge, Vincent.

Re the code style comment: No big deal, we can run it through astyle later.

vtraag:
I've run it through astyle and amended the commit, I'm merging it now.

PR #2541, opened 2024-03-22
Fix efficient st-cuts

### LINKED ISSUE #2539: igraph_minimum_size_separators() produces vertex id's that are not in the graph
**Describe the bug**
igraph_minimum_size_separators() sometimes produces vertices that are not in the graph.

**To reproduce**
Use this graph:
```
directed: false
vcount: 15
edges: {
8 7
6 2
12 2
14 0
5 2
9 3
10 4
13 5
10 7
11 6
11 8
11 4
9 4
12 1
13 6
13 12
1 0
14 3
}
```

This produces these vertex id lists, which includes '19', which is not in the graph.

```
1 14
3 19
1 9 14
3 9 19
1 4 14
4 19
1 4 11 14
1 11 12 14
4 11 19
11 12
1 6 12 14
6 12
0 12
9 14
4 14
9 11 14
0 9 11
4 11 14
0 11
0 6 9
0 6
9 11
3 11
6 9
3 6
8 10
4 8
10 11
4 11
4 6
7 11
3 4
4 7
0 3
2 13
```

**Version information**
commit 2b22f1eae60dc13ce0adb8a13f0921bd200a7c61
This bug seems to have been produced in d3bc59f9efaaaa1aa17d622d4a4ff1bedd6b1bc9

vtraag:
This fixes #2539, which is ultimately not a problem with `minimum_size_separators`, but with `all_st_mincuts`. There were two problems:
- The topological order wasn't properly used to permute the vertices. It used the order directly, while it should use the permutation that results in that sorted order (i.e. the inverse of that order).
- The `NtoL` mapping wasn't correct: 
  - It didn't consider all nodes, but only the (more limited number of) nodes in the condensed residual graph.
  - It didn't use the proper permutation (i.e. the inverse order), but simply the order.

This wasn't caught earlier in other tests, because in those examples, this problem happened to not surface. With this PR, it should now be correct, I believe.

### DIFF
--- a/src/flow/st-cuts.c
+++ b/src/flow/st-cuts.c
@@ -1410,25 +1410,36 @@ igraph_error_t igraph_all_st_mincuts(const igraph_t *graph, igraph_real_t *value
     IGRAPH_CHECK(igraph_simplify(&residual, /*multiple=*/ true, /*loops=*/ true, /*edge_comb=*/ NULL));
 
     /* We relabel the residual graph so that it is in topological sort order. */
+    igraph_integer_t no_of_nodes_residual = igraph_vcount(&residual);
     igraph_vector_int_t order;
     igraph_t tmpgraph;
-    IGRAPH_VECTOR_INT_INIT_FINALLY(&order, no_of_nodes);
+    IGRAPH_VECTOR_INT_INIT_FINALLY(&order, no_of_nodes_residual);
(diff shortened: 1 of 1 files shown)


szhorvat:
I verified that none of the three test cases the fuzzer found crash now.

PR #2549, opened 2024-03-27
Reachability sets of vertices in directed graphs

### LINKED ISSUE #1872: Wishlist: How many vertices are reachable from each vertex?

A function that counts how many vertices are reachable from each vertex in a directed graph.

Here is an example where every vertex is labelled with how many others are reachable from it:

image

**Can we devise a _fast_ (faster than `O(|V| ^2)`) algorithm that accomplishes this?**  

**Use cases for the feature**

 - This value is the same as the out-degree of vertices in the transitive closure of the graph. The transitive closure is often a dense graph that cannot be practically stored. 
 - It helps in finding the largest out-component of a graph.

**References**

 - TODO.

Tagl:
Closes #1872 

The implementation uses bitset logic and popcount operations.
There is an issue regarding compilers and/or CPU architectures to be handled.
Another thing to look at is whether the compiler is doing auto-vectorization for us and compiling to SIMD operations.
If not, it is worth specializing the bitset logic in the code whenever possible to improve performance even further by using the vectorization registers of CPUs.


include/igraph_reachability.h
```diff
@@ -0,0 +1,51 @@
...
+__BEGIN_DECLS
+
+#define BITMASK(b) (1 << ((b) % IGRAPH_INTEGER_SIZE))
+#define BITSLOT(b) ((b) / IGRAPH_INTEGER_SIZE)
+#define BITSET(a, b) ((a)[BITSLOT(b)] |= BITMASK(b))
+#define BITCLEAR(a, b) ((a)[BITSLOT(b)] &= ~BITMASK(b))
+#define BITTEST(a, b) ((a)[BITSLOT(b)] & BITMASK(b))
+#define BITNSLOTS(nb) ((nb + IGRAPH_INTEGER_SIZE - 1) / IGRAPH_INTEGER_SIZE)
+
+IGRAPH_EXPORT igraph_integer_t igraph_bitset_popcount(igraph_vector_int_t* bitset);
```
ntamas, line 40: It needs to be decided whether we want to make bitsets public. If we do, then we need to prefix the macros with `IGRAPH_`, otherwise other packages will complain if our macros conflict with theirs. If we don't, then we cannot put these in a public header; we need to move them to `reachability.c` or an internal `bitset.c` (with a corresponding internal header that is not part of the public interface).

src/misc/reachability.c
```diff
@@ -0,0 +1,122 @@
...
+   along with this program.  If not, see <https://www.gnu.org/licenses/>.
+*/
+
+#include "igraph_adjlist.h"
+#include "igraph_community.h"
+#include "igraph_components.h"
+#include "igraph_interface.h"
+#include "igraph_reachability.h"
+
+void _print_vector_int(const igraph_vector_int_t *v) {
```
ntamas, line 27: I guess this function was meant for debugging purposes. We have `igraph_vector_int_print()` so you can use that. This function has to be removed because it refers to `printf()` and the R interface maintainers will complain. R has very strict rules about what symbols may appear in extension modules and `printf()` is not allowed.

src/misc/reachability.c
```diff
@@ -0,0 +1,122 @@
...
+igraph_integer_t igraph_bitset_popcount(igraph_vector_int_t *bitset)
+{
+    igraph_integer_t i, n, count = 0;
+    n = igraph_vector_int_size(bitset);
+    for (i = 0; i < n; ++i)
+    {
+        // TODO: Check if __cpuid claims the operation is supported by CPU
+        // https://learn.microsoft.com/en-us/cpp/intrinsics/popcnt16-popcnt-popcnt64?view=msvc-170
+        #ifdef _MSC_VER
+        count += __popcnt64(VECTOR(*bitset)[i]);
```
ntamas, line 45: Ultimately we should test for `__popcnt64` and `__builtin_popcountll` with specific tests from CMake at the configuration stage and we should provide `HAVE_POPCNT64` and `HAVE_BUILTIN_POPCOUNTLL` macros in `config.h`.

Also, I was wondeirng whether this takes into account that igraph integers can be configured as 32-bit or 64-bit (these are the two possible values for `IGRAPH_INTEGER_SIZE`).

src/misc/reachability.c
```diff
@@ -0,0 +1,122 @@
...
+    igraph_vector_int_list_t *reach)
+{
+    igraph_integer_t no_of_nodes;
+    igraph_integer_t i, j, k, n;
+    igraph_adjlist_t adjlist, dag;
+    igraph_vector_int_t *dag_neighbours, *neighbours, *from_bitset, *to_bitset;
+
+    no_of_nodes = igraph_vcount(graph);
+
+    igraph_connected_components(graph, membership, csize, no_of_components, IGRAPH_STRONG);
```
ntamas, line 68: An `IGRAPH_CHECK` is missing from here.

src/misc/reachability.c
```diff
@@ -0,0 +1,122 @@
...
+
+    IGRAPH_CHECK(igraph_adjlist_init(graph, &adjlist, IGRAPH_OUT, IGRAPH_LOOPS_ONCE, IGRAPH_MULTIPLE));
+    IGRAPH_FINALLY(igraph_adjlist_destroy, &adjlist);
+
+    IGRAPH_CHECK(igraph_vector_int_list_init(reach, *no_of_components));
```
ntamas, line 73: I think we should handle gracefully when the user is not interested in the number of components and provides `NULL` as `no_of_components`. Actually, we should probably accept `NULL` for all arguments except `graph` and allocate our own data structures if needed.

ntamas: Note that the typical convention in igraph is that incoming arguments are already initialized and are owned by the caller. We should not re-initialize `reach` here; we should call `igaph_vector_int_list_resize()` instead. I believe this may be the reason for the memory leaks in the unit tests.

ntamas (commented):
Initial quick review. I know this is a draft, I just wanted to point out a few obvious places for improvement.

src/misc/reachability.c
Tagl: Removed

src/misc/reachability.c
Tagl: Added

src/misc/reachability.c
Tagl: Yeah that was a leftover from an older version, forgot to change to resize. Memory leak fixed now.

include/igraph_reachability.h
Tagl: Added prefix, will move later to its own file with more general use functions too.

src/misc/reachability.c
Tagl: No, it currently assumes 64-bit but I will change that.

I will add all those checks and fallback to implementation that doesn't require the popcount instruction when unavailable.
ARM and maybe M2 may also need specialization, not sure as I have not developed for those.

src/misc/reachability.c
Tagl: Note that without knowing membership, it's not particularly useful to get the reachability set. For optimization of the average case, I instead work with and return the reachability of each strongly connected component. So to obtain the value for a vertex you need both the reachability set of its component and the membership to determine the component. Perhaps this is not the API you would prefer, but it does save memory in the vast majority of cases.

src/misc/reachability.c
Tagl: The CPU checks may need to happen at runtime. The binary may be built on a different machine than runs it, right?

ntamas: The binary is built for a certain architecture and instruction set so these do not have to be runtime checks. It _may_ happen that you build the binary for an instruction set and run it on a machine that has a _wider_ instruction set (like, you can build for `armv6` and then run it on a machine that supports `armv7`), but this is rare. And I don't think it's even possible to perform this check in runtime.

Tagl: It is possible to perform the checks at runtime using `__cpuid` or `__builtin_cpu_supports` for example.
Then using a dispatch function call to set the function pointer to the correct function on its first run you can ensure that the correct function is called depending on what features are available. This way you do not waste CPU instructions on determining the correct implementation except on the first call to the function.

szhorvat: There's `__builtin_popcountg()` for generic integers (note that `igraph_integer_t` may not be a `long long`).

szhorvat: Annoyingly, `__builtin_popcountg()` is for the unreleased GCC 14 and Clang 19 only. So this is not an option.

include/igraph_reachability.h
Tagl: Move to its own file.

src/misc/reachability.c
Tagl: Luckily the functions are simple enough that the compilers (clang and gcc) are able to autovectorize given the correct compiler flags. Not sure about MSVC but won't spend any effort on it yet.

So now the only outstanding issue is handling when the operations are not available for MSVC.

Tagl: I added some checks in cmake for the MSVC implementation.

Tagl:
@szhorvat @ntamas This is ready for review now.

szhorvat:
Before continuing, could you please do the following?

 - Let's add `\experimental` to the documentation of every new function / macro. This should come below the `\brief` line. The `\brief` line should be followed by an empty line (of course with the usual leading `*`)
 - Let's rename `igraph_count_reachable_directed()` to `igraph_count_reachable()` and add support for undirected graphs. This is for the sake of high-level interfaces, where it's convenient to expose both the directed and undirected versions as a single function. Let's add an `igraph_bool_t directed` parameter: when this is `false`, directed inputs are treated as undirected.  The implementation should be simple using the `membership` and `csize` output parameters of `igraph_connected_components()`.
 - In the docs of this function, mention that each vertex is considered reachable from itself.

szhorvat:
@ntamas Regarding the `igraph_bool_t directed` parameter that I proposed: There are many functions that use such a parameter to "ignore edge directions". Here `igraph_connectedness_t` (weak or strong) would also fit. Which one shall we use? Does `igraph_connectedness_t` even make sense or should it be replaced with an `igraph_bool_t directed` in 1.0, for consistency with other functions?

Tagl:
> Before continuing, could you please do the following?
> 
>     * Let's add `\experimental` to the documentation of every new function / macro. This should come below the `\brief` line. The `\brief` line should be followed by an empty line (of course with the usual leading `*`). With this we reserve the right to make API-breaking changes in the future.
> 
>     * Let's rename `igraph_count_reachable_directed()` to `igraph_count_reachable()` and add support for undirected graphs. This is for the sake of high-level interfaces, where it's convenient to expose both the directed and undirected versions as a single function. Let's add an `igraph_bool_t directed` parameter: when this is `false`, directed inputs are treated as undirected.  The implementation should be simple using the `membership` and `csize` output parameters of `igraph_connected_components()`.
> 
>     * In the docs of this function, mention that each vertex is considered reachable from itself.

Done

ntamas:
> Which one shall we use?

idk; to be honest I'm okay with both. Whichever you choose, just let's stick to it afterwards.

szhorvat:
Would it make sense to introduce a "bit matrix" type, intended to be used to represent adjacency matrices of simple graphs? This is precisely what we have in the reachability function, and LAD uses such an object too internally. I'm not sure how often this would be used, though, so let's keep things as they are in this PR.

src/core/bitset.c
```diff
@@ -0,0 +1,648 @@
...
+    IGRAPH_ASSERT(bitset != NULL);
+    IGRAPH_ASSERT(bitset->stor_begin != NULL);
+    IGRAPH_CHECK(igraph_bitset_reserve(bitset, new_size));
+
+    for (igraph_integer_t i = bitset->size; i % IGRAPH_INTEGER_SIZE != 0; ++i) {
+        IGRAPH_BIT_CLEAR(*bitset, i);
+    }
+    for (igraph_integer_t i = IGRAPH_BIT_NSLOTS(bitset->size); i < IGRAPH_BIT_NSLOTS(new_size); ++i) {
+        VECTOR(*bitset)[i] = 0;
+    }
```
szhorvat, line 372: This could be done more efficiently with a `memset()`

src/core/bitset.c
```diff
@@ -0,0 +1,648 @@
...
+ * \brief Resizes the bitset.
+ *
+ * \experimental
+ *
+ * </para><para>
+ * Note that this function does not free any memory, just sets the
+ * size of the bitset to the given one. It may, on the other hand,
+ * allocate more memory if the new size is larger than the previous
+ * one. In this case the newly appeared elements in the bitset are
+ * \em not set to zero, they are uninitialized.
```
szhorvat, line 343: The docs claim that the newly added bits are uninitialized, but the function actually zeros them. While this is different from what `igraph_vector` functions do, I think it is a good thing, as manual zeroing is not trivial. So let's keep the zeroing (agreed @ntamas ?) and change the docs.

szhorvat:
Coverity is unhappy about shifting by the width of our integer. Can you look into whether the concern is of significance, @Tagl ?

image

szhorvat:
P.S. Travis is up and running again and all checks are passing on exotic platforms (as you can see).

szhorvat:
Last commit fixes the second Coverity issue.

Tagl:
> Last commit fixes the second Coverity issue. Coverity is pretty good, isn't it?
> 
> UPDATE: Actually, it fixes both. I though we were shifting an `int` by 32 bits, as I didn't notice the `-1`. We're good.

Yeah that should fix the issue. I seem to have forgotten popcount when I fixed this issue in the other counting functions.

src/core/bitset.c
```diff
@@ -0,0 +1,649 @@
...
+ *
+ * Time complexity: O(n/w).
+ */
+
+void igraph_bitset_or(igraph_bitset_t *dest,
+                      const igraph_bitset_t *src1, const igraph_bitset_t *src2) {
+    for (igraph_integer_t i = 0; i < IGRAPH_BIT_NSLOTS(dest->size); ++i) {
+        VECTOR(*dest)[i] = VECTOR(*src1)[i] | VECTOR(*src2)[i];
+    }
+}
```
szhorvat, line 576: This and other similar functions should check that the two sources have compatible sizes.

szhorvat, line 576: @ntamas Notice that these bitset functions are different from vector functions in that:

 - `vector_prod(a, b)` does `a *= b`
 - `bitset_or(a, b, c)` does `a = b | c`

An possible advantage is that when one does need the result to go in a new bitset, no copying is necessary. When not, `bitset_or(a, a, b)` is valid. It's unclear how this sort of aliasing interferes with the compiler's ability to optimize, if at all.

The choice between the two kinds of interfaces has an implication on error checking, which is currently missing and should be added. If keeping the current interface, the natural thing to do is to:

 1. check that `src1` and `src2` are of the same size and 
 2. resize `dist` (which is assumed to be initialized) to be the same size

If we do this, that interferes with the ability to use `bitset_or(a, a, b)`. The alternative is to keep these functions very low level and require the user to ensure that sizes are compatible. However, this would be quite different from how other igraph functions work.

----

After thinking this through, I'm leaning towards making the interface be just like `vector_add()`, i.e. `bitset_or(a, b)` should do `a |= b`.

Any comments @ntamas, @Tagl ?

Tagl: There are cases where having three parameters saves you a copy operation. Therefore I believe there should be an option for that. Resizing makes sense to me if we don't leave it to the user to track the sizes. 

Most cases when working with bitsets you specify the size at compile time or runtime but the sizes of all your bitsets are the same. Therefore leaving the size check to the user means that they can simply skip the size checks entirely in most cases.

However, looking at boost::dynamic_bitset, it only provides the in place operators and also says a requirement for calling the functions is the size is equal, so they leave checking to the user.

ntamas: No strong opinions from my side; I don't think we need to keep vectors and bitsets consistent, though, so feel free to do it either way.

szhorvat: The decision was to leave this as-is, treating these functions as "low-level". So the interface can stay the same for flexibility, and there won't be any error checks. This should be clearly document, as it will be surprising the users of the library.

@Tagl Can you write a couple of sentences and paste them into the docs of each relevant function please?

src/core/bitset.c
szhorvat: @Tagl Can you update the doc accordingly please?

szhorvat:
I suggest adding two more functions here:

 - `igraph_bitset_print()`, to print as zeros and ones with no spaces, and document the printing order 
 - `igraph_transitive_closure(const igraph_t *graph, igraph_t *closure);` creates a transitive closure as a graph object `closure`. Create the list of edges and call `igraph_create()` on them. The directedness of the result should be the same as the directedness of the input graph. This can be implemented in a couple of lines at most and provides a bit of extra testing of the bitset functionality.

doc/bitset.xxml
```diff
@@ -0,0 +1,53 @@
...
+<!-- doxrox-include igraph_bitset_xor -->
+<!-- doxrox-include igraph_bitset_not -->
+<!-- doxrox-include igraph_bitset_popcount -->
+<!-- doxrox-include igraph_bitset_countl_zero -->
+<!-- doxrox-include igraph_bitset_countl_one -->
+<!-- doxrox-include igraph_bitset_countr_zero -->
+<!-- doxrox-include igraph_bitset_countr_one -->
+</section>
+
+<section id="bitset-properties"><title>Vector properties</title>
```
szhorvat, line 43:
```suggestion
<section id="bitset-properties"><title>Bitset properties</title>
```

src/core/bitset.c
Tagl: Updated

src/core/bitset.c
Tagl: Updated

src/core/bitset.c
Tagl: Updated

doc/bitset.xxml
Tagl: Fixed

szhorvat:
> `igraph_bitset_print()`, to print as zeros and ones with no spaces, and document the printing order

Let me clarify this. Let's have:

```c
igraph_error_t igraph_bitset_fprint(const igraph_bitset_t *bitset, FILE *file);
igraph_error_t igraph_bitset_print(const igraph_bitset_t *bitset) {
    return igraph_bitset_fprint(bitset, stdout);
}
```

The definition and declaration of `igraph_bitset_print()` should be guarded by 

```c
#ifndef USING_R

#endif
```

because CRAN forbids the use of `stdio`. This does not apply to `igraph_bitset_fprint()`.

It is okay to use `fprintf()` or `fputc()` / `fputs()`.  Only the `f`-less versions are forbidden by CRAN.

szhorvat:
About the transitive closure, a question is:

 1. Do we add edges to (a copy of) the existing graph?
 2. Or do we create an entirely new simple graph?

I suggest to go with (2), and document this. This is simpler and probably sufficient for most applications.

There is a difference between (1) and (2) when the original graph had self-loops or multi-edges, or when it had stored attributes.

szhorvat:
Note: We do seem to have this currently undocumented function:

```
igraph_error_t igraph_transitive_closure_dag(const igraph_t *graph, igraph_t *closure)
```

This wasn't touched for a long time, and worked only for DAG. We can probably remove it in the upcoming 1.0, and replace it with this new implementation that works for all graphs not just DAGs.

ntamas:
Yes, let's go with option (2), i.e. creating an entirely new simple graph, with a signature identical to `igraph_transitive_closure_dag()`. `igraph_transitive_closure_dag()` can be deprecated when we have the new function.

include/igraph_reachability.h
```diff
@@ -0,0 +1,54 @@
...
+
+IGRAPH_EXPORT igraph_error_t igraph_count_reachable(
+    const igraph_t *graph,
+    igraph_vector_int_t *counts,
+    igraph_bool_t directed);
+
+
+IGRAPH_EXPORT igraph_error_t igraph_transitive_closure(
+    const igraph_t *graph,
+    igraph_t* closure);
```
Tagl, line 50:
```suggestion
    igraph_t* closure,
    igraph_bool_t directed);
```

szhorvat:
@Tagl Two things I wanted to point out:

 - Functions should allow omitting an output argument (by setting it to `NULL`) when this makes sense
 - Prefer macros like `IGRAPH_VECTOR_INT_INIT_FINALLY` instead of separate `init` and` FINALLY`

Tagl:
> Two things I wanted to point out:
> 
>     * Functions should allow omitting an output argument (by setting it to `NULL`) when this makes sense
> 
>     * Prefer macros like `IGRAPH_VECTOR_INT_INIT_FINALLY` instead of separate `init` and` FINALLY`

Noted

src/misc/reachability.c
szhorvat: I agree that output parameters should almost always be optional, and this is now implemented. The one exception is `membership`, which is necessary to interpret the result, as Arnar said.

src/connectivity/reachability.c
```diff
@@ -0,0 +1,247 @@
...
+    IGRAPH_CHECK(igraph_adjlist_init_empty(&dag, no_of_comps));
+    IGRAPH_FINALLY(igraph_adjlist_destroy, &dag);
+
+    for (igraph_integer_t v = 0; v < no_of_nodes; v++) {
+        const igraph_vector_int_t *neighbours = igraph_adjlist_get(&adjlist, v);
+        igraph_vector_int_t *dag_neighbours = igraph_adjlist_get(&dag, VECTOR(*membership)[v]);
+        const igraph_integer_t n = igraph_vector_int_size(neighbours);
+        for (igraph_integer_t i = 0; i < n; i++) {
+            if (VECTOR(*membership)[v] != VECTOR(*membership)[VECTOR(*neighbours)[i]]) {
+                igraph_vector_int_push_back(dag_neighbours, VECTOR(*membership)[VECTOR(*neighbours)[i]]);
```
ntamas, line 117: Aren't we missing an `IGRAPH_CHECK()` here?

szhorvat: @Tagl Could you please fix this? All function that return an error code should have an `IGRAPH_CHECK`, _unless_ we know for certain that they cannot fail in this specific use case.

Tagl:
```suggestion
                IGRAPH_CHECK(igraph_vector_int_push_back(dag_neighbours, VECTOR(*membership)[VECTOR(*neighbours)[i]]));
```

ntamas (approved):
Suggest merging upon CI pass. Thanks a lot!

szhorvat:
It would have been better to squash-and-merge this given the messy commit history.

PR #2530, opened 2024-03-09
New feature: maximum matching of general graph

EdgarAlenPoe:
Implementing the blossom algorithm for finding maximum cardinality matchings in non-bipartite graphs.
As per issue [#2403](https://github.com/igraph/igraph/issues/2403)

PR #2550, opened 2024-03-28
 New feature: maximum matching of general graph

### LINKED ISSUE #2403: Find a maximum matching of a non-bipartite graph
A maximum matching (also known as maximum-cardinality) matching is a matching that contains the largest possible number of edges. Now, _igraph/c_ only has `igraph_maximum_bipartite_matching`, not for non-bipartite graphs.

**References**

1. https://en.wikipedia.org/wiki/Blossom_algorithm

elena-schafer:
The new pull request after the issues with #2530

Fixes #2403


szhorvat:
@ntamas The interface here is based on the stub function that was accidentally left over in 0.10 (though not exposed). As I said in chat, I am not sure that using `m[v] = u` to represent the `u-v` match is a good idea. In fact, this representation is plainly insufficient for weighted multigraphs, as we need to distinguish between parallel edges.

We certainly need a way to return specific edge IDs.

We should discuss what representation is best to use, and change the interface before 1.0 is relesaed. This should probably stay an experimental function in 1.0 as well, so we have the opportunity to make changes in the future.

ntamas:
Yes, one would need the edge IDs in weighted multigraphs -- but only in weighted multigraphs and not in weighted simple graphs or non-weighted multigraphs, where the matching vector is enough.

One option would be provide the edge IDs only in the result, and provide a helper function that converts the edge IDs to a matching vector if this is what the user ultimately needs. Another option would be to provide two output arguments: one for the matching vector (as we do it now) and one for the edges in the matching. I think that if we think about higher-level interfaces only, then providing the edge IDs is enough as the higher-level interface can easily wrap the edge IDs in a purpose-built object (like a `Matching` class) that provides easy access to the matching as well if the user wants that. When using igraph in C, the user would need to jump through an extra  hoop to get the matching vector if we provide the edge IDs only. On the other hand, people working directly in C are probably not scared of the extra complexity.

All in all, I'm okay with changing the interface to provide the IDs of the edges in the matching only, as long as we also provide a helper function to convert it quickly to a matching vector.

src/misc/matching.c
```diff
@@ -1018,6 +1018,450 @@ static igraph_error_t igraph_i_maximum_bipartite_matching_weighted(
...
 
+static igraph_error_t igraph_i_maximum_matching_find_aug_path(const igraph_t *graph, igraph_vector_int_t *matching, igraph_vector_int_t *path);
+static igraph_error_t igraph_maximum_matching_unweighted(const igraph_t *graph, igraph_integer_t *matching_size, igraph_vector_int_t *matching);
+static void igraph_i_maximum_matching_get_tree_path(igraph_integer_t *current, igraph_vector_bool_t *f_is_root, igraph_vector_int_t *contraction,
+        igraph_vector_int_t *distance, igraph_vector_int_t *f_parent, igraph_vector_int_list_t *b_lists, igraph_vector_int_t *path);
+
+/**
+ * \function igraph_maximum_matching
+ * \brief Calculates a maximum matching in a graph.
+ *
```
szhorvat, line 1029:
```suggestion
 *
 * \experimental
 *
```

Just so we don't forget.

EdgarAlenPoe:
The algorithm itself uses the `m[v] = u` approach internally, and it's probably necessary.
This makes me kinda favor the two output arguments approach, as we already have the one.
However this isn't the most clean for the weighted multigraph case as there's an argument that becomes unnecessary.

I'm really fine with either, and will get started implementing it once we decided this.

szhorvat:
We didn't forget this PR, just need to find the time to go over it. It's a non-trivial amount of work.

The `m[v] = u` approach is fine internally, but the interface can still be such that returning an edge set is possible. 

Changing the interface is relatively easy. The first step of the review should be verifying correctness and performance.

src/misc/matching.c
```diff
@@ -1007,6 +1007,450 @@ static igraph_error_t igraph_i_maximum_bipartite_matching_weighted(
...
+#ifdef MATCHING_DEBUG
+    debug("Path:");
+    igraph_vector_int_print(&path);
+#endif
+
+    // augment matching along path
+    // repeat until no augmenting path is found
+    while (igraph_vector_int_size(&path) != 0) {
+        //augment
+        for (igraph_integer_t i = 0; i + 2 < igraph_vector_int_size(&path); i = i + 2) {
```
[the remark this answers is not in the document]
szhorvat: Well, this is _almost_ correct. It is indeed good to save the vector length and avoid repeated `igraph_vector_int_size()` calls in the `for` loop. But of course the same shouldn't be done for the `while` loop, as the vector size keep changing there.

src/misc/matching.c
```diff
@@ -1007,6 +1007,450 @@ static igraph_error_t igraph_i_maximum_bipartite_matching_weighted(
...
+    // input validation
+    if (weights && igraph_vector_size(weights) < igraph_ecount(graph)) {
+        IGRAPH_ERROR("weights vector too short", IGRAPH_EINVAL);
+    }
+
+    // delegate to correct subroutine
+    if (weights != NULL) {
+        return IGRAPH_UNIMPLEMENTED;
+    } else {
+        igraph_maximum_matching_unweighted(graph, matching_size, matching);
```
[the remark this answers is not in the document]
szhorvat: Good point.

src/misc/matching.c
```diff
@@ -1007,6 +1007,450 @@ static igraph_error_t igraph_i_maximum_bipartite_matching_weighted(
...
+    igraph_vector_bool_destroy(&b_mark);
+    igraph_matrix_bool_destroy(&e_mark);
+    igraph_vector_int_list_destroy(&b_lists);
+    igraph_vector_int_destroy(&temp_list);
+    IGRAPH_FINALLY_CLEAN(12);
+
+    return IGRAPH_SUCCESS;
+}
+
+void igraph_i_maximum_matching_get_tree_path(igraph_integer_t *current, igraph_vector_bool_t *f_is_root, igraph_vector_int_t *contraction,
```
szhorvat, line 1410: This function is `void` instead of returning `igraph_error_t` and the `push_back()` calls within are not checked for errors. Are we certain that these `push_back()` calls cannot fail, i.e. are they acting on a vector with guaranteed sufficient reserved storage? If yes, this should be explained in a comment. If not, error checks are necessary.

src/misc/matching.c
```diff
@@ -1007,6 +1007,450 @@ static igraph_error_t igraph_i_maximum_bipartite_matching_weighted(
...
+                                // do by traversing again, but stopping when blossom marks stop
+                                current = v;
+                                while (current != -1 && VECTOR(b_mark)[current]) {
+                                    VECTOR(contraction)[current] = blossom_root;
+                                    VECTOR(b_mark)[current] = false;
+                                    igraph_vector_int_push_back(list, current);
+                                    current = VECTOR(f_parent)[current];
+                                }
+                                current = w;
+                                igraph_vector_int_resize(&temp_list, 0);
```
szhorvat, line 1372: Generally, all functions that return an error code must be protected using `IGRAPH_CHECK`.

src/misc/matching.c
```diff
@@ -1007,6 +1007,450 @@ static igraph_error_t igraph_i_maximum_bipartite_matching_weighted(
...
+
+    // edge mark matrix
+    igraph_matrix_bool_t e_mark;
+    IGRAPH_CHECK(igraph_matrix_bool_init(&e_mark, n, n));
+    IGRAPH_FINALLY(igraph_matrix_bool_destroy, &e_mark);
+
+    // blossom mark vector, used to find the root of a blossom
+    igraph_vector_bool_t b_mark;
+    IGRAPH_CHECK(igraph_vector_bool_init(&b_mark, n));
+    IGRAPH_FINALLY(igraph_vector_bool_destroy, &b_mark);
```
szhorvat, line 1190: Such pairs can be replaced with the more concise:

```suggestion
    IGRAPH_VECTOR_BOOL_INIT_FINALLY(&b_mark, n);
```

Boolean vector that are not resized can be replaced with a more space-efficient `igraph_bitset_t`. This is a minor nitpick that is better fixed only when everything else is addressed.

examples/simple/igraph_maximum_matching.c
```diff
@@ -0,0 +1,45 @@
+/* -*- mode: C -*-  */
+/*
+   IGraph library.
+   Copyright (C) 2022  The igraph development team <igraph@igraph.org>
```
szhorvat:
```suggestion
   Copyright (C) 2025  The igraph development team <igraph@igraph.org>
```

tests/unit/igraph_maximum_matching.c
```diff
@@ -0,0 +1,216 @@
+/*
+   IGraph library.
+   Copyright (C) 2022  The igraph development team <igraph@igraph.org>
```
szhorvat:
```suggestion
   Copyright (C) 2025  The igraph development team <igraph@igraph.org>
```

szhorvat:
I had a quick look while testing Copilot, but unfortunately I still don't have the capacity to review this PR in full at this moment. I hope that you are interested in coming back to it eventually, but there is no need to address the comments just yet.

ntamas:
This is not API-breaking so we do not need to merge this before the release of 1.0. Unscheduling from the milestone to keep things tidy.

PR #2553, opened 2024-03-31
Fix crash in igraph_rewire() on graph with too few edges

### LINKED ISSUE #2552: `rewire()` sometimes crashes when used on edgeless graph
**Describe the bug**

`rewire()` sometimes crashes when used on edgeless graph (depending on the random seed).

**To reproduce**

In Mathematica,

```
In[2]:= IGRewire[IGEmptyGraph[10], 1]

During evaluation of In[2]:= IGraphM::error: src/graph/type_common.c:123 - Invalid edge ID when retrieving edge endpoints.

During evaluation of In[2]:= IGraphM::error: igraph returned with error: Invalid value.

Out[2]= $Failed
```

This may actually result in a proper crash instead of an error.

A sanitizer version in C sometimes gives:

```
Assertion failed: (h >= l), function igraph_rng_get_integer, file random.c, line 571.
```

from

```
    #8 0x1024b0a20 in igraph_i_rewire rewire.c:101
    #9 0x1024b2578 in igraph_rewire rewire.c:266
```

**Version information**
Current `master`

szhorvat:
Fixes #2552

Furthermore, `igraph_rewire()` no longer throws an error for graphs with fewer than 4 vertices. It simply returns them unchanged, just like any other graph in which no rewiring is possible. These are graphs which are a unique realization of their degree sequence, called threshold graphs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,11 @@
  - `igraph_all_st_mincuts()` now correctly returns all minimum cuts. This also fixes a problem with `igraph_minimum_size_separators()`.
  - Corrected minor error in `igraph_community_label_propagation()` when adding labels to isolated nodes with some fixed labels present.
  - `igraph_community_spinglass()` no longer crashes when passing an edgeless graph and an empty weight vector.
+ - `igraph_rewire()` no longer crashes on graphs with fewer than two edges.
+
+### Changed
+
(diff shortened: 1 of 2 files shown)


ntamas (approved):
In some sense the fact that we no longer return with `IGRAPH_EINVAL` for graphs with less than 4 vertices is a breaking change, but I don't think we should move this to `develop` just because of this.

PR #2572, opened 2024-04-17
Fix weighted graph support for `igraph_community_edge_betweenness()`

### LINKED ISSUE #2229: Proper weight support for igraph_community_edge_betweenness()

szhorvat:
Fixes #2229

### DIFF
--- a/src/community/edge_betweenness.c
+++ b/src/community/edge_betweenness.c
@@ -319,18 +319,28 @@ igraph_error_t igraph_community_eb_get_merges(const igraph_t *graph,
     return IGRAPH_SUCCESS;
 }
 
-/* Find the smallest active element in the vector */
-static igraph_integer_t igraph_i_vector_which_max_not_null(const igraph_vector_t *v,
-                                                   const bool *passive) {
-    igraph_integer_t which, i = 0, size = igraph_vector_size(v);
+/* Find the index i for which v[i] / w[i] is the largest
+ * and the corresponding element is active (i.e. !passive[i]).
+ * If w is a null pointer then all w[i] are assumed to be 1,
+ * and the ranking is done solely based on v.
+ *
+ * This function requires that at least one element is active. */
+static igraph_integer_t igraph_i_which_max_active_ratio(
+        const igraph_vector_t *v,
+        const igraph_vector_t *w,
+        const bool *passive) {
+
+    const igraph_integer_t size = igraph_vector_size(v);
+    igraph_integer_t which, i = 0;
     igraph_real_t max;
+
     while (passive[i]) {
         i++;
     }
     which = i;
-    max = VECTOR(*v)[which];
+    max = VECTOR(*v)[which] / (w ? VECTOR(*w)[which] : 1.0);
     for (i++; i < size; i++) {
-        igraph_real_t elem = VECTOR(*v)[i];
+        igraph_real_t elem = VECTOR(*v)[i] / (w ? VECTOR(*w)[i] : 1.0);
         if (!passive[i] && elem > max) {
             max = elem;
             which = i;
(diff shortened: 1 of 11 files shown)


src/community/edge_betweenness.c
```diff
@@ -319,18 +319,26 @@ igraph_error_t igraph_community_eb_get_merges(const igraph_t *graph,
...
 
-/* Find the smallest active element in the vector */
-static igraph_integer_t igraph_i_vector_which_max_not_null(const igraph_vector_t *v,
-                                                   const bool *passive) {
-    igraph_integer_t which, i = 0, size = igraph_vector_size(v);
+/* Find the index i for which v[i] / w[i] is the smallest
+ * and the corresponding element is active (i.e. !passive[i]).
+ * If w is a null pointer then all w[i] are assumed to be 1.0.
+ * Assumes that at least one element is active. */
+static igraph_integer_t igraph_i_vector_which_max_not_null(
```
ntamas, line 326: Should be renamed to `igraph_i_vector_pair_find_max_ratio` or something like that as we now have two vectors and we are searching for the maximum in the ratio of them.

ntamas (changes requested):

ntamas:
LGTM now. Feel free to merge.

szhorvat:
@ntamas I force pushed two changes. You can use the "compare" link on the GitHub interface to see what changed.

The first one renames the function, and adds a second entry in the changelog. The second rebases on `develop` and fixes conflicts.

szhorvat:
@vtraag I merged this, but if you have concerns, we can still make changes.

PR #2574, opened 2024-04-17
refactor!: minimum_size_separators() no longer returns any separators for complete graphs

### LINKED ISSUE #1910: minimum_size_separators() questionable result on complete graphs
**Describe the bug**

`igraph_minimum_size_separators()` returns questionable results on complete graphs. It returns all size-`(n-1)` vertex subsets.

In Mathematica:

```
In[114]:= IGMinimumSeparators[CompleteGraph[3]]
Out[114]= {{2, 3}, {1, 3}, {1, 2}}

In[115]:= IGMinimumSeparators[CompleteGraph[4]]
Out[115]= {{2, 3, 4}, {1, 3, 4}, {1, 2, 4}, {1, 2, 3}}
```

I would not consider these separators and indeed `igraph_all_minimal_st_separators()` does not return them. I would expect the result of `igraph_minimum_size_separators()` to be a subset of that of `igraph_all_minimal_st_separators()`.

Thus a good result would be empty. However, the result returned here is consistent with that of `igraph_vertex_connectivity()`

Opinions @vtraag?

Before starting to work on this issue, be sure to get familiar with #1480.

**Version information**
Tested with `master` through the Mathematica interface.

### LINKED ISSUE #2518: `minimum_size_separators()` returns no results for disconnected graphs
`igraph_minimum_size_separators()` returns no separators for disconnected graphs.

It checks the vertex connectivity at the beginning (which is BTW a severe bottleneck, see #1328), and if it is zero, it returns immediately.

This is not invalid, but it is inconsistent with the behaviour of `igraph_all_minimal_st_separators()` and `igraph_is_separator()`.

We should investigate the algorithms, and either:

 1. Document the current behaviour clearly, and point out that the size of the returned separators are always the same as the vertex connectivity, or:
 2. Fix the function so that it works on disconnected graphs.

I'm inclined to go with (1) for lack of time.

If we choose to go with (1), that will cast the following issue in a different light:

 - #1910 

If we don't return $n-1$ size subsets for $K_n$ complete graphs, then the result sizes won't always match the vertex connectivity.

szhorvat:
- Fixes #1910, no longer returns separators for complete graphs
 - Fixes #2518, we document that no separators are returned for disconnected graphs

The first of these is a change in behaviour so it goes into the 1.0 milestone.


ntamas:
LGTM. If you are happy with it, feel free to merge. (I noticed that you were force-pushing ~50 mins ago so I'm not sure whether it is ready yet).

ntamas (approved):

szhorvat:
It's ready, it was a rebase. Merging.

PR #2588, opened 2024-04-22
Minimize weight buffer size when creating weighted graphs

szhorvat:
This is more of a discussion starter, we need to decide whether to do this vector size minimization in all similar functions, or none of them. This PR brings consistency between the newly added `igraph_weighted_biadjacency()` and its `adjacency` counterparts.

Thinking more about it, the weight vector is likely to be copied to an attribute anyway, so the minimization may be pointless.

Let me know what you think @ntamas, and I'll either merge this or remove the minimization from `igraph_weighted_biadjacency()`.

### DIFF
--- a/src/constructors/adjacency.c
+++ b/src/constructors/adjacency.c
@@ -788,6 +788,10 @@ igraph_error_t igraph_weighted_adjacency(
         IGRAPH_ERROR("Invalid adjacency mode.", IGRAPH_EINVAL);
     }
 
+    /* The weighted graph produced by this function is likely to be used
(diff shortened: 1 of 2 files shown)


ntamas:
> we need to decide whether to do this vector size minimization in all similar functions, or none of them

For me it seems like premature optimization with no real evidence that this has caused performance problems anywhere ever. I would not spend time adding this in other similar functions yet.

szhorvat:
OK, then I will remove this from `igraph_weighted_biadjacency()` and mention in the docs that `weights` will usually have more storage allocated than what it needs.

PR #2591, opened 2024-04-23
perf: iterate through directed weighted adjacency matrix in storage order

szhorvat:
Before going further with these sorts of changes, I want to make sure that we are in agreement. The idea is to iterate through (bi)adjacency matrices in storage order for the sake of better performance.

We never formally guaranteed a certain edge ordering here, but neither did we spell out that we don't. Thus some people may have been relying on it. The best time to make this change is therefore 1.0.

Another argument for not guaranteeing an edge order is that those matrix -> graph conversions which do access both $A_{ij}$ and $A_{ji}$ at the same time could be sped up in the future by a similar blocking technique to what was used in `igraph_matrix_transpose()`. Block sizes should be tunable, and would affect the edge ordering.

If there is a serious concern about the non-stable edge ordering produced by these functions, I suggest adding a new function that canonicalizes the edge order in-place. People who need to rely on the edge order can call this function explicitly.  Note that some level of consistency is already achievable using the `order` parameter of `igraph_es_all()`.


ntamas:
Looks good to me in general and I agree with spelling out that we are not guaranteeing any specific edge ordering.

szhorvat:
This is it for now, I won't implement cache blocking for the undirected case yet.

For the rest, the performance gains are significant.

ntamas (approved):

PR #2592, opened 2024-04-26
refactor!: igraph_eigenvector_centrality() now supports granular edge direction mode control

### LINKED ISSUE #2095: Add 'mode' argument for eigenvector_centrality

`eigenvector_centrality` computes the left eigenvector of the adjacency matrix. In order to have more control over its behaviour, the `directed` argument should be replaced with a `mode` argument.

**Use cases for the feature**

This is one of the few functions that takes edge directions into account yet cannot treat edges as "reversed".

**References**

 - https://github.com/igraph/igraph/issues/1470#issuecomment-694974841

szhorvat:
Closes #2095

The `mode` reversal is intentional, and ensures that `mode=IGRAPH_OUT` will have the same meaning as `directed=true` had before. In this picture, edge directions represent the "propagation of influence". Vertex `a` influences `b` if `a -> b`; in other words, `b`'s scores are determined by incoming connections.

It's a bit uncomfortable that C compilers won't usually give a warning when passing a boolean where an enum is expected. So if people forget to update their calls to this function, there will be no warning. On the upside, passing `true` will be interpreted as `IGRAPH_OUT` (both being equivalent to the integer `1`), so the code will continue to behave as before. If someone passes a `false`, then an "Invalid mode" error will be thrown. In summary, people are protected from bad results even if they forget to update some calls to this function.


szhorvat:
Regarding the warning for (weighted) directed graphs:

Suppose all weights are positive. It seems to me that if the graph is not (strongly) connected, then some eigenvector centralities will usually be zero. The reverse is certainly true: if some are zero, then the graph must be disconnected. This is because the operation $y=xA$ propagates values along the edges. If $x=xA$ and some elements of $x$ are zero, then the corresponding vertices can't be reachable from the vertices that have non-zero values.

Strictly speaking, if the graph is disconnected, we may not get any zeros from ARPACK. It may happen that the graph is the disjoint union of two components, each of which has the same leading eigenvalue. Then this will be a degenerate eigenvalue of the entire graph as well. But if not all components have the same eigenvalue, then we _will_ get some zeros.

Thus I propose issuing a warning whenever ARPACK gives us some eigenvector centrality values that are (nearly) zero.

szhorvat:
Now this function warns if any of the centrality values are zero. It's ready for a review, and it does actually need a review in case I got something wrong.

szhorvat:
@ntamas Let's discuss this warning when we meet next time. I am having second thoughts. The warning might be too aggressive given what real world networks look like (typically having a giant component, but also some small components which the user might consciously want to ignore, and accept a zero centrality for).

The same issue is also present for hub and authority scores BTW, and can be much worse. If we warn about eigenvector centralities, it makes sense to also warn about hubs and authorities. However, that would result in a warning on basically every real-world network. 

To illustrate, recall that eigenvector centralities are the eigenvectors of $A$, while hubs/authorities are eigenvectors of $AA^T$ and $A^T A$. We are looking for non-negative solutions.  For an undirected graph, every connected component yields a non-negative eigenvector of $A$, and there is no good reason to prefer one over the other, especially when the components are of similar size. One might however argue that when one component dominates, it's fine to accept its corresponding eigenvector for practical purposes.  Now, to better understand hubs and authorities, let's look at the graph whose adjacency matrix is $AA^T$.  Suppose the original graph is weakly connected. Does it mean that $AA^T$ is as well? No! It doesn't even mean that $AA^T$ has a giant component. Take for example this sparse graph:

image

Then $AA^T$ looks like this (doesn't have a dominant component):

image

In this case hubs and authorities are meaningless as there is no good argument to choose the eigenvector corresponding to the largest component when the second largest is barely smaller.

But if we make the original graph a bit denser, we get these:

image

image

Now there is a dominant component, in fact all other components are singletons.

Anyway, let's discuss what to do when we meet.

szhorvat:
What I have in this PR now is a warning when there are _any_ zero eigenvector centralities. Perhaps a practical solution is to warn when more than a certain fraction of values is zero?

szhorvat:
I implemented comparison with the same tolerance as the default of `igraph_vector_zapsmall()`, which is about `3.67e-11` $\approx 10^{-10}$ , or 1/3 of the representable digits of a `double`.

Choosing the threshold is always going to be a judgment call, but what justifies this somewhat is:

 - After this PR, eigenvector centralities are always scaled so that the largest one is 1. This sets the order of magnitude we're working with.
 - The threshold is relative to `DBL_EPSILON`, which in the end should determine the magnitude of the numerical errors.
 - Mathematica's equivalent of `zapsmall` uses $10^{-10}$ by default, so I'm personally have some experience with this threshold, both generally, and specifically with `igraph_eigenvector_centrality()` ... _and_ ... it's not bad, but not perfect.  Let me do some more experiments, and I'll report back.

szhorvat:
With non-weighted graphs, the errors tend to be on the order of `1e-15`, i.e. `DBL_EPSILON` with naively generated ER random graphs.

With weighted graphs, things get a bit messy. A difference in edge weights gets exponentially amplified as we move through the graph. Consider e.g. a path graph with the central edge having a weight of 10 and the rest 1:

image

Now eigenvector centralities drop exponentially:

image

This sort of thing tends to occur in sparse graphs with tree-like appendages, and easily reaches the numerical accuracy threshold.

As an example, let's take an ER random graph with 1000 vertices and 1000 edges.  This has a large giant component, but also has disconnected parts which should have an eigenvector centrality of 0.  If we use a uniform edge weight distribution, we get the following distribution for the order of magnitude (i.e. $\log_{10}$) of computed eigenvector centrality values:

image

Notice the two groups: those with magnitude below $10^{-15}$ should be zeros, and are non-zero only because of numerical errors. The rest are true values, and they do go down all the way to $10^{-10}$.

Now let's use a power-law weight distribution. Now the numerical errors bleed together with the true non-zero values because of this exponential amplification:

image

This happens even though the largest weight is only 30 times larger than the smallest in this graph. Note that this is the same graph as before, just with different weights.

Now let's try an exponential weight distribution with mean value of 1:

image

The separation between numerical errors and true non-zero values still seems to be around $10^{-15}$, even though they bleed together.

----

Now coming back to unweigthed ER graphs and trying to maximize the numerical errors, it seems that random trees can have a very skewed eigenvector centrality distribution, probably due to a similar exponential amplification as in the weighted case, but now due to different degrees rather than weights. The values easily go below `1e-15`.

Now let's try very sparse ER random graphs just above the giant component transition threshold, i.e. slightly more than half as many edges as vertices. Here components tend to be treelike. For example, `vcount=10000` and `ecount=6000` yields

image

But we do know which values should be exact zero (by decomposing into connected components), so now let's try look at the histogram of the worst errors across many such random G(10000,6000) graphs, relative to `DBL_EPSILON`:

image

This is cut off at 50 times `DBL_EPSILON`, but I think the picture is relatively clear.

**Proposal:** Let's make the cutoff threshold approximately 10x to 100x `DBL_EPSILON`. I would just go with 10x. This catches most issues while minimizing false positives.

szhorvat:
Change the threshold to `10*DBL_EPSILON` and rephrased the warning to say that the graph _may_ not be connected (instead of _is_ not connected).

ntamas (approved):

ntamas:
LGTM. I think we can merge if the CI passes (I have fixed a conflict with `develop` hence the need to re-check). Would you prefer to merge or squash?

szhorvat:
Simple merge, don't squash.

szhorvat:
Any comment on the proposal of a 30%-zero warning threshold for hub/authority scores?

ntamas:
> Any comment on the proposal of a 30%-zero warning threshold for hub/authority scores?

Seems fine to me. Do you want to do it as part of this PR or separately?

PR #2611, opened 2024-05-15
Chung-Lu model

### LINKED ISSUE #2610: Chung-Lu model implementation

An implementation of the famous Chung-Lu model.

**Use cases for the feature**

This is a very-well known model that's commonly references. It is useful both for teaching and because people commonly look for it. Note that we already have a very similar model in `igraph_static_fitness_game()`.

In both models a weight $w_i$ is assigned to vertex $i$, and the probability of connection between $i$ and $j$ is proportional to $w_i w_j$. While in the Chung-Lu model computes the probability between all vertex pairs and connects them accordingly, leading to a variable final edge count, `igraph_static_fitness_game()` fixes the edge count sharply, and chooses the endpoints of each edge with probabilities proportional to their weights.

**Interface and implementation**

```c
igraph_error_t igraph_chung_lu_game(
    igraph_t *graph, 
    const igraph_vector_t *expected_out_deg,
    const igraph_vector_t *expected_in_deg,
    igraph_bool_t loops);
```

If `expected_in_deg == NULL`, generate an undirected graph. Otherwise require that the expected in-degree sum equals the expected out-degree sum.

To decide: What to do when the formula probabilities gives a value larger than 1? Clip it to 1 and issue a warning (probably best) or throw an error?

**References**

 - Chung-Lu: http://link.springer.com/10.1007/PL00012580
 - Linear-time undirected Chung-Lu implementation: https://scholar.harvard.edu/files/joelmiller/files/fast_chung_lu_generation.pdf

szhorvat:
Closes #2610

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
  - `igraph_bitset_fill()` sets all elements of a bitset to the same value.
  - `igraph_bitset_null()` clears all elements of a bitset.
+ - `igraph_chung_lu_game()` implements the Chung-Lu models (experimental function).
 
 ### Fixed
 
--- a/doc/generators.xxml
+++ b/doc/generators.xxml
@@ -57,6 +57,7 @@
 <!-- doxrox-include igraph_k_regular_game -->
 <!-- doxrox-include igraph_static_fitness_game -->
 <!-- doxrox-include igraph_static_power_law_game -->
+<!-- doxrox-include igraph_chung_lu_game -->
 <!-- doxrox-include igraph_forest_fire_game -->
 <!-- doxrox-include igraph_rewire -->
 <!-- doxrox-include igraph_growing_random_game -->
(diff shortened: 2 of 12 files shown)


szhorvat:
This is a slightly non-trivial linear-in-edges implementation instead of the straightforward quadratic-in-vertices version. It would probably make sense to provide a generic version where the connection probability $p(w_s, w_t)$ has an arbitrary dependence on the source and target vertex weights $w_s$ and $w_t$. The approach works for as long as $p$ is monotonic in $w$.

szhorvat:
The other two variants commonly discussed in the literature are now implemented. This is ready to merge.

ntamas (approved):
Looks good to me, assuming that the math works out (which I haven't checked as there are published papers with these models).

szhorvat:
> Looks good to me, assuming that the math works out (which I haven't checked as there are published papers with these models).

I tested it using the Mathematica interface, and made sure that the average degree of each vertex is what it should be.

PR #2614, opened 2024-05-23
feat: `igraph_erdos_renyi_game_gnp()` can now generate graphs with more than 100 million vertices

### LINKED ISSUE #2613: Lift vertex count restriction for G(n,p) implementation
`igraph_erdos_renyi_game_gnp()` only works up to about a 100 million vertices due to this (necessary) check:

https://github.com/igraph/igraph/blob/f097d91ebd6f64332a3f263fccef2ff1b32e3936/src/games/erdos_renyi.c#L128-L130

This issue is to lift this restriction by generating connections separately for each vertex, i.e. effectively filling the adjacency matrix row-by-row instead of all at once. This is analogous to how `igraph_chung_lu_game()` currently works, and might involve a small performance penalty, but it will come with reduced memory usage.

We already have benchmarks (`tests/benchmarks/erdos_renyi.c`) to test performance. If the new implementation is faster, it should be used in all cases. Otherwise, it should be used for large vertex counts only.

Should anyone not yet familiar with igraph come across this issue, I recommend reading Section 2 of https://scholar.harvard.edu/files/joelmiller/files/fast_chung_lu_generation.pdf for the general method.

Note: The same limitation applies to `igraph_erdos_renyi_game_gnm()`, but fixing that is significantly more complicated, and requires modifying the algorithm in `igraph_random_sample()`.

szhorvat:
Closes #2613 

I benchmarked the new implementation, and it tends to be slower than the original one. Therefore it is only used when the original one would overflow.

I tested this alternative implementation on small graphs manually, and verified that it produces the expected graph density. Including a test in the test suite is not feasible because of high runtime and memory usage. We need over 100 million vertices for the alternative implementation to become necessary.

### DIFF
--- a/src/games/erdos_renyi.c
+++ b/src/games/erdos_renyi.c
@@ -42,6 +42,66 @@
  * fixed degrees, etc.)</para>
  */
 
+/* This implementation is used only with very large vertex counts, above
+ * sqrt(MAX_EXACT_REAL) ~ 100 million, when the default implementation would
+ * fail due to overflow. While this version avoids overflow and uses less memory,
+ * it is also slower than the default implementation. */
+static igraph_error_t gnp_large(
+    igraph_t *graph, igraph_integer_t n, igraph_real_t p,
+    igraph_bool_t directed, igraph_bool_t loops, igraph_integer_t ecount_estimate
+) {
+
+    igraph_vector_int_t edges;
+    int iter = 0;
+
+    /* Necessitated by floating point arithmetic used in the implementation. */
+    if (n >= IGRAPH_MAX_EXACT_REAL) {
+        IGRAPH_ERROR("Number of vertices is too large.", IGRAPH_EOVERFLOW);
+    }
+
+    if (ecount_estimate > IGRAPH_ECOUNT_MAX) {
+        ecount_estimate = IGRAPH_ECOUNT_MAX;
+    }
+
+    IGRAPH_VECTOR_INT_INIT_FINALLY(&edges, 0);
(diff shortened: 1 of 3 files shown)


src/games/erdos_renyi.c
```diff
@@ -42,6 +42,68 @@
...
+ * it is also slower than the default implementation. */
+static igraph_error_t gnp_large(
+    igraph_t *graph, igraph_integer_t n, igraph_real_t p,
+    igraph_bool_t directed, igraph_bool_t loops, igraph_integer_t ecount_estimate
+) {
+
+    igraph_vector_int_t edges;
+    int iter = 0;
+
+    printf("Yay!\n");
```
ntamas, line 57: This line should be removed :)

ntamas (changes requested):
Just a minor issue (see comment above). I noticed that the new implementation is not covered by unit tests, but I guess it would be hard to produce such a large graph in a unit test that would necessitate triggering the large implementation.

src/games/erdos_renyi.c
szhorvat: 🤦 There's a reason I'm asking for reviews!

szhorvat:
> but I guess it would be hard to produce such a large graph in a unit test that would necessitate triggering the large implementation

Yes, that's right. The "Yay!" was a leftover indicator that this implementation was actually being triggered when I was experimenting with unit tests. Memory usage is already a concern, though not a huge one with today's machines. But running time is several seconds (maybe 10 s) at best, even on an Apple M2 with full optimizations turned on. So including it in unit tests is not feasible.

PR #2615, opened 2024-05-23
Add `igraph_mean_degree()` and `igraph_count_loops()`

szhorvat:
While this is trivial functionality that is easily implemented by users, it is so commonly needed that I think it makes sense to have it in the library. It is of a similar nature to `igraph_density()`, but in practice it is even more useful / commonly needed.

### DIFF
--- a/interfaces/functions.yaml
+++ b/interfaces/functions.yaml
@@ -804,6 +804,9 @@ igraph_maxdegree:
 igraph_density:
     PARAMS: GRAPH graph, OUT REAL res, BOOLEAN loops=False
 
+igraph_mean_degree:
+    PARAMS: GRAPH graph, OUT REAL res, BOOLEAN loops=True
+
 igraph_neighborhood_size:
     PARAMS: |-
         GRAPH graph, OUT VECTOR_INT res, VERTEX_SELECTOR vids, INTEGER order,
--- a/interfaces/functions.yaml
+++ b/interfaces/functions.yaml
@@ -859,6 +859,9 @@ igraph_has_loop:
 igraph_has_multiple:
     PARAMS: GRAPH graph, OUT BOOLEAN res
 
+igraph_count_loops:
+    PARAMS: GRAPH graph, OUT INTEGER loop_count
+
 igraph_count_multiple:
     PARAMS: GRAPH graph, OUT VECTOR_INT res, EDGE_SELECTOR es=ALL
     DEPS: es ON graph
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@
  - `igraph_bitset_fill()` sets all elements of a bitset to the same value.
  - `igraph_bitset_null()` clears all elements of a bitset.
  - `igraph_chung_lu_game()` implements the classic Chung-Lu model, as well as a number of its variants (experimental function).
+ - `igraph_mean_degree()` computes the average of vertex degrees (experimental function).
+ - `igraph_count_loops()` counts self-loops in the graph (experimental function).
 
 ### Fixed
 
--- a/doc/structural.xxml
+++ b/doc/structural.xxml
@@ -241,6 +241,7 @@
 
 <section id="other-operations"><title>Other operations</title>
 <!-- doxrox-include igraph_density -->
+<!-- doxrox-include igraph_mean_degree -->
 <!-- doxrox-include igraph_reciprocity -->
 <!-- doxrox-include igraph_diversity -->
 <!-- doxrox-include igraph_is_mutual -->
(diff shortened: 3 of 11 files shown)


szhorvat:
Should we factor out the self-loop counting and cache handling portion into an `igraph_count_loops()` function that complements the existing `igraph_has_loop()` and `igraph_is_loop()`?

interfaces/functions.yaml
```diff
@@ -804,6 +804,9 @@ igraph_maxdegree:
 igraph_density:
     PARAMS: GRAPH graph, OUT REAL res, BOOLEAN loops=False
 
+igraph_mean_degree:
+    PARAMS: GRAPH graph, OUT REAL res, BOOLEAN loops=True
```
ntamas, line 808: Here's an inconsistency with the default value of `loops` for `igraph_density`. If we are heading towards consistent and predictable defaults, we should change one or the other.

interfaces/functions.yaml
```diff
@@ -856,6 +859,9 @@ igraph_has_loop:
 igraph_has_multiple:
     PARAMS: GRAPH graph, OUT BOOLEAN res
 
+igraph_count_loops:
+    PARAMS: GRAPH graph, OUT INTEGER loop_count
```
ntamas, line 863: Can we imagine a failure mode for `igraph_count_loops()` for now or the future? If we can't, we can make `loop_count` a return value just like we do with `igraph_vcount()` or `igraph_ecount()`. This would also let us change `igraph_mean_degree()` to just use a return value.

interfaces/functions.yaml
szhorvat: I know this _looks_ inconsistent (it grabbed my attention as well), but in reality it isn't. I think we should leave the default as they are now, for both functions.

These two `loops` parameters have very different meanings.

 - In `igraph_density()` it controls whether the graph is assumed to _potentially_ have self-loops, and has an impact on computing the largest possible number of edges. This function doesn't check if the graph _actually_ has self-loops.
 - In `igraph_mean_degree()` it controls whether self-loops are discarded if they happen to be present. `loops=false`, i.e. discarding loops, causes the runtime to be linear in the edge count. `loops=true` is constant time, and suitable for the majority of use cases.

In the typical practical use case, people work with a network that _does not_ have self-loops, and _cannot_ have self-loops. Thus in `igraph_density()`, `loops=false` is mandatory for the result to make sense. `igraph_mean_degree()` would give the same result regardless of whether `loops=true` of `loops=false` is used, but `loops=true` does a constant-time calculation, while `loops=false` is linear.

szhorvat: Something I thought about was whether `igraph_density()` should issue a warning when we know from the cache that the graph does have loops, but `loops=false` was set. In this case the result is meaningless, so a warning makes sense. I did not add such a warning, conditional on what is cached, because users may take a lack of warning as assurance that there are no problems.

interfaces/functions.yaml
szhorvat: I thought about this, and actually expected that you'd vote for using an error code anyway. This was one reason I chose this. The other was that `igraph_density()` also returns an error code, even though it doesn't really need it with the current implementation.

There are two ways I can imagine that a future version of these functions could fail:

 - We change the underlying graph representation and checking for loops is no longer a straightforward iteration through edges without memory allocation. But then it is likely that many more interface changes will become necessary.
 - The caching functions become more complicated and start potentially requiring memory allocation. I don't _think_ this will happen, and I would hope that it doesn't happen, but I am not 100% sure. The next major change to caching is the "deduction engine", which I _think_ wouldn't require memory allocation (I'd definitely want to avoid it given how expensive `malloc` is).

I'm on the fence about this. I'll leave the decision to you. Let me know if I should refactor.

Note that `igraph_has_loop()` can also return its result directly, but `igraph_has_multiple()` currently cannot.

szhorvat: I am very slightly leaning towards keeping the error code, partly out of caution (future stability), and partly for reasons for symmetry/consistency with other functions. But I do recognize that this "convenience function" for a mean degree is not much more convenient in C than `2.0 * igraph_ecount(g) / igraph_vcount(g)` if we have to use an output parameter.

ntamas:
Looks OK to me now. Feel free to decide whether to merge or squash.

ntamas (approved):

szhorvat:
@ntamas One advantage of making these functions return their result is that we can mark them with the `pure` attribute, which allows extra optimizations. For example, multiple calls on the same graph are optimized away (unless the graph was changed). Not too late to change this. What do you think? I'm still on the fence.

ntamas:
Feels like premature micro-optimization to me. I don't really see a use-case where this would bring significant gains that would make it worth to change things. If performance is really a concern to someone, they can fall back to `2.0 * igraph_ecount(g) / igraph_vcount(g)` easily.

PR #2620, opened 2024-05-28
Fixes and robustness improvements to eigenvector-like centrality computation

szhorvat:
This PR includes the following:

 1. The initial ARPACK vector for the hub score computation is now the out-strength, as hub scores correlate with this. Previously it was the total degree.
 2. The initial vector computation is now fully consistent between hub/authority and eigenvector centrality computations.
 3. When vertex (out/in)-degrees are zero, the corresponding element of the starting ARPACK vector is allowed to be zero. If a vertex degree is zero, the corresponding eigenvector centrality _must_ be zero. **I need a careful review for this.**
 4. Scaling these values to a max of 1 is now factored out into a separate function. This is theoretically slower than before due to an extra loop but very likely negligible. This fixes the scaling of authority scores when large negative scores are present.
 5. Warnings are given for negative weights for hub/authority scores.
 6. Negative hub scores (which only occur with negative edge weights) were incorrectly clipped to zeros. This is now fixed.
 7. `igraph_i_vector_mostly_negative()` was already unnecessary since on `develop` we always scale these scores to a max of 1. Its redundant use is now removed. The function is kept for the time being, even though it's unused.
 8. Fixed DNEUPD() error reporting in the ARPACK interface.
 9. Tweaks automatic NCV-selection in ARPACK based on experimentation
 10. Increases NCV in the directed eigenvector centrality calculation, as this seems to fail to converge much less frequently
 11. A new `eigen_stress` test, disabled by default, helped with verifying that nothing broke, and helped make tweaks 9 and 10.

These go on `develop`, which already had multiple changes to this functionality. Fixes 6, 8, 9 and 10 will be backported to `master`.

The hope is that these changes will fix https://github.com/conda-forge/python-igraph-feedstock/issues/77

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -288,7 +288,10 @@ add_legacy_tests(
   components
   coreness
   efficiency
-  eigen_stress
+# The eigen_stress test is disabled by default. The chance of failure on at
+# least some platforms, or with some BLAS/LAPACK/ARPACK versions is high
+# due to ARPACK's fickleness. The test is still useful for local testing.
+  # eigen_stress
   global_transitivity
   harmonic_centrality
   hub_and_authority


szhorvat:
I tested `igraph_hub_and_authority_scores()` and `igraph_eigenvector_centrality()` on all graphs in the graph atlas, as well as some random directed orientations of these graphs.

`igraph_hub_and_authority_scores()` never failed, and `igraph_eigenvector_centrality()` never failed on undirected graphs.

`igraph_eigenvector_centrality()` did fail on some directed graphs, but then it also did this before this PR. It has a tendency to fail on cycles in particular, which is not surprising.

szhorvat:
@ntamas Error reporting seems to be messed up on `develop`. When I get failures, it looks like this:

```
Warning at src/linalg/arpack.c:821 : ARPACK solver failed to converge (3001 iterations, 0/1 eigenvectors converged)
Error at src/linalg/arpack.c:1309 : Maximum number of iterations reached - Matrix-vector product failed.
fish: Job 1, 'tests/test_hub_and_authority' terminated by signal SIGABRT (Abort)
```

Notice `Matrix-vector product failed`. This comes from here: 

https://github.com/igraph/igraph/blob/d6115bf06a7fee0c27b227922b4196c981d77712/src/core/error.c#L104-L106

I thought this error code (15) was no longer in use, but somehow it is still being passed to the error handler as `IGRAPH_EARPACK`, then the wrong error message text is shown for it. Can you have a look at this? This is not related to this PR.

szhorvat:
@ntamas Please see #2621

szhorvat:
I increased the NCV value for directed eigenvector centrality computations, which seems to make it be more robust.

szhorvat:
@ntamas Not perfect, but better than before. It fails less frequently.

ntamas (approved):

PR #2621, opened 2024-05-28
Fix ARPACK error text, protect against use of removed error codes

szhorvat:


ntamas (commented):
Not sure whether users of the C core who wrote code for themselves are not using these error codes on their own. If they do, and they also use `igraph_strerror`, their code would start to fail.

I think we should probably print a warning instead when using a deprecated error code but should not abort completely.

szhorvat:
I disagree here. The constants were removed, so people _cannot_ return these error codes from C code. The only way they can is if they compile their library with an old igraph header but try to link to a new igraph library—precisely the kind of mistake that assertions are meant to protect from.

I understand that there might be very rare use cases where keeping the same numerical codes as before makes compatibility less of a headache. An example is if a high-level interface checks the numerical codes directly. It would do this as it's not easy to use C enums constants directly from another language. But this would only be _checking_ them, not returning them.

szhorvat:
I notice just now that the documentation of error codes wasn't updated—it's still the same as for 0.10. Could you take care of this please?

ntamas:
Error codes updated in dddd79dab

PR #2623, opened 2024-05-31
feat: igraph_hypercube()

### LINKED ISSUE #2511: Wishlist: Hypercube graph

A function that constructs a $k$-cube.

This can be implemented in terms of `igraph_square_lattice()`

**Use cases for the feature**

Useful for graph theory demonstrations / teaching.

**References**

 - https://en.wikipedia.org/wiki/Hypercube_graph

szhorvat:
Closes #2511

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
  - `igraph_stack_capacity()` returns the allocated capacity of a stack.
  - `igraph_is_clique()` checks if all pairs within a set of vertices are connected (experimental function).
  - `igraph_is_independent_vertex_set()` checks if no pairs within a set of vertices are connected (experimental function).
+ - `igraph_hypercube()` creates a hypercube graph (experimental function).
 
 ### Fixed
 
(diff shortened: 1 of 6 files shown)


ntamas (approved):

ntamas:
LGTM, but I really see no reason to mark this as experimental. Right now it seems to me that we routinely mark new functions as experimental just to leave the vague possibility open that we _might_ change the function signature in the future, and I'm a bit afraid that experimental markers will just stay there and get forgotten over time. This way it is no better than not having a stable API at all.

Can we come up with clear policies about when we are marking a function as experimental and for how long (how many versions) it needs to stay in the codebase before it "graduates" to being non-experimental? I don't mean any strict set of rules but a few guidelines would be good. Like, can we say that before each minor release we go through the list of experimental functions and remove the marker if the function was already there in the previous minor release with the same signature, _unless there is a good reason not to do so_?

This would be something to discuss in one of the upcoming dev meetings.

szhorvat:
I was planning to drop most experimental markers in 1.0.

I was thinking of the following policy, starting from 1.0:

 - All new functions are experimental for at least one 1.x release. 
 - For functions introduced in 1.n, aim to drop the experimental marker in 1.(n+1) unless there's a reason not to.

ntamas:
Sounds good to me!

PR #2624, opened 2024-06-01
feat: tests bitsets for all/any zeros/ones

szhorvat:
This is useful for checking if subsets represented as bitsets overlap/intersect. It will be helpful for the later Kneser graph implementation.

Any concerns about the function naming?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
  - `igraph_bitset_fill()` sets all elements of a bitset to the same value (experimental function).
  - `igraph_bitset_null()` clears all elements of a bitset (experimental function).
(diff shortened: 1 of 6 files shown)


ntamas:
The code seems okay; not sure about the singular / plural in case of `all`, i.e. should it be `are_all_zeros()` (as in, "are all the elements zeros") or `is_all_zero()` (as in, "is this an all-zero bitset?").

Feel free to merge if you are okay with this.

szhorvat:
If you're fine with the naming, I'd keep it as-is. It's more predictable, it's just tuples of the same words. One can think "all-zero bitset" instead of "all bits are zeros". Maybe it makes more sense that way.

It was inspired by `igraph_vector_is_any_nan`, and partly by Mathematica's `AnyTrue` / `AllTrue` / `NoneTrue` functions (but those are without `is`).

PR #2625, opened 2024-06-02
refactor: undirected hub/authority score calculation now falls back to eigenvector centrality

szhorvat:
Fall back to eigenvector centrality calculation when passing an undirected graph to `igraph_hub_and_authority_scores()`. The explanation is in the code comments. If something is unclear, it needs to be improved there.

When using undirected graphs, a warning is issued now. This is useful because it makes little sense to compute hub and authority scores on undirected graphs. People who do it are usually confused and "just want to find hubs". I have seen this more than once. The warning is also necessary because further warnings (such as about negative weights) will now come from `igraph_eigenvector_centrality()`, which would otherwise be confusing.


szhorvat:
To be fair, the equivalence will break down when weights are negative and the most negative eigenvalue can be larger in magnitude than the most positive. But that case is very problematic anyway, and support should be considered at best experimental. The properties that make these centralities useful are no longer guaranteed when weights are negative.

ntamas (approved):

Issue #2331, opened 2023-04-05
Yifan Hu graph layouts


szhorvat:
This is a request to implement Yifan Hu's approach to graph layout. My understanding is that this method can minimize various energy functions, including the Kamada-Kawai and Fruchterman-Reingold ones, but performs much better than a naive force-based moving of nodes. It should do well on difficult cases where naive energy minimization gets stuck (e.. #2292)

This method is implemented in Mathematica and produces excellent quality results for small-to-mid size graphs.

It also also implemented [in GraphViz as sfdp](https://graphviz.org/docs/layouts/sfdp/) and in Gephi as "Yifan Hu multilevel".

References:

 - Efficient and high quality force-directed graph drawing http://yifanhu.net/PUB/graph_draw_small.pdf

leuc:
https://github.com/leuc/igraph/blob/feat/yifan_hu_graphviz/src/layout/yifan_hu.c 

WIP  2D + 3D variants based on the graphviz code with shared Barnes-Hut. 

Code is AI generated.

Edit: 
Parameters are a bit difficult. 
Paper doesn't match GraphViz doesn't match Gephi implementation.
(Jacobi vs. Gauss-Seidel on the Vertex updates, etc)

Got `jagmesh1` in 3D partially unfolding, but 2D fails to converge. Smaller basic graphs work though.

<img width="3440" height="1384" alt="Image" src="https://github.com/user-attachments/assets/37c47017-39c4-4d93-a3b7-ae273e5d4aa2" />

```c
igraph_int_t maxiter = 500;
igraph_real_t repulsive_exponent = 2.0;
igraph_real_t natural_length = -1.0;
igraph_real_t step = 0.3;
igraph_bool_t adaptive_cooling = 1;
igraph_real_t tolerance = 0.01;
igraph_quadtree_scheme_t quadtree_scheme = IGRAPH_QUADTREE_NORMAL;
igraph_int_t max_qtree_level = 10;
igraph_bool_t beautify_leaves = 0;
```

PR #2631, opened 2024-06-08
refactor: spanning tree is not needed for LGL

szhorvat:
I don't know how the LGL layout works in details, but creating the spanning tree seems not only unnecessary, but actually wrong. 

The code creates a spanning tree graph object, then traversed that using BFS, placing vertices on successively more distance layers _in the tree_.  First, this is pointless as the BFS already effectively produces a tree. Second, traversing an arbitrary tree won't give proper layers in the original graph. In principle, the tree could be anything at all. The spanning tree function doesn't _guarantee_ that we get a BFS tree, which is what the algorithm likely needs. 

Note: currently the spanning tree function does actually do a BFS, so the behaviour shouldn't change from this refactoring.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,7 @@
  - `igraph_is_graphical()` and `igraph_is_bigraphical()` are now linear-time in all cases, and generally several times faster than before (thanks to @gendelpiekel, contributed in #2605).
  - `igraph_erdos_renyi_game_gnp()` can now generate graphs with more than a hundred million vertices.
  - `igraph_hub_and_authority_scores()` now warns when negative edge weights are present.
+ - `igraph_layout_lgl()` now uses a BFS tree rooted in the vertex specified as `proot` to guide the layout. Previously it used an unspecified (arbitrary) spanning tree.
  - Updated the internal heuristics used by igraph's ARPACK interface, `igraph_arpack_rssolve()` and `igraph_arpack_rnsolve()`, to improve the robustness of calculations.
  - Updated the initial vector construction in `igraph_hub_and_authority_scores()`, `igraph_eigenvector_centrality()` and `igraph_(personalized_)pagerank()` with `IGRAPH_PAGERANK_ALGO_ARPACK`. This improves the robustness and convergence of calculations.
 
--- a/src/layout/large_graph.c
+++ b/src/layout/large_graph.c
@@ -26,7 +26,6 @@
 #include "igraph_interface.h"
 #include "igraph_progress.h"
 #include "igraph_random.h"
-#include "igraph_structural.h"
 #include "igraph_visitor.h"
 
 #include "core/grid.h"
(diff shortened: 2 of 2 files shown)


ntamas:
This is the original paper, FYI:

http://www.marcottelab.org/paper-pdfs/jmb-lgl.pdf

In general, we try to stick as close to the original implementation or description of the algorithm as possible (that's what happened with the KK layout in the past -- our implementation was not following the original description closely, then people complained, then Gabor reimplemented things according to the original paper, and then a different set of people started to complain). So, unless the original implementation is clearly wrong or nonsensical, I'd advise not to spend too much time on it.

ntamas:
In particular, the original LGL algorithm was designed for protein networks where edges are weighted by the BLAST E-value of the alignment of a pair of protein sequences (smaller score = better alignment). Using the (minimum weight) spanning tree ensures that you are laying out the "backbone" of the network where pairs of proteins in the network are very similar to each other.

szhorvat:
I fully agree on not spending time on this. It came up because of a refactoring of the spanning tree interface, and this was the simplest fix. As far as I can tell, it doesn't change the behaviour, since the original spanning tree was  BFS tree anyway.

> Using the (minimum weight) spanning tree ensures that you are laying out the "backbone" of the network

But the current implementation does not make use of weights at all, and does not produce a minimum weight spanning tree, just a an unweighted BFS tree. Perhaps the intention was to add weight support at some point?

szhorvat:
Correction: This likely does change the behaviour as `igraph_minimum_spanning_tree_unweighted()` produces a BFS tree centred at vertex `0`.  Now we centre the BFS tree at the actual root vertex, which makes a lot more sense.

szhorvat:
Here's a before/after comparison, using the R interface and the `immuno` network from `igraphdata`. The layout is centred on the highest-betweenness node in both cases.


Before this PR:

image

After this PR:

image

**Comparison**

With KK:

image

With FR:

image

szhorvat:
Overall, I'd call this a minor fix, though the layout is pretty bad (and much slower than KK/FR) no matter how we look at it.

ntamas:
IMHO the correct way to go ahead would be to implement proper support for weights (as the original algorithm was for weighted networks anyway), and fall back to your proposal (using a BFS instead of an explicit spanning tree) only in the unweighted case. If you go ahead with this PR, please document in the code that the BFS is adequate only because the BFS tree is one possible instance of a minimum weight spanning tree mentioned in the original publication in the absence of weights, so we know later that this needs to be replaced with a proper minimum weight spanning tree computation once we introduce support for weights. (I don't even understand why this function was added without weights, but it was before my time in the project).

szhorvat:
Removed from milestone.

I don't think it's worth spending time on this function. It may not follow the paper closely, and I could never get nice layouts from it, similar to what's shown in the paper. Just researching if that's because the implementation is different, or because the parameters need careful tuning, is a lot of work. The risk is that there'll be no improvement in the end.

If anyone will spend a lot of time on layouts, it'd make more sense to spend it on proven methods that we know perform well, such as #2331.

szhorvat:
If I hear no objections, I'm going to merge this for 0.10.13 to ease my work on the later spanning tree API refactoring.

PR #2641, opened 2024-06-28
silence format-truncation warnings in benchmarks

jgmbenoit:
Description: upstream: silence format-truncation warnings: benchmarks
 This patch silences format-truncation warnings as detected
 by gcc (Debian 13.3.0-1) in `tests/benchmarks/erdos_renyi.c'.
Origin: Debian
Last-Update: 2024-06-28

### DIFF
--- a/tests/benchmarks/erdos_renyi.c
+++ b/tests/benchmarks/erdos_renyi.c
@@ -24,7 +24,7 @@ void run_bench(igraph_integer_t vcount, igraph_real_t meandeg, igraph_integer_t 
     igraph_t g;
     igraph_real_t p = meandeg / vcount;
     igraph_vector_t outdeg, indeg;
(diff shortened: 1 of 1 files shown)


szhorvat:
What do I need to do to reproduce this warning?

jgmbenoit:
I can reproduce the warnings with the option  `-Wall`  with `gcc` version 13.3.0 which appears to be the most advanced release. You can see the issue by hand. In the `bench_run` function: since the string `msg` includes the string `msg2`, the string `msg` must bee longer. The warnings shows the calculation. I just doubled it following your convention: 511=2*(255+1)-1. As a matter of fact, I can not see why you choose 255 instead of 256.

szhorvat:
Thanks!

PR #2643, opened 2024-07-03
Versioning and stability policy after 1.0

szhorvat:
The major feature of 1.0 is better API stability. Let's help our users understand what we mean by this. I wrote it up a `VERSIONING.md` file. I treat this as informational rather than strict policy.

@ntamas I need your input here. What's missing:

 - Do we promise not to remove deprecated functions except in major releases? Or do we go for a "continuous improvement" approach where deprecated functions can be removed even in minor release, provided that they were kept in a deprecated state for a while?
 
     With the first alternative (which is currently in these notes) we change functions in major versions only, but this may force major versions more often than ideal. With the second alternative we add new variants of functions with different names, deprecate the old variant, and eventually remove it. This way the library evolves gradually, and a major breaking release is very rarely necessary.
 
 - What about the SOVERSION? We should have a policy for this, and include information for package maintainers. I never fully understood how SOVERSION worked, and how the linker uses it. There are differences in shared library versioning between platforms (Linux/macOS), so we shouldn't base this solely on Linux practices.
 - Do we want to promise zero changes to experimental functions in patch releases?

### DIFF
--- a/VERSIONING.md
+++ b/VERSIONING.md
@@ -1,0 +1,61 @@
+# Versioning and stability
+
+This document is provided for informational purposes only, to help igraph users understand how igraph's programming interface is evolving, and what stability guaratees you can rely on.  It concerns the igraph C library only.  igraph's high level interfaces (R, Python, Mathematica) have their own separate versioning schemes and compatibility policies.
+
+
+## Versioning
+
+Starting with version 1.0, igraph follows the spirit of semantic versioning, with some differences, as described below.  The version number consists of three parts in the `MAJOR.MINOR.PATCH` format.
+
+ - `MAJOR` is incremented after making _incompatible changes_ to the stable programming interface.  Major releases are intended to be infrequent, and are accompanied by release notes where we make the effort to provide detailed guidance on adapting to incompatible changes.
+ - `MINOR` is incremented after making _additions_ to the stable programming interface.  Minor releases are issued regularly.
+ - `PATCH` is incremented when making changes that do not affect compatibility (usually bug fixes, documentation improvements, or build systems changes).
+
+The three version parts are available at compile time as the macros `IGRAPH_VERSION_MAJOR`, `IGRAPH_VERSION_MINOR` and `IGRAPH_VERSION_PATCH`, or at runtime through the `igraph_version()` function.
+
+The majority of public, documented functions are considered to be part of the _stable programming interface_, but there are some notable exceptions:
+
+ - **Experimental functions** may change at any time without notice.  These are clearly marked in their documentation ([example](https://igraph.org/c/html/0.10.13/igraph-Generators.html#igraph_chung_lu_game)).  Most newly added functions start out as _experimental_, and stay in this state until we are confident in their design, typically for one or two minor releases.  User feedback about experimental functions is particularly welcome.  We make the effort to avoid changes to experimental functions in patch releases, but do not guarantee this.
+ - **Internal and undocumented functions** are not part of the stable programming interface, not even if they are present in public headers.  They may change at any time. The names of internal functions usually start with the prefix `igraph_i_` (capitalized for macros), while public functions start with `igraph_`.
+
+
+## Symbol lifecycle
+
+Most new symbols start out as _experimental_ in minor releases.  Eventually, their API is declared stable, and the experimental marker is removed from their documentation in the next minor release.
+
+Symbols go through a deprecation phase before they are removed.  Deprecated symbols are marked in their documentation ([example](https://igraph.org/c/html/0.10.13/igraph-Structural.html#igraph_clusters)), and functions are prefixed with `IGRAPH_DEPRECATED` in the public headers ([example](https://github.com/igraph/igraph/blob/997f59ad742892fff199824a248fab382b40f526/include/igraph_components.h#L45-L47)). With GCC-compatible compilers, use the `-Wdeprecated` flag to get a warning for the use of deprecated functions, but keep in mind that deprecation warnings are not supported for all symbol types (e.g. macros) with all compilers.  The ultimate reference for deprecations is the [changelog][1].
+
+
+## Stability of behaviour
+
+Whether changes in function behaviour are considered breaking is somewhat subjective, and is decided on a case-by-case basis.  Expect some changes within minor releases.  For example, a function that ignored edge multiplicities may gain support for multigraphs in a new minor release.  Do not rely on details of behaviour that are not explicitly documented.
+
+A notable case is stochastic functions: we do not guarantee that the same output is returned across different releases (even patch releases) for the same random seed.  We only guarantee the same statistical properties.  
+
+
+## Advice to users and package maintainers
+
+**Users:**
+
+For as long as you don't use _experimental_ functions, you can be confident that your code will continue to work with future releases having the same major version.  If you do use _experimental_ functions, it is your responsibility to check the [release notes][1] of each igraph release and adapt accordingly.  The use of _internal_ functions is completely unsupported: if you feel you need them, please talk to us.
+
+If you do use _experimental_ functions, make this clear in your README file for the benefit of package maintainers.
+
+While igraph comes with multiple header files, only `#include <igraph.h>` is supported. The rest of the headers exist for internal organizational purposes only, and may change without notice.
+
+**Package maintainers:**
+
+Software that does not use experimental functions from igraph can safely link to future igraph versions with the same major version.  Ask the developer of any software you are packaging if they are using experimental igraph functions.
+
+The high-level interfaces of igraph do use both experimental and internal functions. Each high-level interface release is only guaranteed to be compatible with one specific release of C/igraph. As of this writing, this is a concern only for the Python interface, as the other interfaces (R and Mathematica) cannot link dynamically to C/igraph.
+
+
+## Notes
+
+For the purposes of this document, "API compatibiltiy" means that the same sources can be compiled with different igraph versions of igraph headers.  "ABI compatibility" means that a program that only uses stable API can be linked to a different version of the igraph shared library than what it was compiled with.
+
+We strive to maintain both API and ABI compatibility.
+
+However, it must be pointed out that we do not support manipulating the same in-memory igraph data structures with different igraph versions (for example, if two libraries that exchange data are each statically linked to different igraph versions).
+
+ [1]: https://github.com/igraph/igraph/blob/master/CHANGELOG.md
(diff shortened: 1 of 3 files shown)


ntamas:
> Do we promise not to remove deprecated functions except in major releases?

I would definitely be doing that. First, this is consistent with semver. Second, we shouldn't expect users to keep up with every single minor release of igraph, and if we decide to deprecate and _then_ remove a function a few more releases later, this could be confusing for people who suddenly upgrade from a minor version where the function was not deprecated to another minor version where it was already removed.

> What about the SOVERSION? We should have a policy for this, and include information for package maintainers.

This is what I have in my personal notes about the SOVERSION:

"The first component has to be bumped for any kind of ABI-breaking change (i.e. when an already compiled and linked piece of code would not run correctly when replacing libigraph.so.X silently behind its back). The second and third components have no strict meaning; I propose using them like semver minor and patch components when possible."

The only thing that's unclear to me is whether we should bump the first number even if we change an _experimental_ function only. Distro maintainers might have automated checks for that.

> Do we want to promise zero changes to experimental functions in patch releases?

No; the very point of having experimental functions is to make our job easier to evolve these functions as we please without unnecessary management overhead. I think we can allow ourselves to change these even in patch releases.

szhorvat:
OK, it sounds like we're on the same page. The only thing left to figure out is the SOVERSION. I would not bump the major SOVERSION for changes to experimental functions. For example, I'm pretty sure Debian will have a `libigraph4` package for igraph 1.0. We don't want them to do a `libigraph5` for 1.0.3 ...

szhorvat:
An alternative is to decide that this won't serve any practical purpose (if the linker is not using it automatically) and don't bother ever updating the 2nd and 3rd parts.

szhorvat:
@jgmbenoit @limburgher @yurivict Do you have any input on this topic? What sort of versioning is useful for you, as package maintainers on Debian, Fedora and FreeBSD?

Let's make the question simple and practical. We need to decide on how to manage the two versions numbers (VERSION is 3-component, SOVERSION is 1-component) for igraph in the long term, when setting it this way in CMake:

https://github.com/igraph/igraph/blob/997f59ad742892fff199824a248fab382b40f526/src/CMakeLists.txt#L332-L334

SOVERSION and the first digit of VERSION will be incremented for any major release, in accordance with the document I added in this PR. They will both be 4 for igraph 1.0, i.e. one greater than the current value of 3.

Do your operating systems use the other two components of VERSION in any way, and what would be the most useful for you?

Please see the `VERSIONING.md` document in this PR for an explanation of "experimental" functions in igraph. These are functions with a non-stable API, released for testing purposes, with the intention to eventually stabilize them.

limburgher:
Thank you for asking! In Fedora, with 0.10.12 on my system, the igraph package contains, among other things:
lrwxrwxrwx. 1 root root      14 May  6 19:00 /usr/lib64/libigraph.so -> libigraph.so.3
lrwxrwxrwx. 1 root root      18 May  6 19:00 /usr/lib64/libigraph.so.3 -> libigraph.so.3.1.7
-rwxr-xr-x. 1 root root 2751096 May  6 19:00 /usr/lib64/libigraph.so.3.1.7

The .so is used for compilation, and is a symlink to a symlink to .so.3.17

The package's metadata Provides include:
libigraph.so.3()(64bit)

Consuming packages built agains igraph, in this case, python3-igraph, Require libigraph.so.3()(64bit)

So I think what you've describes should work for us.

jgmbenoit:
I have just got a quick view on VERSIONING.md . The following statement will be a source of troubles unless the SONAME  increment each time an experimental or internal function is removed from the ABI:

> The high-level interfaces of igraph do use both experimental and internal functions. Each high-level interface release is only guaranteed to be compatible with one specific release of C/igraph. As of this writing, this is a concern only for the Python interface, as the other interfaces (R and Mathematica) cannot link dynamically to C/igraph.

An obviously easy way is to keep all experimental and internal functions until the next increment of the SONAME.
However, I will encourage to follow the ``tradition'': a clear ABI with neither  experimental nor internal functions.
For experimentations, may be you can play with ``release canditates'' (RC) versions.

jmroot:
I was asked in the MacPorts IRC channel to post some information about macOS library versioning here. It wasn't clear to me what exactly needs explaining, so I guess I'll start by linking to Apple's documentation: https://developer.apple.com/library/archive/documentation/MacOSX/Conceptual/BPFrameworks/Concepts/VersionInformation.html
(The title is Framework Versions but most of the information applies equally to dylibs.)

neverpanic:
To expand on how CMake on macOS uses the SONAME — it will become part of the library's filename `libsomething.$SONAME.dylib`, and will also be referenced in the library's ID load command (typically an absolute path that tells the linker what to insert into other libraries when linking against the library).

So let's say the igraph library is installed as `/usr/local/lib/libigraph.3.dylib`, then any binary that links against that library would copy `/usr/local/lib/libigraph.3.dylib` (with a compatibility version of 3.0.0 and a current version of 3.1.8) into it's list of required libraries.

Any time you change the `SOVERSION`, programs that linked against the old one will stop working (if the old library is removed) or require rebuilding against the new version. As long as the `SOVERSION` is kept the same the loader will load an updated version of the library.

In other words: Keep the `SOVERSION` stable unless you remove or change symbols in an incompatible way.

ntamas:
@szhorvat What if we add a switch at compilation time that removes all experimental functions from the library? Package maintainers can then decide not to include these in the built library so there would be no API breakages. People can still use the experimental functions if they build the library themselves or if they use a higher-level interface that bundles the library statically.

szhorvat:
I am not at all in favour of this. I'd hope to keep things simple for our users, and I'd hope to be able to refer users to distros like Ubuntu or Debian. Having different configurations of igraph float around is confusing.

I want people to actually use experimental functions as soon as possible so that we can get the feedback necessary to finalize them.

For the same reason, we absolutely _should_ expose such functions in high-level interfaces. That means that distros should link python-igraph statically to C/igraph.  They can do the same with any other package that uses experimental functions, and manage this manually, without having to have a cumbersome enforcement mechanism that disables experimental functions.

jgmbenoit:
> I want people to actually use experimental functions as soon as possible so that we can get the feedback necessary to finalize them.

Is this really a good idea ? You may creates here a grey zone that may confuse users. I would rather encourage experienced volunteers to play with an experimental (git) branch.

szhorvat:
Yes, given how people use igraph, I'm convinced that this is the right way to go. The primary userbase is researchers who use igraph for analysis, not developers who build software that expects a stable API. If we don't make this functionality easily accessible (i.e. expose it in high-level interfaces), there will be no feedback at all. The expertise needed for useful feedback is in network science / graph theory, and the specifics will often differ from function to function. People who have this expertise are frequently less comfortable and less willing to compile packages from source.

jgmbenoit:
I have experts in Network Science in mind. The software part should be transparent for network scientists. If a grey zone is introduced, this transparency may disappear and the software part may become a burden for users less comfortable with building packages (``With which version may I play ?'').  Distributions as Debian are meant for stability with a life cycle counted in years. What you want is quick changes, so quite the opposite. As a scientist I see your point, and I would install my own build of the package and install it the `/usr/local' hierarchy; `/usr` is meant to be manged by the distribution. I do that for some TeX packages and I have done that for some scientific libraries. If you look for quick changes and quick responses, you may have a look on how the graph-tool maintainer (who you may know) do that (may be for a different purpose): basically he maintains several repositories (Debian, Ubuntu, ...). I guess that every thing is automatized. Actually I came to maintain a `private` repository to shortcut the relatively long Debian life cycle: I had done it for igraph for a while, but now I just maintained a stable-backports version for the benefice of every one. There is no ready-to-use way to set up you own Debian repository (I have no clue for non debian-derived distributions). The set-up may take some time, but it will respond to your wish. The end users would have just to add your repository in their  apt sources.list(5) (or the equivalent for non debian-derived distributions). In short all the burden would be on your side, at least for the set up.

szhorvat:
What practical problem do you encounter in Debian specifically that you want to solve by excluding experimental functions? Let's find a reasonable compromise that alleviates that problem sufficiently.

We need this setup to be able to produce quality software in a dynamic way, without too much red tape. We are already very low on resources and can't afford imposing restrictions that create additional work.  Experimental functions reduce workload and reduce stress during development as we don't have to get things perfect the first time.

It is possible to add an `__attribute__((warning("Message here"))` to function prototypes, although only GCC and Clang support it. This makes it possible for you to identify packages that use experimental functions in igraph, and ensure that that they link to igraph statically. `python-igraph` will be one of these. Does this suffice for your purposes?

If yes, I propose something along these lines:

```c
#if !defined(IGRAPH_USE_EXPERIMENTAL) && defined(__GNUC__)
#define IGRAPH_EXPERIMENTAL __attribute__((warning("This function is experimental.")))
#else
#define IGRAPH_EXPERIMENTAL /* empty */
#endif
```

`IGRAPH_EXPERIMENTAL` is then used in public headers to mark experimental functions.

What this does:

 - Structured way to mark experimental functions (not just the docs)
 - Easy way for users to know if they are accidentally using experimental functions
 - Consistency: Same library with same function on all systems (i.e. experimental functions are not forcibly removed)
 - Opt-in to eliminate "experimental" warnings by defining `IGRAPH_USE_EXPERIMENTAL` before `#include <igraph.h>`

jgmbenoit:
> What practical problem do you encounter in Debian specifically that you want to solve by excluding experimental functions?

Dynamic libraries must have an increasing table of functions. That is, if a function is removed or modified the SO major version must be incremented. This allows to upgrade softly. An easy way would be to keep the names of the experimental functions and to remove the dead ones at the next major release f the library.  This could be the library part of your approach.

The igraph python Debian package is currently linked against the C igraph library: this is highly encourage by Debian but not mandatory.  We can reverse this, but I will not encourage to do so.

Having say that, my guess is that you are not the first one to come to this approach. On my side I have no example to furnish: my ``model'' when I create a C library for my own usage is `GSL` and I write a lot of tests to check the good design of the implemented algorithm and of the API. But I am dealing with well establised mathematics, not with mathematics on the edge. I agree to say that this model might be somehow obsolete in the sense that it was created at a pioneering time far before the `git` collaborative age. Via Google I could see that boost has an experimental part as well as the [GCC libc++ library](https://gcc.gnu.org/projects/cxx-status.html). This leads me to think that there is some kind of traditional usage to manage experimental features. And that this traditional usage is certainly supported by the Debian machinery (dh(1)). Do you have examples to furnish on your side ?

PR #2646, opened 2024-07-04
Update infomap

### LINKED ISSUE #2600: Update InfoMAP to a recent version
Updating InfoMAP will likely be a breaking change, so it needs to happen before 1.0.

vtraag:
This is an initial draft PR to fix #2600.

The objectives are:
- Upgrade the previously included version of Infomap to a more recent version
- Make the inclusion of Infomap optional
- Support only a few basic parts of the full Infomap possiblities, most notably:
  * Basic
  * Regularisation approach
  * Multilevel / hierarchical

The current draft PR implements only a basic version, with a rudimentary implementation of Infomap.

What still needs to be done/checked is the following:
- [ ] More advanced capabilities
- [ ] Documentation
- [x] Unit testing
- [x] Interruptability
- [x] Error handling
- [ ] Update CHANGELOG
  - [x] Note CXX14

I already wanted to prepare a draft PR to allow for early comments while in progress.


CMakeLists.txt
```diff
@@ -52,7 +52,7 @@ set(CMAKE_CXX_VISIBILITY_PRESET hidden)
 # Set C and C++ standard version
 set(CMAKE_C_STANDARD 99)
 set(CMAKE_C_STANDARD_REQUIRED True)
-set(CMAKE_CXX_STANDARD 11)
+set(CMAKE_CXX_STANDARD 14)
```
szhorvat, line 55: FYI I noted this in the changelog. Better now to include the changelog update in this PR as it will cause endless conflicts. Please do rebase on the latest develop though for ease of review.

szhorvat:
Can you please review the igraph-side API and check if we want to take the opportunity to make improvements for 1.0?

Regarding error handling and interruptibility: Infomap is slow, so interruptibility is going to be very valuable. Can we really achieve it without patches? If not, then a git submodule will not be appropriate.

CMakeLists.txt
vtraag: You mean this should be noted in the CHANGELOG at the very end after having finished this PR, and not now? I'll include it in the updated to-do list.

This PR is already rebased on the latest `develop`, so I'm not sure what you are referring to?

vtraag:
> Can you please review the igraph-side API and check if we want to take the opportunity to make improvements for 1.0?

This will change anyway because the idea was to include at least the hierarchical version (which gives much nicer results typically) and the regularisation. So, the igraph-side API will change anyway, but that will come as I implement these.

> Can we really achieve it without patches? If not, then a git submodule will not be appropriate.

Yes, fair point, I'm not sure whether this will be achievable indeed.

> Generally, I'm not sure how I feel about the submodule approach since it brings in a lot of irrelevant files that I'd rather not distribute in igraph's release source archives.

I thought your idea was not to include this in the `igraph` release source archives anyway, and only make it available in the binary wheels. The preference for a submodule is just that it makes it easier to update and more clear what version was referenced. But perhaps it just turns out not to be possible indeed.

CMakeLists.txt
szhorvat: I mean that you don't need to worry about it because I've already done it.

vtraag: Ah, right, like that, perfect, thank you!

szhorvat:
Reminder that `ACKNOWLEDGEMENTS.md` needs an update (I'm not sure why it doesn't already have Infomap).

vtraag:
I've now deinited the submodule and just included the files.

It seemed that there were still some problems with building, I didn't yet properly include everything yet.

Unfortunately, it seems I still need to add something somewhere, as it now does not link properly, and I'm getting errors such as

```terminal
[build] /usr/bin/ld: src/libigraph.a(InfomapBase.cpp.o): warning: relocation against `_ZN7infomap3Log8s_silentE' in read-only section `.text._ZN7infomap3Log8isSilentEv[_ZN7infomap3Log8isSilentEv]'
[build] /usr/bin/ld: src/libigraph.a(infomap.cc.o): in function `igraph_community_infomap':
[build] [...]/igraph/src/community/infomap.cc:122:(.text+0x2dd): undefined reference to `infomap::InfomapIterator::operator++()'
[build] /usr/bin/ld: src/libigraph.a(infomap.cc.o): in function `infomap::InfomapIterator::InfomapIterator(infomap::InfoNode*, int)':
[build] [...]/igraph/vendor/infomap/src/core/iterators/InfomapIterator.h:39:(.text._ZN7infomap15InfomapIteratorC2EPNS_8InfoNodeEi[_ZN7infomap15InfomapIteratorC5EPNS_8InfoNodeEi]+0x1a): undefined reference to `vtable for infomap::InfomapIterator'
```

I can't directly find where is the problem. If anyone has any suggestions, happy to hear it. I'll try to look at this again at some later point.

vtraag:
It seems I just didn't supply all the necessary sources.

szhorvat:
Can you please rebase onto (or just merge) the latest `develop`? I just fixed a CI failure.

vtraag:
> Can you please rebase onto (or just merge) the latest `develop`? I just fixed a CI failure.

Will do once I've got a properly working version locally.

vtraag:
@danieledler, I believe I've now got most of the code working, and I get similar test results in terms of partitioning as the previous code. There is one consistent difference however, and that is the code length, in particular, it is always lower (see test results below). 

<details>
<summary>Test results</summary>

```diff
 # Two triangles connected by one edge
-Codelength: 2.51787 (in 2 modules)
+Codelength: 2.32073 (in 2 modules)
 Membership: 0 0 0 1 1 1 
 # Two 4-cliques (0123 and 4567) connected by two edges (0-4 and 1-5)
-Codelength: 2.94884 (in 2 modules)
+Codelength: 2.74930 (in 2 modules)
 Membership: 0 0 0 0 1 1 1 1 
 # Zachary Karate club
-Codelength: 4.60606 (in 3 modules)
+Codelength: 4.31179 (in 3 modules)
 Membership: 0 0 0 0 1 1 1 0 2 0 1 0 0 0 2 2 1 0 2 0 2 0 2 2 2 2 2 2 2 2 2 2 2 2 
 # Flow (from infomap_dir.tgz)
-Codelength: 3.32773 (in 4 modules)
+Codelength: 3.15098 (in 4 modules)
 Membership: 0 0 0 0 1 1 1 1 2 2 2 2 3 3 3 3 
 # MultiphysChemBioEco40W_weighted_dir.net (from infomap_dir.tgz)
-Codelength: 3.87095 (in 5 modules)
-Membership: 0 0 0 0 0 0 0 0 0 0 1 1 1 0 1 1 0 1 1 1 2 2 2 2 2 2 2 2 2 2 3 3 3 3 3 3 3 4 3 3 
+Codelength: 3.20258 (in 5 modules)
+Membership: 0 0 0 0 0 0 0 0 0 0 1 2 1 1 1 1 1 1 1 2 3 3 3 3 3 3 3 3 3 3 4 4 4 4 4 4 4 4 4 4 
 # Wiktionary english verbs (synonymy 2008)
 |V|=7339 |E|=8293 directed=0
-Codelength OK.
+Codelength was 4.55453, expected 5.70800
-- -------------------------------------------------------
```
</details>

Is simply calling `codelength()` indeed correct as below?

https://github.com/igraph/igraph/blob/fe4f2e70d21b760b13f288da801a05bf2158486b/src/community/infomap.cc#L176

Additionally, just to double check, do I correctly extract the partition for a simple (no multiplex, no higher order) graph like this:

https://github.com/igraph/igraph/blob/fe4f2e70d21b760b13f288da801a05bf2158486b/src/community/infomap.cc#L114-L130

This was based on the `writeClu` function, and it appeared that I had to ignore the leafNodes (not exactly sure I understand why), so that's just why I wanted to double check with you.

vtraag:
Daniel let me know that the code-length differences are expected since the old code uses coded teleportation.

vtraag:
Rebased on the latest develop.

vtraag:
@szhorvat, if you can want to take a look, I believe the code is in principle fine as is. Documentation, comments and CHANGELOGs still need to be done, but they can be done once the code is OK.

I don't understand why the build fails. Locally, all tests pass. In particular, I don't understand why the `null_communities` test fails with `Network is empty - Failed` since the branch of the code that raises that exception is only run when `igraph_vcount(graph) > 0`:

https://github.com/igraph/igraph/blob/30a1108889de693c66311fd91bd48ed3c568d77d/src/community/infomap.cc#L190-L200

(in particular, the exception is raised in L193).

szhorvat:
Let's first figure out why CI fails (I'll check that first).

If you're not in a hurry, I'll look at the other outstanding PRs first. I'm a but overwhelmed as teaching has just started here.

If there's anything that's blocking you that you want me to look at sooner, please just ping me.

vtraag:
OK, it seems to work correctly now, I've adapted the code to work around a particular issue (https://github.com/mapequation/infomap/issues/364).

The test results seem to differ on some platforms. The random seed generator from igraph is guaranteed to provide identical results across platforms, right? Perhaps the RNG that infomap relies on does not provide the same guarantee.

vtraag:
Infomap relies on `std::mt19937`, which according to [this post](https://stackoverflow.com/questions/34903356/c11-random-number-distributions-are-not-consistent-across-platforms-what-al) indeed might not necessarily produce identical random numbers across platforms for all distributions.

szhorvat:
> The random seed generator from igraph is guaranteed to provide identical results across platforms, right?

Yes, identical even between 32-bit and 64-bit.


> Perhaps the RNG that infomap relies on does not provide the same guarantee.

One of the adaptations we should make is to use igraph's RNG for everything. Hopefully this is possible without a performance hit that's too large.

szhorvat:
> Infomap relies on `std::mt19937`

OK, so it uses C++11's standard RNG framework? Then it shouldn't be too hard to adapt it to use igraph's own. We just need a wrapper for igraph's RNG that follows C++'s standard RNG API.

If Infomap's RNG is hard-coded, perhaps it's best to submit an upstream patch that makes it configurable?  Note that I haven't looked at any code yet.

danieledler:
We are happy to replace the standard RNG in Infomap. The inconsistent result across platforms is a problem for us too. See https://github.com/mapequation/infomap/issues/365.

vtraag:
> One of the adaptations we should make is to use igraph's RNG for everything. Hopefully this is possible without a performance hit that's too large.

Why would you consider this necessary? I just seed the RNG from infomap now based on igraph's RNG. I don't think it matters that infomap then uses a different RNG, right?

> If Infomap's RNG is hard-coded, perhaps it's best to submit an upstream patch that makes it configurable?

I believe that option it shouldn't be too difficult indeed, I believe it's only a limited uniform integer sampling procedure.

> We are happy to replace the standard RNG in Infomap.

Great! It might perhaps still be useful though to have it configurable though, if we would indeed prefer to use igraph's RNG.

One would expect there to be a simple existing library that ensures cross-platform reproducible results. There seem to be plenty of options, but I'm not sure if any of them are any good.

danieledler:
> Daniel let me know that the code-length differences are expected since the old code uses coded teleportation.

They are expected, and I have double-checked the codelength on the two triangle network with a codelength calculator outside Infomap.

szhorvat:
> > One of the adaptations we should make is to use igraph's RNG for everything. Hopefully this is possible without a performance hit that's too large.
> 
> Why would you consider this necessary? I just seed the RNG from infomap now based on igraph's RNG. I don't think it matters that infomap then uses a different RNG, right?

Doing what you suggest is not a terrible thing, but it makes me a bit uncomfortable for a number of minor reasons. The clean and nice solution would be to use igraph's own RNG.

The first question is: is the InfoMAP code parallelized, and in particular: does it generate random numbers from multiple parallel threads? If yes, then using igraph's own RNG becomes complicated, and it's probably a good idea to just pass a random seed to InfoMAP, as you suggested.

Otherwise, using igraph's RNG is better because:

 - The user gets the RNG they choose, with all its quality tradeoffs.
 - We get cross-platform consistency for free (even between 32-bit and 64-bit platforms).
 - I'm not sure that simply passing a random seed results in the same randomness (lack of correlation) with all RNG types.

Yes, the practical relevance of these concerns is relatively low, and I wouldn't block merging this PR due to this.

If you don't want to do this refactoring, I would still prefer to do it myself at a later stage (perhaps after this PR got merged).

szhorvat:
@vtraag Getting this into a mergable state (even if not perfect, e.g. regarding the RNG) would be helpful for the GSoC project of @BeaMarton13 on developing a community detection tutorial.

ntamas:
I have replaced the RNG usage with igraph's own RNG in an attempt to fix the CI failures (which seem to be due to platform-dependent inconsistencies between the RNG implementations).

vtraag:
Not sure why the CI doesn't run?

vtraag:
Thanks @ntamas for the implementation of the RNG. I believe it is now in a sufficiently developed state to be taken forward. Documentation should still be updated. In order to get 1.0 out, it would probably be best to ignore the more advanced capabilities for now. These can always be added later.

vendor/CMakeLists.txt
```diff
@@ -5,3 +5,36 @@ add_subdirectory(lapack)
 add_subdirectory(mini-gmp)
 add_subdirectory(pcg)
 add_subdirectory(plfit)
+
+# We add the infomap vendored target here, so that
+# we can keep infomap as a clean submodule.
+
+file(GLOB_RECURSE infomap_src CONFIGURE_DEPENDS "infomap/src/*.h" "infomap/src/*.cpp")
```
ntamas: I'm not really a fan of globs in makefiles because build systems often cannot detect when new files are added. It is probably okay here as it is a vendored project, but I might replace this with an explicit list of files instead later.

ntamas (commented):
I would probably clean up `vendor/infomap` a bit so it does not include files that are vendored directly from upstream but are not required for building Infomap. (Examples, Github workflow files, all the JS/TypeScript/R stuff and so on). It makes the PR easier to review. I think we can do this safely now that the folder is not a submodule.

ntamas:
Cleaned up the vendored sources a bit. One thing that would be nice to have before integration is interruption support - then I think it's good to go as a first iteration. We can always improve it later by adding some of the more advanced stuff under different function names.

ntamas:
@szhorvat What shall we do with the coverage tests for this PR? Clearly it is not feasible to achieve the same level of coverage on the infomap project as we have in the rest of the library, partly because the more complex parts of the Infomap library are not exposed in igraph yet. I'm inclined to blacklist the entire `vendor/infomap` folder from coverage calculations. What do you think?

szhorvat:
Yes, blacklist.

ntamas:
Okay, it looks like the interruption handler is actually added (`conf.interruptionHandler = &igraph_allow_interruption`) so this is good to go -- I'll update the changelog.

szhorvat:
@ntamas Does this / should this support linking to an external infomap? 

Since this is not a submodule anymore, should we remove some irrelevant files such as `.gitignore`?

PR #2648, opened 2024-07-05
refactor: igraph_get_stochastic_sparse() no longer rejects graphs with zero out/in degrees.

### LINKED ISSUE #2502: Inconsistent handling of zero degrees by `get_stochastic()` vs `get_stochastic_sparse()`
**Describe the bug**

The stochastic matrix $S$ of a graph is its adjacency matrix, with columns normalized to sum to 1: $S_{ij} = A_{ij} / \sum_i A_{ij}$. Then the probability distribution of the location of a random walker in the graph, $p_i$, is updated as $p:=pS$ after each step.

If a vertex has zero out-degree, the column normalization is not possible. What should igraph do then? Currently:

 - `igraph_get_stochastic_sparse()` throws an error
 - `igraph_get_stochastic()` leaves the column as all-zeros

This is inconsistent. The sparse and dense versions should behave the same.

**To reproduce**

In R,

```
> g<-make_graph(c(1,2))
> stochastic_matrix(g)
Error in stochastic_matrix(g) : 
  At vendor/cigraph/src/core/sparsemat.c:3408 : Rows with zero sum are not allowed. Invalid value
> stochastic_matrix(g, sparse=F)
     [,1] [,2]
[1,]    0    1
[2,]    0    0
> 
```

**Version information**
This behaviour was present for many years. The check for zero out-degree was added 13 years ago.

szhorvat:
Closes #2502

@vtraag, I wouldn't mind a 3rd opinion on this. See the issue for details.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -53,6 +53,7 @@
  - `igraph_eigenvector_centrality()` now warns about eigenvector centralities equal to zero, as these indicate a disconnected graph, for which eigenvector centrality is not meaningful.
  - `igraph_hub_and_authority_scores()` now warns when a large fraction of centrality scores are zero, as this indicates a non-unique solution, and thus the returned result may not be meaningful.
  - `igraph_hub_and_authority_scores()` now warns when providing an undirected graph as input, and falls back to the equivalent eigenvector centrality computation.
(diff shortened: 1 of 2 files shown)


vtraag (approved):
Looks good to me and the underlying principle is also fine with me. As long as it is consistent with other functions, I don't mind what choice is taken exactly.

PR #2657, opened 2024-07-09
fix: crash in igraph_layout_umap() when passing distances == NULL and distances_are_weights == true

szhorvat:


szhorvat:
I still see crashes when interrupting the computation. Marking as draft until I can find the issue.

PR #2666, opened 2024-07-23
feat: igraph_vector_reverse_section() and igraph_vector_rotate_left()

szhorvat:
This PR adds two functions:

 - `igraph_vector_reverse_section()` reverses a section of a vector. Modelled after `vector_reverse()` and `vector_remove_section()`. This is added to help with implementing rotation.
 - `igraph_vector_rotate_left()` rotates by the given number of steps. This will help in canonicalizing cycles, and test PR #2181 (CC @GenieTim): the cycle `(4, 3, 9)` is of course the same as `(3, 9, 4)`. Canonicalization will be possible using something like `igraph_vector_int_rotate_left(v, igraph_vector_int_which_min(v))` which shifts the smallest element to the front. EDIT: Actually, complete canonicalization might also require a reversal, but rotation is needed as well.

Neither return an error code. I plan to remove the error code from `igraph_vector_reverse()` as well on `develop`. Any objections to this?

I did not bother to mark these simple functions as experimental.

I will update the changelog separately after this is merged.


szhorvat:
Merging this to make progress. We can still make changes before the 0.10.14 release. A post-review is still appreciated @ntamas.

PR #2670, opened 2024-08-07
feat: igraph_feedback_vertex_set()

szhorvat:
This is a naive generalization to feedback vertex sets of the exact minimum FAS algorithm I implemented in #2668.

Perhaps this is not the best way to do it, but it works, and it wasn't too much effort on top of #2668. 

I give a short summary of the feedback vertex set (FVS) problem so people can give better feedback about the interface of this functionality:

  - It is NP-complete both for undirected and directed graphs (FAS is hard only for directed).
  - There seems to be more work for heuristic methods for the undirected case, as this is more generally useful. When we add heuristic methods in the future, some methods may work _only_ for the undirected case.
  - There is a weighted generalization, with vertex weights (not edge weights). This implementation supports weights.

PR notes:

 - The interface mirror that of `igraph_feedback_arc_set()`, even at the cost of deviating from [the standard parameter ordering](https://github.com/igraph/igraph/wiki/Guidelines-for-function-argument-ordering).
 - This is an exact (thus slow) method based on integer programming. Eventually it will be nice to have some heuristic methods, as that will be useful for speeding up the minimum cycle basis calculation.
 - I called the method `IGRAPH_FVS_EXACT_IP`, and did NOT add an `IGRAPH_FVS_EXACT_IP_CG`, even though this is analogous to `IGRAPH_FAS_EXACT_IP_CG`. We can add a more fine-grained method selection at the point when we add more IP methods (which may never happen)
 
Travis ppc64le and s390x has been broken for [everyone](https://travis-ci.community/t/ppc64-s390x-failing-on-startup/14358) (not just igraph) for the past few days; this CI failure is unrelated to the PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
  - `igraph_vector_rotate_left()` applies a cyclic permutation to a vector.
  - `igraph_strvector_swap_elements()` swaps two strings in an `igraph_strvector_t`.
  - `igraph_find_cycle()` finds a single cycle in a graph, if it exists (experimental function).
+ - `igraph_feefback_vertex_set()` finds a minimum feedback vertex set in a directed or undirected graph (experimental function).
 
 ### Changed
 
--- a/doc/structural.xxml
+++ b/doc/structural.xxml
@@ -224,6 +224,7 @@
 <!-- doxrox-include igraph_is_dag -->
 <!-- doxrox-include igraph_topological_sorting -->
 <!-- doxrox-include igraph_feedback_arc_set -->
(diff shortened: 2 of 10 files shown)


ntamas (approved):
Haven't checked the code in detail but it seems like it's essentially the same as the FAS problem and the bulk of the work is carried out by the solver; it is only the setup of the problem that needs to be done in igraph so I think it's okay.

Only one minor comment in the Stimulus specification. Actually, I'll probably just fix this myself.

ntamas:
Feel free to merge upon CI pass.

PR #2672, opened 2024-08-17
feat: allow NULL for cut_prob parameter of motif finding functions

### LINKED ISSUE #2555: `igraph_motifs_randesu()` should accept `NULL` for `cut_prob`
`igraph_motifs_randesu()` should accept a `NULL` value for its `cut_prob`, and should interpret it as "find all motifs instead of sampling".  This can be implemented by internally constructing an all-zero vector of length `size` and using it.

szhorvat:
Closes #2555

@ntamas The only two potentially controversial changes here are:

 - This is done for 0.10.14 instead of 1.0. I point this out in the docs so people don't inadvertently write code with 0.10.14 that they (incorrectly) expect to work with <= 0.10.13
 - The change in `functions.yaml`

I don't need a code review, just an agreement for these two changes.

### DIFF
--- a/interfaces/functions.yaml
+++ b/interfaces/functions.yaml
@@ -1953,15 +1953,15 @@ igraph_write_graph_dot:
 #######################################
 
 igraph_motifs_randesu:
-    PARAMS: GRAPH graph, OUT VECTOR hist, INTEGER size=3, VECTOR cut_prob
+    PARAMS: GRAPH graph, OUT VECTOR hist, INTEGER size=3, OPTIONAL VECTOR cut_prob
 
 igraph_motifs_randesu_estimate:
     PARAMS: |-
-        GRAPH graph, OUT INTEGER est, INTEGER size=3, VECTOR cut_prob,
+        GRAPH graph, OUT INTEGER est, INTEGER size=3, OPTIONAL VECTOR cut_prob,
         INTEGER sample_size, OPTIONAL VECTOR_INT sample
 
 igraph_motifs_randesu_no:
-    PARAMS: GRAPH graph, OUT INTEGER no, INTEGER size=3, VECTOR cut_prob
+    PARAMS: GRAPH graph, OUT INTEGER no, INTEGER size=3, OPTIONAL VECTOR cut_prob
 
 igraph_dyad_census:
     PARAMS: GRAPH graph, OUT REAL mut, OUT REAL asym, OUT REAL null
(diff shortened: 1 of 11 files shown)


interfaces/functions.yaml
```diff
@@ -1953,15 +1953,15 @@ igraph_write_graph_dot:
 #######################################
 
 igraph_motifs_randesu:
-    PARAMS: GRAPH graph, OUT VECTOR hist, INTEGER size=3, VECTOR cut_prob
+    PARAMS: GRAPH graph, OUT VECTOR hist, INTEGER size=3, VECTOR cut_prob=NULL
```
ntamas, line 1956: I think we should mark `cut_prob` as `OPTIONAL` just to be clear.

interfaces/functions.yaml
```diff
@@ -1953,15 +1953,15 @@ igraph_write_graph_dot:
...
 
 igraph_motifs_randesu_estimate:
     PARAMS: |-
-        GRAPH graph, OUT INTEGER est, INTEGER size=3, VECTOR cut_prob,
+        GRAPH graph, OUT INTEGER est, INTEGER size=3, VECTOR cut_prob=NULL,
```
ntamas, line 1960: Same here.

interfaces/functions.yaml
```diff
@@ -1953,15 +1953,15 @@ igraph_write_graph_dot:
...
         INTEGER sample_size, OPTIONAL VECTOR_INT sample
 
 igraph_motifs_randesu_no:
-    PARAMS: GRAPH graph, OUT INTEGER no, INTEGER size=3, VECTOR cut_prob
+    PARAMS: GRAPH graph, OUT INTEGER no, INTEGER size=3, VECTOR cut_prob=NULL
```
ntamas, line 1964: Same here.

tests/unit/igraph_motifs_randesu_estimate.out
```diff
@@ -8,7 +8,7 @@ Full graph of 50 vertices, motif size 3, sample all, (50 choose 3 = 19600):
 Estimate: 19600
 
 Full graph of 50 vertices, motif size 3, sample all, cut_prob 0.1 at each level:
-Estimate: 15990
+Estimate: 16034
```
ntamas, line 11: Any idea why the estimate has changed? Your changes should only affect the case of `cut_prob=NULL`, right? Is it because _earlier_ tests in the same test case took a different code path?

ntamas (changes requested):

tests/unit/igraph_motifs_randesu_estimate.out
szhorvat: Yes, I investigated this. Take a look at line 339 in the new version. This line isn't run anymore when using `NULL`, so there are fewer invocations to the RNG. This affects the subsequent random numbers that are generated.

interfaces/functions.yaml
szhorvat: Fixed.

interfaces/functions.yaml
szhorvat: Fixed.

PR #2674, opened 2024-09-09
fix: eigenvector centralization now always assumes scaled vertex-level centrality scores

### LINKED ISSUE #2673: `igraph_centralization_eigenvector_centrality_tmax()` makes no sense with `scale=false`
The centralization tmax functions make sense only in cases when an absolute scale exists for the centrality considered. This is not the case with eigenvector centrality, unless using a specific type of normalization. With `scale=false`, we do NOT guarantee any kind of normalization in `igraph_eigenvector_centrality()`. We simply return whatever ARPACK produces, which is usually L2-normalized, but I believe this is not guaranteed. In some edge cases we even return results which are explicitly not L2-normalized.

In igraph 1.0, the scale parameter is removed, and the output is always scaled. So this issue goes away with 1.0.

Should we do anything about it in 0.10.14? 

I would recommend starting to ignore the `scale` parameter in `igraph_centralization_eigenvector_centrality_tmax()` and `igraph_centralization_eigenvector_centrality()` already in 0.10.14, behaving as if `scale=true` was passed, and printing a warning when passing `scale=false`. This way we can protect people from using invalid results.

This shouldn't be a big issue, as centralization functions are not exposed in Python, and in R/Mathematica the default is `scale=true`. It is likely to cause some revdepcheck failures, @krlmlr, but I consider that a good thing (protecting users from mistakes, and preparing for the removal of the `scale` parameter).

What do you think @ntamas ?

szhorvat:
Fixes #2673

Note that the `scale` parameter has already been removed for igraph 1.0.

@vtraag If you have input on this before the imminent release, let us know.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@
 
  - `igraph_feedback_arc_set()` uses a much faster method for solving the exact minimum feedback arc set problem. The new method (`IGRAPH_FAS_EXACT_IP_CG`) is used by default (i.e. with `IGRAPH_FAS_EXACT_IP`), but the previous method is also kept available (`IGRAPH_FAS_EXACT_IP_TI`).
  - `igraph_motifs_randesu()`, `igraph_motifs_randesu_callback()`, `igraph_motifs_randesu_estimate()` and `igraph_motifs_randesu_no()` now accept `NULL` for their `cut_prob` parameter, signifying that a complete search should be performed.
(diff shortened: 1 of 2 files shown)


szhorvat:
@ntamas I finished the changelog and updated the docs. You can merge before release.

szhorvat:
@mbojan If you have input about this, or the forum post that started this and asked about why `sna` is not producing the same result as igraph, let us know! https://igraph.discourse.group/t/clarification-on-eigenvector-centralization-calculation-in-igraph/1867?u=szhorvat

szhorvat:
Thanks for the feedback @mbojan !

PR #2687, opened 2024-10-25
feat: resolution parameter for community_optimal_modularity()

### LINKED ISSUE #2678: Resolution support for `igraph_community_optimal_modularity()`
Add resolution support for `igraph_community_optimal_modularity()`.

A breaking change that is easy to implement—we should do it for 1.0.

szhorvat:
`igraph_community_optimal_modularity()` now takes a resolution parameter. Fixes #2678.

Since a new parameter was added, I took the opportunity to standardize the parameter ordering according to https://github.com/igraph/igraph/wiki/Guidelines-for-function-argument-ordering


szhorvat:
Ideally someone would add some proper tests that make sure that the modularity obtained with heuristic optimizers is no greater than what this function produces. Running a stochastic optimizer a couple of hundred times on such a small graph is likely to hit close to the global maximum.

szhorvat:
@ntamas Any comments on this before I merge?

ntamas:
Looks good to me.

PR #2690, opened 2024-10-31
refactor: use OPTIONAL instead of =NULL in interfaces

szhorvat:
This PR ensures that all optional parameters are marked as `OPTIONAL` in the interfaces, and that none that are marked as `OPTIONAL` have the redundant `=NULL`.

Is this safe to go ahead with, @ntamas @Antonov548 @krlmlr ?

### DIFF
--- a/interfaces/functions.yaml
+++ b/interfaces/functions.yaml
@@ -1225,7 +1225,7 @@ igraph_get_biadjacency:
     DEPS: types ON graph
 
 igraph_is_bipartite:
-    PARAMS: GRAPH graph, OUT BOOLEAN res, OPTIONAL OUT BIPARTITE_TYPES type
+    PARAMS: GRAPH graph, OUT BOOLEAN res, OPTIONAL OUT BIPARTITE_TYPES types
 
 igraph_bipartite_game_gnp:
     PARAMS: |-
--- a/interfaces/functions.yaml
+++ b/interfaces/functions.yaml
@@ -2545,11 +2545,16 @@ igraph_eulerian_cycle:
 #######################################
 
 igraph_fundamental_cycles:
-    PARAMS: GRAPH graph, OUT EDGESET_LIST basis, OPTIONAL VERTEX start, INTEGER bfs_cutoff, OPTIONAL EDGEWEIGHTS weights
+    PARAMS: |-
+        GRAPH graph, OUT EDGESET_LIST basis, OPTIONAL VERTEX start,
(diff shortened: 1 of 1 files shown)


szhorvat:
The R interface is not quite ready:

```
> devtools::test()
ℹ Testing igraph
Starting 2 test processes
✔ | F W  S  OK | Context
✔ |         44 | vs-es                                                                                                                                           
✔ |        227 | vs-operators [3.2s]                                                                                                                             
✔ |         17 | weakref                                                                                                                                         
✔ |         50 | watts.strogatz.game                                                                                                                             
✔ |      1  47 | adjacency                                                                                                                                       
✔ |         32 | scan [5.8s]                                                                                                                                     
✔ |          7 | all.st.cuts                                                                                                                                     
✔ |          1 | articulation.points                                                                                                                             
✔ |          6 | assortativity                                                                                                                                   
✔ |         78 | adjacency.spectral.embedding [1.1s]                                                                                                             
✔ |         11 | average.path.length                                                                                                                             
✔ |         25 | ba.game                                                                                                                                         
✔ |         20 | biconnected.components                                                                                                                          
⠸ [ FAIL 0 | WARN 0 | SKIP 1 | PASS 618 ] @ attributes                                                                                                           

Error in `private$handle_error()`:
! testthat subprocess exited in file `test-bipartite.projection.R`
Caused by error:
! R session closed the process connection, killed
Run `rlang::last_trace()` to see where the error occurred.
```

szhorvat:
Some parameters were set to have a default of `=NULL`, even though they shouldn't be optional. This was an R-specific thing. I fixed it in this repo and will re-add the defaults of `NULL` on the R side.

szhorvat:
OK, the R side is cleaned up, see https://github.com/igraph/rigraph/pull/1567

szhorvat:
@ntamas Do you see anything that is of concern here? Anything that might break the new Python interface? Can I go ahead and merge?

There are some parameter types which are effectively always optional, such as `EXTRA`. Should these always have `OPTIONAL` or never?

szhorvat:
If I don't hear back by Tuesday, I'm going ahead with merging this. That will make it necessary to also merge https://github.com/igraph/rigraph/pull/1567

ntamas:
@szhorvat Feel free to go ahead and merge, and then take care of the R-related PR as well. It's all OK from my side.

> There are some parameter types which are effectively always optional, such as EXTRA. Should these always have OPTIONAL or never?

I think being explicit is better than being implicit so these should always have `OPTIONAL`.

PR #2693, opened 2024-11-05
Simple cycle detection: requested test

### LINKED ISSUE #2692: Undirected self-loops are sometimes not detected by the cycle finder
@GenieTim, it looks like there's an issue I missed before merging the PR. Sometimes, self-loops are not detected.

This is fine, all 3 cycles are found:

```
        igraph_small(&g, 7, IGRAPH_UNDIRECTED,
                     0,1, 1,2, 2,0,
                     0,0,
                     0,3, 3,4, 4,5, 5,0,
                     -1);
```

But the self-loop is not found with this graph:

```
        igraph_small(&g, 7, IGRAPH_UNDIRECTED,
                     0,1, 1,2, 2,0,
                     1,1,
                     0,3, 3,4, 4,5, 5,0,
                     -1);
```

The only difference is putting the self-loop on vertex 1 instead of vertex 0.

Could you please have a look? This would require opening a new PR.

In the very worst case, we'll drop self-loops and do a separate pass to detect them. But hopefully, this can be fixed in the algorithm without special-casing.

UPDATE: This affects only undirected graphs. I updated the example code.

GenieTim:
This PR aims to fix #2692 

So far, I cannot reproduce the claim - the tests I added here seem to pass just fine on my device, let's see, maybe I made a typo in the test and it's not actually what was requested, or there is a device dependency that will show up in CI?


szhorvat:
I'm sorry, I didn't post the complete code in the issue I opened. It seems that this affects only undirected graphs. I had `IGRAPH_ALL` in my test.

szhorvat:
Not sure what's going on with the test failures. They do pass on my machine.

PR #2700, opened 2024-11-14
feat: igraph_vertex_path_from_edge_path() can determine the start vertex automatically

szhorvat:
@ntamas, following @GenieTim's comment on wanting both the edges and vertices of cycles, I would like to make it easier to expose and use the `igraph_vertex_path_from_edge_path()` function from high-level interfaces. In 6fd3903c84d741d4bed0284c85db277836a4af79, I improved the docs to make it clear that it works with repeated vertices (meaning that cycles are a valid input, and so are arbitrary walks). But in practical use cases one would often want to avoid having to provide a starting vertex. This PR makes it so that the function can auto-detect the start vertex. 

One ugliness in this approach is that it isn't possible if the path has length zero: in this case this function still returns one vertex, the start vertex. Thus we cannot auto-detect that. Existing igraph functions do rely on this behaviour.

After noticing this issue, I became a bit reluctant about the auto-detection feature. Should we go ahead with this PR?  It can probably still be quite useful.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,10 +12,12 @@
 
  - `igraph_neighborhood_size()`, `igraph_neighborhood()` and `igraph_neighborhood_graphs()` now accept a negative `order` value and interpret it as infinite order. Previously, a negative `order` value was disallowed.
  - `igraph_famous()` now accepts `Groetzsch` as an alias of `Grotzsch`.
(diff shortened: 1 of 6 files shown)


szhorvat:
@ntamas I'd like a review on this please.

szhorvat:
@ntamas Ping again for a review. Let's decide if this will go into the next 0.10.x release.

ntamas (approved):

Issue #1641, opened 2021-01-30
PageRank: ARPACK method sometimes fails and produces negative values


szhorvat:
**Describe the bug**
`igraph_pagerank()` with the ARPACK method sometimes fails, produces an eigenvalue that is not 1, and produces negative results or even non-finite values (not sure if NaN or Infinity).

**To reproduce**
This is easy to reproduce with fuzzing: create a directed random graph with at least 8 vertices, and run the function many times. Be sure to check if the eigenvalue differs from 1. Eventually you will find a case where it does.  I see no correlation with the graph structure that causes the failure.  Here are some examples:

One:
<img width="360" alt="image" src="https://user-images.githubusercontent.com/1212871/106362264-2c684180-6322-11eb-9a8c-953de3d5d903.png">

Two:
<img width="360" alt="image" src="https://user-images.githubusercontent.com/1212871/106362266-2ffbc880-6322-11eb-9da7-2d300589c63f.png">

Three:
<img width="360" alt="image" src="https://user-images.githubusercontent.com/1212871/106362270-3427e600-6322-11eb-888e-01a2df380eec.png">

Four:
<img width="360" alt="image" src="https://user-images.githubusercontent.com/1212871/106362272-35f1a980-6322-11eb-9a8e-d1be8f87f4bd.png">

The failure is non-deterministic and is somehow related to the *random* (!!) way that `igraph_pagerank` initialized ARPACK's starting vector. 

Example in Mathematica:

```mathematica
Do[
 Check[
  g = RandomGraph[BernoulliGraphDistribution[8, p] (* G(8,p) model *), DirectedEdges -> True];
  r = IGPageRank[g, Method -> "Arnoldi"];, Break[]],
 {p, 0.1, 1, 1./5000} (* loop and scan range of p from 0.1 to 1, but this is really reproducible with any p *)
]
```

Output:

```
IGraphM::warning: The PageRank eigenvalue should be 1, actual eigenvalue is -0.425. Possible convergence problem.
```

PageRank values:
```
{0.0367487,  7.48975*10^14, 0.191093, -0.628403, -0.058798, -7.48975*10^14, -0.110246, -0.407911}
```

Graph:

<img width="360" alt="image" src="https://user-images.githubusercontent.com/1212871/106362350-97b21380-6322-11eb-890d-b8bd0f59a024.png">

For completeness, this specific graph's edge list in C format:

```
[...]

PR #2705, opened 2024-11-21
Re-translate BLAS/LAPACK/ARPACK

szhorvat:
This re-translates BLAS/LAPACK/ARPACK with the latest f2c and updates to BLAS 3.12.0 and ARPACK-NG 3.7.0.

The ARPACK update alleviates problems like one reported at https://igraph.discourse.group/t/eigenvector-centrality-is-equal-to-0-for-all-nodes-even-in-a-strongly-connected-network/1898

Some unused f2c files are removed. A complete copy of f2c wasn't present in igraph anyway.

The thread-safety patch (mainly for ARPACK) is updated programmatically by replacing `static` with `IGRAPH_F77_SAVE` in all instances when the variable name does not start with `c_`, which I presume denotes a constant. A blanket replacement is not possible without compilation errors about constant initializers.  This approach probably introduces some unnecessary `IGRAPH_F77_SAVE` (which is a thread-local specifier) but better safe than sorry.


szhorvat:
@ntamas I think this is good now. I am hoping that it will make ARPACK-based functions more robust.

The Windows issue was due to not patching this:

https://github.com/igraph/igraph/blob/e0dbe374e30adc654ad55a50d5aaf47ebc753908/tools/lapack/getlapack.sh#L141-L144

How did you do this last time? I just did it by a lot of manual find and replacing.

szhorvat:
This PR significantly improves #1641.

vendor/lapack/arscnd.c
```diff
@@ -28,15 +24,16 @@
     Purpose   
     =======   
 
-    SECOND returns the user time for a process in seconds.   
-    This version gets the time from the system function ETIME. */
+    SECOND returns the user time for a process in arscnds.
```
ntamas, line 27: This replacement is probably unintended.

ntamas (approved):
Looks good to me but I had to admit that I did not check all the changes in the generated code thoroughly - I just assumed that `f2c` is doing its job correctly.

Issue #2635, opened 2024-06-15
Speed up `igraph_induced_subgraph()` through better automatic method selection


szhorvat:
This is a good issue for first-time contributors to igraph.

**Purpose:** Speed up `igraph_induced_subgraph()` by developing a better method selection strategy for `igraph_induced_subgraph()`.

**Details:** `igraph_induced_subgraph()` can use one of two implementations, controlled by the `impl` parameter. One is more efficient for small subgraphs, the other for subgraphs comparable to the size of the original graph. Automatic method selection is possible too, but currently it uses an arbitrary, unfounded approach. We need a method selection heuristic that is backed by benchmark results

**Tasks**

 - [ ] Write a benchmark for each method of `igraph_induced_subgraph()`, covering various vertex count ratios between the subgraph and the original graph. Benchmark graphs with different vertex counts and mean degrees. Try both random graphs (G(n,m) models) and graphs with high degree heterogeneity (preferential attachment model).

    Example benchmark for inspiration: https://github.com/igraph/igraph/blob/master/tests/benchmarks/intersection.c

 - [ ] Use the benchmark to investigate at what vertex count ratio the performance of the two methods becomes equal. Does this value depend on the vertex count of the graph? Does it depend on the mean degree? Write up the results in this issue for future reference.

    Tip: plotting the benchmark results makes it easier to interpret them.

 - [ ] Implement a new method selection heuristic.

**Additional optional tasks**

 - [ ] Also benchmark creating the subgraph using a combination of `igraph_induced_subgraph_edges()` and `igraph_subgraph_from_edges()`.

carlos-planelles:
Opened PR #2708 
Here is a summary of the benchmark results for future reference:
![image](https://github.com/user-attachments/assets/e1f5ecd7-ef36-4f7d-b795-a9f4903e8c7b)
The line shows the chosen heuristic for implementation selection.

Time Difference represents time taken by `create_from_scratch` minus time taken by `copy_and_delete`.

| Graph Size | Proportion of Vertices | Time Difference |
| ---------- | ---------------------- | --------------- |
|        100 |                     20 |          -0.841 |
|        100 |                     25 |           0.161 |
|        100 |                     30 |           0.201 |
|        100 |                     35 |           0.561 |
|        100 |                     40 |           1.061 |
|        100 |                     45 |           0.791 |
|        100 |                     50 |           0.921 |
|        100 |                     55 |           1.051 |
|       1000 |                     20 |          -0.641 |
[...]

PR #2708, opened 2024-12-09
benchmark: igraph_induced_subgraph()

carlos-planelles:
This PR addresses #2635, updating the automatic implementation selection heuristic for `igraph_induced_subgraph`. This involved adding a benchmark for the two possible methods for various graph sizes and subgraph ratios.

- Added a benchmark to compare the two methods used by `igraph_induced_subgraph` (copy and delete vs create from scratch)
- Evaluated the performance difference between the two methods for graph sizes between 100 vertices and 100000 vertices and 20%-55% subsets
- Updated the implementation selection heuristic based on findings

Sample results are summarized here:
image
The line shows the chosen heuristic for implementation selection.

Time Difference represents time taken by `create_from_scratch` minus time taken by `copy_and_delete`.

| Graph Size | Proportion of Vertices | Time Difference |
| ---------- | ---------------------- | --------------- |
|        100 |                     20 |          -0.841 |
|        100 |                     25 |           0.161 |
|        100 |                     30 |           0.201 |
|        100 |                     35 |           0.561 |
|        100 |                     40 |           1.061 |
|        100 |                     45 |           0.791 |
|        100 |                     50 |           0.921 |
|        100 |                     55 |           1.051 |
|       1000 |                     20 |          -0.641 |
|       1000 |                     25 |          -0.391 |
|       1000 |                     30 |          -0.111 |
|       1000 |                     35 |           0.111 |
|       1000 |                     40 |           0.281 |
|       1000 |                     45 |           0.461 |
|       1000 |                     50 |           0.501 |
|       1000 |                     55 |           0.641 |
|      10000 |                     20 |          -1.051 |
|      10000 |                     25 |          -0.671 |
|      10000 |                     30 |          -0.671 |
|      10000 |                     35 |          -0.111 |
|      10000 |                     40 |           0.101 |
|      10000 |                     45 |           0.371 |
|      10000 |                     50 |           0.401 |
|      10000 |                     55 |           0.631 |
|     100000 |                     20 |          -1.741 |
|     100000 |                     25 |          -1.471 |
|     100000 |                     30 |          -0.941 |
|     100000 |                     35 |          -0.561 |
|     100000 |                     40 |          -0.261 |
|     100000 |                     45 |          -0.041 |
|     100000 |                     50 |           0.151 |
|     100000 |                     55 |           0.351 |

ChatGPT and Claude AI were used for this PR, primarily for helping to debug and help me understand the igraph functions I needed to use, as it is my first time working closely with this project.

### DIFF
--- a/src/operators/subgraph.c
+++ b/src/operators/subgraph.c
@@ -338,8 +338,7 @@ static igraph_error_t igraph_i_induced_subgraph_suggest_implementation(
         ratio = (igraph_real_t) num_vs / igraph_vcount(graph);
     }
 
-    /* TODO: needs benchmarking; threshold was chosen totally arbitrarily */
-    if (ratio > 0.5) {
+    if (ratio > log(graph->n) / log(10) / 10) {
         *result = IGRAPH_SUBGRAPH_COPY_AND_DELETE;
     } else {
         *result = IGRAPH_SUBGRAPH_CREATE_FROM_SCRATCH;
--- a/src/operators/subgraph.c
+++ b/src/operators/subgraph.c
@@ -338,7 +338,10 @@ static igraph_error_t igraph_i_induced_subgraph_suggest_implementation(
         ratio = (igraph_real_t) num_vs / igraph_vcount(graph);
     }
 
-    if (ratio > log(graph->n) / log(10) / 10) {
+    /* The threshold of 0.5 is justified by the benchmarking done in
+     * https://github.com/igraph/igraph/pull/2708
+     * Small improvements may be possible by using better heuristics. */
+    if (ratio > 0.5) {
         *result = IGRAPH_SUBGRAPH_COPY_AND_DELETE;
     } else {
         *result = IGRAPH_SUBGRAPH_CREATE_FROM_SCRATCH;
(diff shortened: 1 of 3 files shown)


szhorvat:
@carlos-planelles Are you still interested in working on this PR? If yes, would you be able to check if the estimate for the threshold is valid even for graphs with a higher mean degree?  Currently, the benchmark is for graphs with a mean degree of `2*ecount / vcount = 10`.

It would also be useful to check thresholds for graphs with a heterogeneous degree distribution, such as the ones produced by `igraph_barabasi_game()`. In the undirected case, the mean degree is approximately `2*m` for these graphs. I expect these to behave differently from $G(n,m)$ graphs.

szhorvat:
Thanks @carlos-planelles! I merged the benchmark you wrote but opted to stay with the current threshold of 0.5. Even though we kept the threshold for now, this work was very useful in that now we know that this is a good choice. Previously it was selected arbitrarily.

I think that tightening the threshold should be done with care, and we should quantify the impact of the potential slowdown from using `COPY_AND_DELETE` even for small graphs. If you'd like to work more on this, you are very welcome.

PR #2717, opened 2025-02-23
Allow infinite order in neighborhood functions

szhorvat:
This PR changes `igraph_neighborhood_size()`, `igraph_neighborhood()` and `igraph_neighborhood_graphs()` to allow an infinite order, indicated by a negative `order` value. Previously, a negative order was treated as an error.

It also changes error checking on `mindist`: `mindist > order` throws an error only if the order is not infinite. Do we want to change this to _never_ throw an error for `mindist > order`?

If this PR is accepted, the Python and R docs should be updated as well (CC @maelle).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
  - `igraph_count_triangles()` counts undirected triangles in a graph.
  - `igraph_count_adjacent_triangles()` (rename of `igraph_adjacent_triangles()`)
 
+### Changed
+
+ - `igraph_neighborhood_size()`, `igraph_neighborhood()` and `igraph_neighborhood_graphs()` now accept a negative `order` value and interpret it as infinite order. Previously, a negative `order` value was disallowed. 
+
 ### Deprecated
 
  - The undocumented function `igraph_vector_sumsq()` is deprecated. Use `igraph_blas_dnrm2()` to compute the Euclidean norm of real vectors.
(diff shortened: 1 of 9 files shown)


maelle:
Thanks. If this gets accepted, please open an issue in the rigraph repo so we don't lose track. In R I imagine we'd use `Inf` not a negative value? @schochastics

schochastics:
Yes I think `Inf` would make more sense in R

schochastics:
TIL about the negative values. Guess I never used that feature. I think that should be changed for all functions. But maybe in a systematic way in a separate PR? So leave negative values here and then do a collective change? Does this make sense @maelle?

schochastics:
TIL about the negative values. Guess I never used that feature. I think that should be changed for all functions. But maybe in a systematic way in a separate PR? So leave negative values here and then do a collective change? Does this make sense @maelle?

maelle:
Yes it does :-)

ntamas:
Hmmm I guess this PR is okay from me but it requires a lot of work in the higher level interfaces to update everything in a smooth and backward-compatible manner? Is this worth all the fuss? Realistically, one could simply pass `IGRAPH_INTEGER_MAX` as the order and achieve the same effect on the C side, right?

szhorvat:
@ntamas Does it require fuss in the high-level interfaces? Similar functions just accept `-1` for infinity in current high-level interfaces (except Mathematica), so this would need no other change initially than the documentation.

szhorvat:
This was also inspired by practical teaching experience. If there are no objections, I'd much prefer to merge. Let me know.

ntamas:
As discussed via Slack yesterday, this is good to go.

PR #2738, opened 2025-03-24
Skeleton for rich-club coefficient

zarazong:
This is a draft PR with the basic skeleton for the rich club calculations in properties/rich_club.c (function declarations have been added to igraph_structural.h, and the new .c file to CMakeLists.txt). The parameters and details are still being worked out, especially the options for weighted graphs, etc. - would love feedback on this!

Used ChatGPT to help me understand the igraph library's structure and figure out where best to place the new files. 

### DIFF
--- a/include/igraph_structural.h
+++ b/include/igraph_structural.h
@@ -135,6 +135,19 @@ IGRAPH_EXPORT igraph_error_t igraph_feedback_vertex_set(
     const igraph_t *graph, igraph_vector_int_t *result,
     const igraph_vector_t *vertex_weights, igraph_fvs_algorithm_t algo);
 
+IGRAPH_EXPORT igraph_error_t igraph_rich_club_density_sequence(
+    const igraph_t *graph,
+    const igraph_vector_int_t *vertex_order,
+    igraph_vector_t *res
+    /* other parameters here */
+    );
+
+IGRAPH_EXPORT igraph_error_t igraph_rich_club_coefficient(
+    const igraph_t *graph,
+    igraph_vector_t *res
+    /* other parameters (e.g. bool weighted) here */
+);
+
 /* -------------------------------------------------- */
 /* Spectral Properties                                */
 /* -------------------------------------------------- */
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -121,7 +121,8 @@
 ### Added
 
  - `igraph_count_triangles()` counts undirected triangles in a graph.
- - `igraph_count_adjacent_triangles()` (rename of `igraph_adjacent_triangles()`)
+ - `igraph_count_adjacent_triangles()` (rename of `igraph_adjacent_triangles()`).
+ - `igraph_rng_get_bool()` and `RNG_BOOL()` produce a single random boolean.
 
 ### Changed
 
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -147,7 +147,9 @@ jobs:
 
   - job: documentation
     steps:
-      - script: sudo apt-get install ninja-build xmlto texinfo source-highlight libxml2-utils xsltproc fop docbook2x -y
+      - script: |
+          sudo apt-get update
+          sudo apt-get install ninja-build xmlto texinfo source-highlight libxml2-utils xsltproc fop docbook2x -y
         displayName: Install dependencies
 
       - task: CMake@1
--- a/doc/random.xxml
+++ b/doc/random.xxml
@@ -24,6 +24,7 @@
 </section>
 
 <section id="generating-random-numbers"><title>Generating random numbers</title>
+<!-- doxrox-include igraph_rng_get_bool -->
 <!-- doxrox-include igraph_rng_get_integer -->
 <!-- doxrox-include igraph_rng_get_unif01 -->
 <!-- doxrox-include igraph_rng_get_unif -->
--- a/examples/simple/igraph_assortativity_degree.out
+++ b/examples/simple/igraph_assortativity_degree.out
@@ -1,17 +1,17 @@
 Forest fire model network with 1000 vertices and 0.2 forward burning probability.
 
 Assortativity before rewiring = 0.164775
-Assortativity after rewiring = -0.0126025
+Assortativity after rewiring = -0.0244963
 
-Assortativity before rewiring = 0.151338
-Assortativity after rewiring = -0.00941098
+Assortativity before rewiring = 0.204558
+Assortativity after rewiring = -0.0115178
 
-Assortativity before rewiring = 0.125732
-Assortativity after rewiring = -0.012306
+Assortativity before rewiring = 0.25205
+Assortativity after rewiring = -0.0198939
 
-Assortativity before rewiring = 0.154392
-Assortativity after rewiring = -0.00993259
+Assortativity before rewiring = 0.136678
+Assortativity after rewiring = 0.00245584
 
-Assortativity before rewiring = 0.154686
-Assortativity after rewiring = -0.00258259
+Assortativity before rewiring = 0.262014
+Assortativity after rewiring = -0.00207304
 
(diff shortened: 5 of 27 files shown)


szhorvat:
The failure was on `develop`, and I fixed it now. If  you rebase on the latest develop, it will go away.

However, I'd like to ask you to target this work onto `master` (despite what the instructions said). This is because this is not a breaking change and it will be easier to test / integrate into high-level interfaces on `master`.

include/igraph_structural.h
```diff
@@ -135,6 +135,19 @@ IGRAPH_EXPORT igraph_error_t igraph_feedback_vertex_set(
...
+    const igraph_t *graph,
+    const igraph_vector_int_t *vertex_order,
+    igraph_vector_t *res
+    /* other parameters here */
+    );
+
+IGRAPH_EXPORT igraph_error_t igraph_rich_club_coefficient(
+    const igraph_t *graph,
+    igraph_vector_t *res
+    /* other parameters (e.g. bool weighted) here */
```
szhorvat, line 148: Other parameters I suggest including: 

 - `const igraph_vector_t *weights` (if `NULL`, it's unweighted)
 - `igraph_neimode_t mode` (whether to order vertices based on out-degree, in-degree or total degree)
 - `igraph_bool_t loops` (whether self-loops are assumed possible — later we might think about renaming this to avoid confusion with parameters that control ignoring self-loops)

Something to think about is whether we want a single output vector `phi` where `phi[k]` gives $\phi(k)$, or two output vectors $phi$ and $k$, where some degree values are skipped. The latter may be useful with networks that contain a few very large degree vertices (typical with scale-free networks).

For example, if the degrees are `1, 2, 5`, two possible representations are:

 - `phi[k] = [0.05, 0.1, 0.2, 0.2, 0.2, 0.3]`

or

 - `phi = [0.05, 0.1, 0.2, 0.3]`
 - `k` = [0, 1, 2, 5]`

We can think about what is better, and how the choice affects performance, when you begin thinking about the implementation / algorithm.

zarazong: Got it. I changed the base branch from `develop` to `master` as per your previous comment. However, it's now showing that I'm merging 552 commits into master - is that expected?

During the rebase, I also ran into a merge conflict in `tests/CMakeLists.txt` involving lad, modularity, and spanning_tree. I checked the current version of the file in `develop`, which includes all three, so I ended up keeping all of them. I just want to double-check that this is what was intended?

I'm still quite new to contributing to open source, so I really appreciate your guidance and want to make sure I haven't unintentionally broken anything 😅

zarazong: > Something to think about is whether we want a single output vector `phi` where `phi[k]` gives ϕ ( k ) , or two output vectors ϕ and k , where some degree values are skipped. The latter may be useful with networks that contain a few very large degree vertices (typical with scale-free networks).
> 
> For example, if the degrees are `1, 2, 5`, two possible representations are:
> 
> * `phi[k] = [0.05, 0.1, 0.2, 0.2, 0.2, 0.3]`
> 
> or
> 
> * `phi = [0.05, 0.1, 0.2, 0.3]`
> * `k = [0, 1, 2, 5]`

Would it make sense to consider the typical use cases for a rich club coefficient function in this case --- like its application in brain network analysis which you mentioned in the issue description? I did a quick search on brain networks and other applications, and [one of the first papers I found](https://doi.org/10.1016/j.physleta.2024.129910) was about applying RCC to scale-free networks. Are scale-free networks generally a common use case for RCC? If so, I see why the two-vector approach might be better.

My initial concern was that using two vectors might be unintuitive, with users needing to loop through the `k` vector to find the index before looking up the corresponding `phi` value. But it also seems like igraph is known for being more performant in terms of speed/flexibility compared with NetworkX, and NetworkX already has an implementation, so maybe this extra something is okay?

szhorvat:
> Got it. I changed the base branch from develop to master as per your previous comment. However, it's now showing that I'm merging 552 commits into master - is that expected?

This is because you branched off `develop` in your fork of the repo. Can you please rebase your changes to `master`, either manually or with `git rebase`? Manual transfer of changes, or even just starting a new branch may be easier now.

Alternatively I can help you tomorrow and rebase / force push here.

> Are scale-free networks generally a common use case for RCC? If so, I see why the two-vector approach might be better.

Let's go with this then.

> Just to clarify, by 'assumed possible,' do you mean that:
> 
>     * `loops = TRUE`: If the graph contains any self-loops, they are treated as valid, and we include them in the calculation; but if there are no self-loops, it doesn’t affect the calculation
> 
>     * `loops = FALSE`: If the graph contains self-loops, they are ignored for the calculation even if they exist


The behaviour should be the same as in `igraph_density()`.

It's late here, I'll respond in more detail tomorrow.

zarazong:
Sounds good. I went for the function that computes densities after each vertex removal. What I currently have is below!

I'm running into a small issue with `igraph_delete_vertices()`. My initial idea was to just loop through the vector `vertex_order` given and call `igraph_delete_vertices()` on `igraph_vss_1(VECTOR(*vertex_order)[i]))` on each loop, but then I realized that the function changes the IDs of the remaining vertices so it no longer maps to ones provided by `vertex_order`. 

What is a good way to go about this? I was thinking:
- mark vertices by their original IDs using attributes before doing the removals; or
- working with the old-to-new-IDs mapping provided by `igraph_delete_vertices_idx()`

Or perhaps there is a better strategy?

```
igraph_error_t igraph_i_rich_club_density_sequence(const igraph_t *graph,
                                                  const igraph_vector_int_t *vertex_order,
                                                  igraph_bool_t directed,
                                                  igraph_bool_t loops,
                                                  igraph_vector_t *res) {
    /* TODO:

     * should we check that the vertex_order given is a permutation of the
       vertices of the graph? Or is it valid if vertex_order is only a subset of the vertices?

     * implement directed vs. undirected functionality

    */
    igraph_integer_t numVertices = igraph_vcount(graph);

    /* Error: vertex_order wrong size */
    if (igraph_vector_int_size(vertex_order) != numVertices) {
        IGRAPH_ERROR("Invalid vertex order length.", IGRAPH_EINVAL);
    }

    /* Make sure res is the right size (better way to do this?) */
    IGRAPH_CHECK(igraph_vector_resize(res, numVertices));

    /* create copy of the graph */
    igraph_t graphCopy;
    IGRAPH_CHECK(igraph_copy(&graphCopy, graph));

    /* loop through the number of vertices in the graph */
    for (int i = 0; i < numVertices; i++) {

        /* compute the density */
        igraph_real_t density;
        IGRAPH_CHECK(igraph_density(&graphCopy, &density, loops));

        /* store density in result */
        VECTOR(*res)[i] = density;

        /* remove a vertex */
        IGRAPH_CHECK(igraph_delete_vertices(&graphCopy, igraph_vss_1(VECTOR(*vertex_order)[i])));
    }

    /* destroy */
    igraph_destroy(&graphCopy);

    // TODO: add static back in

    return IGRAPH_SUCCESS;
}
```

szhorvat:
Explicit vertex removal is not going to work. It is much too inefficient. The goal is to devise an efficient algorithm that does not require modifying the graph at all.

For example: Let `m[i]` be the number of edges after `i` vertices have been removed. As a start, iterate through each edge and determine for what `i` it contributes to `m[i]`. Increment `m[i]` accordingly.

What is the complexity of this algorithm? Can we improve on it?

szhorvat:
Did you manage to move to the `master` branch? Do you need me to force-push a change that does this, and if I do it, will you be able to sync that change locally on your machine?

zarazong:
> Did you manage to move to the `master` branch? 
Nope, I left it as is to not further complicate things 😅

> Do you need me to force-push a change that does this, and if I do it, will you be able to sync that change locally on your machine?
Yes and yes, that would be great - thank you so much!

szhorvat:
I'm sorry, it turns out that the current state is such that this would not be easy. Can you please start a new branch? Since so far you've only made a few changes, it should only take a minute to manually re-do them.

Issue #2684, opened 2024-10-22
Investigate faster algorithms for eccentricity-related concepts


szhorvat:
This paper proposes a faster algorithm for diameter/radius, and possibly generalizes more broadly to eccentricity-related concepts:

 * Michele Borassi, Pierluigi Crescenzi, Michel Habib, Walter A. Kosters, Andrea Marino, Frank W. Takes,
Fast diameter and radius BFS-based computation in (weakly connected) real-world graphs: With an application to the six degrees of separation games, Theoretical Computer Science, Volume 586, 2015, Pages 59-80 https://doi.org/10.1016/j.tcs.2015.02.033

NetworkX has an implementation, but is' been de-emphasized, and it's unclear why. See https://github.com/networkx/networkx/issues/7687

We should investigate if this is worth implementing in igraph.

Related: #2683 

@franktakes, let me know if you have any input on the topic. Are there any limitations we should be aware of? Is this method always better than the naive one, and can it serve as the default implementation?

franktakes:
Correct, my algorithm almost always requires fewer BFSes than APSP, and in real-world networks (sparse, with a skewed degree dsitribution, small-world) this reduces the work from _n_ BFSes to a a handful or a few dozen BFSes required to find the diameter. This means orders of magnitude improvement in computation time if you think about large networks. 
To achieve this speedup in computing diamter and radius, there is some extra _O(n)_ bookkeeping going on after each BFS. That time is almost always neglectable compared to the BFSes themselves, which run in _O(m)_ each.

There are some worst-case examples, such as regular graphs (all nodes have the same degree, such as circle graphs) and (near)-complete graphs, where there is little to no speedup compared to the naive implementation. I would imagine that the bookkeeping then makes it a tiny bit slower than a naive implementation. 

Indeed, I once had the algorithm incorporated into networkx, and there it's only used if you use `useBounds = True`. Thanks for pointing me at the thread where they now discuss this, I'll keep an eye on this one and that one as well!

GrishaVar:
@szhorvat I'll start work on this

szhorvat:
Thanks @GrishaVar !

Note for observers: @GrishaVar is a student in my network science class and we are communicating about this project offline as well.

PR #2739, opened 2025-03-25
Bound Diameter

GrishaVar:
Implements #2684


GrishaVar:
@szhorvat This PR now also implements `igraph_set_remove`; doing that was the path of least resistance. See #2287: I guess these sets will be deprecated/replaced at some point (which I think is a good idea), but you'll need an interface for the remove function either way. The implementation has the same issues with iteration as `igraph_set_add`, though I think in my specific case this should only affect performance.

I was hoping everything I do would stay in one PR, let me know if you want the sets in a different one (or if you want a different solution entirely)

src/paths/distances.c
```diff
@@ -1049,3 +1052,131 @@ igraph_error_t igraph_graph_center_dijkstra(
...
+    printf("start: %li\n", vid_start);
+    if (vid_start < 0) {
+        // no given start vertex, so we choose one ourselves
+        igraph_maxdegree_arg(graph, &vid_start, igraph_vss_all(), IGRAPH_ALL, IGRAPH_LOOPS);
+        // TODO: not sure about this vids selector
+        // TODO: something about loop being slower?
+        printf("start: %li\n", vid_start);
+    }
+
+    // initialise set W (line 3)
```
szhorvat, line 1096: Which part of the paper does this refer to?

szhorvat (commented):
Just a few small comments now, we can discuss more in person.

src/paths/distances.c
GrishaVar: Discussed offline; these refer to the "Determining the Diameter of Small World Networks" paper.

GrishaVar:
@szhorvat Added more tests, including Barabasi; they pass. Added clean-up for containers as well, without it the address sanitiser kept complaining. Also, I found an OBOE in the `set_remove` - it was accessing one element too many, but the unit tests didn't detect it because it was the element that got removed at the end. Wouldn't have found it without the sanitiser!

Remaining big TODOs:
- [ ] Deal with disconnected graphs
- [ ] Deal with directed graphs
- [ ] Add vertex-choosing heuristic

tests/unit/set.c
```diff
@@ -69,6 +69,36 @@ int main(void) {
     if (igraph_set_contains(&set, 42) || !igraph_set_contains(&set, 7)) {
         return 3;
     }
+    
+    /* igraph_set_remove */
+    // remove element in the middle
+    igraph_set_remove(&set, 10);
+    if (igraph_set_size(&set) != 20 || igraph_set_contains(&set, 10)) {
+        return 5;
```
szhorvat, line 77: For future tests, it's fine to use `IGRAPH_ASSERT`. Returning a non-zero exit code is an old style that we're abandoning. This test will eventually get rewritten to use `IGRAPH_ASSERT`.

szhorvat:
I wrote up some tips on efficient development: https://github.com/igraph/igraph/wiki/Tips-for-quicker-development

tests/unit/igraph_diameter_bound.c
```diff
@@ -0,0 +1,69 @@
...
+    IGRAPH_ASSERT(result == 12);
+    igraph_destroy(&g);
+
+    // random barabasi - compare to default diameter
+    igraph_rng_seed(igraph_rng_default(), 1234);
+    igraph_barabasi_game(&g, 100, 1, 30, 0, 0, 1, IGRAPH_UNDIRECTED, IGRAPH_BARABASI_BAG, NULL);
+    igraph_diameter_bound(&g, &result, 0, NULL, NULL, IGRAPH_UNDIRECTED, 0);
+    igraph_diameter(&g, &reference, NULL, NULL, NULL, NULL, IGRAPH_UNDIRECTED, 1);
+    IGRAPH_ASSERT(result == reference);  // no reference value to compare to
+    igraph_destroy(&g);
```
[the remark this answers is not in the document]
GrishaVar: The m value is 30 here. I lowered it a bit, otherwise the diameter was always very small

src/paths/distances.c
```diff
@@ -1049,3 +1052,131 @@ igraph_error_t igraph_graph_center_dijkstra(
...
+    while (!IGRAPH_VIT_END(vit)) {
+        igraph_set_add(&W, IGRAPH_VIT_GET(vit));
+        IGRAPH_VIT_NEXT(vit);
+    }
+    // TODO: fill W with nodes from V. Can this be done in a better way than loop?
+    // TODO: clean up set!!
+
+    // initialise upper/lower diameter bounds (line 3)
+    igraph_integer_t dia_lower = INT_MIN;
+    igraph_integer_t dia_upper = INT_MAX;
```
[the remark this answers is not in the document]
szhorvat: We discussed this in person, and agreed to keep `0`.

szhorvat:
@GrishaVar Do you need any input from me right now?

szhorvat:
@GrishaVar Can you please push what you have so far? This is the equivalent of the draft report submission for you.

szhorvat:
> @GrishaVar Can you please push what you have so far? This is the equivalent of the draft report submission for you.

Please don't forget to push.

GrishaVar:
@szhorvat update:
- I've added a heuristic for choosing the next vertex to explore. As the paper suggests, the first vertex is of max degree, after than we take largest upper and smallest lower eccentricity bound in a loop.
- Disconnected graphs are supported now. We explore one component at a time until none are left; infinities from distance calculations are always ignored except to detect the components themselves, once per component. There is one new test for disconnected graphs, the same one `igraph_diameter` uses. Since the algorithm for doing this is not in the paper, the comments in that section are a bit more verbose.
- The adjacency list thing is done, *but* only _unweighted_ graphs. It looks like doing it for weighted graphs is a bit annoying because they use `igraph_lazy_inclist_t` instead. I think we discussed this at some point but I don't remember much about it; we should sync on this.
- Remaining todos:
    - [ ] Convert everything to (optionally) weighted diameter
    - [ ] Get rid of all the prints
    - [ ] Figure out weird maxdegree_arg issue
    - [ ] Add `igraph_set_difference`, would be an O(nlogn) -> O(n) speedup
    - [ ] Documentation and more tests (if needed?)

These are all pretty minor now, hopefully I can get this wrapped up soon :slightly_smiling_face:

szhorvat:
> * Convert everything to (optionally) weighted diameter

Let's put this to the bottom of the priority list, as it is very straightforward.

> * Get rid of all the prints

Optionally, you can use a debug macro that makes it possible to turn this on or off, on demand. You can grep the source tree for `DEBUG` to see a number of variations on this. The style used by `MATCHING_DEBUG` is probably one of the better ones.

It is up to you whether you want to do this. I don't have a strong opinion. You can judge best if people might feel the need to turn on debug logging in the future for this function.

----

Something that would be quite nice to have is a basic benchmark. You can see how these work in `tests/benchmarks`. I suggest looking at the more recent benchmark additions, and copying one.

The purpose of this is to make future tuning of this function easier.

I would put this second-to-last on the priority list, just above the weighted version.

szhorvat:
- [ ] Decide if set_remove needs to return an error code.

szhorvat:
Could you please factor out the cleanup of `debug()` functions into a separate PR?

szhorvat:
Copilot reviews got support for C a few days ago. Let's see if produces anything useful.

ntamas:
@szhorvat Is this ready for merging, do you still want to have another review round, or shall I review it before merging?

szhorvat:
It'd be great if you could do a review, but I want to go through it again in detail before merging. It's best if I do that only after I'm back in Reykjavík. Note: I did not check the set removal function at all yet.

ntamas:
Retargeting for 1.1 as it is not API-breaking and we need to clear the issue list for the 1.0 milestone now that we are getting closer to the release date.

Issue #2726, opened 2025-03-18
Wishlist: Rich-club function


szhorvat:
**What is the feature or improvement you would like to see?**

This issue is to implement functionality for computing the rich-club coefficient / function of a network. The functionality should be sufficiently flexible to support most variations on the rich-club coefficient concept. NetworkX already has an implementation, but it lack flexibility.

There are two parts to this task: (1) literature search to determine what variants are in common use (2) implementation. Once we settle on a concept, the implementation will be easy, and suitable as a first project for a new contributor.

Note that the weighted concept described in Opsahl et al is entirely different from the unweighted one of Colizza et al. This issue is for the unweighted concept from Colizza et al, as well as the straightforward weighted generalization of this unweighted concept (which is different from that of Opsahl).

Only the "unnormalized" version should be implemented, as there are many reasonable null models that could be used for normalization.

Two functions should be implemented:

 1. A function that takes a vertex ordering as input, effectively removes vertices in this order, and computes the density of the remaining graph. The output is a vector of the same length as the vertex count, giving the remaining densities after each removal.
 2. A function that computes $$\phi(k)$$ as a function of the degree, not as a function of the number of removed vertices. Vertices are removed in order of degree.

Things to pay attention to:

 - When computing densities, it matters whether self-loops are considered possible (regardless of whether the graph includes them).
 - When ordering by degree, there should be an option to choose out-, in- or total-degree.
 - If edge weights are given, the density should be defined as the total weight divided by the number of vertex pairs.

**Use cases for the feature**

This is a common network analysis technique that even has a Wikipedia page. It is particularly popular in connectomics (brain network analysis).

**References**

 1. https://en.wikipedia.org/wiki/Rich-club_coefficient
 2. Vittoria Colizza, Alessandro Flammini, M. Ángeles Serrano, and Alessandro Vespignani. ‘Detecting Rich-Club Ordering in Complex Networks’. Nature Physics 2, no. 2 (15 February 2006): 110–15. https://doi.org/10.1038/nphys209.
 3. Tore Opsahl, Vittoria Colizza, Pietro Panzarasa, and José J. Ramasco. ‘Prominence and Control: The Weighted Rich-Club Effect’. Physical Review Letters 101, no. October (2008): 1–4. https://doi.org/10.1103/PhysRevLett.101.168702.
[...]

PR #2740, opened 2025-03-25
feat: igraph_rich_club_sequence()

minifinity:
New PR for [rich-club coefficient](https://github.com/igraph/igraph/issues/2726) --- branching off `master` instead of `develop`. 

Currently working on vertex removal discussed [here](https://github.com/igraph/igraph/pull/2738#issuecomment-2752616899).


zarazong:
Update - new implementation of `igraph_rich_club_density_sequence()` does the following:
- gets list of edges from `graph` via `igraph_get_edgelist()`
- loops through edgelist
  - for each edge, gets `edgeRemovalIndex` of whichever of its endpoints gets removed first
  - increments `edgesRemainingAfter[i]` with `i` being the number of vertices removed, from 0 up to and including `edgeRemovalIndex`
  - derives the vector of densities after each removal from `edgesRemainingAfter` by dividing each value by the number of total possible edges $\frac{n(n-1)}{2}$

Also added a very simple unit test to verify the output. I tested a couple of variations and the outputs seem right so far, but should add more robust tests later.

**Complexity:** I think it's $O(EV)$ - from the below nested for-loop:
```
for (int i = 0; i < numEdges; i++) {  // O(E)
    /* variables for obtaining edgeRemovalIndex */
    igraph_integer_t edgeRemovalIndex = (orderV1 < orderV2 ? orderV1 : orderV2);

    for (int j = 0; j <= edgeRemovalIndex; j++) { // O(V/2)
        VECTOR(edgesRemainingAfter)[j]++; // add to edge remaining count
    }
}
```
Am thinking about how this can be improved by an accumulation procedure!

szhorvat:
> ```
>     for (int j = 0; j <= edgeRemovalIndex; j++) { // O(V/2)
>         VECTOR(edgesRemainingAfter)[j]++; // add to edge remaining count
>     }
> ```
>
> Am thinking about how this can be improved by an accumulation procedure!

I believe the inner loop can be eliminated: instead of incrementing `edgesRemainingAfter` in the `0..i` range, we just increment at `i`. When we're done, we accumulate from the end of the vector towards the beginning. This should give the right result.

I have not taken out pen and pencil to check this carefully (due to lack of time), but I believe something like this should work.

I hope it's clear what I mean. Can you please investigate along these lines? You can prototype in your favourite language (Python or whatever you like) if that makes it easier.

src/properties/rich_club.c
```diff
@@ -0,0 +1,108 @@
...
+                                                 igraph_vector_t *res) {
+    /* TODO:
+     * implement directed vs. undirected functionality
+     */
+
+    /**
+    // Error: vertex_order wrong size
+    if (igraph_vector_int_size(vertex_order) != numVertices) {
+        IGRAPH_ERROR("Invalid vertex order length.", IGRAPH_EINVAL);
+    }*/
```
szhorvat, line 58: This is good. You can uncomment it.

Do not use `/**` for comments unless you are writing documentation, as this is interpreted specially. Use `/*` only. We can discuss documentation later.

szhorvat:
This is a good start!

I suggest you use [python-igraph](https://python.igraph.org/) (or another high-level interface) to verify the result here. Use the naive method: delete vertices one by one and recompute the density after each step. As you discovered, vertex indices changes after removal, but you can give vertices string names, and refer to them this way.

The naive method is slow, so it is not suitable for a C implementation. But it is also simple, so it is a good way to test your C implementation.

src/properties/rich_club.c
```diff
@@ -0,0 +1,108 @@
...
+        for (int j = 0; j <= edgeRemovalIndex; j++) {
+            VECTOR(edgesRemainingAfter)[j]++; // add to edge remaining count
+        }
+    }
+
+    igraph_integer_t totalPossibleVertices = numVertices * (numVertices - 1) / 2;
+
+    // density calculation
+    for (int i = 0; i < numVertices; i++) {
+        VECTOR(*res)[i] = VECTOR(edgesRemainingAfter)[i] / (igraph_real_t) totalPossibleVertices;
```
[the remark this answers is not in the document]
zarazong: > This should be recomputed for each step in this loop, as each step involves the removal of a vertex.

Oops I overlooked this - fixed now!
For graphs with 1 vertex or 0 vertices, is the density 0 or undefined?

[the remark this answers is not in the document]
szhorvat: When the graph has $n$ vertices let's produce a density vector `rho` of length `n`, where `rho[i]` gives the density after removing `i` nodes. Thus the first element, `rho[0]`, is the density of the original graph. The last element, `rho[n-1]`, is the density when a single vertex remains. There is no element for the 0-vertex graph, as this is not meaningful.

When working with simple graphs, `rho[n-1]` will be calculates as 0/0 = NaN. We don't need to special-case this, NaN is fine. When allowing self-loops, the density of a single-vertex graph can be either 0 or 1, depending on whether the loop is present.

zarazong:
> I believe the inner loop can be eliminated: instead of incrementing `edgesRemainingAfter` in the `0..i` range, we just increment at `i`. When we're done, we accumulate from the end of the vector towards the beginning. This should give the right result and reduce complexity to O ( E + V ) .
> 
> I have not taken out a pen and pencil to check this carefully (due to lack of time), but I believe something like this should work.

That's very cool! I hadn't thought about it that way. I ran through a simple example on paper, and I think since every index holds the number of edges for which that index is the last index where the edge remains, summing from end to beginning collapses the repeated counts from the initial inner loop. I added this example into the unit test and the output matched, but I'll verify this and the program as a whole more extensively using python-igraph.

> * `igraph_vector_int_init()` can fail, therefore anything that calls it can also fail and must return an error code.

Quick question: what’s the difference between `IGRAPH_VECTOR_INT_INIT_FINALLY()` and `igraph_vector_int_init()`? I noticed tha in some other files, vectors are intialized with `IGRAPH_VECTOR_INT_INIT_FINALLY()` without being wrapped in `IGRAPH_CHECK()` macros. Is one fail-able and the other not?

Lastly, the latest commit includes all the other fixes I made per your feedback (thank you!) They include:
- the accumulation method for `edgesRemainingAfter vector`
- added `IGRAPH_CHECK()` macros for fail-able functions
- helper function `igraph_i_get_vertex_order_map()` now returns its map as an output argument instead of a return value
- fixed the density calculation error and updated the unit test results (which had the same logic error)

src/properties/rich_club.c
```diff
@@ -0,0 +1,116 @@
...
+    igraph_integer_t total = 0;
+    for (igraph_integer_t i = numVertices - 1; i >= 0; i--) { // O(V)
+        total += VECTOR(edgesRemainingAfter)[i];
+        VECTOR(edgesRemainingAfter)[i] = total;
+    }
+
+    // density calculation
+    for (igraph_integer_t i = 0; i < numVertices; i++) { // O(V)
+        // (numVertices - i) = the number of vertices left on this loop
+        igraph_real_t totalPossibleEdges = (numVertices - i) * ((numVertices - i) - 1) / 2;
```
szhorvat, line 96: This is still calculated using integer arithmetic, which risks overflow. Either cast the first term to real, or introduce reals in the expression, with decimal points, to force floating point arithmetic. Example of doing both:

```suggestion
        igraph_real_t totalPossibleEdges = (igraph_real_t) (numVertices - i) * ((numVertices - i) - 1.0) / 2.0;
```

You can just commit this suggestion.

src/properties/rich_club.c
```diff
@@ -0,0 +1,116 @@
...
+    igraph_integer_t numEdges = igraph_ecount(graph);
+    igraph_vector_int_t edges;
+    igraph_vector_int_t edgesRemainingAfter;
+
+    IGRAPH_CHECK(igraph_vector_int_init(&edges, 0));
+    IGRAPH_CHECK(igraph_vector_int_init(&edgesRemainingAfter, numVertices));
+
+    IGRAPH_CHECK(igraph_vector_resize(res, numVertices)); // resize res
+
+    IGRAPH_CHECK(igraph_get_edgelist(graph, &edges, 0)); // get list of edges
```
szhorvat, line 68: To iterate over edges, simple write a for loop for `eid = 0 .. ecount-1`. Use `IGRAPH_FROM(graph, eid)` and `IGRAPH_TO(graph, eid)` to get endpoints.

src/properties/rich_club.c
```diff
@@ -0,0 +1,116 @@
...
+
+    // get map of vertex ordering -> orderOf[id] = index of vertexID in vertex_order
+    igraph_vector_int_t orderOf;
+    IGRAPH_CHECK(igraph_vector_int_init(&orderOf, numVertices));
+    igraph_i_get_vertex_order_map(vertex_order, &orderOf);
```
szhorvat, line 73: Also needs an error check.

src/properties/rich_club.c
```diff
@@ -0,0 +1,116 @@
...
+    // density calculation
+    for (igraph_integer_t i = 0; i < numVertices; i++) { // O(V)
+        // (numVertices - i) = the number of vertices left on this loop
+        igraph_real_t totalPossibleEdges = (numVertices - i) * ((numVertices - i) - 1) / 2;
+        VECTOR(*res)[i] = VECTOR(edgesRemainingAfter)[i] / totalPossibleEdges;
+    }
+
+    igraph_vector_int_destroy(&edges);
+    igraph_vector_int_destroy(&orderOf);
+    igraph_vector_int_destroy(&edgesRemainingAfter);
```
szhorvat, line 102: The code is in good shape so we can start working on error handling a bit.

Add here:

```
IGRAPH_FINALLY_CLEAN(3);
```

This removed the 3 destructors from the "finally" stack.

Please also order destructor calls in the opposite order of init calls. This is for readability and clarity.

src/properties/rich_club.c
```diff
@@ -0,0 +1,116 @@
...
+    if (igraph_vector_int_size(vertex_order) != numVertices) {
+        IGRAPH_ERROR("Invalid vertex order length.", IGRAPH_EINVAL);
+    }
+
+    igraph_integer_t numEdges = igraph_ecount(graph);
+    igraph_vector_int_t edges;
+    igraph_vector_int_t edgesRemainingAfter;
+
+    IGRAPH_CHECK(igraph_vector_int_init(&edges, 0));
+    IGRAPH_CHECK(igraph_vector_int_init(&edgesRemainingAfter, numVertices));
```
szhorvat, line 64: Please change all vector inits to `IGRAPH_VECTOR_INT_INIT_FINALLY`. Then the `IGRAPH_CHECK` is not needed.

tests/unit/igraph_rich_club.c
```diff
@@ -0,0 +1,60 @@
...
+
+#include <stdio.h>
+#include <igraph.h>
+#include "test_utilities.h"
+
+int main(void) {
+    igraph_t graph;
+    igraph_vector_t res;
+    igraph_vector_int_t vertex_order;
+    const igraph_integer_t N_VERTICES = 6;
```
szhorvat, line 27: Please do not use capitals for variables. We reserve all-caps for macros.

tests/unit/igraph_rich_club.c
```diff
@@ -0,0 +1,60 @@
...
+        VECTOR(vertex_order)[i] = i;
+    }
+
+    IGRAPH_CHECK(igraph_rich_club_density_sequence(&graph, &vertex_order, 0, 0, &res));
+
+    printf("Test 3 (disjoint graph):\n");
+    for (int i = 0; i < igraph_vector_size(&res); i++){
+        printf("%.4f ", VECTOR(res)[i]);
+    }
+    printf("\n");
```
szhorvat, line 49: Please replace this with `print_vector()`. This function is defined in `test_utilities.h`, which you already included. It not only prints more readably, it also ensures that NaN is printed in a platform-independent way.

tests/unit/igraph_rich_club.c
```diff
@@ -0,0 +1,60 @@
...
+    /* GRAPH FORMAT
+       1-4 0-3-2-5
+     */
+    IGRAPH_CHECK(igraph_small(&graph, N_VERTICES, IGRAPH_UNDIRECTED,
+                              1, 4, 2, 3, 0, 3, 2, 5, -1));
+
+    igraph_vector_init(&res, 0);
+    igraph_vector_int_init(&vertex_order, N_VERTICES);
+
+    // vertex removal order: 0, 1, 2, 3, 4, 5
```
szhorvat, line 38: Let's use `igraph_vector_sort_ind` to sort vertices by degree. Create a degree vector, apply `igraph_vector_sort_ind` to it. It will give you a permutation that you can pass directly to `igraph_rich_club_density_sequence()`

szhorvat (commented):
Good work so far! I left a few comment and will look again tomorrow.

szhorvat:
Please also add a test for the null graph (zero vertices) and the singleton graph (one vertex).

szhorvat:
Can you please also try to get the directed / loopy case done?  The only difference it makes is in how you calculate `totalPossibleEdges`, the rest is the same.

When all this is done, can you add weight support? Take a pointer to an `igraph_vector_t `weight vector. If it's `NULL`, we do the same thing as now (unweighted). If there are weights, increment `edgesRemainingAfter` not by 1, but by the weight. All other things are the same, but we need to switch to using real values instead of integers throughout. You can even use something like `+= weighted ? VECTOR(*weights)[eid] : 1`.

Finally, let's add a safety check: if `loops=false` and you encounter a self-loop, issue a warning (`IGRAPH_WARNING`), but otherwise proceed as normal. Make sure that no more than a single self-loop warning is printed, even if there are multiple self-loops. Use a flag to keep track of whether the warning was already issued.

tests/unit/igraph_rich_club.c
szhorvat: I'm sorry, scratch that. This was a bad idea on my part as the ordering won't be consistent when there is more than one vertex with the same degree.

Still you can use `igraph_vector_int_init_seq()` to directly create a properly ordered vector.

It would be good to have a test that does not use the original vertex order, but something else. For example, also run the function with a reversed vector (see `igraph_vector_int_reverse()`).

zarazong:
Thanks again for the detailed feedback - that's been _really_ helpful! I have a bit more on my plate tomorrow, but will get on this later tomorrow night or the following day.

szhorvat:
I wrote up some tips on efficient development: https://github.com/igraph/igraph/wiki/Tips-for-quicker-development

src/properties/rich_club.c
```diff
@@ -0,0 +1,116 @@
+/*
+   IGraph library.
+   Copyright (C) 2022  The igraph development team <igraph@igraph.org>
```
szhorvat, line 3:
```suggestion
   Copyright (C) 2025  The igraph development team <igraph@igraph.org>
```

zarazong:
Added:
- `totalPossibleEdges` for directed + loops cases:

    |                            | **no loops**          | **loops**                |
    | -----------        | -----------             | ----                         |
    | **undirected**  | $\frac{n(n-1)}{2}$ | $\frac{n(n+1)}{2}$ |
    | **directed**      | $n(n-1)$                 | $n^2$                    |
- `weights` parameter --- `edgesRemainingAfter` should increment by edge weight instead of 1 if `weights` is not null
- safety check for when `loops = false` and there is a self-loop
   - to clarify --- if `loops = false` and there is a self-loop, we calculate the denominator `totalPossibleEdges` as if there were no loops, but still count the edge (and its weight if weighted)? Is that the distinction between "ignoring self-loops" and "assuming they aren't possible"?
- null graph and singleton graph tests, and vertex removal order test (in-order vs reversed) for a more typical case. Currently, these still only test for the no loop + undirected + unweighted case --- I will add tests to cover the rest next.

szhorvat:
> * to clarify --- if `loops = false` and there is a self-loop, we calculate the denominator `totalPossibleEdges` as if there were no loops, but still count the edge (and its weight if weighted)?

Yes, correct. Let's make it the user's responsibility to deal with this situation correctly. They might want to remove loops from the graph, or they might want to keep loops and set `loops=true` in this function. This is consistent with how the `igraph_density()` works.

szhorvat:
This is close to being done.

Can you please:

 - Make sure the CI tests pass.
 - Remove `igraph_rich_club_coefficient()` completely from this PR. I'm thinking it's better if we finish the first function, merge this PR, then implement the second function in a new PR. If you intend to submit a GSoC application, this will give you time to work on that.
 - Make sure the input is validated. We need an error check for the length of the weights vector.
 - Clean up and finalize the code. For example at this time you are not using the `edges` vector: remove it.
 - Add documentation.
 - Add the new function to `interfaces/functions.yaml`.

If it's not clear how to do some of these, let me know. I'll take a detailed look at the code later.

zarazong:
Commenting to say I saw this - working on it!

zarazong:
Updated:
> * [x] Make sure the CI tests pass.
> * [x] Remove `igraph_rich_club_coefficient()` completely from this PR. I'm thinking it's better if we finish the first function, merge this PR, then implement the second function in a new PR. If you intend to submit a GSoC application, this will give you time to work on that.
> * [x] Make sure the input is validated. We need an error check for the length of the weights vector.
> * [x] Clean up and finalize the code. For example at this time you are not using the `edges` vector: remove it.

Some additional questions:
> * [x] Add documentation.

- Should I leave my comments for the helper functions in the file, or take them out completely? And if they are left in, should I use `/*` instead of `/**` since they are not part of the documentation?
- I also have the following as a part of the main description:
  ```
   * Vertex order can be sorted by degree so that the resulting density sequence helps
   * reveal how interconnected a graph is across different degree levels, or the presence
   * of a "rich-club" effect.
  ```
  I was wondering whether I should include this, since the output vector of `igraph_rich_club_density_sequence()` is not exactly that of $\phi(k)$ as a function of degree. Would it be confusing?

> * [ ] Add the new function to `interfaces/functions.yaml`.

-  Sorry, I'm having trouble understanding the format of the file. Would this be correct:
  ```
  igraph_rich_club_density_sequence:
    PARAMS: |-
        GRAPH graph, 
        VECTOR_INT vertex_order, 
        BOOLEAN directed,
        BOOLEAN loops, 
        VECTOR weights,
        OUT VECTOR res
  ```
  Additionally, is there "reasonable" default value for `directed` and `loops`?

zarazong:
added a few more tests for the weighted vs unweighted and loops vs no loops cases, and one for a weighted graph!

szhorvat:
Could you please have a look at the failing CI tests?

Tips: 

 - It's usually best to look at the `linux_static_vendored` test first, as it has sanitizers enabled. It seem there's a memory leak here, can you track it down?
 - Since only a few CI tests are failing, likely it's an issue that is caught by sanitizers.
 - Do not worry about the Travis test failing.

I'll take a detailed look as soon as I can.

szhorvat:
See the commit I just pushed: this includes the function in the docs.

src/properties/rich_club.c
```diff
@@ -0,0 +1,167 @@
...
+        }
+        igraph_integer_t orderV1 = VECTOR(orderOf)[v1]; // order of endpoints
+        igraph_integer_t orderV2 = VECTOR(orderOf)[v2];
+
+        igraph_integer_t edgeRemovalIndex = (orderV1 < orderV2 ? orderV1 : orderV2);
+        VECTOR(edgesRemainingAfter)[edgeRemovalIndex] += (weights ? VECTOR(*weights)[eid] : 1);
+    }
+
+    // edgesRemainingAfter vector: edges (or total edge weight) remaining after i removals
+    igraph_integer_t total = 0;
```
szhorvat, line 148: Attention, now we have reals since we're dealing with weights. This is a common mistake when generalizing a non-weighted implementation to a weighted one. I'll commit the fix, you don't need t worry about it now.

```suggestion
    igraph_integer_t total = 0;
```

szhorvat:
Note: the usual meaning of a `directed` parameter is to control whether edge directions are considered. An undirected graph should never be treated as directed (not possible for most algorithms), but a directed graph may sometimes be treated as undirected.

We'll discuss within the team whether the `directed` parameter is a good idea or should be removed, but you don't need to worry about this.

See my last commit that deal with this.

szhorvat:
I just reordered the function parameters to comply with our standard, documented here:

https://github.com/igraph/igraph/wiki/Guidelines-for-function-argument-ordering

If in the future we ask you to rename variable or reorder parameters, use an IDE, as it makes such tasks easy. I used CLion for this.

szhorvat:
@zarazong This is basically ready, and looks good to me. Thank you for implementing this! Could you please help with two minor tasks?

 - Make sure that weighted test uses at least some non-integer weights. This would catch the bug I pointed out earlier.
 - Please use `NULL` instead of `0` for pointers, and `true`/`false` instead of `1`/`0` for booleans. This is purely for readability.

szhorvat:
> * Sorry, I'm having trouble understanding the format of the file. Would this be correct:
> 
> 
> ```
> igraph_rich_club_density_sequence:
>   PARAMS: |-
>       GRAPH graph, 
>       VECTOR_INT vertex_order, 
>       BOOLEAN directed,
>       BOOLEAN loops, 
>       VECTOR weights,
>       OUT VECTOR res
> ```

I pushed a commit that adds this.

For `directed`, we use a default of true, meaning: treat directed graphs as directed by default. For `loops`, we use a default of false, for consistency with `igraph_density()`. These defaults will be used in high-level interfaces, such as igraph's R interface.

szhorvat:
@ntamas Can you please review this? Please do not merge, just review.

@schochastics @ntamas What do you think about including the `directed` parameter? Its use would be very niche, but I can imagine situations where we want to compute the density of a directed graph as undirected. For example, in some applications it is useful/necessary to orient the edges of an undirected graph, but the graph still described an undirected network (e.g. when computing Kirchhoff flows, we need a reference direction for flows). Being able to pass `directed=false` provides a minor convenience in such cases.

szhorvat:
I don't want to step on your toes and create merge conflicts, so I won't push this fix, but I just noticed that `igraph_vector_t *weights` should be `const igraph_vector_t *weights`. Can you fix this as well please?

zarazong:
@szhorvat Thank you so much for all your guidance this past week - I have so much to learn both conceptually about graphs and about open-source contributing itself, but your feedback made all the difference, and I've already learned a ton from this PR. I’m going to shift focus to the GSoC proposal and will follow up on that soon. Thanks again!

Tagl (approved):
Looks great!

My one comment would be that having some test that hits the error handling and/or warning would be nice.

szhorvat:
@zarazong You do not need to act on the comments below, it's just a bit of brainstorming.

----

I keep thinking about ways to generalize this, and make it useful for many different kinds of applications.

Essentially, we are just  computing how many edges remain after certain vertices are removed, and normalizing these edge counts in a specific way to obtain densities.

I propose the following modification:

 - Remove the `loops` parameter, but keep `directed`.
 - Introduce a `normalization` parameter which controls how edge counts are normalized. Possible options: `UNNORMALIZED` keeps raw edge counts. `DENSITY` computes the density, assuming no loops. `DENSITY_LOOPS` assumes that loops are possible. `MEAN_DEGREE` computes the mean degree of the remaining graph instead of its density. Of these, both density and mean degree calculations would be affected by the `directed` parameter, so it makes sense to keep that.

If we normalize by a null model later, transforming the edge count to a density does not matter at all, and could be skipped anyway. Since rich club analysis almost always involves normalizing with a null model, one almost never needs to look at density values. 

If we do need to look at these values, then one might argue that whether we want to look at the density or the mean degree is application-specific. However, it's good to note that the density is always in $[0,1]$ (for simple graphs), while the mean degree's upper bound is constrained by the graph size. Thus when vertices are removed in order of increasing degree, the density tends to increase monotonically, while the mean degree first increases then decreases. When removing vertices in a different order (e.g. to simulate an attack on the network), it is unclear which measure is best to look at.

src/properties/rich_club.c
```diff
@@ -0,0 +1,172 @@
...
+
+#include "igraph_interface.h"
+
+/* Given a vertex order, computes a mapping such that for each index = vertex ID, the
+ * corresponding value is its position in the order.
+ *
+ * Index: ordering  -> Index: vertex ID
+ * Value: vertex ID    Value: ordering
+ */
+static igraph_error_t get_vertex_order_map(const igraph_vector_int_t *vertex_order,
```
ntamas, line 29: This is also implemented in Cliquer in `cliquer/reorder.c` in `reorder_invert` (although it does the inversion in-place by copying the results back into the original vector). I wonder whether this operation is common enough so that it would warrant the addition of `igraph_vector_int_invert()` or something like that. Just an idea, this is not a blocker for merging the current PR, I'm just a tad bit worried that we'll keep on implementing this for other functions as well if we keep this function internal.

tests/CMakeLists.txt
```diff
@@ -378,6 +378,11 @@ add_legacy_tests(
   bug_2517
 )
 
+add_legacy_tests(
+  FOLDER tests/unit NAMES
+  igraph_rich_club
```
ntamas, line 383: There's another command that adds tests from the `tests/unit` folder, starting from line 282. `igraph_rich_club` should be added there instead of starting a new section here.

ntamas (changes requested):
Only one important comment to address and then this is good to go.

src/properties/rich_club.c
szhorvat: Good point about this function. This is why I like to have a second pair of eyes on PRs.

This is really part of the "cleaning up permutations and mappings" project, #1930

One decision to be made is if we think of this vector as a permutation (meaning that the values are indices) or as a list of vertices (meaning that the values are vertex IDs). Do we want a `vs_t` here or a `vector_int_t`? It's better to think of it as a permutation, as that is a more general concept, and can be obtained with functions like `igraph_vector_sort_int()`, by sorting degrees or any other vertex property. We should note this in the docs (I will do it).

Finally, we must validate the permutation vector and forbid duplicate indices. This is an additional argument in favour of having a separate function for this.

zarazong:
Sidenote: I think the [link](https://igraph.org/c/pdf/latest/igraph-docs.pdf) to the PDF version of the documentation broke?

ntamas:
The link is now fixed, thanks!

src/properties/rich_club.c
szhorvat: I replaced it with `igraph_invert_permutation()`, which also checks for duplicate vertices.

tests/unit/rich_club.c
```diff
@@ -0,0 +1,288 @@
...
+#include <igraph.h>
+#include "test_utilities.h"
+
+void null_graph(void) {
+    igraph_t graph;
+    igraph_vector_t result;
+    igraph_vector_int_t vertex_order;
+
+    igraph_vector_init(&result, 0);
+    igraph_vector_int_init(&vertex_order, 0);
```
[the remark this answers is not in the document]
szhorvat: @Copilot This is not necessary in tests.

szhorvat:
@ntamas, can you please review, then squash-and-merge? Be sure to squash!

szhorvat (approved):

PR #2741, opened 2025-03-30
Mycielskian construction and Mycielski graph

### LINKED ISSUE #2730: Wishlist: Mycielskian

Implement the Mycielski construction.

Suggested prototype:

```c
igraph_error_t igraph_mycielskian(igraph_t *res, const igraph_t *graph, igraph_integer_t k);
```

The `k` parameter controls how many times the Mycielski construction should be applied to the input graph.

A second functions should construct the Mycielski graph $M_k$ by applying the construction `k` times to the singleton graph.

```c
igraph_error_t igraph_mycielski_graph(igraph_t *graph, igraph_integer_t k);
```

This functionality is already in IGraph/M, which can be used as a reference.

Review how other systems handle this (NetworkX, SageMath, Maple, Magma, etc.) to collect ideas before starting the implementation.

**Use cases for the feature**

Useful for the study of graph colouring.

**References**

 - https://en.wikipedia.org/wiki/Mycielskian
 - http://szhorvat.net/mathematica/IGDocumentation/#igmycielskian

gulshan-123:

I have first pre calculated the number of vertices and edges required. Then copied the original edges to the new vector.
Carefully updating the edge vector.
Then at once in the last all the edges are added from the vector.

Basic structure and some code for the issue #2730  is written.
Simple tests are tested locally by comparing outputs from the function and NetworkX python implementation.
Need to do some more error handling may be. To add documentations and test.

igraph_mycielski_graph remains to be done..

UPDATE: igraph_mycielski --> also implemented

Fixes #2730


szhorvat:
> igraph_mycielski_graph remains to be done..

This can just call `igraph_mycielskian()` on a singleton graph.

I'll try to review it by tonight UTC+0. On first skim, it looks very neat! :-)

src/constructors/mycielskian.c
```diff
@@ -0,0 +1,119 @@
+/*
+   IGraph library.
+   Copyright (C) 2022  The igraph development team <igraph@igraph.org>
```
szhorvat, line 3:
```suggestion
   Copyright (C) 2025  The igraph development team <igraph@igraph.org>
```

src/constructors/mycielskian.c
```diff
@@ -0,0 +1,119 @@
...
+#include "igraph_conversion.h"
+
+#include "igraph_interface.h"
+
+#include "core/interruption.h"
+#include "math/safe_intop.h"
+
+igraph_error_t igraph_mycielskian(igraph_t *res, const igraph_t *graph, igraph_integer_t k) {
+    if (k < 0) {
+        IGRAPH_ERROR("Number of iterations (k) must be non-negative", IGRAPH_EINVAL);
```
szhorvat, line 29: All error messages must be punctuated. When writing messages, keep in mind that the error message will be shown by different high-level interfaces, within different host languages. Let's try to avoid it looking confusing in some of these contexts, and remove the mention of `k`. 

```suggestion
        IGRAPH_ERROR("The number of Mycielski iterations must not be negative.", IGRAPH_EINVAL);
```

src/constructors/mycielskian.c
```diff
@@ -0,0 +1,119 @@
...
+    }
+    if (graph == NULL || res == NULL) {
+        IGRAPH_ERROR("Invalid input/output graph", IGRAPH_EINVAL);
+    }
```
szhorvat, line 33:
```suggestion
```

We usually do not check for errors that can only be committed in C. The principle is to validate input that can be passed from high-level interfaces.

_If_ we added a check here, it should be an `IGRAPH_ASSERT` instead, but in this case let us not add any check.

src/constructors/mycielskian.c
```diff
@@ -0,0 +1,119 @@
...
+    if (igraph_is_directed(graph)) {
+        IGRAPH_ERROR("Mycielski's construction is not defined for directed graphs", IGRAPH_EINVAL);
```
szhorvat, line 35: Please make sure that all error messages have a fullstop at the end, including the ones I don't comment on.

src/constructors/mycielskian.c
```diff
@@ -0,0 +1,119 @@
...
+    // igraph_small(&g, 1, 0, -1);
+
+    if (k <= 0) {
+        IGRAPH_ERROR("Number of iterations (k) must be positive", IGRAPH_EINVAL);
+    }
+    if (graph == NULL) {
+        IGRAPH_ERROR("Invalid Output graph", IGRAPH_EINVAL);
+    }
+    if (k == 1) {
+        igraph_small(&g, 1, IGRAPH_UNDIRECTED, -1); // single vertex
```
szhorvat, line 109: Please avoid `igraph_small()` in library code unless there is no convenient alternative. Do use `igraph_small()` in tests.

You can use `igraph_empty()` here.

szhorvat (changes requested):
I left a few small comments but did not check everything carefully yet.

Can you please add tests and documentation before we do a full review? The tests are especially important.

Make sure that you check that you include a check that applying the Mycielski construction to the singleton graph 3 times gives a graph isomorphic to the Grötzsch graph. You can create the Grötzsch graph using `igraph_famous()`.

Let me know if you need pointers for how to add tests and docs.

gulshan-123:
Thank you for the feedback! I'll add the tests and documentation before the full review.
I will read the wiki pages on how to add documentations and unit tests. Also, I will look through docs of IGraphM and NetworkX to get some idea. Will do these by Tomorrow Noon.

szhorvat:
The docs don't need to be too detailed. I will update them before merging the PR. Just make sure to have the basics structure ready, include a basic explanation (without making it too long) and ensure that the two functions are included in the appropriate xxml file for the "Graph generators" and "Graph operators" chapters, respectively.

Make sure that all CI tests pass. Note that the Travis test occasionally fail; don't worry about that.

szhorvat:
I looked a bit more carefully at the code. Here are a few more things that should be changed.

 - We need to make a decision about computing the Mycielskian of the singleton graph. IGraph/M currently special cases this: it adds a single extra vertex and connects it to the existing vertex. NetworkX and SageMath go on with the standard version of the construction, but this yields a disconnected graph, which is somewhat pointless. Adding a single vertex and connecting it produces a graph that carries most properties that we'd get from the standard construction (incremented chromatic number, triangle free, etc.). I'm currently leaning towards the special case, but a literature search would be in order.
 - The number of vertices and edges increases exponentially with the number of iterations, so we must do overflow checks using the facilities of the `safe_intop.h` header, which you already included. Coming up for closed-form formulas for the graph size might be helpful here. Here's the result: The vertex count goes from $n$ to $2n+1$ in one step. The $k$-th iterated Mycielskian has $n_k = (n + 1) 2^k  - 1$ vertices The edge count goes up from $m$ to $3m+n$, as we add 2 edges for every existing edge and 1 edge for every vertex. One can show then that $m_k = \frac{1}{2} \left( (2m + 2n + 1) 3^k - n_{k+1} \right)$.
 - Currently you are sometimes using `igraph_copy()`, and sometimes take the edges of the original graph and use them to construct a new one. These are not equivalent, as `igraph_copy()` also copies attributes. I suggest to drop all attributes, i.e. do not use `igraph_copy()` at all.
 - To create a $P_2$ graph in `igraph_mycielski_graph()`, you can use `igraph_full()` with `n=2` (or even `igraph_ring()`).

gulshan-123:
Ok, I will take all of this feedback into account and make the necessary changes


> * We need to make a decision about computing the Mycielskian of the singleton graph. IGraph/M currently special cases this: it adds a single extra vertex and connects it to the existing vertex. NetworkX and SageMath go on with the standard version of the construction, but this yields a disconnected graph, which is somewhat pointless. Adding a single vertex and connecting it produces a graph that carries most properties that we'd get from the standard construction (incremented chromatic number, triangle free, etc.). I'm currently leaning towards the special case, but a literature search would be in order.

maintaining connectivity would be a better idea, so I will handle this case specially. Should I directly use the mycielski graph function? (k+1 th term in mycielski graph for k iteration of construction for singleton graph)
 Also, IGraph/M handles special case of empty graph returning a single vertex for one construction, should I handle it here too? (k iteration of construction gives the kth term in mycielski graph)

gulshan-123:
> * The number of vertices and edges increases exponentially with the number of iterations, so we must do overflow checks using the facilities of the `safe_intop.h` header, which you already included. Coming up for closed-form formulas for the graph size might be helpful here.

should we update new_vcount and new_ecount in a loop but with overflow checking? As in the closed form:
1) There is 3^k term which is not directly computable with inbuilt igraph overflow checking, (need to use pow(), as in constructors/de_bruijn.c)
2) In the calculation of ecount, there is a minus sign between two terms, so may be, both terms overflow but the difference terms do not.
But checking overflow in each loop will involve a lot of work. Please suggest what should I do.

szhorvat:
> maintaining connectivity would be a better idea, so I will handle this case specially

OK, let's do this.

> Should I directly use the mycielski graph function? (k+1 th term in mycielski graph for k iteration of construction for singleton graph)

If you're asking whether `igraph_mycielskian()` should call `igraph_mycielski_graph()`, please don't do that.

> Also, IGraph/M handles special case of empty graph returning a single vertex for one construction, should I handle it here too? (k iteration of construction gives the kth term in mycielski graph)

Good point, let's do that.

> should we update new_vcount and new_ecount in a loop but with overflow checking?

Yes, let's go with the simple solution for now (i.e. calculation in a loop). There a linear number of overflow checks in $k$, but an exponential number of edges added. Thus the overflow checks' will be overall negligible.

But let us include the edge/vertex count of the result in the documentation, for the user's reference.

gulshan-123:
Hi @szhorvat , I have added the tests: for mycielski graph, I added and asserted for k=1,2,3 and 4.
For mycielski, I verified manually about some small graphs and then added for a path (iterations=2 giving Grotzsch graph), a single vertex (iterations=3 giving again grotzsh graph).
Please review.
I am yet to add documentations.

src/constructors/mycielskian.c
```diff
@@ -0,0 +1,148 @@
...
+    }
+
+    if (igraph_is_directed(graph))
+        IGRAPH_CHECK(igraph_empty(res, new_vcount, IGRAPH_DIRECTED));
+    else
+        IGRAPH_CHECK(igraph_empty(res, new_vcount, IGRAPH_UNDIRECTED));
+
+    igraph_vector_int_t edges;
+    IGRAPH_VECTOR_INT_INIT_FINALLY(&edges, 0);
+    IGRAPH_CHECK(igraph_get_edgelist(graph, &edges, false)); // copy the edges from the original graph to the new vector
```
szhorvat, line 87:
```suggestion

    // Copy the edges from the original graph to the new vector
    IGRAPH_CHECK(igraph_get_edgelist(graph, &edges, /* bycol */ false)); 
```

Let's keep line length reasonable for readability. When the line would get very long, put comments on a separate line.

src/constructors/mycielskian.c
```diff
@@ -0,0 +1,148 @@
...
+            VECTOR(edges)[edge_index++] = j;
+            VECTOR(edges)[edge_index++] = w;
+        }
+
+        // Update offset for next step
+        offset = offset * 2 + 1;
+    }
+
+    // Add all edges in one go
+    IGRAPH_CHECK(igraph_add_edges(res, &edges, 0));
```
szhorvat, line 121:
```suggestion
    IGRAPH_CHECK(igraph_add_edges(res, &edges, NULL));
```

Please use `NULL` for null pointers and `true` / `false` for Boolean constants. This is more readable than 1 and 0.

src/constructors/mycielskian.c
```diff
@@ -0,0 +1,148 @@
...
+        if (k <= 1) { // 0-> return null graph, 1-> return a single vertex
+            IGRAPH_CHECK(igraph_empty(res, k, IGRAPH_UNDIRECTED));
+            return IGRAPH_SUCCESS;
+        }
+        igraph_t g;
+        // create a path 0---1
+        IGRAPH_CHECK(igraph_ring(&g, 2, IGRAPH_UNDIRECTED, 0, 0));
+        IGRAPH_FINALLY(igraph_destroy, &g);
+
+        igraph_mycielskian(res, &g, k - 2);
```
szhorvat, line 48: Can you please write the two special cases without recursion? I know that this is a two-step recursion at most, so it won't cause problems, but it is not technically necessary, nor much simpler. I have my tools set up to warn about recursion, as unbounded recursion can (and will) cause the finally stack to overflow. Thus I'd rather not use easily avoidable recursion, to avoid future confusion.

You can do the following:

 - create the `edges` vector earlier
 - decrement `k` as necessary
 - increase `vcount` and `ecount` as necessary
 - append to the edges vector as needed, you can use `igraph_vector_int_push_back()`

Let's have two code blocks, the first one handles `vcount == 0`, the second `vcount == 1` (which may have been created by the earlier block).

Note that if we did stick to the recursion, the `IGRAPH_CHECK` would still be necessary on `igraph_mycielskian()`.

src/constructors/mycielskian.c
```diff
@@ -0,0 +1,148 @@
...
+    if (k == 1) {
+        IGRAPH_CHECK(igraph_empty(graph, 1, IGRAPH_UNDIRECTED));
+        return IGRAPH_SUCCESS;
+    }
+
+    // create a path 0---1
+    IGRAPH_CHECK(igraph_ring(&g, 2, IGRAPH_UNDIRECTED, 0, 0));
+    IGRAPH_FINALLY(igraph_destroy, &g);
+
+    igraph_mycielskian(graph, &g, k - 2);
```
szhorvat, line 143: missing IGRAPH_CHECK

tests/unit/mycielski_graph.c
```diff
@@ -0,0 +1,78 @@
...
+    igraph_mycielski_graph(&res, 3);
+    igraph_small(
+        &expected_res, 5, /* directed = */ 0,
+        /* edge 0 */ 0, 1,
+        /* edge 1 */ 1, 2,
+        /* edge 2 */ 2, 3,
+        /* edge 3 */ 3, 4,
+        /* edge 4 */ 4, 0,
+        -1
+    );
```
szhorvat, line 61: You can create this with `igraph_ring()`.

src/constructors/mycielskian.c
```diff
@@ -0,0 +1,148 @@
...
+
+    return IGRAPH_SUCCESS;
+}
+
+igraph_error_t igraph_mycielski_graph(igraph_t *graph, igraph_integer_t k) {
+    igraph_t g;
+
+    if (k <= 0) {
+        IGRAPH_ERROR("The Mycielski graph order must be a positive integer.", IGRAPH_EINVAL);
+    }
```
szhorvat, line 133: Can we please generalize this function to allow `k=0` and return a null graph?

src/constructors/mycielskian.c
```diff
@@ -0,0 +1,148 @@
...
+
+        // Add edges connecting each `ui` to `w` (forming a star)
+        for (igraph_integer_t j = prev_vcount; j < w; j++) {
+            VECTOR(edges)[edge_index++] = j;
+            VECTOR(edges)[edge_index++] = w;
+        }
+
+        // Update offset for next step
+        offset = offset * 2 + 1;
+    }
```
szhorvat, line 118:
```suggestion

        IGRAPH_ALLOW_INTERRUPTION();
    }
```

src/constructors/mycielskian.c
```diff
@@ -0,0 +1,148 @@
...
+
+#include "igraph_constructors.h"
+#include "igraph_conversion.h"
+
+#include "igraph_interface.h"
+
+#include "core/interruption.h"
+#include "math/safe_intop.h"
+
+igraph_error_t igraph_mycielskian(igraph_t *res, const igraph_t *graph, igraph_integer_t k) {
```
szhorvat, line 27: Please change the order of the `graph` and `res` parameters to be consistent with other igraph functions.

szhorvat (changes requested):
This is looking good, thank you for implementing it!

Let's also do the documentation.

src/constructors/mycielskian.c
```diff
@@ -0,0 +1,148 @@
...
+        IGRAPH_SAFE_ADD(new_ecount, new_vcount, &new_ecount); // new edges = 3 * old edges + old vertices
+
+        IGRAPH_SAFE_MULT(new_vcount, 2, &new_vcount);
+        IGRAPH_SAFE_ADD(new_vcount, 1, &new_vcount); // new vertices = 2 * old vertices + 1
+    }
+
+    if (igraph_is_directed(graph))
+        IGRAPH_CHECK(igraph_empty(res, new_vcount, IGRAPH_DIRECTED));
+    else
+        IGRAPH_CHECK(igraph_empty(res, new_vcount, IGRAPH_UNDIRECTED));
```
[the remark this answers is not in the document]
gulshan-123: used igraph_create at the end

gulshan-123:
Hi @szhorvat , I have taken care of all the feedback. In particular,
I have now changed the recursion to iterative implementation. Also, changed the parameter order in accordance with igraph.

src/constructors/mycielskian.c
```diff
@@ -0,0 +1,253 @@
...
+        // Update offset for next step
+        offset = offset * 2 + 1;
+
+        IGRAPH_ALLOW_INTERRUPTION();
+    }
+
+    // Add all edges in one go
+    IGRAPH_CHECK(igraph_create(res, &edges, 0, igraph_is_directed(graph)));
+    IGRAPH_FINALLY_CLEAN(1);
+    igraph_vector_int_destroy(&edges);
```
szhorvat, line 189: For clarity, `IGRAPH_FINALLY_CLEAN()` should always come after the destructors. I'll fix this.

src/constructors/mycielskian.c
```diff
@@ -0,0 +1,253 @@
...
+        }
+
+        // Update offset for next step
+        offset = offset * 2 + 1;
+
+        IGRAPH_ALLOW_INTERRUPTION();
+    }
+
+    // Add all edges in one go
+    IGRAPH_CHECK(igraph_create(res, &edges, 0, igraph_is_directed(graph)));
```
szhorvat, line 187: Pass the actual vertex count here, otherwise the result won't be correct when isolated vertices were present. I'll fix this.

gulshan-123:
Ok, I will fix the documentation wherever I have used latex.

gulshan-123:
I have changed the documentation, used plain text instead of latex.
I have successfully build the html documentation locally, but it is failing here, i don't know why.

szhorvat:
@ntamas Can you review this please?

I'll fix up the docs before merging.

src/constructors/mycielskian.c
```diff
@@ -0,0 +1,245 @@
...
+ * \param graph Pointer to the input graph.
+ * \param res Pointer to an uninitialized graph object where the Mycielskian
+ *        of the input graph will be stored.
+ * \param k Integer, the number of Mycielskian iterations (must be non-negative).
+ * \return Error code.
+ *
+ * \sa \ref igraph_mycielski_graph().
+ *
+ * Time complexity: Exponential in \p k.
+ */
```
szhorvat, line 90: Something important is missing: a discussion of the special case for the null graph and singleton graphs.

We also must mention the behaviour with directed graphs, non-simple graphs and disconnected graphs.

src/constructors/mycielskian.c
```diff
@@ -0,0 +1,245 @@
...
+ * which is \c M_4, is a triangle-free graph with chromatic number 4.
+ * 
+ * \param graph Pointer to an uninitialized graph object. The generated
+ *        Mycielski graph will be stored here.
+ * \param k Integer, the order of the Mycielski graph (must be non-negative).
+ * \return Error code.
+ *
+ * \sa \ref igraph_mycielskian().
+ *
+ * Time complexity: Exponential in \p k.
```
szhorvat, line 220: $O(3^k)$.

gulshan-123:
Thank you, @szhorvat , for your guidance. I have learned a lot about **igraph** and open-source contributions while working on this PR.  

I am now preparing my **GSoC proposal** for the *Spatial Analysis for igraph* project. I am highly interested in contributing to **igraph** before, during, and after GSoC.

szhorvat:
@gulshan-123 I'm pretty overwhelmed at the moment while trying to catch up on all igraph tasks. Would you be able to help me by resolving conflicts and updating this PR for the latest `master`?  Otherwise I'll get to it sometime next week.

gulshan-123:
@szhorvat, I have resolved the conflicts and merged the latest master into this PR

src/constructors/mycielskian.c
```diff
@@ -0,0 +1,245 @@
...
+
+        // new vertices = 2 * old vertices + 1
+        IGRAPH_SAFE_MULT(new_vcount, 2, &new_vcount);
+        IGRAPH_SAFE_ADD(new_vcount, 1, &new_vcount);
+    }
+
+    IGRAPH_CHECK(igraph_vector_int_resize(&edges, new_ecount * 2));
+
+    igraph_integer_t edge_index = 2 * ecount;  // Current last edge index in edge vector
+    igraph_integer_t offset = vcount;          // Tracks where new vertices start
```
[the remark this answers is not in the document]
szhorvat: Invalid suggestion.

szhorvat:
@ntamas Can you please review this?

tests/unit/mycielskian.c
```diff
@@ -0,0 +1,128 @@
...
+        /* edge 3 */ 2, 4,
+        /* edge 4 */ 2, 3,
+        /* edge 5 */ 3, 4,
+        -1
+    );
+    igraph_mycielskian(&g, &res, 0);
+    print_graph(&res);
+    igraph_destroy(&res);
+    igraph_destroy(&g);
+    printf("\n\n");
```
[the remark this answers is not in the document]
szhorvat: tested through the printed output

tests/unit/mycielskian.out
```diff
@@ -0,0 +1,111 @@
...
+null graph, k=3
+directed: false
+vcount: 5
+edges: {
+1 0
+3 0
+2 1
+4 2
+4 3
+}
```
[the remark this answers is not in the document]
szhorvat: The result is correct.

szhorvat:
@greptileai

PR #2755, opened 2025-04-22
docs: add a glossary

### LINKED ISSUE #2754: Add a glossary to the documentation
Let's add a glossary to the documentation. It should contain terms that are used repeatedly in the documentation, but their meaning may not be obvious to all users, or they are igraph-specific and not standard in graph theory or network science.

Let's write this glossary here on GitHub first. The team can edit my posts, so can contribute. Then we can convert the markdown to docbook with pandoc and add it to the docs as the last chapter.

@ntamas, let me know what you think.

szhorvat:
Closes #2754


ntamas:
DocBook has specific tags for glossaries so if we want to be super pedantic we should use those instead: https://tdg.docbook.org/tdg/5.0/glossary

szhorvat:
Good point. Let's agree on the text first, then I can refactor this into proper glossary format.

szhorvat:
This is what the glossary would look like in HTML. It might require a bit of extra CSS.

image

I'm not sure if this is worth the trouble. It looks very formal, and perhaps a bit too much for such a short list. Do we need sorting by letters and cross-referencing? Let me know what you think.

schochastics:
What is the scope of the glossary? For example what about `clique` or `subgraph`?

szhorvat:
That's up for debate. Loosely, I aimed to include terms such as the following:

 - Non-standard terms that someone unfamiliar with igraph would not know. Example: "game".
 - Technical language specific to C/igraph, e.g. "finally stack".
 - Frequently used terms, even if they are standard. Someone coming from the network analysis side may not immediately understand the difference between "adjacent" and "incident", even if these are fairly standard in graph theory.
 - Terms that are specific to software (not graph theory or network science), and a person coming form the theory side may not immediately understand. Example: "attribute". I've seen other terms for the same concept, such as "property" or "annotation", so it's good to establish that in igraph we use "attribute".
 - Sometimes there are multiple terms in the graph theory literature for the same concept, or the same term is used differently by different authors. The glossary helps us standardize over a consistent subset. For example, it is intentional that I included "self-loop", "multi-edge" and "simple graph", and omitted "pseudograph" and even "multigraph". The latter
 - There are often-used concepts (not even terms) that are defined differently by different authors. The reason why "connected graph" is included is to make it clear that igraph does not consider the null graph connected, and is consistent about this choice in all cases where considerations of connectedness come up (e.g. the null graph is also not a tree, but it is a forest).

Some other examples: 

 - "complete graph" included only to create an opportunity for mentioning "full graph", which is a non-standard term used by igraph. Personally, I don't like "full graph" very much.
 - "bipartite graph" is included because R/igraph uses this term differently, thus there's a chance that users may assume that C/igraph or other interfaces also do so (they don't).
 - "biadjacency matrix" both because it's not an entirely standard term, and because old versions used "incidence matrix" for the same concept. "incidence matrix" is included for the same reason: going forward it will apply to the edge-vertex incidence matrix only.
 - "vertex" and "edge" are included for two reasons: (1) igraph prefers "vertex" to "node" (2) opportunity to define "vertex ID" and "edge ID" as well.

I know that this is not exactly a real glossary, but I hope that:

 - It will be practically useful to readers
 - It will help AI tools understand igraph better and hallucinate less incorrect information.

ntamas:
Looks good to me, but if we have a Markdown version of the glossary anyway, I wonder whether it would make sense to implement a Python script that generates the DocBook glossary from the Markdown version. I can implement a script like that if neeeded; I just worry that the two glossaries will eventually go out of sync if we duplicate them like this.

(And if we manage the DocBook glossary with a Python script, then we can probably generate proper DocBook glossary tags without having to rewrite the current version manually).

schochastics (approved):
LGTM

ntamas (approved):

szhorvat:
Should we start calling the stack of temporary objects the **finalizer stack**? "finally stack" doesn't sound good.

szhorvat:
> Looks good to me, but if we have a Markdown version of the glossary anyway, I wonder whether it would make sense to implement a Python script that generates the DocBook glossary from the Markdown version. I can implement a script like that if neeeded; I just worry that the two glossaries will eventually go out of sync if we duplicate them like this.
> 
> (And if we manage the DocBook glossary with a Python script, then we can probably generate proper DocBook glossary tags without having to rewrite the current version manually).

This of course sounds like a good idea, but I don't have the spare capacity to write that script ...

Do we merge as-is or do we wait until someone has time for the script?  If we merge as-is, do we merge the Markdown version, or discard that and edit only the XML from now on?

Let me know so I can go ahead. Also let me know what name you prefer for the stack with the destructors:

 - "finally" stack
 - finally stack
 - finalizer stack

ntamas:
Finalizer stack is good for me.

I won't have time for the Markdown-to-DocBook conversion script, but the Markdown version is easier to read and edit so I would keep it. Just add an HTML comment at the top of the Markdown file to remind the user that there is a DocBook version and that is the authoritative one, so if the Markdown version is edited then the edits should be followed up in the DocBook XML source as well.

PR #2761, opened 2025-05-01
Multigraph support for the G(n,p) generator

### LINKED ISSUE #2757: Multigraph support for G(n,p) model

Multigraph support for G(n,p) model.

$p$ is reinterpreted as not the probability of connections between two vertices, but as the expected number of connections.  This is a direct equivalent of the class simple-graph G(n,p) within the MaxEnt framework.

This should be implemented both for the general and the bipartite cases.

Extensions of our Chung-Lu implementation should also be investigated, or Chung-Lu kept as "experimental" until this can be sorted out.

Since it's  API-breaking, it must go into 1.0 (or at least the API must be prepared for a later implementation).

**Use cases for the feature**

Fundamental multigraph model. As far as I am aware, this model is not implemented in any widely used package.

**References**

 - See the "bosonic" model in section D of Park & Newman: Statistical mechanics of networks (2004).

szhorvat:
Closes #2757

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -72,6 +72,10 @@ This section gives detailed on breaking changes you need to consider when updati
 
  - `igraph_barabasi_game()`, `igraph_barabasi_aging_game()`, `igraph_recent_degree_game()` and `igraph_recent_degree_aging_game()` no longer interprets an empty `outseq` vector as a missing out-degree sequence. Pass `NULL` if you don't wish to specify an out-degree sequence.
  - `igraph_degree_sequence_game()` no longer interprets an empty in-degree vector as a request for generating undirected graphs. To generate undirected graphs, pass `NULL` for in-degrees.
+ - `igraph_erdos_renyi_game_gnm()` uses an `igraph_edge_type_sw_t allowed_edge_types` parameter instead of `loops` and `multiple`, and can uniformly sample multigraphs. It also gained an `edge_labeled` Boolean parameter which controls whether to sample from the set of ordered edge lists (equivalent to `igraph_iea_game()` for multigraphs). 
+ - `igraph_erdos_renyi_game_gnp()` uses an `igraph_edge_type_sw_t allowed_edge_types` parameter instead of `loops` and `multiple`, and can now sample multigraphs from a maximum entropy model with a prescribed _expected_ edge multiplicity. It also gained an `edge_labeled` Boolean parameter which controls whether to sample from the set of ordered edge lists.
+ - `igraph_bipartite_game_gnm()` gained an `igraph_edge_type_sw_t allowed_edge_types` parameter, and can uniformly sample multigraphs. It also gained an `edge_labeled` Boolean parameter which controls whether to sample from the set of ordered edge lists (equivalent to `igraph_bipartite_iea_game()` for multigraphs). 
+ - `igraph_bipartite_game_gnp()` gained an `igraph_edge_type_sw_t allowed_edge_types` parameter, and can now sample multigraphs from a maximum entropy model with a prescribed _expected_ edge multiplicity. It also gained an `edge_labeled` Boolean parameter which controls whether to sample from the set of ordered edge lists.
  - `igraph_lcf()` was renamed to `igraph_lcf_small()` and `igraph_lcf_vector()` was renamed to `igraph_lcf()`. Now `igraph_lcf()` takes shifts as a vector input, while `igraph_lcf_small()` accepts a shorthand notation where shifts are given as a variable number of function arguments.
  - `igraph_sbm_game()` uses an `igraph_edge_type_sw_t allowed_edge_types` parameter instead of `igraph_bool_t loops` and supports generating graphs with multi-edges. The parameter determining the total number of vertices (`n`) was removed as it was redundant.
  - `igraph_rewire_edges()` uses an `igraph_edge_type_sw_t allowed_edge_types` parameter instead of `loops` and `multiple`.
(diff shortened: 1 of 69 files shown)


CHANGELOG.md
```diff
@@ -49,7 +49,9 @@
 
  - `igraph_int_t` may now be used as an alias to `igraph_integer_t`.
  - `igraph_erdos_renyi_game_gnm()` gained a `multiple` Boolean argument to uniformly sample G(n,m) graphs with multi-edges.
+ - `igraph_erdos_renyi_game_gnp()` gained a `multiple` Boolean argument, and can now sample multigraphs from a maxium entropy model with a presrcibed _expected_ total edge count.
  - `igraph_bipartite_game_gnm()` gained a `multiple` Boolean argument to uniformly sample bipartite G(n,m) graphs with multi-edges.
+ - `igraph_bipartite_game_gnp()` gained a `multiple` Boolean argument, and can now sample bipartite multigraphs from a maxium entropy model with a prescribed _expected_ total edge count.
```
ntamas, line 54:
```suggestion
 - `igraph_bipartite_game_gnp()` gained a `multiple` Boolean argument, and can now sample bipartite multigraphs from a maximum entropy model with a prescribed _expected_ total edge count.
```

examples/simple/igraph_decompose.c
```diff
@@ -25,7 +25,7 @@ int main(void) {
     igraph_graph_list_clear(&complist);
 
     /* Random graph with a giant component */
-    igraph_erdos_renyi_game_gnp(&g, 100, 4.0 / 100, IGRAPH_UNDIRECTED, 0);
+    igraph_erdos_renyi_game_gnp(&g, 100, 4.0 / 100, IGRAPH_UNDIRECTED, IGRAPH_NO_LOOPS , IGRAPH_NO_MULTIPLE);
```
ntamas, line 28:
```suggestion
    igraph_erdos_renyi_game_gnp(&g, 100, 4.0 / 100, IGRAPH_UNDIRECTED, IGRAPH_NO_LOOPS, IGRAPH_NO_MULTIPLE);
```

src/misc/bipartite.c
```diff
@@ -1109,6 +1115,7 @@ igraph_error_t igraph_is_bipartite(const igraph_t *graph,
  *     \c IGRAPH_ALL, then each edge direction is considered
  *     independently and mutual edges might be generated. This
  *     argument is ignored for undirected graphs.
+ * \param Whether to generate multi-edges.
```
ntamas, line 1118:
```suggestion
 * \param multiple Whether to generate multi-edges.
```

szhorvat:
@ntamas Here's what I'm planning to do here. Let me know if this is fine, so we can be efficient and avoid redoing work.

 - Add an `edge_labeled` parameter to both G(n,m), G(n,p) and their bipartite versions. This means that the model is defined over the set of edge-labelled graphs, i.e. ordered edge lists. The classic models are over unordered edge lists.
 - Remove `igraph_iea_game()` and `igraph_bipartite_iea_game()`. These are the same as G(n,m) with `multi=true` and `edge_labeled=true`.
 - The only case for which I did not fully work out the sampler yet is G(n,p), `multi=false`, `edge_labeled=true`. I can do this, but it requires more work than I have time for before 1.0. So for the time being this will through `IGRAPH_UNIMPLEMENTED`.

I have a document explaining all these model variations and working out their statistics. I am hoping to turn this into a publication, but in the worst case it can always go on arXiv. This'll explain what people want to know, and more.

----

This way igraph gains some pretty advanced features which are not implemented anywhere else, including the multigraph G(n,p) as well as considering whether edges are distinguishable.

ntamas:
I think it would make more sense to have the "edge labelled" variants as separate functions. All these functions with an assortment of boolean parameters start to seem like the cockpit of a Boeing, and to be honest the edge labelled variants seem like a niche use-case to me; adding these arguments to the base functions would just confuse users who are not familiar with edge-labelled graphs. (Such an argument would make even me stand confused for a while even though I've been working with graphs for quite a while now).

szhorvat:
What I forgot to write was that I also intend to replace `loops` / `multiple` with a single `allowed_edge_types` parameter, which controls parameter proliferation.

Note that if we add new functions for this, that is no less than 4 new functions: G(n,m), G(n,p), both bipartite and non-bipartite.

ntamas:
I don't mind the 4 extra functions - what I would mind is if users who are not familiar with edge-labelled graphs start reading the documentation of `igraph_erdos_renyi_game_gnp` and get tripped up by one of the boolean arguments, not really knowing what to put there in the absence of deeper background knowledge.

szhorvat:
@ntamas 

I was at first going to agree with you, but the more I think about it, the better it seems to just add that parameter. Some versions of the edge labelled model is of relatively little use, so it's cleaner to control this using a parameter.

This parameter's documentation will have a clear statement that it should be set to false to get the classic ER model, and a clear explanation of what setting to true does (pick edges independently for G(n,m), more generally: sample from the set of unordered vs ordered edge lists).

It will have a default value in the interface definition and it will be very easy to ignore in high-level interfaces.

How much do you object to adding this parameter? I think it would make for a much nicer presentation, and the parameter count would not increase after consolidating loops/multi into an allowed_edge_types parameter.

ntamas:
Okay, go ahead and add the new parameter :) I don't have a strong opinion here.

ntamas:
@szhorvat I would like to go through the release process of igraph 1.0.0-rc1 during the weekend, _before_ September 1. Is there a chance to merge this PR before that date?

szhorvat:
I put a lot of work into this and would like to finish it before 1.0. I expect to finish _during_, not _before_ the weekend. I hope for it this week, but other things came up, including the eigenvector centrality bugfix, which I think we must undo before release ... And the GSoC deadline are also Sep 1, which students need some support with.

Can you please do the RC next week, not during the weekend? We're so close to the finish line now, let's do it right!

If you have time for video chat before the RC that'd be useful and help me complete stuff faster.

ntamas:
I will release the RC at the end of the weekend without this PR but will allow this PR to be merged after the first RC, before the final release. I hope this is okay for you. (I need to draw a line somewhere, otherwise there will always be "just one more PR" to merge).

In the worst case, we merge the API changes and just return `IGRAPH_UNIMPLEMENTED` if the user passed in `multiple=true`, and the final implementation can come in 1.1. We only need to get the API changes right.

szhorvat:
@ntamas I think this is about as far as I have time to take this PR if we want an 1.0 release this week. Documentation could certainly use improvement, the code could be refactored to be a bit nicer, and a few more tests would be nice. But I think it can be merged as-is for now.

Things to note:

 - Bipartite generators still use an `allowed_edge_types` parameter, and not just a Boolean `multiple`. Note that bipartite graphs cannot have self-loops, so a single Boolean `multiple` would suffice. But doing it this way is more consistent. `realize_bipartite_degree_sequence()` also does this.
 - Now that the `edge_labeled` parameter was added, we might as well drop the `iea_game()` functions as they are redundant. I opted not to drop them for now. They are still experimental.

szhorvat:
I need to go out for a while, but if tests pass, a squash-and-merge is okay on this.

PR #2768, opened 2025-05-10
chore: allow coverage to work with lcov 2.x and macOS

szhorvat:


szhorvat:
@ntamas I made some changes to allow the `coverage` target to work on more systems, but it is still very fragile. Currently it works only with the Ninja generator, not the Unix Makefiles generator.

This made it possible to build the `coverage` target for Codecov, as we did originally (instead of just the `check` target). This bumps back the coverage from ~75% to 83.8%, but I can't really tell which of these two numbers is the accurate one. Some lines that were reported as "partial" are now reported as fully covered. It never made much sense to me why these were only partially covered. You can have a look at the Covecov report to see the changes.

If you think this is better, please merge.

It'd be good if you could test it locally as well (with Ninja).

szhorvat:
It seems to me that the new value (83.8%) is the valid one and this should be merged.

PR #2746, opened 2025-04-07
Implement giant component and percolation

Zepeacedust:
- [x] By submitting this pull request, I assign the copyright of my contribution to _The igraph development team_.

Closes #2736

PR #2778, opened 2025-06-24
Percolation

### LINKED ISSUE #2736: Wishlist: Giant component and percolation

Port `IGPercolationCurve[]` from IGraph/M to the C core, with potential generalizations.

See http://szhorvat.net/mathematica/IGDocumentation/#igpercolationcurve

This can be implemented using a union-find data structure.

Can be extended to support both bond percolation (edge removal) and site percolation (vertex removal). These should be two different functions.

**Use cases for the feature**

Percolation is among the most studies processes on networks. This functionality is useful both for teaching and research.

**References**

–

Zepeacedust:

Closes #2736

Newer version of #2746 on correct branch.

----

Note: If you can't check off items in this list, let me know and I'll copy it into the top post, which you can edit.

 - [x] Can you please finish the documentation of the two functions? Be concise but also thorough. Try to explain what these functions do to a naive user who does not even know the word "percolation".  I suggest phrasing things in terms of *removing* edges/vertices, even though they are given in reverse order of removal.  You'll see whether the explanation seems clear once you've written it up.  Writing good documentation is not easy, and this will be a good exercise :-)
 - [x] Add `\experimental` at the top of the documentation text (see other examples in the codebase for what this should look like).
 - [x] Please add a test for the null graph, and another test for the singleton graph (singleton with or without edges).
 - [x] Please print the whole sequence for the karate club network, when edges/vertices are added in storage order (i.e. the order sequence is 0, 1, 2, ...). The expected results for bond and edge percolation are below.
 - [x] Please generalize the tests a little bit: There should be a function that runs the percolation computation on a given graph with a given vertex/edge order, then does basic sanity checks on the result: Is the output length correct? Are all values positive (0 is not allowed, but 1 is possible for bond percolation when the first edge is a self-loop)? Is the sequence non-decreasing? Is the last/largest value the size of the largest component?  You can use a flag to control whether the result is printed, and it should be printed for all non-random orders.
 - [x] Add `igraph_rng_seed(igraph_rng_default(), 1234)` (or your favourite seed) to the beginning of the test.
 - [x] Rename the test from `igraph_percolation.c` to just `percolation.c` as the test name is not the same as a function name in this case.
 - [x] Be sure to mark all conversations that require no followup as "resolved".

```
g = IGFamousGraph["Zachary"]
```

Expected output for karate club with bond percolation:

```mathematica
In[32]:= Reverse@Table[
  VertexCount@
   IGGiantComponent@EdgeDelete[g, Take[Reverse@EdgeList[g], k - 1]],
  {k, 1, EdgeCount[g]}
  ]

Out[32]= {2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 17, \
17, 17, 17, 17, 17, 17, 18, 18, 18, 19, 20, 21, 22, 22, 22, 22, 22, \
22, 22, 22, 22, 22, 23, 23, 23, 23, 24, 24, 24, 25, 25, 26, 26, 27, \
27, 27, 28, 28, 29, 29, 29, 31, 31, 31, 32, 33, 33, 33, 33, 34, 34, \
34, 34, 34, 34, 34, 34, 34, 34, 34, 34}
```

Expected output for karate club with edge percolation:

```mathematica
In[33]:= Reverse@Table[
  VertexCount@
   IGGiantComponent@VertexDelete[g, Take[Reverse@VertexList[g], k - 1]],
  {k, 1, VertexCount[g]}
  ]

Out[33]= {1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 14, 14, 15, \
16, 16, 17, 17, 18, 18, 18, 18, 18, 18, 22, 23, 25, 26, 27, 33, 34}
```


tests/unit/igraph_bond_percolation.c
```diff
@@ -0,0 +1,109 @@
...
+#include "igraph_constructors.h"
+#include "igraph_error.h"
+#include "igraph_games.h"
+#include "igraph_interface.h"
+#include "igraph_sparsemat.h"
+#include "igraph_vector.h"
+#include "test_utilities.h"
+#include <igraph.h>
+#include <string.h>
+#include "igraph_components.h"
```
szhorvat: Please remove all igraph headers except `<igraph.h>` and make sure that `"test_utilities.h"` comes afterwards, then any system headers (such as `string.h`) come last.

tests/unit/igraph_bond_percolation.c
```diff
@@ -0,0 +1,109 @@
...
+  
+  igraph_vector_int_destroy(&edge_ids);
+  igraph_destroy(&k_3);
+  igraph_destroy(&c_4);
+
+  igraph_famous(&karate, "Zachary");
+  // TODO: verify that it counts the vertices correctly.
+  printf("Karate graph biggest component size, no edge list given: ");
+  largest_component(&karate, &size);
+  printf("%li\n",size);
```
szhorvat:
```suggestion
  printf("%" IGRAPH_PRId "\n",size);
```

This is the reason for one of the CI failures.

include/igraph_epidemics.h
```diff
@@ -62,7 +62,6 @@ IGRAPH_EXPORT void igraph_sir_destroy(igraph_sir_t *sir);
 IGRAPH_EXPORT igraph_error_t igraph_sir(const igraph_t *graph, igraph_real_t beta,
                              igraph_real_t gamma, igraph_integer_t no_sim,
                              igraph_vector_ptr_t *result);
-
```
szhorvat: Please remove this change, so that only those files are affected by the PR in which you made relevant changes. This avoids conflicts.

src/CMakeLists.txt
```diff
@@ -308,6 +308,7 @@ add_library(
   misc/scan.c
   misc/sir.c
   misc/spanning_trees.c
+  misc/percolation.c
```
szhorvat: Can you move this up to keep alphabetic order?

tests/unit/igraph_site_percolation.c
```diff
@@ -0,0 +1,125 @@
...
+#include "igraph_constructors.h"
+#include "igraph_error.h"
+#include "igraph_games.h"
+#include "igraph_interface.h"
+#include "igraph_sparsemat.h"
+#include "igraph_vector.h"
+#include "test_utilities.h"
+#include <igraph.h>
+#include <string.h>
+#include "igraph_components.h"
```
szhorvat: Same comment about headers are before.

tests/unit/igraph_site_percolation.c
```diff
@@ -0,0 +1,125 @@
+/*
+/*
```
szhorvat: Make sure there's no double comment opening. This causes failures with some compilers.

```suggestion
/*
```

src/misc/percolation.c
```diff
@@ -0,0 +1,272 @@
...
+ * \param b The other vertex incident to the edge.
+ */
+
+static igraph_error_t igraph_i_percolate_edge(igraph_vector_int_t *links,
+					      igraph_vector_int_t *sizes,
+					      igraph_integer_t* biggest,
+					      igraph_integer_t a,
+					      igraph_integer_t b) {
+    // find head of each tree
+    // TODO: Path compression
```
szhorvat: Do you plan to work on this?

src/misc/percolation.c
```diff
@@ -0,0 +1,272 @@
...
+ * \param links Vector representing parents.
+ * \param sizes sizes[i] is the number of children of links[i]
+ * \param biggest The biggest value in sizes, is updated if a bigger cluster is created.
+ * \param a A vertex incident to the edge.
+ * \param b The other vertex incident to the edge.
+ */
+
+static igraph_error_t igraph_i_percolate_edge(igraph_vector_int_t *links,
+					      igraph_vector_int_t *sizes,
+					      igraph_integer_t* biggest,
```
szhorvat:
```suggestion
					      igraph_integer_t *biggest,
```

Consistency nitpick.

src/misc/percolation.c
```diff
@@ -0,0 +1,272 @@
...
+    igraph_vector_int_t links;
+    IGRAPH_VECTOR_INT_INIT_FINALLY(&links, upper+1);
+
+    for (igraph_integer_t i = 0; i < upper+1; i++) {
+        VECTOR(sizes)[i] = 1;
+        VECTOR(links)[i] = i;
+    }
+
+    igraph_integer_t edge_count = igraph_vector_int_size(edges);
+    if (edge_count % 2 == 1) IGRAPH_ERROR("Invalid edge list, odd number of elements", IGRAPH_EINVAL);
```
szhorvat:
```suggestion
    if (edge_count % 2 == 1) {
        IGRAPH_ERROR("Invalid edge list, odd number of elements", IGRAPH_EINVAL);
    }
```

Style consistency and readability. Can you please fix any other instances of this? When the `if (...)` is followed by something very brief, it's okay to go single line, but in these cases let's keep it multi-line with brackets.

We can also just reformat with `astyle` in a single go.

src/misc/percolation.c
```diff
@@ -0,0 +1,272 @@
...
+    edge_count >>= 1;
+    IGRAPH_CHECK(igraph_vector_int_resize(output, edge_count));
+
+    for (igraph_integer_t i = 0; i < edge_count; i++) {
+        igraph_i_percolate_edge(&links, &sizes, &biggest, VECTOR(*edges)[2*i], VECTOR(*edges)[2*i+1]);
```
szhorvat: If a function returns an error code, it must always be wrapped in `IGRAPH_CHECK`. In some special cases, we avoid this for performance reasons but then add a clear comment about why it's okay to omit the `IGRAPH_CHECK`.

Unless I'm missing something, it's not actually necessary for `igraph_i_percolate_edge` to return an error code (can it ever fail?) Perhaps it's better to just make it `void` then.

src/misc/percolation.c
```diff
@@ -0,0 +1,272 @@
...
+ * \function igraph_i_percolate_edge
+ * \brief Percolates a single edge.
+ * \param links Vector representing parents.
+ * \param sizes sizes[i] is the number of children of links[i]
+ * \param biggest The biggest value in sizes, is updated if a bigger cluster is created.
+ * \param a A vertex incident to the edge.
+ * \param b The other vertex incident to the edge.
+ */
+
+static igraph_error_t igraph_i_percolate_edge(igraph_vector_int_t *links,
```
szhorvat: Please remove `igraph_i_` from the names of functions that are `static`, i.e. local to this translation unit. It is true that there may be some leftovers of this naming convention, but what we currently use is:

 - `igraph_` for public functions.
 -  `igraph_i_` for internal functions that are used in multiple translation units (and therefore appear in some internal header)
 - no naming prefix for local functions (i.e. those that are `static`)

szhorvat: I hope that Emacs makes this refactoring easy. You should use an IDE that supports automatic refactoring.

src/misc/percolation.c
```diff
@@ -0,0 +1,272 @@
...
+						     igraph_vector_int_t * edge_indices) {
+  if (edge_indices == NULL) {
+    // generate random vertex list
+    igraph_integer_t size = igraph_ecount(graph);
+    igraph_vector_int_t new_edges;
+    IGRAPH_CHECK(igraph_vector_int_init_range(&new_edges, 0, size));
+
+    RNG_BEGIN();
+    IGRAPH_CHECK(igraph_vector_int_shuffle(&new_edges));
+    RNG_END();
```
szhorvat:
```suggestion
    IGRAPH_CHECK(igraph_vector_int_shuffle(&new_edges));
```

These are already present in the shuffle function. Generally, you only need them when using raw RNG functions, such as the `RNG_` macros of the functions that these macros expand to.

src/misc/percolation.c
```diff
@@ -0,0 +1,272 @@
...
+
+    IGRAPH_FINALLY(igraph_vector_int_destroy, &new_edges);
+    igraph_bond_percolation(graph, output, &new_edges);
```
szhorvat: This would need an `IGRAPH_CHECK`.

But what should be done here is to refactor to avoid recursion (even though it's a single level recursion) and use only conditionals.

Reason: We use static checkers that report any use of recursion and this creates extra noise. This single-level recursion here is harmless but true recursion can lead to stack overflow, which is why the static checker verifies this. The checker is not smart enough to notice that this one is fine.

src/misc/percolation.c
```diff
@@ -0,0 +1,272 @@
...
+  IGRAPH_FINALLY(igraph_vector_int_destroy, &edgelist);
+  
+  for (igraph_integer_t i = 0; i < igraph_vector_int_size(edge_indices); i++) {
+    IGRAPH_CHECK(igraph_edge(graph,
+			      VECTOR(*edge_indices)[i],
+			     &VECTOR(edgelist)[2*i],
+			     &VECTOR(edgelist)[2*i+1]));
+  }
+  
+  igraph_i_edge_list_percolation(&edgelist, output);
```
szhorvat: Don't forget the IGRAPH_CHECKs!

src/misc/percolation.c
```diff
@@ -0,0 +1,272 @@
...
+  igraph_vector_int_t edgelist;
+  IGRAPH_CHECK(igraph_vector_int_init(&edgelist, 2*igraph_vector_int_size(edge_indices)));
+  IGRAPH_FINALLY(igraph_vector_int_destroy, &edgelist);
+  
+  for (igraph_integer_t i = 0; i < igraph_vector_int_size(edge_indices); i++) {
+    IGRAPH_CHECK(igraph_edge(graph,
+			      VECTOR(*edge_indices)[i],
+			     &VECTOR(edgelist)[2*i],
+			     &VECTOR(edgelist)[2*i+1]));
+  }
```
szhorvat: It should be possible to use `igraph_edges()` (note the plural in edgeS) instead of this loop.

src/misc/percolation.c
```diff
@@ -0,0 +1,272 @@
...
+  igraph_vector_int_destroy(&edgelist);
+  IGRAPH_FINALLY_CLEAN(1);
+  
+  return IGRAPH_SUCCESS;
+}
+
+static igraph_error_t igraph_i_percolate_site(const igraph_t *graph,
+                                              igraph_vector_int_t *links,
+                                              igraph_vector_int_t *sizes,
+					      igraph_integer_t *biggest,
```
szhorvat: To deal with minor misformattings, can you run the whole file through astyle? Install astyle, then:

```
astyle --project=.astylerc src/misc/percolation.c 
```

I can also do this for you if needed, let me know.

src/misc/percolation.c
```diff
@@ -0,0 +1,272 @@
...
+  // if vertex list is null, generate a random one.
+  if (vertices == NULL) {
+    // generate random vertex list
+    igraph_integer_t size = igraph_vcount(graph);
+    igraph_vector_int_t new_vertices;
+    IGRAPH_CHECK(igraph_vector_int_init_range(&new_vertices, 0, size));
+
+    RNG_BEGIN();
+    IGRAPH_CHECK(igraph_vector_int_shuffle(&new_vertices));
+    RNG_END();
```
szhorvat:
```suggestion
    IGRAPH_CHECK(igraph_vector_int_shuffle(&new_vertices));
```

See the other similar comment for the reason.

src/misc/percolation.c
```diff
@@ -0,0 +1,272 @@
...
+
+    IGRAPH_FINALLY(igraph_vector_int_destroy, &new_vertices);
+    igraph_site_percolation(graph, output, &new_vertices);
```
szhorvat: Please refactor to avoid recursion (please see the analogous comment for `bond_percolation`)

src/misc/percolation.c
```diff
@@ -0,0 +1,272 @@
...
+#include "igraph_conversion.h"
+#include "igraph_error.h"
+#include "igraph_iterators.h"
+#include "igraph_structural.h"
+#include "igraph_random.h"
+#include "igraph_interface.h"
+#include "igraph_types.h"
+#include "igraph_vector.h"
+#include "igraph_vector_list.h"
+#include <stdio.h>
```
szhorvat:
```suggestion
#include "igraph_components.h"
#include "igraph_constants.h"
#include "igraph_error.h"
#include "igraph_random.h"
#include "igraph_interface.h"
#include "igraph_types.h"
#include "igraph_vector.h"
```

Remove unused headers. A good IDE can help with this.

`<stdio.h>` should not be needed, and is generally a red flag we're looking for, as it indicates the presence of `printf` and other similar functions that are forbidden by [CRAN](https://cran.r-project.org/). The core igraph library should not print to stdio directly. Instead it should use mechanisms that can redirect to other output channels (e.g. into a Mathematica notebook).

src/misc/percolation.c
```diff
@@ -0,0 +1,272 @@
...
+
+/**
+ * \function igraph_i_edge_shuffle
+ * \brief In-place Fisher-Yates shuffle for edge lists.
+ * \param edges List of edges, will be permuted.
+ */
+static igraph_error_t igraph_i_edge_shuffle(igraph_vector_int_t *edges) {
```
szhorvat: This function does not seem to be used.

szhorvat (changes requested):
Can you please move `percolation.c` from `misc` to `connectivity`?

This is not a complete review yet, only things that immediately jumped out.

szhorvat:
Let's also try a Copilot review and see if it finds anything useful. It's not necessary to fix everything that Copilot finds. Its suggestions are not always reasonable.

src/misc/percolation.c
Zepeacedust: It was used with an older implementation, and hadn't been removed yet.

src/misc/percolation.c
Zepeacedust: In an older implementation there were checks for validity, but those have been removed.
Will change to void.

src/misc/percolation.c
Zepeacedust: Yes, I implemented tests first so I could verify that it does not change functionality.

src/CMakeLists.txt
Zepeacedust: Resolved when moving the file to connectivity.

tests/unit/igraph_percolation.c
```diff
@@ -0,0 +1,198 @@
...
+    igraph_full(&k_3, 3, false, false);
+
+    igraph_small(&c_4, 4, IGRAPH_UNDIRECTED, 0, 1, 1, 2, 2, 3, 3, 0, -1);
+
+    printf("K_3 graph percolation curve, no provided edge sequence:\n");
+    IGRAPH_CHECK(percolate_b(&k_3, NULL));
+
+    igraph_vector_int_t edge_ids;
+    IGRAPH_CHECK(igraph_vector_int_init_int(&edge_ids, 4, 0, 2, 1, 3));
+    VECTOR(edge_ids);
```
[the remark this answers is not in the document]
szhorvat: @Zepeacedust What happened here?

src/connectivity/percolation.c
```diff
@@ -0,0 +1,281 @@
...
+    if (edge_indices == NULL) {
+        // Use random edge order
+        generated_edges = true;
+        igraph_integer_t size = igraph_ecount(graph);
+
+        IGRAPH_CHECK(igraph_vector_int_init_range(&new_edges, 0, size));
+
+        IGRAPH_CHECK(igraph_vector_int_shuffle(&new_edges));
+
+        IGRAPH_FINALLY(igraph_vector_int_destroy, &new_edges);
```
szhorvat:
```suggestion
        IGRAPH_CHECK(igraph_vector_int_init_range(&new_edges, 0, size));
        IGRAPH_FINALLY(igraph_vector_int_destroy, &new_edges);
        IGRAPH_CHECK(igraph_vector_int_shuffle(&new_edges));
```

There must be no possibility of an error being thrown between the `init` call and the corresponding `FINALLY` call.

Theoretically, if `igraph_vector_int_shuffle()` fails, `new_edges` won't be freed as it wasn't placed on the "finally" stack yet.

I know that in this case, `igraph_vector_int_shuffle()` can't _really_ fail. In fact, in the next major version the return type of this function will be changed to `void`: https://github.com/igraph/igraph/blob/4d325eeb39558583748b1d5d378ba5e11c02f156/CHANGELOG.md?plain=1#L13  Nevertheless, we should always keep the `init` and the `FINALLY` calls together, so that when we read the code later, it doesn't look like something's wrong.

Please fix this (you can simply commit my suggestion using the button above).

src/connectivity/percolation.c
```diff
@@ -0,0 +1,281 @@
...
+
+    if (VECTOR(*sizes)[vertex] != 0) {
+        IGRAPH_ERROR("Vertex already added.", IGRAPH_EINVAL);
+    }
+
+    VECTOR(*sizes)[vertex] = 1;
+
+    igraph_vector_int_t neighbors;
+    IGRAPH_CHECK(igraph_vector_int_init(&neighbors, 0));
+    IGRAPH_FINALLY(igraph_vector_int_destroy, &neighbors);
```
szhorvat:
```suggestion
    IGRAPH_VECTOR_INT_INIT_FINALLY(igraph_vector_int_init(&neighbors, 0));
```

Please use the convenience macro whenever possible.

src/connectivity/percolation.c
```diff
@@ -0,0 +1,281 @@
...
+    if (vertices == NULL) {
+        // Use random vertex order
+        generated_vertices = true;
+        igraph_integer_t size = igraph_vcount(graph);
+
+        IGRAPH_CHECK(igraph_vector_int_init_range(&new_vertices, 0, size));
+
+        IGRAPH_CHECK(igraph_vector_int_shuffle(&new_vertices));
+
+        IGRAPH_FINALLY(igraph_vector_int_destroy, &new_vertices);
```
szhorvat:
```suggestion
        IGRAPH_CHECK(igraph_vector_int_init_range(&new_vertices, 0, size));
        IGRAPH_FINALLY(igraph_vector_int_destroy, &new_vertices);
        IGRAPH_CHECK(igraph_vector_int_shuffle(&new_vertices));
```

See the analogous comment above.

src/connectivity/percolation.c
```diff
@@ -0,0 +1,281 @@
...
+    }
+    igraph_integer_t biggest = 1;
+
+    igraph_vector_int_t sizes;
+    IGRAPH_CHECK(igraph_vector_int_init(&sizes, size));
+    IGRAPH_FINALLY(igraph_vector_int_destroy, &sizes);
+
+    igraph_vector_int_t links;
+    IGRAPH_CHECK(igraph_vector_int_init(&links, size));
+    IGRAPH_FINALLY(igraph_vector_int_destroy, &links);
```
szhorvat: Please use the convenience init+finally macros.

tests/unit/igraph_percolation.c
```diff
@@ -0,0 +1,198 @@
...
+    IGRAPH_VECTOR_INT_INIT_FINALLY(&outputs, 0);
+
+    IGRAPH_CHECK(igraph_bond_percolation(graph, &outputs, edge_indices));
+
+    print_vector_int(&outputs);
+
+    igraph_vector_int_destroy(&outputs);
+    IGRAPH_FINALLY_CLEAN(1);
+    return IGRAPH_SUCCESS;
+}
```
szhorvat: In tests, we don't need to implement error handling.

You don't have to remove this here, but it's not really necessary.

tests/unit/igraph_percolation.c
```diff
@@ -0,0 +1,198 @@
...
+    // test with normal graph and provided edge list
+
+    igraph_t k_3, c_4, karate, random;
+    igraph_integer_t size = 0;
+    igraph_full(&k_3, 3, false, false);
+
+    igraph_small(&c_4, 4, IGRAPH_UNDIRECTED, 0, 1, 1, 2, 2, 3, 3, 0, -1);
+
+    printf("K_3 graph percolation curve, no provided edge sequence:\n");
+    IGRAPH_CHECK(percolate_b(&k_3, NULL));
```
szhorvat: `IGRAPH_CHECK` should only be used in functions that return `igraph_error_t`. Let's remove the error checks and make the function return `void`.

Note that when running tests, if an error occurs, the program aborts immediately and the test fails.

igraph supports "error handlers". The handler decides what to do upon error. The default handler prints the error and aborts. The handler used in Python will instead throw an appropriate Python exception and runs the function that calls all destructors on the "finally" stack.

Since the default behaviour is to abort, we don't need to do error handling in tests.

tests/unit/igraph_percolation.c
```diff
@@ -0,0 +1,198 @@
...
+
+    igraph_vector_int_destroy(&components);
+    IGRAPH_FINALLY_CLEAN(1);
+
+    igraph_vector_int_t bad_vert_list_repeat, bad_vert_list_too_big, bad_vert_list_missing;
+    igraph_vector_int_init_int(&bad_vert_list_too_big, 6, 0, 1, 2, 3, 4, 5);
+    igraph_vector_int_init_int(&bad_vert_list_missing, 3, 0, 1, 2);
+    igraph_vector_int_init_int(&bad_vert_list_repeat,  5, 0, 0, 0, 0, 0);
+    // should error due to being too big
+    CHECK_ERROR(percolate_s(&k_5, &bad_vert_list_too_big), IGRAPH_EINVAL);
```
szhorvat: No need to change anything here, just an explanation:

As I said above, the default behaviour is to abort on error. `CHECK_ERROR` temporarily sets an error handler that ignores errors (doesn't abort), so it can verify that the expected error code is returned. It restores the default error handler before returning.

src/connectivity/percolation.c
```diff
@@ -0,0 +1,281 @@
...
+    igraph_vector_int_t edgelist;
+    IGRAPH_CHECK(igraph_vector_int_init(&edgelist, 2 * igraph_vector_int_size(internal_edge_indices)));
+    IGRAPH_FINALLY(igraph_vector_int_destroy, &edgelist);
+
+    for (igraph_integer_t i = 0; i < igraph_vector_int_size(internal_edge_indices); i++) {
+        IGRAPH_CHECK(igraph_edge(graph,
+                                 VECTOR(*internal_edge_indices)[i],
+                                 &VECTOR(edgelist)[2 * i],
+                                 &VECTOR(edgelist)[2 * i + 1]));
+    }
```
szhorvat: As a rule, we avoid calling `size` functions directly in the `for` loop. I'm referring to `i < igraph_vector_int_size(internal_edge_indices);`. Instead, retrieve the size, save it into a variable, then use that variable.

Why?

Because functions calls are expensive. If the body of the for loop is very tight (and the body of _this_ loop could be refactored to avoid functions calls entirely) then time taken by the call to the `size` function will dominate.  Instead of benchmarking these, we take the simple approach and just not call the `size` function unnecessarily.

If this were C++, we'd be using inline functions defined in headers. But it's C, and the `size` function is defined in a separate compilation unit, so the compiler is unable to inline it.

There are ways to mitigate this performance impact, such as using [LTO](https://en.wikipedia.org/wiki/Interprocedural_optimization) or using non-standard GCC features that allow marking the `size` function as "constant" (which we do). But these solutions can't be applied in all situations (all configurations, all platforms, all compilers).

----

Now all this said, I suggest replacing the entire for loop by a call to `igraph_edges()` (as I recommended yesterday). Is there a reason I'm missing why this is not possible?

src/connectivity/percolation.c
```diff
@@ -0,0 +1,281 @@
...
+    IGRAPH_FINALLY(igraph_vector_int_destroy, &neighbors);
+
+    IGRAPH_CHECK(igraph_neighbors(graph, &neighbors, vertex, IGRAPH_IN));
+
+    for (igraph_integer_t i = 0; i < igraph_vector_int_size(&neighbors); i++) {
+        if (VECTOR(*sizes)[VECTOR(neighbors)[i]] == 0) {
+            continue;
+        }
+        percolate_edge(links, sizes, biggest, vertex, VECTOR(neighbors)[i]);
+    }
```
szhorvat: Please see the long comment I made above about why `size` functions should not be called repeatedly in `for` loops.

szhorvat:
I just pushed a commit that shows how to add functions to the documentation. The documentation comment must be present, with the appropriate `\function` tag.

Can you please do the following?

 - Draft documentation for the two functions.
 - Also include `site_percolation()` in the docs.

Instructions on building the docs: https://igraph.org/c/html/latest/igraph-Installation.html#igraph-Installation-building-the-documentation

Guidelines for writing docs: https://github.com/igraph/igraph/wiki/Documentation-guidelines

szhorvat:
Please see the very last commit I pushed on how to add the function interfaces. This is a YAML file that contains semantic information about the function parameters. It is used for auto-generating raw bindings for high-level interfaces.

There is a CI test (called "Stimulus") that checks that the interface matches the function prototype declared in C. This is indeed failing now as the parameters are not `const` as expected. See the two following comments I made below:

src/connectivity/percolation.c
```diff
@@ -0,0 +1,281 @@
...
+    igraph_vector_int_destroy(&neighbors);
+    IGRAPH_FINALLY_CLEAN(1);
+
+    return IGRAPH_SUCCESS;
+}
+
+igraph_error_t igraph_site_percolation(
+        const igraph_t *graph,
+        igraph_vector_int_t *output,
+        igraph_vector_int_t *vertices) {
```
szhorvat: Please make the last parameter be `const igraph_vector_int_t *`, as we don't modify it. You will need to make `internal_vertices` a pointer-to-const as well.

Let's rename this parameter to `vertex_order`.

src/connectivity/percolation.c
```diff
@@ -0,0 +1,281 @@
...
+ * \function igraph_bond_percolation
+ * \brief calculates the bond-percolation curve of a graph.
+ * \param graph The input graph
+ * \param output output[i] is the size of the largest component after adding i+1 edges, will be created.
+ */
+
+igraph_error_t igraph_bond_percolation(
+        const igraph_t *graph,
+        igraph_vector_int_t * output,
+        igraph_vector_int_t * edge_indices) {
```
szhorvat: Let's make this param `const` and rename it to `edge_order`. See my last comment above.

src/connectivity/percolation.c
Zepeacedust: The reason why is that it uses `igraph_es_t` instead of a vector, and I couldn't find any documentation about it.

szhorvat: You can use `igraph_ess_vector(your_edge_index_vector)` to create an `igraph_es_t` based on a vector of indices. The `ess` version returns the result directly, and is much more convenient. There's also an `es` version (single `s`) which uses an output parameter, but we don't want that inconvenience.

szhorvat: I fixed this.

src/connectivity/percolation.c
```diff
@@ -0,0 +1,290 @@
...
+
+    if (VECTOR(*sizes)[vertex] != 0) {
+        IGRAPH_ERROR("Duplicate vertex in vertex order vector.", IGRAPH_EINVAL);
+    }
+
+    VECTOR(*sizes)[vertex] = 1;
+
+    igraph_vector_int_t neighbors;
+    igraph_integer_t neighbor_count;
+    IGRAPH_VECTOR_INT_INIT_FINALLY(&neighbors, 0);
```
szhorvat: The `neighbors` vector is initialized and destroyed on every call to this function.  This is expensive (`malloc` takes significant time).

Create a `neighbors` vector once, and re-use it for each call to `igraph_neighbors()`.

You will need to do the initialization in the caller and pass the vector to this function every time, for re-use.

tests/unit/igraph_percolation.c
```diff
@@ -0,0 +1,200 @@
...
+
+    *size = VECTOR(outputs)[igraph_vector_int_size(&outputs) -1];
+
+    igraph_vector_int_destroy(&outputs);
+    IGRAPH_FINALLY_CLEAN(1);
+
+    return IGRAPH_SUCCESS;
+}
+
+int test_site(void) {
```
szhorvat: Please remove all `IGRAPH_CHECK` from `test_site` and `test_bond`, and make these functions `void`. Only those functions are allowed to call `IGRAPH_CHECK` which return `igraph_error_t`.

szhorvat: Within both `test_site` and `test_bond`, start with printing what we are testing (site or bond percolation) so that when things start to fail, the output is easy to interpret.

szhorvat (changes requested):
Note: If you can't check off items in this list, let me know and I'll copy it into the top post, which you can edit.

 - [ ] Can you please finish the documentation of the two functions? Be concise but also thorough. Try to explain what these functions do to a naive user who does not even know the word "percolation".  I suggest phrasing things in terms of *removing* edges/vertices, even though they are given in reverse order of removal.  You'll see whether the explanation seems clear once you've written it up.  Writing good documentation is not easy, and this will be a good exercise :-)
 - [ ] Add `\experimental` at the top of the documentation text (see other examples in the codebase for what this should look like).
 - [ ] Please add a test for the null graph, and another test for the singleton graph (singleton with or without edges).
 - [ ] Please print the whole sequence for the karate club network, when edges/vertices are added in storage order (i.e. the order sequence is 0, 1, 2, ...). The expected results for bond and edge percolation are below.
 - [ ] Please generalize the tests a little bit: There should be a function that runs the percolation computation on a given graph with a given vertex/edge order, then does basic sanity checks on the result: Is the output length correct? Are all values positive (0 is not allowed, but 1 is possible for bond percolation when the first edge is a self-loop)? Is the sequence non-decreasing? Is the last/largest value the size of the largest component?  You can use a flag to control whether the result is printed, and it should be printed for all non-random orders.
 - [ ] Add `igraph_rng_seed(igraph_rng_default(), 1234)` (or your favourite seed) to the beginning of the test.
 - [ ] Rename the test from `igraph_percolation.c` to just `percolation.c` as the test name is not the same as a function name in this case.
 - [ ] Be sure to mark all conversations that require no followup as "resolved".

```
g = IGFamousGraph["Zachary"]
```

Expected output for karate club with bond percolation:

```mathematica
In[32]:= Reverse@Table[
  VertexCount@
   IGGiantComponent@EdgeDelete[g, Take[Reverse@EdgeList[g], k - 1]],
  {k, 1, EdgeCount[g]}
  ]

Out[32]= {2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 17, \
17, 17, 17, 17, 17, 17, 18, 18, 18, 19, 20, 21, 22, 22, 22, 22, 22, \
22, 22, 22, 22, 22, 23, 23, 23, 23, 24, 24, 24, 25, 25, 26, 26, 27, \
27, 27, 28, 28, 29, 29, 29, 31, 31, 31, 32, 33, 33, 33, 33, 34, 34, \
34, 34, 34, 34, 34, 34, 34, 34, 34, 34}
```

Expected output for karate club with edge percolation:

```mathematica
In[33]:= Reverse@Table[
  VertexCount@
   IGGiantComponent@VertexDelete[g, Take[Reverse@VertexList[g], k - 1]],
  {k, 1, VertexCount[g]}
  ]

Out[33]= {1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 14, 14, 15, \
16, 16, 17, 17, 18, 18, 18, 18, 18, 18, 22, 23, 25, 26, 27, 33, 34}
```

src/connectivity/percolation.c
```diff
@@ -0,0 +1,290 @@
...
+}
+
+/**
+ * \function igraph_i_edge_list_percolation
+ * \brief Gives the size of the largest connected component as edges are added.
+ * \param edges Vector of edges, will be added in order.
+ * \param output output[i] is the size of the largest connected component after edges[i] is added
+ */
+
+static igraph_error_t edge_list_percolation(
```
szhorvat: Let's make this function public (including adding it to headers, adding docs, renaming, adding to interfaces).

tests/unit/igraph_percolation.c
```diff
@@ -0,0 +1,200 @@
...
+    igraph_full(&k_5, 5, false, false);
+
+    igraph_small(&c_4, 4, IGRAPH_UNDIRECTED, 0, 1, 1, 2, 2, 3, 3, 0, -1);
+
+    printf("K_5 graph percolation curve, no provided vertex sequence:\n");
+    IGRAPH_CHECK(percolate_s(&k_5, NULL));
+
+    igraph_vector_int_t edge_ids;
+    IGRAPH_CHECK(igraph_vector_int_init_int(&edge_ids, 4, 0, 2, 1, 3));
+    VECTOR(edge_ids);
```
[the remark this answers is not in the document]
szhorvat: Hm, I thought we fixed this already ...

src/connectivity/percolation.c
Zepeacedust: Since each vertex is only checked once, it would not help to cache the calls, but I can switch to `adjlist` which I think allocates the entire list at once.

tests/unit/igraph_percolation.c
Zepeacedust: Making them void would cause issues with testing for expected errors, since `IGRAPH_ASSERT` expands to contain a return statement, is there another way to test for errors or should I no longer do that?

include/igraph_components.h
```diff
@@ -66,6 +66,7 @@ IGRAPH_EXPORT igraph_error_t igraph_bridges(const igraph_t *graph, igraph_vector
 
 IGRAPH_EXPORT igraph_error_t igraph_bond_percolation(const igraph_t *graph, igraph_vector_int_t *output, const igraph_vector_int_t *edge_order);
 IGRAPH_EXPORT igraph_error_t igraph_site_percolation(const igraph_t *graph, igraph_vector_int_t *output, const igraph_vector_int_t *vertex_order);
+IGRAPH_EXPORT igraph_error_t igraph_edge_list_percolation(const igraph_vector_int_t *edge_list, igraph_vector_int_t *output);
```
szhorvat: Can you please use `edges` as the parameter name here, and make sure to include this function in:

 - the documentation
 - the interface definition file

Don't forget to make the `edges` parameter `const`, as it's passed by reference and it is an input parameter (never modified, not used for output).

tests/unit/igraph_percolation.c
szhorvat: `IGRAPH_ASSERT` does NOT return. It invokes a "fatal error" (`IGRAPH_FATAL`) which is required not to even return to the point where it was called from.

Zepeacedust: I see, I was reading error messages incorrectly. This looks like an easy fix then.

src/connectivity/percolation.c
szhorvat: What I mean is not to cache results, but to allocate a single vector and re-use it in the `igraph_neighbors()` call. Allocation is expensive and this makes a real difference.

Each time you call `igraph_neighbors()`, it will be with a different vertex, so it returns a different results. But is it faster (and not just theoretically) to write the result into the same pre-allocated vector instead of allocating a new vector each time. You don't have to worry about the allocation size, as that vector is resized naturally by `igraph_neighbors()`. Note that when resizing a vector to be smaller, no storage is freed up! The storage is still available internally for later use (when the vector becomes bigger again).

szhorvat:
I marked this as draft for now. Please hit the "Ready for review" button when you believe this is ready to merge.

doc/structural.xxml
```diff
@@ -108,6 +108,11 @@
 <!-- doxrox-include igraph_is_biconnected -->
 </section>
 
+<section id="percolation"><title>Percolation</title>
+<!-- doxrox-include igraph_bond_percolation -->
+<!-- doxrox-include igraph_site_percolation -->
+</section>
```
szhorvat: Don't forget to add the last function here!

interfaces/functions.yaml
```diff
@@ -1304,6 +1304,19 @@ igraph_is_biconnected:
...
+#######################################
+
+igraph_bond_percolation:
+    PARAMS: GRAPH graph, OUT VECTOR_INT output, OPTIONAL EDGE_INDICES edge_order
+
+igraph_site_percolation:
+    PARAMS: GRAPH graph, OUT VECTOR_INT output, OPTIONAL VERTEX_INDICES vertex_order
+
+igraph_edge_list_percolation:
+    PARAMS: IN VECTOR_INT edges, OUT VECTOR_INT output
```
szhorvat:
```suggestion
    PARAMS: VERTEX_INDEX_PAIRS edges, OUT VECTOR_INT output
```

This is one of the more esoteric parts of igraph, and you had no way of knowing this. I'll fix up the interfaces as we go, please just make this change for now.

src/CMakeLists.txt
```diff
@@ -309,6 +310,7 @@ add_library(
   misc/sir.c
   misc/spanning_trees.c
 
+
```
szhorvat: Extra line to be removed.

src/connectivity/percolation.c
```diff
@@ -0,0 +1,309 @@
...
+
+#include "igraph_components.h"
+#include "igraph_constants.h"
+#include "igraph_error.h"
+#include "igraph_interface.h"
+#include "igraph_types.h"
+#include "igraph_vector.h"
+
+/**
+ * \experimental
```
szhorvat:
```suggestion
```

We don't need detailed documentation for _internal_ functions. Use your judgement about how much explanation is needed so that the code is readable.

However, `\experimental` _should not_ be added to internal functions.

src/connectivity/percolation.c
```diff
@@ -0,0 +1,309 @@
...
+    You should have received a copy of the GNU General Public License
+    along with this program.  If not, see <https://www.gnu.org/licenses/>.
+*/
+
+#include "igraph_components.h"
+#include "igraph_constants.h"
+#include "igraph_error.h"
+#include "igraph_interface.h"
+#include "igraph_types.h"
+#include "igraph_vector.h"
```
szhorvat:
```suggestion
#include "igraph_components.h"

#include "igraph_constants.h"
#include "igraph_error.h"
#include "igraph_interface.h"
#include "igraph_types.h"
#include "igraph_vector.h"
```

Nitpick, please just commit this. The first header is separate and the one containing the functions defined in this file. This is not always followed, but I'm trying to get the codebase a bit more organized for readability.

src/connectivity/percolation.c
```diff
@@ -0,0 +1,309 @@
...
+ *        after adding <code>vertex_order[i]</code>. Will be resised.
+ * \param vertex_order The order the vertices will be added in.
+ *        Will raise error if there are duplicates, or a vertex is missing.
+ *        If null, a random order will be used.
+ * \return Error code:
+ *        \clist
+ *        \cli IGRAPH_EINVAL
+ *             The vertex list is invalid, see above.
+ *        \endclist
+ */
```
szhorvat: Points to pay attention to in the future:

 - Location of `\experimental`
 - The newlines I added below.

TODO: Can you add a note on time complexity?

```suggestion
/**
 * \function igraph_site_percolation
 * \brief The size of the largest component as vertices are added to a graph.
 *
 * \experimental
 *
 * Calculates the percolation curve, or the size of the largest connected component as
 * vertices are added in the order given. If the output is reversed, it is the size of
 * the largest component as vertices are removed in the reverse of the order given.
 * If there is no vertex order given, it will generate one.
 *
 * \param graph The graph that vertices are assumed to be in.
 * \param output <code>output[i]</code> will contain the size of the largest component
 *        after adding <code>vertex_order[i]</code>. Will be resized.
 * \param vertex_order The order the vertices will be added in.
 *        Will raise error if there are duplicates, or a vertex is missing.
 *        If \c NULL, a random order will be used.
 * \return Error code:
 *        \clist
 *        \cli IGRAPH_EINVAL
 *             The vertex list is invalid, see above.
 *        \endclist
 */
```

In the future, it's enough to write `\return Error code.`  Being more details is good, but there are countless modes of failure and there's no need to put in the effort to try to document several of them, unless the behaviour is unusual and unexpected to users. 

This is just for future reference. No change needed in the `\return` part here.

tests/unit/percolation.c
```diff
@@ -0,0 +1,289 @@
...
+    GNU General Public License for more details.
+
+    You should have received a copy of the GNU General Public License
+    along with this program.  If not, see <https://www.gnu.org/licenses/>.
+*/
+
+#include <igraph.h>
+#include "igraph_error.h"
+#include "igraph_vector.h"
+#include "test_utilities.h"
```
szhorvat:
```suggestion
#include <igraph.h>

#include "test_utilities.h"
```

As discussed, no detailed headers in tests. We do not guarantee that the arrangement of these subheaders stays the same even through minor versions. For end-users, only `<igraph.h>` should be considered stable, all the rest is for internal organization.

tests/unit/percolation.c
```diff
@@ -0,0 +1,289 @@
...
+    }
+
+    igraph_vector_int_destroy(&outputs);
+    IGRAPH_FINALLY_CLEAN(1);
+
+    return IGRAPH_SUCCESS;
+}
+
+
+int test_bond(void) {
```
szhorvat:
```suggestion
void test_bond(void) {
```

tests/unit/percolation.c
```diff
@@ -0,0 +1,289 @@
...
+    igraph_destroy(&singleton);
+    igraph_destroy(&null_graph);
+    igraph_destroy(&k_3);
+    igraph_destroy(&c_4);
+    igraph_destroy(&karate);
+    igraph_destroy(&random);
+    VERIFY_FINALLY_STACK();
+    return 0;
+}
+igraph_error_t percolate_s(igraph_t *graph, igraph_vector_int_t *vert_indices, igraph_bool_t printing) {
```
szhorvat:
```suggestion

}

igraph_error_t percolate_s(igraph_t *graph, igraph_vector_int_t *vert_indices, igraph_bool_t printing) {
```

No return for void and empty line between functions for readability.

src/connectivity/percolation.c
```diff
@@ -0,0 +1,309 @@
...
+ * \function igraph_bond_percolation
+ * \brief Calculates the bond-percolation curve of a graph.
+ * Calculates the bond-percolation curve, or the size of the largest component as
+ * edges are added to the graph in the order given. If both the output and the input
+ * are reversed is the size of the largest component as edges are removed.
+ * \param graph The input graph
+ * \param output output[i] is the size of the largest component after adding
+ *    <code>i+1</code> edges, will be created.
+ * \param edges The order of the edges, will be generated at random if null.
+ * */
```
szhorvat: Major points:

 - Missing experimental.
 - Closing `*/`
 - Missing `\return`.

TODO: Can you add a note on time complexity?

TODO: Make sure param names match the function prototype.

```suggestion
/**
 * \function igraph_bond_percolation
 * \brief Calculates the bond-percolation curve of a graph.
 *
 * \experimental
 *
 * Calculates the bond-percolation curve, or the size of the largest component as
 * edges are added to the graph in the order given. If both the output and the input
 * are reversed, it is the size of the largest component as edges are removed.
 *
 * \param graph The input graph
 * \param output output[i] is the size of the largest component after adding
 *    <code>i+1</code> edges, will be created.
 * \param edges The order of the edges, will be generated at random if \c NULL.
 * \return Error code.
 */
```

src/connectivity/percolation.c
```diff
@@ -0,0 +1,309 @@
...
+    }
+
+    igraph_vector_int_t neighbors;
+    IGRAPH_VECTOR_INT_INIT_FINALLY(&neighbors, 0);
+    for (igraph_integer_t i = 0; i < vcount; i++) {
+        IGRAPH_CHECK(percolate_site(graph, &links, &sizes, &biggest, VECTOR(*p_vertex_order)[i], &neighbors));
+        VECTOR(*output)[i] = biggest;
+    }
+    igraph_vector_int_destroy(&neighbors);
+    IGRAPH_FINALLY_CLEAN(1);
```
szhorvat: Yes, this is what I meant with re-using the `neighbors` vector!

tests/unit/percolation.c
```diff
@@ -0,0 +1,289 @@
...
+    percolate_s(&karate, NULL, false);
+    // Generated disconnected graph, 100 vertices, p=0.01
+
+    igraph_vector_int_t storage_order;
+    igraph_vector_int_init_range(&storage_order, 0, igraph_vcount(&karate));
+    printf("Zachary karate graph, vertices in storage order.\n");
+    percolate_s(&karate, &storage_order, true);
+    igraph_vector_int_destroy(&storage_order);
+
+    percolate_b(&random, NULL, false);
```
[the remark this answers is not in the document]
szhorvat: @Zepeacedust Please check if this is what you meant (see Copilot comment).

tests/unit/percolation.c
```diff
@@ -0,0 +1,289 @@
...
+    igraph_full(&k_5, 5, false, false);
+    igraph_small(&c_4, 4, IGRAPH_UNDIRECTED, 0, 1, 1, 2, 2, 3, 3, 0, -1);
+    igraph_famous(&karate, "Zachary");
+    igraph_erdos_renyi_game_gnp(&random, 100, 0.01, false, false);
+    printf("# Site percolation test suite\n");
+    printf("Null graph, no provided vertex order.\n");
+    percolate_s(&null_graph, NULL, true);
+
+    IGRAPH_ASSERT(true);
+
```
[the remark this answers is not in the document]
szhorvat: What happened here? :-) (see Copilot comment).

include/igraph_components.h
```diff
@@ -64,6 +64,10 @@ IGRAPH_EXPORT igraph_error_t igraph_biconnected_components(const igraph_t *graph
 IGRAPH_EXPORT igraph_error_t igraph_is_biconnected(const igraph_t *graph, igraph_bool_t *result);
 IGRAPH_EXPORT igraph_error_t igraph_bridges(const igraph_t *graph, igraph_vector_int_t *bridges);
 
+IGRAPH_EXPORT igraph_error_t igraph_bond_percolation(const igraph_t *graph, igraph_vector_int_t *output, const igraph_vector_int_t *edge_order);
+IGRAPH_EXPORT igraph_error_t igraph_site_percolation(const igraph_t *graph, igraph_vector_int_t *output, const igraph_vector_int_t *vertex_order);
+IGRAPH_EXPORT igraph_error_t igraph_edge_list_percolation(const igraph_vector_int_t *edges, igraph_vector_int_t *output);
```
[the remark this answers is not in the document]
szhorvat: This is fine, no need to act on this comment.

src/connectivity/percolation.c
```diff
@@ -0,0 +1,309 @@
...
+ * \function igraph_i_edge_list_percolation
+ * \brief Gives the size of the largest connected component as edges are added.
+ * Calculates the size of the largest component as edges are added to a graph in the order given.
+ * If the edge-sequence and output are reversed, they are the size of the largest component as
+ * edges are removed.
+ *
+ * \param edges Vector of edges, where the n-th edge has endpoints edges[2n] and edges [2n+1].
+ * \param output output[n] is the size of the largest connected component after edge n is added.
+ *        Will be resized.
+ */
```
szhorvat: First, see the comments I made on the docs of the other two functions.

 - Name in `\function` needs fixing.
 - Move `\experimental` to correct place. (Suggestion: In the future, check how a construct is use in other parts of the codebase, and emulate the usage from there.)
 - Newlines.
 - Missing `\return`
 - Please add time complexity.

szhorvat (changes requested):
This is a very good improvement, but there are a few more things to pay attention to. Please ping me to Discord when ready.

Note: I ran this through a static analyser, which was happy.

tests/unit/percolation.c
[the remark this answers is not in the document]
Zepeacedust: I can't find the exact commit where it happened, but I am fairly sure that it is an artifact of a test that was failing when I didn't think it should have, and I was starting to doubt that the assert was even working. So I added something that should never fail, and due to the fact it never fails I didn't notice that it was still there.

tests/unit/percolation.c
```diff
@@ -0,0 +1,284 @@
...
+    IGRAPH_VECTOR_INT_INIT_FINALLY(&outputs, 0);
+
+    IGRAPH_CHECK(igraph_site_percolation(graph, &outputs, NULL));
+
+    *size = VECTOR(outputs)[igraph_vector_int_size(&outputs) -1];
+
+    igraph_vector_int_destroy(&outputs);
+    IGRAPH_FINALLY_CLEAN(1);
+    return IGRAPH_SUCCESS;
+}
```
szhorvat: @Zepeacedust This function is never used, please have a look at what you intended to do with it.

Zepeacedust: It was used to check that the percolation agreed on the total size of the graph with a pre-existing components, but that check was folded into the larger `percolate_s` function.
It's vestigial now, and can be removed.

szhorvat:
@ntamas, please have a look, but do not merge yet. Check if the documentation is understandable to you.

I'm thinking to relax the requirement that _all_ edges/vertices should be included i the edge/vertex order list.  If we allow partial lists, it's good to think about whether it's better to allow specifying a incomplete list to _add_ to the graph (which matches the implementation) or to _remove_ from the graph (which might or might not match what people want).

Today I found that graph-tool has similar functionality. It interprets partial lists as vertices _added_, not _removed_. I'm including to stick to this, and leave anything more complicated to the user to implement.

----

I would however keep the check for duplicates.

szhorvat:
@Zepeacedust Do we have a check for duplicate edge IDs? I see we do have one for duplicate vertex IDs.

Zepeacedust:
We do not, it would not be too hard to implement though.

src/connectivity/percolation.c
```diff
@@ -0,0 +1,339 @@
...
+        // Use random edge order
+
+        IGRAPH_CHECK(igraph_vector_int_init_range(&i_edge_order, 0, igraph_ecount(graph)));
+        IGRAPH_FINALLY(igraph_vector_int_destroy, &i_edge_order);
+        IGRAPH_CHECK(igraph_vector_int_shuffle(&i_edge_order));
+
+        p_edge_order = &i_edge_order;
+    } else {
+        p_edge_order = edge_order;
+    }
```
szhorvat: This would be an easy place to add a check for duplicate edges, by iterating through the edge ID list and:

 - Checking that IDs are within bounds.
 - Using an `igraph_bitset_t` to mark seen edges, and throw an error when duplicates are found.

szhorvat: @Zepeacedust Unless I'm missing something, this function does not actually check that all edges are specified in the edge order list, correct? Let's leave it like that for now.

src/connectivity/percolation.c
```diff
@@ -0,0 +1,339 @@
...
+        VECTOR(*output)[i] = biggest;
+    }
+    igraph_vector_int_destroy(&neighbors);
+    IGRAPH_FINALLY_CLEAN(1);
+
+    for (igraph_integer_t i = 0; i < vcount; i++) {
+        if (VECTOR(sizes)[i] == 0) {
+            IGRAPH_ERROR("Vertex order vector is missing vertices from graph.", IGRAPH_EINVAL);
+        }
+    }
```
szhorvat: @Zepeacedust Isn't this check redundant with the check above on the size of `vertex_order`?

szhorvat:
I keep thinking about how to make this more generally useful. We often want to compute the mean degree. It would be nice to have an extra output parameter for `igraph_site_percolation()` that records how many edges (not vertices!) have been added so far. So, `edge_count[i]` contains the number of edges present after having added `i+1` vertices.

Though somewhat less useful, analogously we can also add a `vertex_count` parameter to `igraph_bond_percolation()` and `igraph_edgelist_percolation()`. This records the number of vertices that became adjacent to some edge so far.

I'll have to take another look at the rich club function and sync this concept up with what that does.

src/connectivity/percolation.c
```diff
@@ -0,0 +1,339 @@
...
+
+    if (VECTOR(*sizes)[vertex] != 0) {
+        IGRAPH_ERROR("Duplicate vertex in vertex order vector.", IGRAPH_EINVAL);
+    }
+
+    VECTOR(*sizes)[vertex] = 1;
+
+    igraph_integer_t neighbor_count;
+
+    IGRAPH_CHECK(igraph_neighbors(graph, neighbors, vertex, IGRAPH_IN));
```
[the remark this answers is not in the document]
szhorvat: @Zepeacedust Good catch by the bot. What did you intend to achieve with `IN` here?

Zepeacedust:
> Though somewhat less useful, analogously we can also add a `vertex_count` parameter to `igraph_bond_percolation()` and `igraph_edgelist_percolation()`. This records the number of vertices that became adjacent to some edge so far.
> 
> I'll have to take another look at the rich club function and sync this concept up with what that does.

I think that a more useful metric would be to return the total number of components present at each step. Although that does also make sense for the site percolation.
At what point does a single function return too many metrics?

interfaces/functions.yaml
```diff
@@ -1304,6 +1304,19 @@ igraph_is_biconnected:
 igraph_count_reachable:
     PARAMS: GRAPH graph, OUT VECTOR_INT counts, NEIMODE mode
 
+#######################################
+# Percolation
+#######################################
+
+igraph_bond_percolation:
+    PARAMS: GRAPH graph, OUT VECTOR_INT giant_size, OUT VECTOR_INT, connected_vertices OPTIONAL EDGE_INDICES edge_order
```
szhorvat:
```suggestion
    PARAMS: GRAPH graph, OUT VECTOR_INT giant_size, OUT VECTOR_INT connected_vertices, OPTIONAL EDGE_INDICES edge_order
```

szhorvat:
@Zepeacedust In this implementation, if a vertex receives a self-loop, but does not connect to other vertices, it still won't be counted in the output of `bond_percolation()`, correct?

Let's count every vertex that has some edges, for consistency with `site_percolation()`.  Think of it like this: a vertex is counted as soon as it appears for the first time in the edge list of `igraph_edgelist_percolation()`.

Perhaps we can slightly change the encoding in `links` and start with `links[v] == -1` for each vertex. I know that this'll complicate the check in the union-find, but it allows keeping track of whether the vertex was already seen.

include/igraph_components.h
```diff
@@ -43,27 +43,42 @@ __BEGIN_DECLS
 
 /* Deprecated alias to igraph_connected_components; will be removed in 0.11 */
 IGRAPH_EXPORT IGRAPH_DEPRECATED igraph_error_t igraph_clusters(const igraph_t *graph, igraph_vector_int_t *membership,
-                                  igraph_vector_int_t *csize, igraph_integer_t *no,
```
szhorvat: Please do no reformat code that is not relevant for this PR as it will create merge conflicts.

src/connectivity/percolation.c
```diff
@@ -0,0 +1,388 @@
...
+    }
+}
+
+
+// test if node is already connected to something
+// If it is a root, its size will be greater than 1
+// If it is not a root then it will link somewhere.
+static bool is_connected(igraph_vector_int_t *links, igraph_vector_int_t *sizes, igraph_integer_t vert) {
+    return VECTOR(*links)[vert] != vert || VECTOR(*sizes)[vert] > 1;
+}
```
szhorvat: Is this function used? If not, remove.

szhorvat (changes requested):
Let's get this ready for merging ASAP.

As we agreed,

 - Duplicate checks must be made both for edge and vertex percolation. This seems missing from edge percolation.
 - Do not throw an error if not all edges/vertices are specified. I see an error message checking that all vertices are specified.

szhorvat:
Please also check any earlier unresolved  comments and resolve when everything is done (but make sure the comment is addressed)!

Mark this as "Ready for review" and ping me here when you think no more changes are needed before merging.

include/igraph_components.h
Zepeacedust: I think this is happening because the prior code isn't conforming with the astyle file for the project.
I have set emacs to automatically apply it whenever I save a file, so it messes with the whole file.

I have reverted the changes and am looking for a way to control the application of astyle.

szhorvat:
@ntamas If you have time, can you please review this?

szhorvat:
> * Do not throw an error if not all edges/vertices are specified. I see an error message checking that all vertices are specified.

@Zepeacedust I still see this:

```
        IGRAPH_ERRORF("Vertex order vector length (%" IGRAPH_PRId") "
                      "does not match the vertex count (%" IGRAPH_PRId ").",
                      IGRAPH_EINVAL,
                      igraph_vector_int_size(p_vertex_order), vcount);
```

And also this:

```
IGRAPH_ERROR("Vertex order vector is missing vertices from graph.", IGRAPH_EINVAL);
```

Can you relax the restriction on needing to specify all vertices? A partial vertex or edge order vector should be allowed.  Be sure that there's a test that exercises this case.

src/connectivity/percolation.c
```diff
@@ -0,0 +1,393 @@
...
+    igraph_integer_t parent, child;
+    if (VECTOR(*sizes)[a] < VECTOR(*sizes)[b]) {
+        parent = b;
+        child = a;
+    } else {
+        parent = a;
+        child = b;
+    }
+
+    // Make a child of b
```
ntamas: This comment is not relevant any more; you are making `child` a child of `parent` and the selection was done already.

src/connectivity/percolation.c
```diff
@@ -0,0 +1,393 @@
...
+        IGRAPH_CHECK(igraph_vector_int_init_range(&i_edge_order, 0, igraph_ecount(graph)));
+        IGRAPH_FINALLY(igraph_vector_int_destroy, &i_edge_order);
+        IGRAPH_CHECK(igraph_vector_int_shuffle(&i_edge_order));
+
+        p_edge_order = &i_edge_order;
+    } else {
+        // Verify that there are no duplicates.
+        igraph_bitset_t present_edges;
+        igraph_integer_t edge_count = igraph_vector_int_size(edge_order);
+        igraph_bitset_init(&present_edges, edge_count);
```
ntamas: Needs an `IGRAPH_CHECK()`

szhorvat:
@ntamas Thanks for checking it!  I made a bunch of fixes and readability improvements. Please have another check and if you're happy, merge this (or let me know that I can merge). Be sure to squash commits, we don't want to add 200 commits to `master`.

I want to get this off the table ASAP.

I can merge into develop and update accordingly later.

ntamas (approved):

ntamas:
Merged into both `develop` and `master` (because it's not API-changing so why not).

PR #2781, opened 2025-06-28
feat: igraph_layout_align()

### LINKED ISSUE #875: [Feature suggestion] Rotate graph layouts for nicer appearance
I noticed a difference between Mathematica and igraph's layout method: Mathematica always rotates the layouts wide.  Here's a comparison of the same type of layout algorithm in both (Mathematica is on the top):

image

This could be done through a simple principal component analysis and would be a nice feature for the high level interfaces. Just an idea :-) I implemented this for the Mathematica interface.

szhorvat:
Closes #875

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [master]
 
+### Added
+
+ - `igraph_layout_align()` attempts to align a graph layout with the coordinate axes in a visually pleasing manner (experimental function).
+
 ### Fixed
 
  - Fix failure in SIR simulation due to roundoff errors creating slightly negative rates.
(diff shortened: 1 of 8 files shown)


szhorvat:
This is the current state. I need to permute coordinates so as to align "more horizontally" in the 2D case. Comments welcome, @ntamas and @schochastics .

image

image

szhorvat:
@ntamas @schochastics This is ready to go. It'd be great to see this used as the part of the default plot layout in the next R release.

image

image

schochastics:
The result looks great!

szhorvat:
@greptileai

PR #2786, opened 2025-06-30
Near-linear implementation of Havel-Hakimi

minifinity:
I used LLMs to search through documentation/understand programming syntax.

### DIFF
--- a/.github/workflows/arch-emu.yml
+++ b/.github/workflows/arch-emu.yml
@@ -18,12 +18,15 @@ env:
   CMAKE_COLOR_DIAGNOSTICS: ON
   CCACHE_MAXSIZE: 64M
   OMP_NUM_THREADS: 1
-  CTEST_PARALLEL_LEVEL: 0 # Since CMake 3.29, a setting of 0 will try to use all available resources
+  CTEST_PARALLEL_LEVEL: 2 # No more than two tests running at a time to prevent timeouts
 
 jobs:
 
   alpine:
     runs-on: ubuntu-latest
+
+    # Do not run in forks
+    if: github.repository == 'igraph/igraph'
 
     defaults:
       run:
@@ -122,7 +125,7 @@ jobs:
 
       - name: Test
         # Need a longer timeout as emulated runs are slow.
-        run: cd build && ctest --output-on-failure --timeout 240
+        run: cd build && ctest --output-on-failure
 
       - name: Ccache statistics
         continue-on-error: true
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,7 @@
  - The widest path functions `igraph_widest_path_widths_floyd_warshall()`, `igraph_widest_path_widths_dijkstra()`, `igraph_get_widest_paths()`, and `igraph_get_widest_path()` incorrectly ignored edges with positive infinite width. Now they ignore edges with negative infinite width.
  - `igraph_cliques_callback()` would sometimes fail to respect a request to stop (i.e. returning `IGRAPH_STOP`) from the callback. This is now corrected.
  - `igraph_hypercube()` now validates the hypercube dimension and prevents negative values.
+ - `igraph_sparsemat_view()` checks for out-of-memory conditions.
 
 ### Deprecated
 
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -78,19 +78,19 @@ jobs:
           extra_cmake_args: '-DBLA_VENDOR=OpenBLAS'
           build_shared: true
 
-  - job: linux_clang_20
+  - job: linux_clang_22
     steps:
       - script: |
           sudo apt-get install ninja-build ccache -y
           wget https://apt.llvm.org/llvm.sh
           chmod +x llvm.sh
-          sudo ./llvm.sh 20
+          sudo ./llvm.sh 22
         displayName: Install dependencies
 
       - template: .azure/build.yml
         parameters:
           build_type: Debug
-          extra_cmake_args: '-DUSE_SANITIZER=Address\;Undefined -DCMAKE_C_FLAGS="-Og -fno-sanitize-recover=undefined -fno-sanitize=float-divide-by-zero" -DCMAKE_CXX_FLAGS="-Og -fno-sanitize-recover=undefined -fno-sanitize=float-divide-by-zero" -DCMAKE_C_COMPILER=clang-20  -DCMAKE_CXX_COMPILER=clang++-20'
+          extra_cmake_args: '-DUSE_SANITIZER=Address\;Undefined -DCMAKE_C_FLAGS="-Og -fno-sanitize-recover=undefined -fno-sanitize=float-divide-by-zero" -DCMAKE_CXX_FLAGS="-Og -fno-sanitize-recover=undefined -fno-sanitize=float-divide-by-zero" -DCMAKE_C_COMPILER=clang-22  -DCMAKE_CXX_COMPILER=clang++-22'
 
   - job: linux_x87
     steps:
--- a/etc/cmake/compilers.cmake
+++ b/etc/cmake/compilers.cmake
@@ -63,7 +63,7 @@ macro(use_all_warnings TARGET_NAME)
         $<$<BOOL:${IGRAPH_WARNINGS_AS_ERRORS}>:-Werror>
         -Wall -Wextra -pedantic
         -Wstrict-prototypes
-        -Wno-unused-function -Wno-unused-parameter -Wno-unused-but-set-variable -Wno-sign-compare -Wno-constant-logical-operand
+        -Wno-unused-function -Wno-unused-parameter -Wno-unused-but-set-variable -Wno-sign-compare -Wno-constant-logical-operand -Wno-uninitialized-const-pointer
       >
       $<$<BOOL:${COMPILER_SUPPORTS_UNKNOWN_WARNING_OPTION_FLAG}>:-Wno-unknown-warning-option>
       # Intel compiler:
--- a/fuzzing/build.sh
+++ b/fuzzing/build.sh
@@ -8,9 +8,9 @@ mkdir $DEPS_PATH
 # It may be necessary to leave CMAKE_BUILD_TYPE empty and specify LIBXML2_WITH_MODULES=OFF
 # in order for fuzz introspector builds to succeed (details unverified).
 cd $SRC
-wget https://download.gnome.org/sources/libxml2/2.14/libxml2-2.14.5.tar.xz
-tar xf libxml2-2.14.5.tar.xz
-cd libxml2-2.14.5
+wget https://download.gnome.org/sources/libxml2/2.14/libxml2-2.14.6.tar.xz
+tar xf libxml2-2.14.6.tar.xz
+cd libxml2-2.14.6
 mkdir build && cd build
 CFLAGS="$CFLAGS -O2" cmake .. -DCMAKE_INSTALL_PREFIX=$DEPS_PATH -DBUILD_SHARED_LIBS=OFF -DLIBXML2_WITH_ICU=OFF -DLIBXML2_WITH_PYTHON=OFF -DLIBXML2_WITH_TESTS=OFF -DLIBXML2_WITH_ZLIB=OFF -DLIBXML2_WITH_LZMA=OFF -DLIBXML2_WITH_PROGRAMS=OFF -DLIBXML2_WITH_MODULES=OFF
 make install -j$(nproc)
--- a/src/core/sparsemat.c
+++ b/src/core/sparsemat.c
@@ -3202,6 +3202,9 @@ igraph_error_t igraph_sparsemat_view(igraph_sparsemat_t *A, igraph_integer_t nzm
                           igraph_integer_t *p, igraph_integer_t *i, igraph_real_t *x, igraph_integer_t nz) {
 
     A->cs = IGRAPH_CALLOC(1, cs_igraph);
+
+    IGRAPH_CHECK_OOM(A->cs, "Insufficient memory to create sparsemat view.");
+
     A->cs->nzmax = nzmax;
     A->cs->m = m;
     A->cs->n = n;
--- a/tests/unit/igraph_lapack_dsyevr.c
+++ b/tests/unit/igraph_lapack_dsyevr.c
@@ -46,7 +46,7 @@ igraph_bool_t check_ev(const igraph_matrix_t *A,
         igraph_blas_dgemv(/*transpose=*/ 0, /*alpha=*/ 1.0, A, &v,
                                          /*beta=*/ -VECTOR(*values)[i], &y);
         for (j = 0; j < n; j++) {
-            if (fabs(VECTOR(y)[i]) > tol) {
+            if (fabs(VECTOR(y)[j]) > tol) {
                 printf("Matrix:\n");
                 igraph_matrix_print(A);
                 printf("lambda= %g\n", VECTOR(*values)[i]);
--- a/tests/unit/zero_allocs.c
+++ b/tests/unit/zero_allocs.c
@@ -21,6 +21,7 @@
 
 int main(void) {
     int *a;
+    char *b;
 
     /* Macros */
 
@@ -33,15 +34,17 @@ int main(void) {
     IGRAPH_FREE(a);
     IGRAPH_ASSERT(!a); /* IGRAPH_FREE(a) sets 'a' to NULL */
 
-    a = IGRAPH_MALLOC(0);
-    IGRAPH_ASSERT(a);
+    /* We use type 'char' to work around warnings from some moderns compilers such as Clang 22:
+     * error: allocation of insufficient size '1' for type 'int' with size '4' [-Werror,-Walloc-size] */
+    b = IGRAPH_MALLOC(0);
+    IGRAPH_ASSERT(b);
 
-    IGRAPH_FREE(a);
-    IGRAPH_ASSERT(!a);
+    IGRAPH_FREE(b);
+    IGRAPH_ASSERT(!b);
 
     /* Functions */
 
-    a = igraph_calloc(0, sizeof(int));
+    a = igraph_calloc(0, sizeof(*a));
     IGRAPH_ASSERT(a);
 
     a = igraph_realloc(a, 0);


szhorvat:
@minifinity @Tagl Is the plan to add the fully linear implementation to this same PR later, or do you intend to get this merged, then improve using a separate PR?

Always feel free to make use of the CI checks to run analyses on the PRs, even if you don't intend to merge in the current form.

src/misc/degree_sequence.cpp
```diff
@@ -68,66 +75,72 @@ template<typename T> inline bool degree_less(const T &a, const T &b) {
...
     }
 
-    while (! vertices.empty()) {
-        if (largest) {
-            std::stable_sort(vertices.begin(), vertices.end(), degree_less<vd_pair>);
+    igraph_integer_t n_edges_added = 0;
+    for (igraph_integer_t i = 0; i < igraph_vector_int_size(degseq); i++) {
+        // CHECK FOR COMPLETE
+        if (nodes.empty()) return IGRAPH_SUCCESS;
+
```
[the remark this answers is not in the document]
Tagl: This is a fine suggestion since we removed the custom order case from your old code.

src/misc/degree_sequence.cpp
```diff
@@ -68,66 +75,72 @@
...
         }
 
-        if (largest) {
-            for (igraph_integer_t i = 0; i < vd.degree; ++i) {
-                if (--(vertices[vertices.size() - 1 - i].degree) < 0) {
-                    goto fail;
-                }
+        // CONNECT THEM
+        for (const auto node : visited) { // O(E)
+            VECTOR(*edges)[2*n_edges_added] = hub.vertex;
```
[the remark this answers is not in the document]
Tagl: Can we see the input that caused this?
`edges` is initialized by caller
```
IGRAPH_VECTOR_INT_INIT_FINALLY(&edges, deg_sum);
```
So n_edges_added is only large enough to cause this when the algorithm has finished from what I can gather.

Tagl:
> Is the plan to add the fully linear implementation to this same PR later, or do you intend to get this merged, then improve using a separate PR?

~~I think it's good to merge this and make another.~~ Lets update this draft with a linear implementation before merging. However, there are some documentation changes still missing here.

@minifinity The documentation needs to be updated to note the new time complexity.

I think it would also be nice to update the tests so that they output the degree of each node from the construction like we discussed yesterday. This will make it easier to verify that the output of the function is valid even though it doesn't match the graph produced by the previous implementation exactly. @szhorvat Is it a concern that the produced graph is not the exact same as before?

src/misc/degree_sequence.cpp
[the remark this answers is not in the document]
szhorvat: I have not seen `@github-advanced-security` before, so that's new. It looks like it is reacting to the CIFuzz output, which is also failing. These fuzzers are run for 30 seconds only, which means that you should be able to reproduce the issue locally, and generate your own offending input, in a relatively short about of time.

Fuzzing instructions are in the wiki:

https://github.com/igraph/igraph/wiki/Fuzzing

Let me know if they are unclear, so we can improve them.

You will need the open source version of Clang to do this. For @Tagl, who uses Linux, this just means installing Clang from your distro's package manager (i.e. don't use GCC). For @minifinity this means NOT using Apple Clang, but getting it from somewhere else (typically Homebrew or MacPorts).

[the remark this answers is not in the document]
szhorvat: Here's a complete transcript of my fuzzing workflow, including building fuzzers (I use MacPorts, which is why my compiler is called `clang-mp-20`), running the fuzzer, minimizing the crash input further.

https://gist.github.com/szhorvat/0dcbc5fe3221746201028c580588c0b7

Before the last re-build that you can see in the trascript, I looked at the stack trace to see where the crash happens. It says line 94 if `linear_algos_undirected.cpp`. This is the call there:

```cpp
            CHECK_ERROR(
                igraph_realize_degree_sequence(&g, &iv1, NULL, IGRAPH_SIMPLE_SW, IGRAPH_REALIZE_DEGSEQ_LARGEST),
                IGRAPH_EINVAL);
```

I added `igraph_vector_int_print(&iv1);` just before it, re-built the fuzzer, and ran it on the minimized test case. Then it printed this degree sequence:

https://gist.github.com/szhorvat/0dcbc5fe3221746201028c580588c0b7#file-gistfile1-txt-L761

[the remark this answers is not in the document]
szhorvat: When this sort of thing happens, another idea is to run a static analyser, such as Clang Analyzer. See here for reference:

https://github.com/igraph/igraph/wiki/Running-linters-and-static-analysers

On macOS this is very easy to do with Xcode.  Create a new build directory and run this:

```
cmake -GXcode .. -DIGRAPH_WARNINGS_AS_ERRORS=OFF
```

Open the resulting `igraph.xcodeproj` file.

Manually manage schemed, hit `+`, choose the entry named simple `igraph`. (Automatically create schemes is also fine but will slow you down a bit.) Confirm everything.

Choose Product -> Analyze from the many and wait for the build to complete. You want to have your laptop connected to power for this, it'll spin up the fans.

In the left hand side panel, choose this icon:

image

Scroll down to the `igraph` section, and see if there are any warnings for your file.

_In this case there aren't_, but generally, this tool is very good at finding memory access issues, so I thought it's worth letting you know how to use it.

szhorvat:
> Lets update this draft with a linear implementation before merging.

I agree, as there is some overhead to bringing any PR to a mergable state (see all the back and forth we went though in #2778 ;-)

szhorvat:
> @szhorvat Is it a concern that the produced graph is not the exact same as before?

No, that is not a problem at all. It wouldn't be feasible to keep it the same.

src/misc/degree_sequence.cpp
```diff
@@ -75,71 +70,238 @@ template<typename T> inline bool degree_less(const T &a, const T &b) {
...
-// helper function for igraph_i_havel_hakimi()
-vd_pair decrement_degree(std::set<vd_pair>& nodes, std::set<vd_pair>::iterator it) {
-    vd_pair node = *it;
-    nodes.erase(it);
+// "Bucket Node" for nodes of the same degree
+struct BNode {
+    igraph_integer_t count = 0;                // TODO: correct use of uint?
+    std::stack<vd_pair> nodes;
+    BNode* next;       // next bucket (higher degree)
+    BNode* prev;       // prev bucket (lower degree)
```
szhorvat: I have not looked in detail, but I noticed the pointers here. When all nodes are in an array, do not use pointers! Use the _index_ in an array to refer to a node. Keep in mind that an `std::vector` may re-allocate its storage. Indices are stable, pointers are not. We don't want to have to worry about avoiding reallocations. Even if it is obvious to you that this won't be a problem here (due to a fixed-length allocation at the beginning and no re-allocation), it may not be so obvious to future maintainers who are making changes under stress and time constraints.

szhorvat:
@minifinity Were you able to find the issue with the non-graphical degree sequence that caused a crash?

src/misc/degree_sequence.cpp
```diff
@@ -68,66 +70,232 @@ template<typename T> inline bool degree_less(const T &a, const T &b) {
...
-    igraph_integer_t ec = 0; // number of edges added so far
+struct HavelHakimiList {
+    std::vector<BNode> buckets;
 
-    std::vector<vd_pair> vertices;
-    vertices.reserve(n);
-    for (igraph_integer_t i = 0; i < n; ++i) {
-        vertices.push_back(vd_pair(i, VECTOR(*deg)[i]));
+    // Given degree sequence, sets up linked list of BNodes (degree buckets)
+    // O(N)                                       // sentinel BNode [0] and [N] as bookends
```
szhorvat: I think this comment is meant to be on a separate line.

src/misc/degree_sequence.cpp
```diff
@@ -68,66 +70,232 @@ template<typename T> inline bool degree_less(const T &a, const T &b) {
...
-        if (largest) {
-            std::stable_sort(vertices.begin(), vertices.end(), degree_less<vd_pair>);
-        } else {
-            std::stable_sort(vertices.begin(), vertices.end(), degree_greater<vd_pair>);
+    // ----- O(1) convenience functions ----- //
+    BNode head() {return buckets.front();}
+    BNode tail() {return buckets.back();}
+
+    void remove_bucket(igraph_integer_t degree) {            // assuming sentinel nodes
+        igraph_integer_t& prev_idx = buckets[degree].prev;            // TODO: references
```
szhorvat: TODO note: Anything else to do here before merging?

src/misc/degree_sequence.cpp
```diff
@@ -68,66 +70,232 @@ template<typename T> inline bool degree_less(const T &a, const T &b) {
...
+        IGRAPH_ERROR("Custom ordering specified but no order given", IGRAPH_EINVAL);
+    } else if (n_nodes == 0 || (n_nodes == 1 && VECTOR(*degseq)[0] == 0)) {
+        return IGRAPH_SUCCESS;
     }
 
-    return IGRAPH_SUCCESS;
+    for (igraph_integer_t i = 0; i < n_nodes; i++) {
+        igraph_integer_t deg = VECTOR(*degseq)[i];
+        if (deg >= n_nodes) {
+            IGRAPH_ERROR("Invalid degree sequence.", IGRAPH_EINVAL);
```
szhorvat: Please use the following message consistently, and make sure that this is done in all _new_ code.

```
"The given degree sequence cannot be realized as a simple graph."
```

Goals:

 - Use the same message whenever the input is not graphical.
 - Communicate which case were're handling (simple graph, multigraph, etc.)

src/misc/degree_sequence.cpp
```diff
@@ -68,66 +70,232 @@ template<typename T> inline bool degree_less(const T &a, const T &b) {
...
+
+static igraph_error_t igraph_i_havel_hakimi(const igraph_vector_int_t *degseq,
+                                            igraph_vector_int_t *edges,
+                                            Method method,
+                                            const igraph_vector_int_t *hub_order = nullptr) {
+    igraph_integer_t n_nodes = igraph_vector_int_size(degseq);
+
+    // ----- upfront error/graphicality checks ----- //
+    if (method == Method::CUSTOM && hub_order == nullptr) {
+        IGRAPH_ERROR("Custom ordering specified but no order given", IGRAPH_EINVAL);
```
szhorvat: It seems to me that this message will only be triggered if there is a bug in _our_ code. 

General principle:

 - Use `IGRAPH_ERROR` for user-facing messages only, not when the bug is in our code.
 - Use `IGRAPH_ERROR` for errors that could be triggered from a high-level interface (such as Python). Programmer errors that are due to a misuse of a function from C that is not possible from a high-level interface (e..g passing a null pointer where it's documented that this is no allowed) should use other error mechanisms
 - If you want to add a safety check against such programmer errors, use an assert. Typically `IGRAPH_ASSERT`, which is not disabled in the released version. If there is a concern about the performance effect of the assertion, use `assert()` (very rarely!).  If you do want a custom message, use `IGRAPH_FATAL`. TL;DR: Almost always an `IGRAPH_ASSERT`

szhorvat: IGRAPH_ERROR messages must always end in punctuation (i.e. a fullstop)

szhorvat (changes requested):
Just small comments, need to run now.

src/misc/degree_sequence.cpp
```diff
@@ -68,66 +70,232 @@ template<typename T> inline bool degree_less(const T &a, const T &b) {
...
+        return IGRAPH_SUCCESS;
+    }
+};
 
-        ec += vd.degree;
+enum class Method {
+    LARGEST_FIRST,
+    SMALLEST_FIRST,
+    CUSTOM
+};
```
szhorvat: I would prefer not to introduce an extra enum when one is already available in the public API. Is there any reason you are not using that one?

src/misc/degree_sequence.cpp
```diff
@@ -68,66 +70,232 @@ template<typename T> inline bool degree_less(const T &a, const T &b) {
...
+        }
+        else if (method == Method::CUSTOM) {
+            igraph_integer_t index = VECTOR(*hub_order)[i];
+            igraph_integer_t degree = VECTOR(seq)[index];
+            hub = vd_pair{index, degree};
+            vault.buckets[degree].count--;
+        }
+        else {
+            IGRAPH_ERROR("Invalid method.", IGRAPH_EINVMODE);
+        }
```
szhorvat: If this error triggers, that's a bug in the igraph code, so once again this should be an `IGRAPH_ASSERT` / `IGRAPH_FATAL` or skipped. The method is already validated in the top-level function.

src/misc/degree_sequence.cpp
```diff
@@ -68,66 +70,232 @@ template<typename T> inline bool degree_less(const T &a, const T &b) {
...
+        vd_pair hub;
+        if (method == Method::SMALLEST_FIRST) {
+            if (!vault.get_min_node(/* out param */hub)) break;
+            vault.remove_min_node();
+        }
+        else if (method == Method::LARGEST_FIRST) {
+            if (!vault.get_max_node(/* out param */hub)) break;
+            vault.remove_max_node();
+        }
+        else if (method == Method::CUSTOM) {
```
szhorvat: `IGRAPH_REALIZE_DEGSEQ_INDEX` still uses the old implementation. The justification for using the more complex linked-list based method was to be able to support this case with good complexity, as we already have simpler approachers for the LARGEST and SMALLEST cases.

Let's have a performant implementation for `IGRAPH_REALIZE_DEGSEQ_INDEX` as well before merging this PR.

src/misc/degree_sequence.cpp
```diff
@@ -68,66 +70,232 @@ template<typename T> inline bool degree_less(const T &a, const T &b) {
...
 
-// Generate simple undirected realization as edge-list.
-// If largest=true, always choose the vertex with the largest remaining degree to connect up next.
-// Otherwise, always choose the one with the smallest remaining degree.
-static igraph_error_t igraph_i_havel_hakimi(const igraph_vector_int_t *deg, igraph_vector_int_t *edges, bool largest) {
-    igraph_integer_t n = igraph_vector_int_size(deg);
+// "Bucket Node" for nodes of the same degree
+struct BNode {
+    igraph_integer_t count = 0;
+    std::stack<vd_pair> nodes;
```
szhorvat, line 77: An `std::stack` does dynamic allocation, which makes a `BNode` a heavy object that is expensive to create. As I recall, there was a way to avoid having to store this in every node, am I mistaken?

We can merge the PR, as there's a good gain in complexity ... but it's strongly preferable to fix this. Allocations are expensive—we really don't want as many allocations as the vertex count.

src/misc/degree_sequence.cpp
```diff
@@ -68,66 +70,232 @@ template<typename T> inline bool degree_less(const T &a, const T &b) {
...
+    igraph_integer_t next;       // next bucket (higher degree)
```
szhorvat: Style: Please separate line comments from the end of the code by only a single space.

src/misc/degree_sequence.cpp
```diff
@@ -68,66 +70,232 @@ template<typename T> inline bool degree_less(const T &a, const T &b) {
...
+struct HavelHakimiList {
+    std::vector<BNode> buckets;
 
-    std::vector<vd_pair> vertices;
-    vertices.reserve(n);
-    for (igraph_integer_t i = 0; i < n; ++i) {
-        vertices.push_back(vd_pair(i, VECTOR(*deg)[i]));
+    // Given degree sequence, sets up linked list of BNodes (degree buckets)
+    // O(N)                                       // sentinel BNode [0] and [N] as bookends
+    HavelHakimiList(igraph_vector_int_t degseq) : buckets(igraph_vector_int_size(&degseq)+1) {
```
szhorvat: It's good practice to mark single-parameter constructors as `explicit`.

szhorvat (changes requested):
Can you please include a more verbose explanation of how the method (not the HH construction, but this specific implementation) works? This implementation is far from trivial, so we should support future maintainers as much as possible.

src/misc/degree_sequence.cpp
```diff
@@ -68,66 +70,232 @@ template<typename T> inline bool degree_less(const T &a, const T &b) {
...
+        }
+        else {
+            IGRAPH_ERROR("Invalid method.", IGRAPH_EINVMODE);
+        }
+        VECTOR(seq)[hub.vertex] = 0;
+
+        // spoke nodes selection
+        igraph_vector_int_t spokes;
+        IGRAPH_CHECK(igraph_vector_int_init(&spokes, 0));        // init with size hub.degree instead?
+        IGRAPH_CHECK(igraph_vector_int_reserve(&spokes, hub.degree));
```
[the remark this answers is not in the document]
szhorvat: Reserve is fine, but the comment should be removed.

szhorvat:
I added a "thumbs up" on Copilot comments that should be addressed in some way (not necessarily by merging the suggestion verbatim).

szhorvat:
The fuzzer is detecting memory leaks. Can you please check these? Most likely it's because some data structures are not registered on the finalizer stack, so they are not freed upon errors—Copilot already pointed these out.

One last thing: @minifinity, I rebased this PR on the latest `master` using the GitHub interface. When you pull locally, you will need to use `git pull --rebase`. If you have uncommitted changed, stash them first with `git stash`, then re-apply them after the pull with `git stash apply`.

src/misc/degree_sequence.cpp
minifinity: Nope. If I remember correctly that was just a note-to-self to check if I want to declare the variable by reference (I do because I modify it later). I should take out the TODOs before I commit, sorry.

src/misc/degree_sequence.cpp
minifinity: I thought about this more and I think it's best to just get rid of the `hub_order` parameter, and just select the vertices in order by index. That leaves the meaning of `IGRAPH_REALIZE_DEGSEQ_INDEX` unchanged so it's not breaking the API, and also as we discussed last time the user can just reorder their degree sequence.

src/misc/degree_sequence.cpp
minifinity: No reason - I just replaced it with `igraph_realize_degseq_t`. I defined the custom enum when I was testing out the logic in another file without the igraph specific types and carried everything over.

src/misc/degree_sequence.cpp
minifinity: Oh interesting, I don't think Arnar and I talked about this. If I recall correctly, we decided to go with a stack because it keeps the nodes in the same order in which they were inserted as we shift them from degree bucket to degree bucket, and that impacts which graph is created but doesn't really matter. 

Would changing this to a `vector<vd_pair>` be less taxing?

src/misc/degree_sequence.cpp
szhorvat: Sounds good, let's do that!

minifinity:
Another fuzzer failure :/ Is this showing a memory leak again?

@szhorvat I have updated most everything else! Sorry for keeping you waiting. It would be great if you could look over two things again, 1) the verbose comment I added above havel hakimi function explaining how it works, and 2) ~~the stack management and handling the memory leaks~~ I think I'm still missing something here. Thanks!

src/misc/degree_sequence.cpp
szhorvat: Let's not worry about this right now. A vector is the same, but I may have misunderstood about the necessity of this.

.gitignore
```diff
@@ -8,3 +8,4 @@
 /build
 /build-*
 /tools/**/*.pyc
+.DS_Store
```
szhorvat: Do you actually need this on you system? I never had problems with this file on mine. Let's not add it unless it is actually getting committed by accident.

src/misc/degree_sequence.cpp
```diff
@@ -68,123 +70,282 @@ template<typename T> inline bool degree_less(const T &a, const T &b) {
...
+            VECTOR(*edges)[2*n_edges_added] = hub.vertex;
+            VECTOR(*edges)[2*n_edges_added + 1] = spoke_idx;
+            n_edges_added++;
+
+            VECTOR(seq)[spoke_idx]--;
+        }
+        IGRAPH_FINALLY_CLEAN(1);
     }
-
+    IGRAPH_FINALLY_CLEAN(1);
```
szhorvat: I see two FINALLY_CLEANs here, both of which seem to be missing a preceding destructor.

szhorvat:
It's not just just the fuzzer that's failing but also several other CI checks. The one to keep an eye on is `linux_static_vendored`, as this has sanitizers enabled. Check the output here.

See for example here:

https://dev.azure.com/igraph-team/igraph/_build/results?buildId=9516&view=logs&j=4c3a8381-b9de-5463-904f-918810769be8&t=e71559a0-54d3-55f3-64cc-74b67ad70398&l=1817

Make sure you know how to locate this output, and ask for help if you can't find it on your own.

Let's take this as an example:

```
Direct leak of 920 byte(s) in 24 object(s) allocated from:
    #0 0x7fe4f10fc778 in realloc ../../../../src/libsanitizer/asan/asan_malloc_linux.cpp:85
    #1 0x55a7ee32da48 in igraph_vector_int_reserve /home/vsts/work/1/s/src/core/vector.pmt:474
    #2 0x55a7ee397308 in igraph_i_havel_hakimi /home/vsts/work/1/s/src/misc/degree_sequence.cpp:331
    #3 0x55a7ee39808e in igraph_i_realize_undirected_degree_sequence /home/vsts/work/1/s/src/misc/degree_sequence.cpp:749
    #4 0x55a7ee3983d9 in igraph_realize_degree_sequence /home/vsts/work/1/s/src/misc/degree_sequence.cpp:976
    #5 0x55a7ee2ff9e3 in main /home/vsts/work/1/s/tests/unit/igraph_ecc.c:192
    #6 0x7fe4f062a1c9  (/lib/x86_64-linux-gnu/libc.so.6+0x2a1c9) (BuildId: 282c2c16e7b6600b0b22ea0c99010d2795752b5f)
    #7 0x7fe4f062a28a in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x2a28a) (BuildId: 282c2c16e7b6600b0b22ea0c99010d2795752b5f)
    #8 0x55a7ee2fbda4 in _start (/home/vsts/work/1/s/build/tests/test_igraph_ecc+0x23da4) (BuildId: a2d45b150ed8d46fab0d831c10f36923f1524b14)
```

We see it's a memory leak. Why don't you see it on your own machine with ASan? It's because of two things:

 - If you have an old Intel-based Mac, LeakSanitizer is not enabled by default and not even supported by Apple Clang
 - If it is an Apple Silicon-based Mac, LeakSanitizer is not going to work at all, unfortunately.

So you will need to rely on the CI output. I do the same, mostly, since I have an Apple Silicon Mac.

Looking at the stack trace, we locate relevant line. The allocation happens on: `src/misc/degree_sequence.cpp:331`, which is:

```c
IGRAPH_CHECK(igraph_vector_int_reserve(&spokes, hub.degree));
```

So it seems that the memory backing `spokes` is never freed. Perhaps `spokes` itself is never destroyed? Indeed I can't see a destructor call, but to be sure, I'll double check with my IDE. In Qt Creator, I right-click `spokes` and choose Find Refereces With Access Type. Most IDEs have something similar. I get this:

image

We can see that neither is a destructor called on `spokes`, nor is it passed to another function to be destroyed there.  So indeed this is a memory leak bug.

----

I hope this makes it clear how to track down and fix each of the leaks.

Let me know if you need more help!

src/misc/degree_sequence.cpp
```diff
@@ -68,123 +70,282 @@ template<typename T> inline bool degree_less(const T &a, const T &b) {
...
+            hub = vd_pair{i, degree};
+            vault.buckets[degree].count--;
+        }
+        VECTOR(seq)[hub.vertex] = 0;
+
+        // spoke nodes selection
+        igraph_vector_int_t spokes;
+        IGRAPH_VECTOR_INT_INIT_FINALLY(&spokes, 0);
+        IGRAPH_CHECK(igraph_vector_int_reserve(&spokes, hub.degree));
+        IGRAPH_CHECK(vault.get_spokes(hub.degree, seq, spokes));
```
szhorvat: Let's refactor this:

 - `get_spokes()` should not rely on `spokes` being an empty vector. It should use `vector_int_clear()` to make sure it's empty before pushing to it.
 - The storage should be reserved by `get_spokes()`, as that's where the relevant logic is located.

The `spokes` vector is repeatedly allocated and destroyed in a loop. This is inefficient. Keep in mind: a memory allocation is always slow, no matter how little memory we are allocating!

Solution: Allocate `spokes` once outside the loop, and re-use it multiple times.

src/misc/degree_sequence.cpp
```diff
@@ -68,123 +70,282 @@ template<typename T> inline bool degree_less(const T &a, const T &b) {
...
+        igraph_integer_t n_spokes = igraph_vector_int_size(&spokes);
+        for (igraph_integer_t j = 0; j < n_spokes; j++) {
+            igraph_integer_t spoke_idx = VECTOR(spokes)[j];
+            VECTOR(*edges)[2*n_edges_added] = hub.vertex;
+            VECTOR(*edges)[2*n_edges_added + 1] = spoke_idx;
+            n_edges_added++;
+
+            VECTOR(seq)[spoke_idx]--;
+        }
+        IGRAPH_FINALLY_CLEAN(1);
```
szhorvat: It looks like a destructor for `spokes` is missing here.

src/misc/degree_sequence.cpp
szhorvat: This second one seems to be missing a destructor for `seq`.

src/misc/degree_sequence.cpp
szhorvat: To elaborate, creating either an `std::stack` or an `std::vector` involves heap allocations, which are very costly, no matter how slow.

This code will perform as many allocations as the number of nodes.

@Tagl, can we make this work relatively easily with an O(1) amount of allocations?

szhorvat:
@minifinity Let me know if you need help in addressing the comment from above.

minifinity:
> @minifinity Let me know if you need help in addressing the comment from above.

I will. Sorry I've just been really behind on my work lately

szhorvat:
@minifinity I see that all checks pass now! 🎉 

Please go through review comments, and mark those as "resolved" which need no further discussion in your opinion.

When you think that this is ready, please click the "Ready to review" button.

szhorvat:
@minifinity @Tagl I pushed a benchmark for this function. Update this PR to the latest `master` (either by rebase or by merging that branch) and you can check how the performance changed.

szhorvat:
Timings before this change:

```
|> Benchmark file: /Users/szhorvat/Repos/igraph-main/igraph/tests/benchmarks/igraph_realize_degree_sequence.c
| G(n,m), n=  100, m=  200, LARGEST, 10000x                                        0.198s   0.19s  0.008s
| G(n,m), n=  100, m=  200, SMALLEST, 10000x                                       0.154s  0.148s  0.006s
| G(n,m), n=  100, m=  200, INDEX, 10000x                                          0.402s  0.395s  0.007s

| G(n,m), n=  100, m= 1000, LARGEST, 10000x                                        0.435s  0.415s  0.021s
| G(n,m), n=  100, m= 1000, SMALLEST, 10000x                                       0.405s  0.385s   0.02s
| G(n,m), n=  100, m= 1000, INDEX, 10000x                                          0.517s  0.497s   0.02s

| G(n,m), n=10000, m=20000, LARGEST, 1x                                            0.256s  0.256s      0s
| G(n,m), n=10000, m=20000, SMALLEST, 1x                                           0.265s  0.264s      0s
| G(n,m), n=10000, m=20000, INDEX, 1x                                              0.958s  0.957s  0.002s

| G(n,m), n=10000, m=100000, LARGEST, 1x                                           0.262s  0.261s  0.001s
| G(n,m), n=10000, m=100000, SMALLEST, 1x                                          0.273s  0.272s  0.001s
| G(n,m), n=10000, m=100000, INDEX, 1x                                             0.972s   0.97s  0.002s

| BA, n=  100, m=   99, LARGEST, 10000x                                            0.125s   0.12s  0.006s
| BA, n=  100, m=   99, SMALLEST, 10000x                                           0.098s  0.092s  0.006s
| BA, n=  100, m=   99, INDEX, 10000x                                              0.391s  0.379s  0.012s

| BA, n=  100, m=  945, LARGEST, 10000x                                            0.381s  0.355s  0.026s
| BA, n=  100, m=  945, SMALLEST, 10000x                                           0.276s  0.253s  0.023s
| BA, n=  100, m=  945, INDEX, 10000x                                              0.514s  0.484s  0.029s

| BA, n=10000, m= 9999, LARGEST, 1x                                                 0.26s  0.255s  0.005s
| BA, n=10000, m= 9999, SMALLEST, 1x                                               0.263s  0.258s  0.005s
| BA, n=10000, m= 9999, INDEX, 1x                                                  0.742s  0.733s  0.009s

| BA, n=10000, m=99945, LARGEST, 1x                                                0.264s   0.26s  0.004s
| BA, n=10000, m=99945, SMALLEST, 1x                                               0.269s  0.265s  0.004s
| BA, n=10000, m=99945, INDEX, 1x                                                  0.815s  0.805s  0.009s
```

And after:

```
|> Benchmark file: /Users/szhorvat/Repos/igraph-main/igraph/tests/benchmarks/igraph_realize_degree_sequence.c
| G(n,m), n=  100, m=  200, LARGEST, 10000x                                        0.124s  0.116s  0.008s
| G(n,m), n=  100, m=  200, SMALLEST, 10000x                                       0.354s  0.345s  0.009s
| G(n,m), n=  100, m=  200, INDEX, 10000x                                          0.312s    0.3s  0.012s

| G(n,m), n=  100, m= 1000, LARGEST, 10000x                                        0.265s  0.241s  0.024s
| G(n,m), n=  100, m= 1000, SMALLEST, 10000x                                       0.512s  0.485s  0.027s
| G(n,m), n=  100, m= 1000, INDEX, 10000x                                          0.479s  0.448s  0.031s

| G(n,m), n=10000, m=20000, LARGEST, 1x                                            0.001s  0.001s      0s
| G(n,m), n=10000, m=20000, SMALLEST, 1x                                           0.277s  0.276s  0.001s
| G(n,m), n=10000, m=20000, INDEX, 1x                                              0.245s  0.244s      0s

| G(n,m), n=10000, m=100000, LARGEST, 1x                                           0.003s  0.003s      0s
| G(n,m), n=10000, m=100000, SMALLEST, 1x                                          0.279s  0.276s  0.004s
| G(n,m), n=10000, m=100000, INDEX, 1x                                             0.273s  0.271s  0.002s

| BA, n=  100, m=   99, LARGEST, 10000x                                            0.072s  0.068s  0.005s
| BA, n=  100, m=   99, SMALLEST, 10000x                                           0.322s  0.311s  0.011s
| BA, n=  100, m=   99, INDEX, 10000x                                              0.229s  0.222s  0.008s

| BA, n=  100, m=  945, LARGEST, 10000x                                            0.276s   0.25s  0.026s
| BA, n=  100, m=  945, SMALLEST, 10000x                                           0.538s  0.508s   0.03s
| BA, n=  100, m=  945, INDEX, 10000x                                              0.515s  0.484s  0.031s

| BA, n=10000, m= 9999, LARGEST, 1x                                                0.001s  0.001s      0s
| BA, n=10000, m= 9999, SMALLEST, 1x                                               0.266s  0.262s  0.004s
| BA, n=10000, m= 9999, INDEX, 1x                                                  0.168s  0.165s  0.003s

| BA, n=10000, m=99945, LARGEST, 1x                                                0.004s  0.003s      0s
| BA, n=10000, m=99945, SMALLEST, 1x                                               0.307s  0.303s  0.004s
| BA, n=10000, m=99945, INDEX, 1x                                                  0.301s  0.296s  0.005s
```

I wonder why there is virtually no speedup for the SMALLEST and INDEX methods.

szhorvat:
@Tagl, are you able to push to this? If not, can you please give @Tagl access to your fork of this repo, @minifinity ?

Tagl:
> @Tagl, are you able to push to this? If not, can you please @Tagl access to your fork of this repo, @minifinity ?

I think not as I'm not part of the igraph organization on GitHub.

src/misc/degree_sequence.cpp
```diff
@@ -68,123 +70,286 @@ template<typename T> inline bool degree_less(const T &a, const T &b) {
...
+    }
 
-    igraph_integer_t ec = 0; // number of edges added so far
+    // Given degree of selected "hub" node, returns degree many "spoke" nodes to connect to
+    // amortized O(alpha(n))
+    igraph_error_t get_spokes(igraph_integer_t degree, const igraph_vector_int_t& seq,
+                              igraph_vector_int_t& spokes) {
+        std::stack<igraph_integer_t> buckets_req; // stack of needed degree buckets
+        igraph_integer_t num_nodes = 0;
+        igraph_integer_t curr = tail().prev; // starts with max_bucket
```
Tagl: Might be nice and clearer to have a convenience function for getting max_bucket and min_bucket instead of calling tail().prev
Not very important though but might avoid some comments you added.

src/misc/degree_sequence.cpp
```diff
@@ -68,123 +70,286 @@ template<typename T> inline bool degree_less(const T &a, const T &b) {
...
+
+        igraph_vector_int_clear(&spokes);
+        IGRAPH_CHECK(igraph_vector_int_reserve(&spokes, degree));
+
+        while (num_nodes < degree && curr > 0) {
+            num_nodes += buckets[curr].count;
+            buckets_req.push(curr);
+            curr = buckets[curr].prev; // gets next smallest NON-EMPTY bucket
```
Tagl: This is the cause of the slow running time.
You need to implement logic here that actually does the path compression we discussed.
I recommend adding another function for that which actually does get the next non-empty bucket. Your current implementation just gets the next bucket.
It's easier to implement this with a recursive function so start with that. If we feel the recursion is slow we can switch to a stack and a loop later.

szhorvat: The issue with recursion is not that it is slow bit that it can cause stack overflow. Anyway, let's go ahead with the simplest fix first, but it's still good to test it with very large inputs and measure the recursion depth for that case to check the risk.

minifinity:
@szhorvat I think the performance should be improved for `smallest` and `index` methods too now

this was before:
```
| G(n,m), n=  100, m=  200, LARGEST, 10000x                                         2.57s   2.55s  0.019s
| G(n,m), n=  100, m=  200, SMALLEST, 10000x                                        2.88s   2.86s  0.018s
| G(n,m), n=  100, m=  200, INDEX, 10000x                                           6.14s   6.11s  0.024s

| G(n,m), n=  100, m= 1000, LARGEST, 10000x                                         6.85s   6.79s  0.062s
| G(n,m), n=  100, m= 1000, SMALLEST, 10000x                                        7.17s   7.11s   0.06s
| G(n,m), n=  100, m= 1000, INDEX, 10000x                                           7.33s   7.28s  0.051s

| G(n,m), n=10000, m=20000, LARGEST, 1x                                             4.96s   4.86s  0.098s
| G(n,m), n=10000, m=20000, SMALLEST, 1x                                            5.11s      5s   0.08s
| G(n,m), n=10000, m=20000, INDEX, 1x                                               8.83s   8.81s  0.018s

| G(n,m), n=10000, m=100000, LARGEST, 1x                                            5.01s   4.95s  0.056s
| G(n,m), n=10000, m=100000, SMALLEST, 1x                                           5.08s   5.03s  0.043s
| G(n,m), n=10000, m=100000, INDEX, 1x                                              8.95s   8.93s  0.017s

| BA, n=  100, m=   99, LARGEST, 10000x                                             1.97s   1.96s  0.012s
| BA, n=  100, m=   99, SMALLEST, 10000x                                            1.61s    1.6s  0.011s
| BA, n=  100, m=   99, INDEX, 10000x                                               5.91s   5.88s  0.034s

| BA, n=  100, m=  945, LARGEST, 10000x                                             5.69s   5.61s  0.074s
| BA, n=  100, m=  945, SMALLEST, 10000x                                            4.03s      4s  0.033s
| BA, n=  100, m=  945, INDEX, 10000x                                               7.15s   7.09s  0.053s

| BA, n=10000, m= 9999, LARGEST, 1x                                                 4.91s   4.85s  0.057s
| BA, n=10000, m= 9999, SMALLEST, 1x                                                4.91s   4.87s  0.039s
| BA, n=10000, m= 9999, INDEX, 1x                                                   8.52s   8.49s  0.019s

| BA, n=10000, m=99945, LARGEST, 1x                                                 4.92s   4.87s  0.046s
| BA, n=10000, m=99945, SMALLEST, 1x                                                5.05s      5s  0.044s
| BA, n=10000, m=99945, INDEX, 1x                                                   8.77s   8.75s  0.019s
```

and after:
```
| G(n,m), n=  100, m=  200, LARGEST, 10000x                                          2.1s   2.04s  0.061s
| G(n,m), n=  100, m=  200, SMALLEST, 10000x                                         2.4s   2.38s  0.021s
| G(n,m), n=  100, m=  200, INDEX, 10000x                                           2.26s   2.25s  0.016s

| G(n,m), n=  100, m= 1000, LARGEST, 10000x                                         5.31s   5.26s  0.056s
| G(n,m), n=  100, m= 1000, SMALLEST, 10000x                                        5.46s    5.4s  0.054s
| G(n,m), n=  100, m= 1000, INDEX, 10000x                                           5.42s   5.36s  0.056s

| G(n,m), n=10000, m=20000, LARGEST, 1x                                            0.022s  0.021s      0s
| G(n,m), n=10000, m=20000, SMALLEST, 1x                                           0.026s  0.025s  0.001s
| G(n,m), n=10000, m=20000, INDEX, 1x                                              0.023s  0.022s  0.001s

| G(n,m), n=10000, m=100000, LARGEST, 1x                                           0.051s   0.05s  0.001s
| G(n,m), n=10000, m=100000, SMALLEST, 1x                                          0.052s  0.051s  0.001s
| G(n,m), n=10000, m=100000, INDEX, 1x                                             0.051s   0.05s  0.001s

| BA, n=  100, m=   99, LARGEST, 10000x                                             1.63s   1.62s  0.015s
| BA, n=  100, m=   99, SMALLEST, 10000x                                            2.31s    2.3s  0.015s
| BA, n=  100, m=   99, INDEX, 10000x                                                1.8s   1.78s  0.013s

| BA, n=  100, m=  945, LARGEST, 10000x                                             6.01s   5.95s  0.055s
| BA, n=  100, m=  945, SMALLEST, 10000x                                            6.62s   6.56s  0.053s
| BA, n=  100, m=  945, INDEX, 10000x                                               6.26s   6.21s  0.052s

| BA, n=10000, m= 9999, LARGEST, 1x                                                0.016s  0.016s      0s
| BA, n=10000, m= 9999, SMALLEST, 1x                                               0.025s  0.025s      0s
| BA, n=10000, m= 9999, INDEX, 1x                                                  0.018s  0.017s      0s

| BA, n=10000, m=99945, LARGEST, 1x                                                0.053s  0.052s  0.001s
| BA, n=10000, m=99945, SMALLEST, 1x                                               0.057s  0.056s  0.001s
| BA, n=10000, m=99945, INDEX, 1x                                                  0.055s  0.054s  0.001s
```

minifinity:
this is after adding a `get_prev()` function for path compression. The benchmark results don't seem too different, but there could be things I'm still missing.

```
| G(n,m), n=  100, m=  200, LARGEST, 10000x                                         2.22s   2.17s  0.052s
| G(n,m), n=  100, m=  200, SMALLEST, 10000x                                        2.49s   2.48s  0.014s
| G(n,m), n=  100, m=  200, INDEX, 10000x                                           2.34s   2.33s  0.012s

| G(n,m), n=  100, m= 1000, LARGEST, 10000x                                         5.33s   5.29s  0.044s
| G(n,m), n=  100, m= 1000, SMALLEST, 10000x                                        5.53s   5.48s  0.041s
| G(n,m), n=  100, m= 1000, INDEX, 10000x                                           5.49s   5.45s  0.038s

| G(n,m), n=10000, m=20000, LARGEST, 1x                                            0.022s  0.022s      0s
| G(n,m), n=10000, m=20000, SMALLEST, 1x                                           0.029s  0.028s  0.001s
| G(n,m), n=10000, m=20000, INDEX, 1x                                              0.024s  0.024s      0s

| G(n,m), n=10000, m=100000, LARGEST, 1x                                            0.05s  0.049s  0.001s
| G(n,m), n=10000, m=100000, SMALLEST, 1x                                          0.052s  0.051s  0.001s
| G(n,m), n=10000, m=100000, INDEX, 1x                                             0.051s   0.05s  0.001s

| BA, n=  100, m=   99, LARGEST, 10000x                                             1.69s   1.68s   0.01s
| BA, n=  100, m=   99, SMALLEST, 10000x                                            2.49s   2.48s  0.012s
| BA, n=  100, m=   99, INDEX, 10000x                                               1.89s   1.88s   0.01s

| BA, n=  100, m=  945, LARGEST, 10000x                                             5.99s   5.93s  0.056s
| BA, n=  100, m=  945, SMALLEST, 10000x                                            6.05s   5.99s  0.058s
| BA, n=  100, m=  945, INDEX, 10000x                                               6.06s   6.01s  0.048s

| BA, n=10000, m= 9999, LARGEST, 1x                                                0.017s  0.017s      0s
| BA, n=10000, m= 9999, SMALLEST, 1x                                               0.024s  0.024s      0s
| BA, n=10000, m= 9999, INDEX, 1x                                                  0.018s  0.017s      0s

| BA, n=10000, m=99945, LARGEST, 1x                                                0.052s  0.051s  0.001s
| BA, n=10000, m=99945, SMALLEST, 1x                                               0.053s  0.053s  0.001s
| BA, n=10000, m=99945, INDEX, 1x                                                   0.06s  0.059s  0.001s
```

Tagl:
> this is after adding a `get_prev()` function for path compression. The benchmark results don't seem too different, but there could be things I'm still missing.

You will probably have to carefully construct a test case that exhibits worst case behavior for something without path compression or careful stitching of the linked lists.
A good choice to test is something like
```
6 6 6 6 .... 6 200000 200000 200000 200000
```
like we discussed in a meeting. Pick smallest first, then you are forced to go from the bucket with large degrees to a bucket with small degrees

szhorvat:
Be sure to do all benchmarking in release mode, no sanitizer. This is the standard benchmarking configuration we use:

image

Keep in mind that very short timings are not reliable. For large graphs, this implementation is much faster than the original 🎉 , yielding timings on the order of milliseconds. This means that to test further improvements, at least the repetition count should be increased for reliable measurements.

----

I ran the fuzzers for a couple of minutes, which gives confidence in the correctness.

----

I came down with a bad flu, so will be slow to respond. Recovering now, but it takes a few days.

src/misc/degree_sequence.cpp
```diff
@@ -68,123 +70,298 @@ template<typename T> inline bool degree_less(const T &a, const T &b) {
...
+    // O(V + alpha(V) * E)
+    // O(V + E) for the LARGEST_FIRST method
+    igraph_vector_int_t seq;
+    IGRAPH_CHECK(igraph_vector_int_init_copy(&seq, degseq));
+    IGRAPH_FINALLY(igraph_vector_int_destroy, &seq);
+
+    HavelHakimiList vault(&seq);
+
+    igraph_integer_t n_edges_added = 0;
+    igraph_vector_int_t spokes; // TODO: allocate spokes vector once
```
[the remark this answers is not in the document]
szhorvat: Already done? It seems to me that the comment is no longer relevant.

szhorvat:
@minifinity Can you please tidy up comments a bit, and get this ready for merging? Give that it's faster than the previous implementation, and works correctly, I'd like to merge this now. Further improvements can go into a new PR.

src/misc/degree_sequence.cpp
```diff
@@ -68,123 +70,302 @@ template<typename T> inline bool degree_less(const T &a, const T &b) {
...
+        insert_bucket(node.degree); // does nothing if already exists
+        buckets[node.degree].nodes.push(vd_pair{node.vertex, node.degree});
+        buckets[node.degree].count++;
     }
 
-    for (const auto &pt : pointers) {
-        vertices.sort(degree_greater<vd_pair>);
+    bool get_max_node(vd_pair& max_node) {
+        igraph_integer_t max_bucket = get_max_bucket();
+        if (max_bucket <= 0) return false;
```
szhorvat: Can you write concise explanatory comments for when `get_max_bucket()` will return 0 or less?

Generally, there should be a short explanatory comment about what each function does, if it is not entirely obvious. This includes `get_max_bucket()` and `get_prev()`, which `get_max_bucket()` calls. This is important for maintainability, as the algorithm is not trivial, and future maintainers will be looking at the code without having a deep understanding.

src/misc/degree_sequence.cpp
```diff
@@ -68,123 +70,298 @@ template<typename T> inline bool degree_less(const T &a, const T &b) {
...
+    // ----- O(1) convenience functions ----- //
+    const BNode & head() const { return buckets.front(); }
+    const BNode & tail() const { return buckets.back(); }
 
-                VECTOR(*edges)[2 * (ec + i)] = vd.vertex;
-                VECTOR(*edges)[2 * (ec + i) + 1] = vertices[i].vertex;
-            }
+    igraph_integer_t get_prev(igraph_integer_t degree) {
+        igraph_integer_t curr = buckets[degree].prev;
+        // might need to add a check to make sure degree > 0
```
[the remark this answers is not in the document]
szhorvat: Why / when is this necessary? If `degree <= 0` should never occur in your current implementation, use `IGRAPH_ASSERT`.

szhorvat:
@minifinity Are you planning to fix the remaining TODO comment in the code?

@ntamas Are you fine with merging this on `master` (after addressing or removing the TODO)?

ntamas:
Are the test failures unrelated?

szhorvat:
Ugh, I may have broken something, will look tomorrow ...

szhorvat:
Should be fixed, but I need to go sleep, long day tomorrow ... I think we should squash-and-merge this, and leave any polish for later.

PR #2788, opened 2025-07-01
Nanoflann and spatial graphs

Zepeacedust:

Add nanoflann as library, and use its K-d trees to write a nearest neighbor graph and distance cutoff graph generators.


include/igraph_spatial.h
```diff
@@ -0,0 +1,44 @@
...
+   GNU General Public License for more details.
+
+   You should have received a copy of the GNU General Public License
+   along with this program; if not, write to the Free Software
+   Foundation, Inc.,  51 Franklin Street, Fifth Floor, Boston, MA
+   02110-1301 USA
+
+*/
+#ifndef IGRAPH_SPATIAL
+#define IGRAPH_SPATIAL
```
szhorvat:
```suggestion
#ifndef IGRAPH_SPATIAL_H
#define IGRAPH_SPATIAL_H
```

(So we don't have to worry about potential conflicts with constants, and also for consistency.)

include/igraph_spatial.h
```diff
@@ -0,0 +1,44 @@
...
+
+#include "igraph_datatype.h"
+#include "igraph_decls.h"
+#include "igraph_types.h"
+#include "igraph_error.h"
+#include "igraph_matrix.h"
+
+typedef enum {
+    IGRAPH_L2_METRIC
```
szhorvat: Let's make this `IGRAPH_METRIC_L2`, which is more autocomplete friendly, and a more logical "most significant first" naming approach (though less grammatical in English).

src/spatial/nearest_neighbor.cpp
```diff
@@ -1,8 +1,59 @@
 
+#include "igraph_operators.h"
+#include "igraph_spatial.h"
 #include "igraph_datatype.h"
+#include "igraph_matrix.h"
 #include "igraph_error.h"
 #include "igraph_types.h"
+#include "igraph_vector.h"
+#include "nanoflann.hpp"
```
szhorvat, line 9: Where do you keep this header now? In the igraph source tree, or are you relying on a system-wide version? I think Arch doesn't have it as a separate package: https://repology.org/project/nanoflann/versions

vendor/nanoflann/nanoflann.hpp
```diff
@@ -0,0 +1,2741 @@
...
+ *
+ *  nanoflann does not require compiling or installing, just an
+ *  #include <nanoflann.hpp> in your code.
+ *
+ *  See:
+ *   - [Online README](https://github.com/jlblancoc/nanoflann)
+ *   - [C++ API documentation](https://jlblancoc.github.io/nanoflann/)
+ */
+
+#pragma once
```
szhorvat, line 45: Please replace this with a classic header guard, i.e.

```cpp
#ifndef NANOFLANN_HPP
#define NANOFLANN_HPP
...
#endif // NANOFLANN_HPP
```

Zepeacedust: My concern is that, since this library is vendored, that you mentioned it being replaced by libraries present in the compiling system. Should we be making changes that would not be reflected in the libraries people would use?

szhorvat: @ntamas Any opinions here?

I am not sure how widely supported `#pragma once` is. This is a minimal change that doesn't really cause much inconvenience when updating nanoflann. I've been doing it in IGraph/M. It saves us the worry that `#pragma once` may break compilation on some esoteric platform where things worked before.

src/spatial/nearest_neighbor.cpp
```diff
@@ -0,0 +1,95 @@
...
+#include <string.h>
+
+class igraph_point_adaptor {
+    igraph_integer_t dimension;
+    const igraph_matrix_t *points;
+    size_t kdtree_get_point_count() const {
+        return igraph_matrix_nrow(points);
+    }
+    igraph_real_t kdtree_get_pt(const size_t idx, const size_t dim) const {
+        return igraph_matrix_get(points, idx, dim);
```
szhorvat: Do not use `matrix_get()` and `vector_get()`, as the performance is bad. Use `MATRIX` and `VECTOR` instead, which are macros that can be inlined.

src/spatial/nearest_neighbor.cpp
```diff
@@ -0,0 +1,95 @@
...
+    private:
+        igraph_integer_t current_vertex = 0;
+        igraph_t *graph;
+    public:
+        using DistanceType = igraph_real_t;
+        GraphBuildingResultSet(igraph_t * graph, igraph_integer_t neighbors, igraph_real_t distance) {
+           this->graph = graph;
+        }
+        bool addPoint(igraph_real_t distance, igraph_integer_t index) {
+            igraph_add_edge(graph, current_vertex, index);
```
szhorvat: Do not call `add_edge()` repeatedly, as it re-builds the graph each time. The time it takes depends on the graph size, and increases as the graph gets bigger.

Collect the edges into a vector, and create the graph using a single `create()` call at the end.

So how do we do that? One way is to `igraph_vector_push_back()` into an igraph vector, but then we need to deal with the return value, which is going to be annoying.  So let's do something else.

Collect results into an `std::vector<igraph_integer_t> v`. The data is guaranteed to be stored contiguously, and you can get a pointer with `v.data()`.  Use this pointer, along with the vector size, to call `igraph_vector_int_view()`, which creates a read-only `igraph_vector_int_t`.  Note that this must not be modified or destroyed, as it shared storage with the `std::vector` now!  The read-only `igraph_vector_int_t` can then be passed to `igraph_create()` to create the final graph.

src/spatial/nearest_neighbor.cpp
```diff
@@ -0,0 +1,95 @@
...
+            return true;
+        }
+        void select_vertex(igraph_integer_t i) {
+            current_vertex = i;
+        }
+};
+
+extern "C" igraph_error_t igraph_nearest_neighbor_graph(igraph_t *graph,
```
szhorvat:
```suggestion
igraph_error_t igraph_nearest_neighbor_graph(igraph_t *graph,
```

The `__BEGIN_DECL` takes care of this.

szhorvat:
Can you please rebase on `develop` and change `__BEGIN_DECLS` -> `IGRAPH_BEGIN_C_DECLS` etc. ? This is a change I made yesterday.

Zepeacedust:
Since being a nearest neighbor of someone is not symmetric, should the function always create a directed graph and let the user see about clobbering it, or should it be set by a flag?

szhorvat:
> Since being a nearest neighbor of someone is not symmetric, should the function always create a directed graph and let the user see about clobbering it, or should it be set by a flag?

Very good observation! Let's add an `igraph_bool_t directed` flag. If true, we produce a directed graph based on the asymmetric relation. If false, we produce an undirected _simple_ graph. Unless you have an idea about how to keep the graph simple efficiently, just use `igraph_simplify()` at the very end, after it's been constructed.

Reasoning: Even though converting to undirected and simplifying could be done by the user, this is a very common use case so we should support it with a flag.

Thus for these points we'd get these two directed/undirected versions:

image

image

Let's orient edges _towards_ nearest neighbours.

szhorvat:
Please don't merge `master` into this branch! That _WILL_ break things.  Only merge `develop`. Please undo the commit merging `master`. Let me know if you need help with that.

vendor/nanoflann/nanoflann.hpp
ntamas: `#pragma once` is quite widely supported now:

https://en.wikipedia.org/wiki/Pragma_once#Portability

The reason why it's not in the C++ standard is because determining semantic equality of files between filesystems is hard (consider having the same file mounted twice on the same machine, once "directly" and once via an NFS share via loopback). Personally, I would just leave `nanoflann` source untouched until we have a far more important reason to start patching it.

szhorvat:
There is an issue with the design of the `GraphBuildingResultSet` class:

 - It has methods that call igraph functions that may fail, without checking errors. Implementing the error check would be complex in this context.
 - It embeds an `igraph_vector_int_t` and uses the finalizer stack to manage it. Now we have TWO conflicting mechanisms to free allocated resources: the C++ destructor and the finalizer stack.

Solution:

 - Collect the results into an `std::vector<igraph_integer_t> edges`
 - When done, use `igraph_vector_int_view()` to create a read-only `igraph_vector_int_t` from `edges.data()` that you can pass to `igraph_create()`.

Additionally, in the spirit of decoupling C++-specific constructs from C-style igraph code, do not create the graph within the class, and do not hold a reference to the graph within the class. This is unnecessary. Make the vector accessible (in a read-only manner) and use it _outside of the class_ to build the graph.

Generally, let us try to separate C++ code that relies on exceptions from igraph-style code that relies on the finalizer stack, as it will be necessary to have a sort of boundary between them and appropriate convert between exceptions / error codes there.

Zepeacedust:
I have swapped over to the builtin RKNNResultset, which would fix that issue, but that seems to only support open ranges.

Is that sufficient, or is it important to be able to have closed ranges for distances?

szhorvat:
We should definitely have our own result set class to get full control. This is not because of the open/closed range issue, just for flexibility. The result set class should collect into a class igraph-format edge list in from1, to1, from2, to2, ... format, stored in an `std::vector`. The reason for `std::vector` is that we won't have to think about error handling.  Once we have the vector, we wrap its internal storage into an `igraph_vector_int_t` using `igraph_vector_int_view()` and call `igraph_create()` to produce the graph.

Zepeacedust:
I added the function to what I remember of the documentation and interfaces.
I'm not quite sure about the interfaces declarations, but they seem correct to me.
I added the spatial functionality as a separate chapter for now, since I think we will be expanding it shortly.
Other than that, there is still the matter of finding the problems with the arbitrary dimension case, which doesn't seem ti be all that pressing for now, it currently returns unimplemented.

szhorvat:
@ntamas Any preference whether we should name the metrics `IGRAPH_METRIC_L2` / `IGRAPH_METRIC_L1` or `IGRAPH_METRIC_EUCLIDEAN` / `IGRAPH_METRIC_MANHATTAN`? I have a slight preference for the latter.

szhorvat:
@Zepeacedust, let's do the following:

- [ ] Add the Fibonacci spiral test case I sent you in chat. Generate the expected output with Mathematica.
- [ ] Sections of the test output should be visually well separated, e.g. by including a newline (add `\n` at the beginning of strings you print).
- [ ] Ensure that `cutoff < 0` is interpreted as `cutoff == IGRAPH_INFINITY`. Just rewrite negative values.
- [ ] Make sure that the null graph is handled properly, _even if the matrix has zero columns_. This may happen due to the "list of lists" representation in some languages, so we shouldn't error for zero columns when there are zero rows.
- [ ] Finalize Manhattan distance support, including tests.
- [ ] Write a small benchmark.
- [ ] When the benchmark is done, investigate whether using `L2_Simple_Adaptor` instead of `L2_Adaptor` for lower dimensions brings an appreciable speedup.
- [ ] Add overflow checks: When moving data from the `std::vector` to the `igraph_vector_int_t`, check if the size is `<= 2 * IGRAPH_ECOUNT_MAX`. If not, throw an `IGRAPH_EOVERFLOW`.
- [ ] Include this change in our copy of nanoflann: https://github.com/jlblancoc/nanoflann/pull/263
- [ ] I am not in favour of using the finalizer stack and complex error handling in tests, as it is not truly necessary there and reduces maintainability. @ntamas, any opinion here?

You can make a copy of this checklist (copy Markdown sources) in a post of your own, which you will be able to edit (i.e. check boxes as you go).

Is there anything else that you think we need to do here? Mark as "ready for review" and be sure to also ping me when you think it is fully ready to merge.

Check the wiki for guidelines:

 - https://github.com/igraph/igraph/wiki
 - Especially this: https://github.com/igraph/igraph/wiki/Checklist-for-new-%28and-old%29-functions
 - And these two: https://github.com/igraph/igraph/wiki/Documentation-guidelines and https://github.com/igraph/igraph/wiki/Error-reporting-guidelines

Tip: `<` and `<=` are not particularly meaningful with floating point numbers because of roundoff errors. Once more tests are added, you will likely need to make use of `igraph_cmp_epsilon()` or `igraph_almost_equals()` with an appropriate tolerance in any `IGRAPH_ASSERT`. This is not an issue when printing floating point numbers, as not all digits are printed. The precision of a standard `double` is just a little worse than `1e-16`, so tolerances as tight as `1e-14` or `1e-13` will usually work.  Expect some breakages in CI, which exercises igrahp with different floating point systems (64-bit SSE and 80-bit x87).

Zepeacedust:
* [x]  Sections of the test output should be visually well separated, e.g. by including a newline (add `\n` at the beginning of strings you print).

* [x]  Add the Fibonacci spiral test case I sent you in chat. Generate the expected output with Mathematica.

* [x]  Ensure that `cutoff < 0` is interpreted as `cutoff == IGRAPH_INFINITY`. Just rewrite negative values.

* [x]  Make sure that the null graph is handled properly, _even if the matrix has zero columns_. This may happen due to the "list of lists" representation in some languages, so we shouldn't error for zero columns when there are zero rows.

* [x]  Finalize Manhattan distance support, including tests.

* [x]  Write a small benchmark.

* [x]  When the benchmark is done, investigate whether using `L2_Simple_Adaptor` instead of `L2_Adaptor` for lower dimensions brings an appreciable speedup.

* [x]  Add overflow checks: When moving data from the `std::vector` to the `igraph_vector_int_t`, check if the size is `<= 2 * IGRAPH_ECOUNT_MAX`. If not, throw an `IGRAPH_EOVERFLOW`.

* [x]  Include this change in our copy of nanoflann: [do not write to stderr jlblancoc/nanoflann#263](https://github.com/jlblancoc/nanoflann/pull/263)

* [ ]  I am not in favour of using the finalizer stack and complex error handling in tests, as it is not truly necessary there and reduces maintainability. @ntamas, any opinion here?

Copied for easier checking

Zepeacedust:
On the finally stack in the tests, I use it because I like having a single function that sets up the environment and cleans up after itself, and since I use the same function to check for cases that should error, that requires that the finally stack is used to clean up after itself.

It would not be that difficult of a change to have a separate function to check for errors that does the cleanup without using the stack, but I feel that since the stack is already there and used for this purpose in the rest of the codebase that it makes sense to use it here aswell.

src/spatial/nearest_neighbor.cpp
```diff
@@ -0,0 +1,277 @@
...
+        igraph_integer_t dimension,
+        igraph_bool_t directed) {
+
+    switch (dimension) {
+    case 0:
+        if (igraph_matrix_nrow(points) == 0) {
+            // null matrix, should not error
+            igraph_empty(graph, 0, true);
+            return IGRAPH_SUCCESS;
+        }
```
szhorvat: There are some issues here:

 - Return value is not checked.
 - `directed` parameter is not respected.

We usually handle edge cases early on, in the top level function, so we don't have to think about them anymore. I'll fix this.

szhorvat: There are some issues here:

 - Return value is not checked.
 - `directed` parameter is not respected.

We usually handle edge cases early on, in the top level function, so we don't have to think about them anymore. I'll fix this.

ntamas:
> @ntamas Any preference whether we should name the metrics `IGRAPH_METRIC_L2` / `IGRAPH_METRIC_L1` or `IGRAPH_METRIC_EUCLIDEAN` / `IGRAPH_METRIC_MANHATTAN`? I have a slight preference for the latter.

`IGRAPH_METRIC_EUCLIDEAN` and `IGRAPH_METRIC_MANHATTAN` are preferred, but note that you can always add L1 and L2 as aliases in the enum if you want to.

ntamas:
> I am not in favour of using the finalizer stack and complex error handling in tests, as it is not truly necessary there and reduces maintainability.

I don't have any problems with that; feel free to go ahead as is.

src/spatial/nearest_neighbor.cpp
```diff
@@ -0,0 +1,281 @@
...
+    tree.buildIndex();
+
+    igraph_vector_t current_point;
+    IGRAPH_VECTOR_INIT_FINALLY(&current_point, dimension);
+
+    igraph_integer_t neighbor_count = neighbors >= 0 ? neighbors : point_count;
+
+    GraphBuildingResultSet results(neighbor_count, cutoff);
+    std::vector<igraph_integer_t> neighbor_set(neighbor_count);
+    std::vector<igraph_real_t>    distances(neighbor_count);
```
szhorvat: Is there any reason these two vectors can't be made part of the result class? Conceptually, they _are_ the result.  They can still be exposed to be accessible outside the class (even just made public members).

Would it simplify the code if, instead of keeping a fixed-length vector and keeping track of the actually used number of elements in `current_added`, we allowed the vector size to actually change?  After the required amount of storage is reserved for an std::vector, there isn't a cost to shrinking it and then re-growing it again. Internally, it will use the same allocated memory.

src/spatial/nearest_neighbor.cpp
```diff
@@ -0,0 +1,281 @@
...
+    std::vector<igraph_integer_t> edges;
+    for (igraph_integer_t i = 0; i < point_count; i++) {
+
+        std::fill(neighbor_set.begin(), neighbor_set.end(), 0);
+        std::fill(distances.begin(), distances.end(), 0);
```
szhorvat: Do we actually need to null these here?

src/spatial/nearest_neighbor.cpp
```diff
@@ -0,0 +1,281 @@
...
+
+        results.init(distances.data(), neighbor_set.data(), i);
```
szhorvat: `distances` and `neighbor_set` are accessed in the result class through their `.data()` pointer. Since we're writing C++, is there any reason you don't wan to access them in the standard way as `std::vector`s, using `[]`?

src/spatial/nearest_neighbor.cpp
szhorvat: No.

src/spatial/nearest_neighbor.cpp
szhorvat: Done.

src/spatial/nearest_neighbor.cpp
szhorvat: Vectors are made part of the class, but `current_added` is kept.

szhorvat:
@ntamas, please review but do not merge.

@Zepeacedust Please check the changes I made carefully, both for future reference, as well as to make sure nothing got messed up.

Questions:

 - Do we want to care about what happens when there are multiple points at exactly the same distance? I think we can skip this for now.
 - @Zepeacedust Do you want to try to do some performance tuning on max_leaf_size, or have you already look at this? This can also be done at a later time.

I think this is basically good to go.

Zepeacedust:
I tinkered with the leaf size a little, but it didn't seem to be detectable with the benchmarks I had.
I think that is better left to an optimization pass by someone better versed in that sort of thing than me, or for later when I have time to read up on it.

src/spatial/nearest_neighbor.cpp
```diff
@@ -0,0 +1,281 @@
...
+                    points,
+                    k,
+                    cutoff,
+                    dimension,
+                    directed);
+        default:
+            IGRAPH_ERROR("Invalid metric.", IGRAPH_EINVAL);
+    }
+
+IGRAPH_HANDLE_EXCEPTIONS_END;
```
ntamas:
```suggestion
    IGRAPH_HANDLE_EXCEPTIONS_END;
```

PR #2793, opened 2025-07-06
feat: rooted and modular graph products

### LINKED ISSUE #2771: Adding More Graph Products in igraph_product
#2748 added the Cartesian and the Tensor product. We should add more graph products not limited to:
- [ ] [Strong Product](https://en.wikipedia.org/wiki/Strong_product_of_graphs)
- [ ] [Lexicographical Product](https://en.wikipedia.org/wiki/Lexicographical_product_of_graphs)
- [ ] [Modular Product](https://en.wikipedia.org/wiki/Modular_product_of_graphs)
- [ ] [Rooted Product](https://en.wikipedia.org/wiki/Rooted_product_of_graphs)
- [ ] [Modular Product](https://en.wikipedia.org/wiki/Modular_product_of_graphs)
- [ ] [Corona Product](https://en.wikipedia.org/wiki/Corona_product)

**Use cases for the feature**
Useful in research and teachings

**References**
- https://en.wikipedia.org/wiki/Graph_product
- https://networkx.org/documentation/stable/reference/algorithms/operators.html
- https://reference.wolfram.com/language/ref/GraphProduct.html

gulshan-123:

Implemented:
- [x] Rooted Product
This is the paper which introduced the Rooted Product: https://doi.org/10.1017/S0004972700007760. `All graphs considered in this paper are finite, and without loops and multiple or directed edges.`
Rooted product is a subgraph of cartesian product and the implementation is similar to other graph products, handling digraphs, loops as well as multigraphs.
The logic is taken from [NetworkX](https://networkx.org/documentation/stable/_modules/networkx/algorithms/operators/product.html#rooted_product) implementation.

Function Definition: (different from igraph_product, since need a root vertex)
```cpp
igraph_error_t igraph_rooted_product(igraph_t *res,
                              const igraph_t *g1,
                              const igraph_t *g2,
                              const igraph_integer_t root)
 ```
- [x] Modular product:
See: https://en.wikipedia.org/wiki/Graph_product#Overview_table
We restrict this function only for simple graphs.
There are two conditions of adjacency, the first condition is same as the tensor product.
The second condition is equivalent to the first condition but on the complement graphs.

- [ ] Corona Product: The product has v1.v2 + v1 vertex which differs from the other graph products which has v1.v2 vertices only.
---
TODO: 
- [x] error if root not in g2. Add more tests, add docs. check behaviour for multigraphs
- [ ] docs: modular product
- [ ] docs: corona product: write about the vertex number specifically
---
Fixes #2771

### DIFF
--- a/src/operators/products.c
+++ b/src/operators/products.c
@@ -497,7 +497,7 @@ static igraph_error_t modular_product(igraph_t *res,
  * </para><para>
  * Reference:
  *
-  * </para><para>
+ * </para><para>
  * Richard Hammack, Wilfried Imrich, and Sandi Klavžar (2011).
  * Handbook of Product Graphs (2nd ed.). CRC Press.
  * https://doi.org/10.1201/b10959


szhorvat:
Thanks @gulshan-123 ! I'll going to be travelling for a while and will probably come back to this around the 18th of July.

szhorvat:
Just a heads up: When you get to adding new enum values in this PR, please do not reorder them, as code has been added to the R interface that (unfortunately) encodes these are plain integers. This would need to be updated after any reordering. 

I'm actually going to open an issue on how to avoid specifying these are integers, which is fragile and prone to breakage.

gulshan-123:
(Tests are failing due to only boilerplate has been added for Modular Product)
@szhorvat, This is what I did & am planning to do. Give Ideas.
1. Rooted graph product is implemented for directed, multigraphs as well as having loops.

3. The definition of [Corona Product](https://en.wikipedia.org/wiki/Corona_product), we creates several copies of second graph and then add edges from first graph. 
- If the graphs are directed, then we will add directed edges from first graph (central graph) to second graph. (Here we may add some kind of `type` parameter too). 
- For multigraphs, only one new edge between two vertices of different graphs will be added (No multiplicity introduced) .
NOTE: The number of vertices will become v1.v2 + v1, differing from general graph products, having v1.v2 vertices (we'll have to handle in docs).

5. [Modular Products](https://en.wikipedia.org/wiki/Modular_product_of_graphs): 
- [The first set of adjacency condition](https://en.wikipedia.org/wiki/Graph_product#Overview_table) is similar to the Tensor product. 
- But the second set will be handled by complementing both the graphs (excluding loops construction i.e. `loop=false` in `igraph_complementer()`) and then applying the first adjacency condition on complement graphs. 
- I believe, this should not be implemented for multigraphs, as complement graph will not have multiple edges?? 
- However, graphs having loops may be considered. Directed graphs can be handled.
---
NOTE: In all of above, [NetworkX](https://networkx.org/documentation/stable/_modules/networkx/algorithms/operators/product.html) has neither implemented for digraphs nor for multigraphs.

szhorvat:
I'll be able to look into this in ~10 days, just mentioning that if directed or multigraph support is problematic, it is always an option to either return `IGRAPH_UNIMPLEMENTED` (leaving it for the future as you're not sure yet) or `IGRAPH_EINVAL` (when you know it makes no sense).

szhorvat:
I am able to review PRs again. Let me know when you need input @gulshan-123

gulshan-123:
I was busy due to internship season going on. Hope I will be free this week! @szhorvat

gulshan-123:
@szhorvat, I am now ready to work on this!

 https://en.wikipedia.org/wiki/Graph_product#Overview_table

(a1​,a2​)∼(b1​,b2​)
image

For modular product:
image

- Condition 1 matches exactly with the tensor product of the original graphs.

- Condition 2 is identical to condition 1 but applied to the complements of graph 1 and graph 2. ([networkx](https://github.com/networkx/networkx/blob/2573bd3b769e560e070bfaef2d77887c21a28821/networkx/algorithms/operators/product.py#L625))

---
Should we implement the modular product by internally computing the tensor product twice:
- Once on the original graphs,
- Once on their complements,

...and then take the union of these two resulting graphs to obtain the final modular product?

szhorvat:
We could do that as a first implementation, then improve on it later. This is going to be inefficient either way due to the huge number of edges that are generated.

I suggest we explicitly forbid non-simple graphs for this product type, as the concept of the complement graph is problematic for these.

----

Thanks for continuing to work on this! I wanted to let you know that any future PRs after this one should go on `develop`, as we're planning a major release within a month.

gulshan-123:
@szhorvat, Added the rooted and modular product.
Modular product is only implemented for simple graphs, and internally uses tensor product twice as discussed above.

For corona products:
image

1. If the graphs are directed, then the new edges from central graph to the outer graph may have two possible direction. Should we add a `type` parameter to decide the directionality?
2. Total number of vertex is v1.(v2+1), differing from the other products which have v1.v2.

I think a separate new functions for corona product may be added which will have a `type` parameter (separate function is also for rooted product). we will write about new vertex number and indices in its docs. I think this will also keep docs clear. ??

szhorvat:
Let's leave the corona product for a future PR, as I'd like to merge this ASAP.

There's an issue with GitHub's PRs at the moment, and I can't add the following comment directly on the diff. This was meant to be added on the destructor call sequence of `modular_product`:

> 
> When these graphs are created, we need to add an `IGRAPH_FINALLY(igraph_destroy, &the_graph)` for each, then we need the `IGRAPH_FINALLY_CLEAN` here.
> 
> For cosmetic and readability (think future refactor) reasons, let's reorder destructor calls in the reverse order of construction. Why readability? This order will make it clear why it's a bit difficult to to destroy the complement graphs before the `igraph_union()`, which I was going to suggest.

szhorvat:
@gulshan-123 Can you please compute the documentation of the modular product? Then I'll merge this, and merge `master` into `develop`. After that you can get started with the corona product in a new PR targeted at the `develop` branch.

gulshan-123:
@szhorvat, added the docs for modular product

szhorvat:
@gulshan-123 Thank you! Please target any future PRs (such the corona product) on the `develop` branch for now.

PR #2799, opened 2025-07-18
Fix #2798: Correct termination logic in leiden community detection loop

### LINKED ISSUE #2798: Logical error in `igraph_community_leiden` loop for negative `n_iterations`
I believe I've found a logical error in the main iteration loop of the `igraph_community_leiden` function, located in `src/community/leiden.c`.

The issue occurs when the function is called with a negative value for `n_iterations`, which is the documented condition to iterate until the community structure converges (i.e., no longer changes).

### The Problem

The current implementation terminates the loop after the *first* iteration that results in a change, rather than continuing until an iteration results in *no change*.

The current code is:

```c
igraph_bool_t changed = false;
for (igraph_integer_t itr = 0;
        n_iterations >= 0 ? itr < n_iterations : !changed;
        itr++) {
    IGRAPH_CHECK(igraph_i_community_leiden(graph, i_edge_weights, i_node_weights,
                                            resolution_parameter, beta,
                                            membership, nb_clusters, quality, &changed));
}
```

The faulty logic proceeds as follows when `n_iterations` is negative:

1.  `changed` is initialized to `false` ([line 1034](https://github.com/igraph/igraph/blob/f51ea4f217a45f5895e749e7ea919040492cab5e/src/community/leiden.c#L1034)).
2.  The loop condition is `!changed`, which is `true` ([line 1036](https://github.com/igraph/igraph/blob/f51ea4f217a45f5895e749e7ea919040492cab5e/src/community/leiden.c#L1036)), so the loop runs once.
3.  The inner function `igraph_i_community_leiden` runs and finds an improvement, setting the `changed` flag to `true` ([line 180](https://github.com/igraph/igraph/blob/f51ea4f217a45f5895e749e7ea919040492cab5e/src/community/leiden.c#L180)).
4.  On the next check, the loop condition `!changed` becomes `false` ([line 1036](https://github.com/igraph/igraph/blob/f51ea4f217a45f5895e749e7ea919040492cab5e/src/community/leiden.c#L1036)), and the loop incorrectly terminates, even though further improvements might be possible.

### Proposed Solution

The fix involves initializing `changed` to `true` and altering the loop condition to check for `changed` directly. This ensures the loop continues as long as changes are being made.

The proposed corrected code is:

```c
igraph_bool_t changed = true;
for (igraph_integer_t itr = 0;
        n_iterations < 0 ? changed : itr < n_iterations;
        itr++) {
    IGRAPH_CHECK(igraph_i_community_leiden(graph, i_edge_weights, i_node_weights,
                                            resolution_parameter, beta,
                                            membership, nb_clusters, quality, &changed));
}
```

This corrected logic works as follows when `n_iterations` is negative:

1.  `changed` is initialized to `true`, ensuring the loop runs at least once.
2.  The loop condition is `changed`.
3.  After an iteration, if there was a change, `changed` will be `true` and the loop will continue.
4.  If an iteration results in no change, the inner function will set `changed` to `false` ([line 747](https://github.com/igraph/igraph/blob/f51ea4f217a45f5895e749e7ea919040492cab5e/src/community/leiden.c#L747)), and the loop will correctly terminate on the next check.

This change correctly implements the intended "iterate until convergence" behavior while preserving the logic for a fixed number of iterations.

lucaslopes:
The loop for negative n_iterations was terminating prematurely. This fix #2798, ensuring it iterates until convergence as intended.

### DIFF
--- a/src/community/leiden.c
+++ b/src/community/leiden.c
@@ -1031,9 +1031,9 @@ igraph_error_t igraph_community_leiden(const igraph_t *graph,
      * iteration may still find some improvement. This is because
      * each iteration explores different subsets of nodes.
      */
-    igraph_bool_t changed = false;
+    igraph_bool_t changed = true;
     for (igraph_integer_t itr = 0;
-         n_iterations >= 0 ? itr < n_iterations : !changed;
+         n_iterations < 0 ? changed : itr < n_iterations;
          itr++) {
         IGRAPH_CHECK(igraph_i_community_leiden(graph, i_edge_weights, i_node_weights,
                                                resolution_parameter, beta,


szhorvat:
I verified this and will merge it. @vtraag, please take a look nevertheless when you're back from holidays.

Thanks @lucaslopes!

vtraag:
Yes, well-spotted, and indeed correct, thanks @lucaslopes!

PR #2801, opened 2025-07-21
Improvements and documentation for `igraph_setup()`

szhorvat:
@ntamas, the three commits can be reviewed separately.


szhorvat:
@ntamas It seems that there is an issue and the seed generation is never run as the default RNG starts out as "seeded". Did you do this on purpose or is it accidental?

ntamas:
Not 100% sure whether I want to make the existence of `igraph_setup_rng()` public - otherwise it's okay. How about not exposing `igraph_setup_rng()` in the docs and keeping the function itself private (i.e. not exported)?

szhorvat:
> Not 100% sure whether I want to make the existence of `igraph_setup_rng()` public

I agree with this, I'd rather have a single setup function. I'll refactor it within this PR.

Can you please look at why the seed generator is never run, as seen in the coverage report? It appears as if the RNG starts out as fully seeded.

szhorvat:
@ntamas I think this is good to go but be sure to look at thenew commits I added.

PR #2804, opened 2025-07-28
Clean up and finalize Infomap update

szhorvat:


szhorvat:
@danieledler I see a note in the old igraph Infomap docs that there should be no isolated vertices in the graph. Do you see any reason why isolated vertices should still be avoided? Does the Infomap library handle them now?

szhorvat:
@danieledler I also need some help with proper input validation. I assume that the vertex and edge weights are not allowed to be negative, NaN or infinite. Are they allowed to be `0.0`?

Is there anything else that we should validate?  Are multi-edges supported?  How about self-loops?

If self-loops are supported, is an undirected self-loop treated as two reciprocal directed ones? In other words, assume a graph with these connections: `1 - 2 - 3, 2 - 2`. Suppose we have a random walker sitting in node 2. What is the probability that it moved to node 1? Is it 1/4 (i.e. self-loops are traversable in both directions, an adjacency matrix would have 2 on the diagonal, as standard), or is it 1/3 (undirected self-loops can only be traversed in one direction, non-standard)?

szhorvat:
@danieledler What is the suggested default for the number of trials?

szhorvat:
@danieledler Would you be willing to accept the interruption support, added earlier by @vtraag, upstream, to simplify maintenance in igraph?

szhorvat:
> @danieledler I see a note in the old igraph Infomap docs that there should be no isolated vertices in the graph. Do you see any reason why isolated vertices should still be avoided? Does the Infomap library handle them now?

According to tests I did, isolated vertices seem to be handled properly.



> Is there anything else that we should validate? Are multi-edges supported? How about self-loops?
> 
> If self-loops are supported, is an undirected self-loop treated as two reciprocal directed ones? In other words, assume a graph with these connections: `1 - 2 - 3, 2 - 2`. Suppose we have a random walker sitting in node 2. What is the probability that it moved to node 1? Is it 1/4 (i.e. self-loops are traversable in both directions, an adjacency matrix would have 2 on the diagonal, as standard), or is it 1/3 (undirected self-loops can only be traversed in one direction, non-standard)?

I also tested this, and it seems that the weights of multi edges are effectively added up (as we want them to be), self-loops are taken into account, and undirected self-loops are equivalent to two (i.e. "reciprocal") directed self loops (again what we want, i.e. 2s on the adjacency matrix diagonal).

----

@ntamas I think we can merge this now. Please review.

@danieledler, it's still useful if you can answer the remaining questions from above. We can make the necessary adaptations in future PRs.

szhorvat:
Ugh, MSVC chokes on the OpenMP code because apparently it only supports OpenMP 2.0 which requires _signed_ iterator types in for loops.

```
D:\a\1\s\vendor\infomap\src\core\InfomapBase.cpp(1854): error C3016: 'moduleIndex': index variable in OpenMP 'for' statement must have signed integral type
```

@ntamas, could you please help me here? What I suggest is to save the OpenMP version, as detected by CMake, and enable OpenMP for Infomap only if the version is at least 3.  Somewhere we must have a `find_package(OpenMP)`. This sets `OpenMP_VERSION`: https://cmake.org/cmake/help/latest/module/FindOpenMP.html  We need to somehow use that in [vendor/infomap/CMakeLists.txt](https://github.com/igraph/igraph/pull/2804/files#diff-176616fe60f2f9f2bf0053fcc36e2a90e94d9cd861e37dcdcc0697b76d3aa27e)

It's also unclear if this affects igraph's CMake package file.

It's a bit painful for me to google for docs here due to internet speeds ...

ntamas (approved):

ntamas:
@szhorvat Loop indices used in OpenMP-parallelized loop (i.e. `moduleIndex` in `InfomapBase.cpp` must be signed integers on Windows. Apparently it's not a problem for other platforms but this is why the tests are failing on Windows now. Can you take care of this?

PR #2805, opened 2025-07-28
chore: Add deps

krlmlr:


krlmlr:
For some reason I was allowed to merge here, sorry about that -- meant to do a review first. Explanation follows.

krlmlr:
Many entries in `functions-R.yaml` only extend the default definition with `DEPS` . Do they do any harm over here? Do they apply to other interfaces? Are you open in porting those deps here, also for other functions? We could also check in Stimulus that certain types _need_ a dependency.

Might have missed an existing discussion.

My trigger: breaking vendoring workflow, https://github.com/igraph/rigraph/actions/runs/16577296929/job/46884629275 . Easy enough to fix, might be more sustainable if treated here already.

Again: if this is wrong, please just revert.

szhorvat:
To me this all makes sense: Add DEPS here rather than in the R repo, and add an extra check to the Stimulus tests to ensure that deps are present.

But what this also affects is the new Python interface, so it's @ntamas who can respond. I don't know if the new Python interface uses these dependencies.

ntamas:
I'll take care of this in the Python interface if needed.

PR #2806, opened 2025-07-30
Delaunay triangulation with QHull

Zepeacedust:

This pull request will add support delaunay triangulation by using qhull.

----

TODO:

 - [x] Pick up fixes from https://github.com/qhull/qhull/issues/155 (but let's wait for a PR to Qhull first)
 - [x] Pick up https://github.com/qhull/qhull/pull/164


szhorvat:
@Zepeacedust Please resolve conflict and update this to the latest `develop`. That will allow CI checks to run again.

If we don't use C++ after all, let's remove the files, to make sure that we won't accidentally end up with multiple symbols of the same name in the object files.

szhorvat:
@Zepeacedust Before you start modifying Qhull significantly, please ensure that that you took the _latest_ sources from the Qhull repo (not just the latest release, but the latest commit).

Regarding the format-specifier bugs from above (which can cause a crash!), there seems to be a PR in the Qhull repo that fixes these. I asked about merging it. If we don't get a response, let's merge the changes from there into our version.

src/spatial/delaunay.c
```diff
@@ -0,0 +1,214 @@
...
+        IGRAPH_FINALLY_CLEAN(3);
+    } else {
+        qh->NOerrexit = True; /* no more setjmp */
+        qh_freeqhull(qh, !qh_ALL);
+        qh_memfreeshort(qh, &curlong, &totlong);
+
+        char error[100];
+
+        snprintf(error, sizeof(error) / sizeof(error[0]), "Error while computing delaunay triangulation, qhull code = %i", qh->last_errcode);
+        IGRAPH_ERROR("bruh &error[0]", IGRAPH_EINVAL);
```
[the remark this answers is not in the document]
szhorvat: You can `IGRAPH_ERRORF("Something %s.", IGRAPH_FAILURE, error);` or similar. Use `IGRAPH_FAILURE` for Qhull errors. Make the error message as clear as we can manage, given the raw material we're getting from Qhull.

src/spatial/delaunay.c
```diff
@@ -0,0 +1,214 @@
...
+    for (igraph_integer_t i = 2; i < size; i += 2) {
+        if ((VECTOR(*in)[i] != VECTOR(*in)[i - 2] || VECTOR(*in)[i + 1] != VECTOR(*in)[i - 1]) && VECTOR(*in)[i+1] != except) {
+            igraph_vector_int_push_back(out, VECTOR(*in)[i]);
+            igraph_vector_int_push_back(out, VECTOR(*in)[i + 1]);
+        }
+    }
+}
+
+igraph_error_t igraph_i_delaunay_edges(igraph_vector_int_t *edges, igraph_matrix_t *points) {
+    int curlong, totlong; /* used !qh_NOmem */
```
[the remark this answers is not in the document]
szhorvat: Copilot is given instructions to check that people use `igraph_integer_t` where appropriate.

This situation here is special as Qhull itself uses `int`s. Don't take Copilot too literally.

When passing points to Qhull for the first time, do check that the point count doesn't exceed `INT_MAX`. Example check: 

https://github.com/igraph/igraph/blob/5dc266fb8a3adabf3ab3788e30876a39ec27b779/src/community/infomap.cpp#L64-L66

szhorvat:
Re the macOS failures: either use `IGRAPH_PRId` as the format specifier or, if you know that values are small enough, cast to the appropriate type. Please do not ever use `long` as it's width is inconsistent between platforms. Either go with `IGRAPH_PRId` or cast down to `int` if safe (in tests only!)

If you go with `int`, please do not use `%i`, use `%d` for consistency.

src/spatial/delaunay.c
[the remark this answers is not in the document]
Zepeacedust: Scipy just passes the entire qhull error to the user, doing something similar might be a good idea.

Otherwise we are stuck between providing unhelpful error codes that qhull only just about documents, or making special cases for the errors to report them in a more insightful manner.

A middle ground could be practical, where we identify some common errors (all points on a line) and report them well, and then give instructions in the documentation for the others so we would not be swamped with help requests regarding the qhull codes.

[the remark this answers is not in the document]
szhorvat: Let's be practical and make this _useful_ for now instead of _pretty_. So pass the ugly Qhull error back. 

Something like:

```
IGRAPH_ERRORF("Qhull returned with error: %s.", IGRAPH_FAILURE, error);
```

You could even have a separate function for passing Qhull errors up, which will make it easy to prettify them later. But let's not spend time on that now.

vendor/qhull/CMakeLists.txt
```diff
@@ -0,0 +1,44 @@
...
+  libqhull_r/poly2_r.c
+  libqhull_r/qset_r.c
+  libqhull_r/random_r.c
+  libqhull_r/rboxlib_r.c
+  libqhull_r/stat_r.c
+  libqhull_r/user_r.c
+  libqhull_r/usermem_r.c
+  libqhull_r/userprintf_r.c
+  libqhull_r/userprintf_rbox_r.c
+
```
szhorvat:
```suggestion
```

src/spatial/delaunay.c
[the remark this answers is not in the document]
Zepeacedust: Added a size check.

src/spatial/delaunay.c
```diff
@@ -0,0 +1,212 @@
...
+static int edge_comparator(const void *a, const void *b) {
+    igraph_integer_t * A = (igraph_integer_t *)a;
+    igraph_integer_t * B = (igraph_integer_t *)b;
+    if (A[0] < B[0]) {
+        return -1;
+    }
+    if (A[0] > B[0]) {
+        return  1;
+    }
+    //first are equal
```
szhorvat:
```suggestion
    // first are equal
```

keep comments neat and consistent, add that space ;-)

src/spatial/delaunay.c
```diff
@@ -0,0 +1,212 @@
...
+            temp = VECTOR(*in)[i];
+            VECTOR(*in)[i] = VECTOR(*in)[i+1];
+            VECTOR(*in)[i+1] = temp;
+        }
+    }
+    // sort
+    igraph_qsort(VECTOR(*in), size / 2, 2 * sizeof(igraph_integer_t), &edge_comparator);
+
+    igraph_vector_int_push_back(out, VECTOR(*in)[0]);
+    igraph_vector_int_push_back(out, VECTOR(*in)[1]);
```
szhorvat: Either `IGRAPH_CHECK` or explain in a comment why not needed (reserved memory already?)

src/spatial/delaunay.c
```diff
@@ -0,0 +1,212 @@
...
+    // nub and filter out point at infinity
+    for (igraph_integer_t i = 2; i < size; i += 2) {
+        if ((VECTOR(*in)[i] != VECTOR(*in)[i - 2] || VECTOR(*in)[i + 1] != VECTOR(*in)[i - 1]) && VECTOR(*in)[i+1] != except) {
+            igraph_vector_int_push_back(out, VECTOR(*in)[i]);
+            igraph_vector_int_push_back(out, VECTOR(*in)[i + 1]);
+        }
+    }
+}
+
+igraph_error_t igraph_i_delaunay_edges(igraph_vector_int_t *edges, igraph_matrix_t *points) {
```
szhorvat: `points` must be `const`, don't mess up the points passed by the user!

You need extra memory anyway to copy to row-major, we can fix it there.  You can do the transposition _while_ copying in a manual loop.

Recommendation: iterate row-wise while copying.

src/spatial/delaunay.c
```diff
@@ -0,0 +1,212 @@
...
+    igraph_integer_t dim = igraph_matrix_ncol(points);
+    boolT ismalloc = False; // handle memory allocation of points explicitly
+    qhT qh_qh;
+    qhT *qh = &qh_qh;
+
+    igraph_vector_int_t int_edges;
+    igraph_vector_t int_points;
+
+    if (numpoints > INT_MAX) {
+        IGRAPH_ERROR("Too many points for Qhull.", IGRAPH_EOVERFLOW);
```
szhorvat, line 145: I already fix this where I saw it, but noting it for the future:

 - Let's keep messages consistent. Let's spell Qhull with a capital in all user-facing text (messages, docs).
 - Messages must end in punctuation (add that fullstop).

src/spatial/delaunay.c
```diff
@@ -0,0 +1,212 @@
...
+    }
+
+    // get in row-major order
+    IGRAPH_CHECK(igraph_matrix_transpose(points));
+
+    // allocate space for extra point added due to
```
szhorvat: Due to what? Complete comment.

src/spatial/delaunay.c
```diff
@@ -0,0 +1,212 @@
...
+    IGRAPH_VECTOR_INIT_FINALLY(&int_points, numpoints * dim);
+
+    igraph_matrix_copy_to(points, VECTOR(int_points));
```
szhorvat: Let's avoid hacks where data is copied into a vector/matrix using its raw pointer, if we can use structured vector/matrix functions instead.

src/spatial/delaunay.c
```diff
@@ -0,0 +1,212 @@
...
+        qh->SCALElast = True;   /* 'Qbb' */
+        qh->KEEPcoplanar = True; /* 'Qc', to keep coplanars in 'p' */
+        qh->TRIangulate = True;
+        qh_checkflags(qh, qh->qhull_command, "  ");
+        qh_initflags(qh, qh->qhull_command);
+        qh->ATinfinity = True;
+        qh->MERGEvertices = False;
+        qh_init_B(qh, VECTOR(int_points), numpoints, dim, ismalloc); // read points and fiddle with them a little.
+        qh_qhull(qh);
+        qh_triangulate(qh);
```
szhorvat: Explain in a comment why this is needed, for future maintainers.

src/spatial/delaunay.c
```diff
@@ -0,0 +1,212 @@
...
+ * The current implementation uses Qhull.
+ *
+ * \param graph A pointer to the graph that will be created.
+ * \param points A matrix containing the points that will be used to create the graph.
+ *     Each row is a point, dimensionality is inferred from the column count.
+ *
+ * \return Error code.
+ *
+ */
+igraph_error_t igraph_delaunay_triangulation(igraph_t *graph, igraph_matrix_t *points) {
```
szhorvat: Rename to `igraph_delaunay_graph`.

Be sure to mark `points` as `const`—we're not allowed to mess up the user's data.

src/spatial/delaunay.c
Zepeacedust: Added helper function to transpose and copy at the same time.

src/spatial/delaunay.c
```diff
@@ -0,0 +1,247 @@
...
+
+    if (numpoints > INT_MAX) {
+        IGRAPH_ERROR("Too many points for Qhull.", IGRAPH_EOVERFLOW);
+    }
+
+    // For internal use, it returns a complete graph.
+    // This is not nessecarily a delaunay graph, but for nondegenerate point sets it is.
+    if (numpoints <= dim) {
+
+    }
```
szhorvat: @Zepeacedust It seems this is not yet complete. I'm marking the PR as a draft again. Let me know when it's ready to go.

src/spatial/delaunay.c
```diff
@@ -0,0 +1,263 @@
...
+    exitcode = setjmp(qh->errexit); /* simple statement for CRAY J916 */
+    if (!exitcode) {
+        qh->NOerrexit = False;
+        qh_option(qh, "delaunay  Qbbound-last QJ", NULL, NULL);
+        qh->PROJECTdelaunay = True; // project points to parabola to calculate delaunay
+        qh->USEstdout = False;
+        qh->DELAUNAY = True;    /* 'd'   */
+        qh->SCALElast = True;   /* 'Qbb' */
+        qh->KEEPcoplanar = True; /* 'Qc', to keep coplanars in 'p' */
+        qh->TRIangulate = True;  // Set flag to split non-simplicial points into
```
szhorvat: Please complete this comment

szhorvat:
What does Qhull do when some vertices coincide? Does it throw an error?

szhorvat:
@ntamas  I asked for a review, but I think we need a bit more time to finish this properly. Please wait until then.

Zepeacedust:
It seems like qhull ignores all but one of the duplicated vertices.

I seem to recall a function to enumerate them, but I'm not finding it again now.

EDIT: There doesn't seem to be such a function, but detecting duplicates is relatively easy since they don't have any edges.

I don't see a way to resolve this that doesn't result in breaking at least some amount of the nice guarantees that delaunay triangulations have, should I just add a check and raise an error?

The other way I could think of is to find the point that is duplicated, and copying over the edges, but that's not computationally trivial.

EDIT 2: I added an error check.

src/spatial/delaunay.c
Zepeacedust: It was supposed to be a check for when we don't have enough points to make a simplex, but that's now handled as an error.

szhorvat:
@ntamas Please review when you have time. This is the first priority to review, as @Zepeacedust has code ready to go for a next PR that depends on this.

Contents of this PR:

 - Add Qhull as vendored code. All files form `libqhull_r` are added even though not all are used. We can't currently link to an external Qhull because modifications were necessary to prevent printing to stderr. As we understand, such modifications are the recommended way to deal with this. The changes made to Qhull are detailed in `CHANGES.md`
 - `igraph_i_delaunay_edges()` — we're going to build more feature on top of this.
 - `igraph_delaunay_graph()`
 - `igraph_i_simplify_edge_list()` is added as a reusable internal function.

TODO (after merging this PR):

 - Update changelog
 - Update ACKNOWLEDGEMENTS.md
 - Exclude Qhull from coverage (for now, the coverage report is useful for debugging)
 - Deal with the case when there are fewer than $d+1$ points in $d$ dimensions or all points lie in a $<d$ dimensional subspace. I plan to use SVD to deal with this, which requires using LAPACK functions we don't currently have in f2c'd form. I might do this only after the 1.0 release, as it's a lot of fuss to do the f2c translation.

doc/spatial.xxml
```diff
@@ -12,6 +12,7 @@
 </section>
 
 <section id="spatial-constructors"><title>Spatial constructors</title>
+<!-- doxrox-include igraph_delaunay_triangulation -->
```
ntamas: Isn't this called `igraph_delaunay_graph` now?

include/igraph_spatial.h
```diff
@@ -27,6 +27,7 @@
 
 IGRAPH_BEGIN_C_DECLS
 
+IGRAPH_EXPORT igraph_error_t igraph_delaunay_graph(igraph_t *graph, const igraph_matrix_t *points);
```
ntamas, line 30:
```suggestion
IGRAPH_EXPORT igraph_error_t igraph_delaunay_graph(igraph_t *graph, const igraph_matrix_t *points);

```

src/internal/utils.c
```diff
@@ -93,3 +94,96 @@ igraph_error_t igraph_i_matrix_subset_vertices(
...
+        for (igraph_integer_t i = 0; i < size; i += 2) {
+            if (VECTOR(*edges)[i] == VECTOR(*edges)[i + 1]) {
+                skipped += 1;
+            } else {
+                break;
+            }
+        }
+    }
+
+    for (igraph_integer_t i = skipped * 2 + 2 ; i < size; i += 2) {
```
ntamas: This is pretty convoluted; please add a comment explaining how it works. I _think_ I understand why it is correct now, but I had to stop for several minutes to think and even to try and find a counterexample (which I eventually gave up on).

Renaming `self_loops` to `allow_self_loops` and `multi_edges` to `allow_multi_edges` helps a bit with the interpretation of the big `if` condition, but it still needs an explanation.

src/spatial/delaunay.c
```diff
@@ -0,0 +1,260 @@
...
+        igraph_vector_int_destroy(&simplex);
+        destroy_qhull(qh);
+        igraph_free(qhull_points);
+        IGRAPH_FINALLY_CLEAN(4);
+    } else {
+        switch (qh->last_errcode) {
+        /* TODO: More specific error descriptions for common Qhull errors. */
+        default:
+            /* TODO: Report Qhull error text? */
+            IGRAPH_ERRORF("Error while computing delaunay triangulation, Qhull error code %d.", IGRAPH_EINVAL, qh->last_errcode);
```
ntamas:
```suggestion
            IGRAPH_ERRORF("Error while computing Delaunay triangulation, Qhull error code %d.", IGRAPH_EINVAL, qh->last_errcode);
```

src/spatial/delaunay.c
```diff
@@ -0,0 +1,260 @@
...
+        igraph_bitset_destroy(&in_edge_list);
+        igraph_vector_int_destroy(&simplex);
+        destroy_qhull(qh);
+        igraph_free(qhull_points);
+        IGRAPH_FINALLY_CLEAN(4);
+    } else {
+        switch (qh->last_errcode) {
+        /* TODO: More specific error descriptions for common Qhull errors. */
+        default:
+            /* TODO: Report Qhull error text? */
```
ntamas, line 229: That would be nice but not necessary if it is hard to extract (e.g. if we need to duplicate Qhull's error strings here).

src/spatial/delaunay.c
```diff
@@ -0,0 +1,260 @@
...
+            IGRAPH_ERRORF("Error while computing delaunay triangulation, Qhull error code %d.", IGRAPH_EINVAL, qh->last_errcode);
+        }
+    }
+
+    return IGRAPH_SUCCESS;
+}
+
+
+/**
+ * \function igraph_delaunay_triangulation
```
ntamas: It's called `igraph_delaunay_graph` now.

src/spatial/spatial_internal.h
```diff
@@ -0,0 +1,33 @@
...
+#define IGRAPH_SPATIAL_INTERNAL_H_
+
+#include "igraph_decls.h"
+#include "igraph_error.h"
+#include "igraph_vector.h"
+#include "igraph_matrix.h"
+
+IGRAPH_BEGIN_C_DECLS
+
+igraph_error_t igraph_i_delaunay_edges(igraph_vector_int_t *edges, const igraph_matrix_t *points);
```
ntamas, line 29: Why do we need this function to be declared in a separate header? The header is not imported from anywhere else within igraph.

ntamas (changes requested):

src/internal/utils.c
szhorvat: @Zepeacedust Can you please address this, and use `igraph_bool_t remove_multiple, igraph_bool_t remove_loops` instead, just like in `igraph_simplify()`? This means both renaming the parameters and negating the conditions.

src/spatial/spatial_internal.h
szhorvat: For the next PR that is coming, which filters these edges down to a smaller subset before constructing a graph called a [β-skeleton](https://en.wikipedia.org/wiki/Beta_skeleton). The very reason for adding Delaunay graphs is to enable the implementation of β-skeletons.

src/internal/utils.c
Zepeacedust: I refactored the checks into two guard statements, and in the process removed the earlier for loop for self loops.
The new version should be easier to understand and maintain.

src/spatial/delaunay.c
szhorvat: This might be improved based on experience with actual use.

szhorvat:
@ntamas Please have another look. The duplicate removal is now idiomatic.

I also added a helper function:

```c
igraph_error_t igraph_i_check_spatial_points(const igraph_matrix_t *points);
```

Once this PR is merged I will add this check to `nearest_neighbor_graph()`, where it is necessary to avoid crashes with non-finite float values.

ntamas (approved):

PR #2810, opened 2025-07-31
Reindex membership vectors that have cluster indices outside of 0 .. vcount-1.

szhorvat:
@ntamas Before completing this PR, I'd like to hear your opinion.

Currently we have a function called `igraph_reindex_membership()` which:

  - Expects a membership vector with cluster indices strictly in the range `0 .. vertex_count-1`
  - It replaces these with indices in the range `0 .. cluster_count - 1` in such a way that new indices appear in increasing order. It does two conceptually different things: (1) it ensures that the largest index is `cluster_count - 1` (2) it ensured that _new_ indices appear in increasing order. It certainly guarantees (1) but it is unclear whether it is meant to guarantee (2) as well.

It is useful to have a function that supports indices that are not in the `0 .. vertex_count-1` range. This PR contains an implementation of this. What we need to decide is how to expose this:

  - A completely separate function (what should it be called?)
  - Just let `igraph_reindex_membership()` support indices outside of `0 .. vertex_count-1` and drop guarantee (2) (it's unclear if this was ever guaranteed).
  - Add an internal function that has an `igraph_bool_t check_range` flag that helps retain the old functionality, i.e. it throws an error with indices outside of the expected range. This can be useful. I actually caught an Infomap bug thanks to this check, as Infomap was shifting indices to be 1-based instead of 0-based and we forgot to shift them back.
  - Same as above, but make this flag public (on `develop`, of course).

Notes:

 - There _might_ be other functions which require cluster indices to be `0 <= c < vcount`. If so, we must clearly document this requirement _or_ use this new functionality to bring them into the proper range.
 - What I'm calling "cluster indices" may be called "vertex types" in some contexts, e.g. graph colouring. See the recent `is_coloring()` PR.
 - Note the failing CI test: `igraph_split_join_distance()` was relying on `igraph_reindex_membership()` to throw an error when indices were out of range. Of course this is easy to relax (and IMO beneficial to relax).

TODO:

 - [ ] Finalize interface
 - [ ] Review what other functions are affected (see Notes above)
 - [ ] Update docs with changes requirements and potentially changed time complexity.

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -686,6 +686,7 @@ add_legacy_tests(
   igraph_le_community_to_membership
   igraph_modularity
   igraph_modularity_matrix
+  igraph_reindex_membership
   igraph_split_join_distance
   levc-stress
   null_communities
--- a/tests/unit/igraph_reindex_membership.c
+++ b/tests/unit/igraph_reindex_membership.c
@@ -1,0 +1,86 @@
+/*
+   IGraph library.
+   Copyright (C) 2025  The igraph development team <igraph@igraph.org>
+
+   This program is free software; you can redistribute it and/or modify
+   it under the terms of the GNU General Public License as published by
+   the Free Software Foundation; either version 2 of the License, or
+   (at your option) any later version.
+
+   This program is distributed in the hope that it will be useful,
+   but WITHOUT ANY WARRANTY; without even the implied warranty of
+   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+   GNU General Public License for more details.
+
+   You should have received a copy of the GNU General Public License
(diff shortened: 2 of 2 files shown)


szhorvat:
I think this is fine in it's current state. Let's merge it so I can make use of it elsewhere.

ntamas:
> (2) it ensured that new indices appear in increasing order. It certainly guarantees (1) but it is unclear whether it is meant to guarantee (2) as well.

It is not meant to guarantee (2). If it was meant to do so, it would have been documented.

I think that this function was added primarily for the sake of the label propagation community detection, where the function started out from a membership vector containing every integer from 0 to |V|-1 and then it started merging clusters into others such that _one_ of the cluster indices was kept and the other was overwritten. As a result, the vector ended up with large gaps between cluster indices. `igraph_reindex_membership()` was meant to fix that but nothing else.

PR #2813, opened 2025-08-03
feat: igraph_community_leiden() now supports directed graphs

szhorvat:
@ntamas If you have time on Monday, a quick rough review would be appreciated. I'll be on the way back home during Monday.

Please do not merge yet. I did this during the flight, and tested it by comparing the result on some small random directed graphs with `igraph_community_optimal_modularity()`.

@vtraag Please review this as well. I will likely merge this before you're back from holidays, so that Bea can make progress on the GSoC project, but we can still make fixes before 1.0 is released.

### DIFF
--- a/src/community/leiden.c
+++ b/src/community/leiden.c
@@ -873,13 +873,8 @@ static igraph_error_t community_leiden(
     i_vertex_in_weights = directed ? vertex_in_weights : NULL;
     i_membership = membership;
 
-    /* Clean membership and count number of *clusters */
-
+    /* Clean membership: ensure that cluster indices are 0 <= c < n. */
     IGRAPH_CHECK(igraph_reindex_membership(i_membership, NULL, nb_clusters));
-
-    if (*nb_clusters > n) {
-        IGRAPH_ERROR("Too many communities in membership vector.", IGRAPH_EINVAL);
-    }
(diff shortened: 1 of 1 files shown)


ntamas:
Skimmed through it and it seems okay at a superficial level, but I'm not familiar enough with the internals of the directed Leiden algorithm to judge the correctness.

szhorvat:
@ntamas Let's merge this then. If you have time, can you update the Python interface to include this change? Note that in the directed case and the "modularity" objective function, we use $\gamma/m$ instead of $\gamma/(2m)$ as the resolution, where $m$ is the total edge weight / edge count.

ntamas:
> Note that in the directed case and the "modularity" objective function, we use γ / m instead of γ / ( 2 m ) as the resolution, where m is the total edge weight / edge count.

Does that have any consequences for the Python interface?

vtraag:
This also looks good to me. I only checked the core of the implementation, not all other related files.

I just would like to note that the directedness obviously is only relevant when `vertex_out_weights != vertex_in_weights`. In particular, this means that the directedness is only relevant for modularity in practice, not for CPM. This is probably something that could be documented, also so that it is clear how higher-level interface could work with this.

PR #2815, opened 2025-08-05
Simplified interface to Leiden communities and weighted density calculation

szhorvat:
This PR:

 - Generalizes density calculation to weighted graphs
 - Adds a simplified interface to Leiden to avoid re-implementing this separately in each high-level interface

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -74,6 +74,7 @@
  - `igraph_matrix_copy_to()` gained an `igraph_matrix_storage_t storage` parameter that specifies whether the data should be written in column-major or row-major format.
  - `igraph_sbm_game()` gained a `multiple` parameter and now supports generating multigraph with prescribed expected edge multiplicities. The parameter determining the total number of vertices (`n`) was removed as it was redundant.
  - The `bfs_cutoff` parameter of the experimental functions `igraph_fundamental_cycles()` and `igraph_minimum_cycle_basis()` is now of type `igraph_real_t`.
+ - `igraph_density()` now takes an optional `weights` parameter.
 
 ### Added
 
@@ -91,6 +92,7 @@
  - `igraph_nearest_neighbor_graph()` computes a neighborhood graph of spatial points based on a neighbor count, cutoff distance, and chosen metric (experimental function). Thanks to Arnór Friðriksson @Zepeacedust for implementing this in #2788!
  - `igraph_delaunay_graph()` computes a Delaunay graph of a spatial point set (experimental function). Thanks to Arnór Friðriksson @Zepeacedust for implementing this in #2806!
  - `igraph_spatial_edge_lengths()` computes edges lengths based on spatial vertex coordinates (experimental function).
+ - `igraph_community_leiden_simple()` is a simplified interface to `igraph_community_leiden()` that allows selecting the objective function to maximize directly.
 
 ### Changed
 
--- a/include/igraph_community.h
+++ b/include/igraph_community.h
@@ -230,6 +230,12 @@ IGRAPH_EXPORT igraph_error_t igraph_community_multilevel(const igraph_t *graph,
                                               igraph_matrix_int_t *memberships,
                                               igraph_vector_t *modularity);
 
+typedef enum {
+    IGRAPH_LEIDEN_OBJECTIVE_MODULARITY = 0,
+    IGRAPH_LEIDEN_OBJECTIVE_CPM,
+    IGRAPH_LEIDEN_OBJECTIVE_ER
+} igraph_leiden_objective_t;
+
 IGRAPH_EXPORT igraph_error_t igraph_community_leiden(
         const igraph_t *graph,
         const igraph_vector_t *edge_weights,
@@ -241,6 +247,18 @@ IGRAPH_EXPORT igraph_error_t igraph_community_leiden(
         igraph_integer_t n_iterations,
         igraph_vector_int_t *membership,
         igraph_integer_t *nb_clusters, igraph_real_t *quality);
+
+IGRAPH_EXPORT igraph_error_t igraph_community_leiden_simple(
+        const igraph_t *graph,
+        const igraph_vector_t *weights,
+        igraph_leiden_objective_t objective,
+        igraph_real_t resolution,
+        igraph_real_t beta,
+        igraph_bool_t start,
+        igraph_integer_t n_iterations,
+        igraph_vector_int_t *membership,
+        igraph_integer_t *nb_clusters,
+        igraph_real_t *quality);
 
 /* -------------------------------------------------- */
 /* Community Structure Comparison                     */
--- a/src/community/leiden.c
+++ b/src/community/leiden.c
@@ -26,6 +26,7 @@
 #include "igraph_memory.h"
 #include "igraph_random.h"
 #include "igraph_stack.h"
+#include "igraph_structural.h"
 #include "igraph_vector.h"
 #include "igraph_vector_list.h"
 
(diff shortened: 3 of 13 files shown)


szhorvat:
@ntamas This is good to go. Do NOT squash, just merge with rebase.  You can use this in py-igraph.

src/community/leiden.c
```diff
@@ -1084,6 +1085,8 @@ static igraph_error_t community_leiden(
  * optimized when you supply the degrees (out- and in-degrees with directed graphs)
  * as the vertex weights and by supplying as a resolution parameter
  * <code>1/(2m)</code> (<code>1/m</code> with directed graphs).
+ * Use the \ref igraph_community_leiden_simple() convenience function to
+ * compute vertex weights automatically to for modularity maximization.
```
ntamas:
```suggestion
 * compute vertex weights automatically for modularity maximization.
```

include/igraph_community.h
```diff
@@ -230,6 +230,13 @@ IGRAPH_EXPORT igraph_error_t igraph_community_multilevel(const igraph_t *graph,
                                               igraph_matrix_int_t *memberships,
                                               igraph_vector_t *modularity);
 
+typedef enum {
+    IGRAPH_LEIDEN_OBJECTIVE_MODULARITY = 0,
+    IGRAPH_LEIDEN_OBJECTIVE_CPM,
+    IGRAPH_LEIDEN_OBJECTIVE_ER,
+    IGRAPH_LEIDEN_OBJECTIVE_ER_LOOPS
+} igraph_leiden_objective_t ;
```
ntamas:
```suggestion
} igraph_leiden_objective_t;
```

src/community/leiden.c
```diff
@@ -1263,3 +1271,239 @@ igraph_error_t igraph_community_leiden(
...
+ * \sa \ref igraph_community_leiden() for a more flexible interface that
+ * allows specifying raw vertex weights.
+ */
+igraph_error_t igraph_community_leiden_simple(
+        const igraph_t *graph,
+        const igraph_vector_t *weights,
+        igraph_leiden_objective_t objective,
+        igraph_real_t resolution,
+        igraph_real_t beta,
+        igraph_bool_t start,
```
ntamas, line 1358: If this is meant to be a simpler interface, I wonder whether we should get rid of the `start` argument and just start the algorithm from the singleton clustering unconditionally. How often do people want to use a non-singleton clustering as a starting point?

src/community/leiden.c
```diff
@@ -1263,3 +1271,239 @@ igraph_error_t igraph_community_leiden(
...
+ *    the assumption that self-loops may be present in the graph.
+ *    \endclist
+ *    In the above formulas, \c A is the adjacency matrix, \c m is the total
+ *    edge weight, \c k are the (out- and in-) degrees, \c γ is the resolution
+ *    parameter, and <code>δ(c_i, c_j)</code> is 1 if vertices \c i and \c j
+ *    are in the same community and 0 otherwise.
+ * \param resolution The resolution parameter, which is represented by γ in
+ *    the objective functions detailed above.
+ * \param beta The randomness used in the refinement step when merging. A small
+ *    amount of randomness (\c beta = 0.01) typically works well.
```
ntamas, line 1329: Can we use `beta = 0.01` in `functions.yaml` as a reasonable default for `beta`? I'm trying to provide most of the arguments with a default value if this is indeed meant as a simpler alternative to the full Leiden function.

tests/unit/igraph_density.c
```diff
@@ -27,7 +27,7 @@
 void test_density(const igraph_t *graph, igraph_bool_t loops) {
     igraph_real_t density;
 
-    if (igraph_density(graph, &density, loops)) {
+    if (igraph_density(graph, NULL, &density, loops)) {
```
ntamas, line 30: Should we add tests for the weighted calculation as well?

CHANGELOG.md
```diff
@@ -69,6 +69,7 @@
  - The deprecated `igraph_sparsemat()` and `igraph_weighted_sparsemat()` functions were removed; use `igraph_get_adjacency_sparse()` instead.
  - `igraph_is_simple()` gained an extra `igraph_bool_t` argument that decides whether edge directions should be considered. Directed graphs with a mutual edge pair are treated as non-simple if this argument is set to `IGRAPH_UNDIRECTED` (which treats the graph as if it was undirected).
  - `igraph_community_leiden()` now takes two `vertex_out_weights` and `vertex_in_weights` parameters in order to support directed graphs, instead of the previou single `node_weights` parameter. To obtain the old behavior for undirected graphs, pass the vertex weights as `vertex_out_weights` and set `vertex_in_weights` to `NULL`.
+ - `igraph_density()` now takes a `weights` parameter.
```
ntamas:
```suggestion
 - `igraph_density()` now takes an optional `weights` parameter.
```

ntamas (changes requested):

src/community/leiden.c
szhorvat: It's already there, but we also need to add a default for `start`, i.e. `start=False`. I'll do that.

src/community/leiden.c
szhorvat: I assumed that the high-level interfaces are already using this, and didn't want to break that functionality. It seems that the Python interface already has an `initial_membership` parameter.

This is meant for attempting to improve on an existing partitioning.

I don't really mind removing it, but given that I've already implemented all the error checks for this, I vote for leaving it in.

tests/unit/igraph_density.c
szhorvat: Done.

szhorvat:
@ntamas I'm going to merge this and change the Python interface code so that it selects between `igraph_community_leiden()` and `igraph_community_leiden_simple()` based on whether the objective function was provided.

Is this fine?

Currently, in Python, we have a different interface between `Graph.community_leiden()` and `GraphBase.community_leiden()`. The `Graph` version takes a friendly objective function, the `GraphBase` version a somewhat cryptic `normalize_resolution` boolean parameter. Is the `GraphBase` version considered proper public API, and do we need to keep it as-is?  Is it fine to have both `GraphBase.community_leiden()` and `GraphBase.community_leiden_simple()` and dispatch to the appropriate one from Python code? What do you suggest as the best way to handle this?

vtraag:
It makes sense to me to use the `community_leiden_simple` in higher level interfaces, otherwise the implementation is of little use.

The `GraphBase` and `Graph` distinction to me is at least that most users will never use a `GraphBase` function directly when overloaded by a `Graph` function.

szhorvat:
Thanks for the feedback @vtraag ! I clarified in the docs that CPM and ER are not sensitive to edge directions.

@ntamas Shall we merge this in its current form? This makes sense if we can make use of it in high-level interfaces: do you think we can do it in Python? Basically it'd switch between `community_leiden()` and `community_leiden_simple()` depending on whether it is vertex weights or the objective function that is given. The two shouldn't be given at the same time. The switching logic within the Python interface can be implemented either in C or in Python, depending on your preference.

An alternative is to go back to having a _single_ function in C that does this switching, but I no longer prefer that.

Let me know so we can make progress here.

src/community/leiden.c
```diff
@@ -1263,3 +1271,229 @@ igraph_error_t igraph_community_leiden(
...
 }
+
+/**
+ * \function igraph_community_leiden_simple
+ * \brief Finding community structure using the Leiden algorithm, simple interface.
+ *
+ * This is a simplified interface to \ref igraph_community_leiden() for
+ * convenience purposes. Instead of requiring vertex weights, it allows
+ * choosing from a set of objective functions to maximize, and calculates
+ * the vertex weights and resolution parameter to pass to
```
ntamas: Is it calculating the resolution parameter automatically? The function seems to have a resolution parameter that needs to be specified explicitly.

ntamas (approved):
Just one quick question regarding the wording of the docstring - it is not entirely clear to me whether the `resolution` parameter is supposed to be provided by the user or is calculated automatically.

src/community/leiden.c
szhorvat: Good point!  I will clarify the wording.  It rescales the resolution parameter before passing it to the original Leiden code. I'll try to find a way to explain what's going on without making it confusing.

PR #2816, opened 2025-08-05
refactor!: igraph_rewire() now takes an igraph_edge_type_sw_t parameter instead of igraph_rewiring_t

### LINKED ISSUE #2554: `igraph_rewire()` should take an `igraph_edge_type_sw_t` parameter instead of `igraph_rewiring_t`
In igraph 1.0, `igraph_rewire()` should take an `igraph_edge_type_sw_t` parameter instead of `igraph_rewiring_t`. This way we will be able to control separately whether multi-edges and self-loops are allowed to be created. 

I plan to implement rewiring procedures for multigraphs. This change prepares the ground for this feature. For now, `IGRAPH_UNIMPLEMENTED` should be returned for unsupported combinations.

Related: #1062.

szhorvat:
Closes #2554 

1.0 milestone one step closer.


src/misc/graphicality.h
```diff
@@ -0,0 +1,10 @@
+
```
ntamas, line 18: Copyright header should be added here.

ntamas (changes requested):
Just one minor nitpick, then feel free to merge.

PR #2817, opened 2025-08-07
Feat: Adding Corona Graph Product

gulshan-123:

---
TODO: 1. Update docs, once direction parameter for digraph is confirmed


gulshan-123:
@szhorvat, (taken from previous PR)
For corona products:
image

1. If the graphs are directed, then the new edges from central graph to the outer graph may have two possible direction. Should we add a `direction` parameter to decide the directionality? We may use `IGRAPH_IN` and `IGRAPH_OUT` (or new constants?).
2. Total number of vertex is v1.(v2+1), differing from the other products which have v1.v2.

I think a separate new functions for corona product may be added which will have a `direction` parameter (separate function is also for rooted product). we will write about new vertex number and indices in its docs. I think this will also keep docs of normal graph product clean. ??

szhorvat:
Definitely a separate function. I'll look at the directedness issue when I get the time, but this should be a separate function just because of the different vertex ID mapping.

gulshan-123:
@szhorvat, I have added the corona product. Currently, I have added a parameter `mode` specifying about direction from central to outer graph in digraph product (to be discussed).
I have also added the vertex id mapping in the docs.
(No rush for review, just keeping you informed)

szhorvat:
@ntamas This is not a blocker for 1.0, but let me take a look if I can review it in time ... it'd be nice to include it.

gulshan-123:
Hi @szhorvat, any more work to do/suggestions to this PR? As far as I remember, we were discussing about the direction parameter to be included for directed graphs.

Issue #2694, opened 2024-11-05
Managing combinatorial explosion: Limit number of returned results


szhorvat:
We discussed that the ideal way to manage combinatorial explosion in the number of results is to implement "iterator interfaces". Doing this is a long way off. For igraph 1.0, I propose a simpler approach that is easy enough to implement, and can be practical for many applications: 

_Limit the number of results._

For example, a clique finder function currently lists _all_ cliques within the provided size range. If we only want one such clique, or a few, the only option is to use the callback version. Many of the callback interfaces are not exposed in high-level languages, as doing so is brittle and difficult. Even when they are, repeatedly calling a host-language function can cause slowdowns.

I propose that the following functions should receive an extra parameter which limits the number of results to return, making it easy to get one or a few results from high-level languages.

 - `igraph_cliques()`
 - `igraph_maximal_cliques()`
 - `igraph_largest_cliques()` *
 - `igraph_independent_vertex_sets()`
 - `igraph_weighted_cliques()`
 - `igraph_largest_weighted_cliques()` *
 - `igraph_maximal_independent_vertex_sets()`
 - `igraph_largest_independent_vertex_sets()` *
 - `igraph_get_all_simple_paths()`
 - `igraph_simple_cycles()`
 - `igraph_get_isomorphisms_vf2()`
 - `igraph_get_subisomorphisms_vf2()`
 - `igraph_subisomorphic_lad()` *

All of these list objects of which there can be exponentially many.

Perhaps the ones with a * don't need this, as there aren't usually that many results, or the implementation is not as straightforward as for the rest.

vtraag:
Instead of adding an argument to these functions, could it perhaps be handled more gracefully by simply returning whatever results are already found when the function is interrupted? This also does not necessitate any API changes, so could be implemented after 1.0.

szhorvat:
I don't agree that this approach will be useful, but I acknowledge the practical time constraints. Please leave this to me. What I can do before the release deadline will be in 1.0, the rest left for 2.0.

The real solution IMO is to (1) move to C++ first (2) implement iterators using C++20's coroutines.

ntamas:
> Instead of adding an argument to these functions, could it perhaps be handled more gracefully by simply returning whatever results are already found when the function is interrupted? 

[...]

PR #2824, opened 2025-08-14
Limit the number of returned results by various functions

szhorvat:
Refs #2694.

@ntamas, any comments on the approach before I proceed to other functions?

### DIFF
--- a/interfaces/functions.yaml
+++ b/interfaces/functions.yaml
@@ -1325,8 +1325,9 @@ igraph_is_clique:
 
 igraph_cliques:
     PARAMS: |-
-        GRAPH graph, OUT VERTEX_INDICES_LIST res, INTEGER min_size=0,
-        INTEGER max_size=0
+        GRAPH graph, OUT VERTEX_INDICES_LIST res, 
+        INTEGER min_size=UNLIMITED, INTEGER max_size=UNLIMITED,
+        INTEGER max_results=UNLIMITED
     DEPS: res ON graph
 
 igraph_cliques_callback:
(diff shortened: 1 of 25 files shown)


szhorvat:
Idea from Tamás in chat: `IGRAPH_UNLIMITED` (or `IGRAPH_UNLIMITED_RESULTS`) defined to `-1` to aid readability.

src/cliques/maximal_cliques_template.h
```diff
@@ -220,6 +228,12 @@ igraph_error_t FUNCTION(igraph_maximal_cliques, SUFFIX)(
         IGRAPH_WARNING("Edge directions are ignored for maximal clique calculations.");
     }
 
+    PREPARE;
+
+    if (max_results == 0) {
+        return IGRAPH_SUCCESS;
+    }
+
```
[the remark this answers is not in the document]
szhorvat: Yes, there is a reason: output vectors must be cleared.

interfaces/functions.yaml
```diff
@@ -1325,8 +1325,9 @@ igraph_is_clique:
 
 igraph_cliques:
     PARAMS: |-
-        GRAPH graph, OUT VERTEX_INDICES_LIST res, INTEGER min_size=0,
-        INTEGER max_size=0
+        GRAPH graph, OUT VERTEX_INDICES_LIST res, 
+        INTEGER min_size=UNLIMITED, INTEGER max_size=UNLIMITED,
+        INTEGER max_results=UNLIMITED
```
szhorvat, line 1330: @ntamas Do we have facilities in Stimulus to indicate, within this repo, that the C value for `UNLIMITED` is `IGRAPH_UNLIMITED`?

ntamas: No, but why would we need that? C does not have default values.

PR #2826, opened 2025-08-15
perf: optimize vector_scale() by 1 and vector_add_constant() with 0

szhorvat:
@ntamas I need a sanity check please. I'd find it convenient when implementing betweenness normalization, to have simpler code without an unnecessary pass over the vector. But perhaps I didn't sleep enough and this is not good for long-term maintainability.

`igraph_vector_scale()` now checks if the scaling factor is 1, and if so it skips scaling. Same for `igraph_vector_add_constant()`.


szhorvat:
It's interesting to see that these short-circuit lines _are_ covered, and it's not due to a test that calls these functions with a 1 / 0.

PR #2827, opened 2025-08-17
Beta skeletons

Zepeacedust:

Working off from #2806 this will implement [Beta-skeletons](https://en.wikipedia.org/wiki/Beta_skeleton).

Functions to be implemented:
Lune and circle based beta skeletons, beta-weighted gabriel graph.

Ref: #2828

### TODO

 - [x] Gabriel graph function
 - [x] Relative Neighborhood Graph function, using open exclusion region
 - [x] Review code readability, add comments where necessary, use typedefs as appropriate
 - [x] Ensure that code is cleanly formatted (use of empty lines, indentation, etc), follows conventions, no extra headers present, etc.
 - [x] Mark comments resolved when appropriate


Zepeacedust:
I really should have rebased this, but since you squash commits I think it will be ok.
Do let me know if you prefer a rebase.

szhorvat:
Yes, please rebase on `develop`.

src/spatial/nanoflann_internal.hpp
```diff
@@ -0,0 +1,120 @@
+#ifndef NANOFLANN_INTERNAL_H_
```
szhorvat: For internal headers, let's prefix the header guard with the directory name, and let's not use a trailing `_`.

So `SPATIAL_NANOFLANN_INTERNAL_H`

src/spatial/beta_skeleton.cpp
```diff
@@ -0,0 +1,474 @@
...
+        ab2(get_sqr_distance(ai, bi, ps)) {
+        init();
+    }
+
+    void init() {
+        clear();
+    }
+
+    void clear() {
+        smallest_beta = INFINITY; // TODO: find proper igraph infinity
```
szhorvat: `IGRAPH_INFINITY`

src/spatial/beta_skeleton.cpp
```diff
@@ -0,0 +1,474 @@
...
+ *
+ * \return Error code.
+ */
+igraph_error_t igraph_lune_beta_skeleton(igraph_t *graph, const igraph_matrix_t *points, igraph_real_t beta) {
+    igraph_vector_int_t potential_edges;
+    IGRAPH_VECTOR_INT_INIT_FINALLY(&potential_edges, 0);
+
+    IGRAPH_CHECK(beta_skeleton_edge_superset(&potential_edges, points, beta));
+    if (beta >= 1) {
+        IGRAPH_CHECK((filter_edges<edge_is_present<standard_r, construct_lune_centres, intersection_predicate >> (&potential_edges, points, beta)));
```
szhorvat: Note to self: Double parentheses are necessary because otherwise there are issues with macro expansion. Check why!

src/spatial/beta_skeleton.cpp
```diff
@@ -36,6 +37,118 @@
 
 
 
+igraph_real_t get_sqr_distance(igraph_integer_t a, igraph_integer_t b, const igraph_matrix_t *points) {
+    igraph_real_t distance = 0;
+    igraph_real_t temp;
+    igraph_integer_t dims = igraph_matrix_ncol(points);
+    for (igraph_integer_t i = 0; i < dims; i++) {
+        temp = abs(MATRIX(*points, a, i) - MATRIX(*points, b, i));
```
szhorvat: Careful, `abs` is not `fabs`! It's also not necessary here.

In C++, `abs()` is overloaded and technically fine. But in C, `abs()` is for integers only.  To reduce confusion, in a mixed C/C++ project like this, let's stick to `fabs()` only.

src/spatial/beta_skeleton.cpp
szhorvat: @Zepeacedust I removed the double parentheses and nothing broke.

src/spatial/nanoflann_internal.hpp
```diff
@@ -0,0 +1,120 @@
...
+        return MATRIX(*points, idx, dim);
+    }
+    template <typename BoundingBox>
+    bool kdtree_get_bbox(BoundingBox &bb) const {
+        IGRAPH_UNUSED(bb);
+        return false; // indicates that it should use default
+    }
+};
+
+IGRAPH_BEGIN_C_DECLS
```
szhorvat: This is a C++-only header containing classes, so do not use these macros.

src/spatial/beta_skeleton.cpp
```diff
@@ -0,0 +1,701 @@
...
+// Some methods to get distances between two vectors, including when one or both are embedded in a matrix.
+static inline igraph_real_t ind_ind_sqr_distance(igraph_integer_t a, igraph_integer_t b, const igraph_matrix_t *points) {
+    igraph_real_t distance = 0;
+    igraph_real_t temp;
+    igraph_integer_t dims = igraph_matrix_ncol(points);
+    for (igraph_integer_t i = 0; i < dims; i++) {
+        temp = MATRIX(*points, a, i) - MATRIX(*points, b, i);
+        distance += temp * temp;
+    }
+    return distance;
```
szhorvat: Please prefer the following in functions similar to this one:

```suggestion
    igraph_real_t distance = 0;
    igraph_integer_t dims = igraph_matrix_ncol(points);
    for (igraph_integer_t i = 0; i < dims; i++) {
        igraph_real_t temp = MATRIX(*points, a, i) - MATRIX(*points, b, i);
        distance += temp * temp;
    }
    return distance;
```

src/spatial/beta_skeleton.cpp
```diff
@@ -0,0 +1,701 @@
...
+    }
+    void clear() {
+        found = 0;
+    }
+    size_t size() const {
+        return found;
+    }
+    bool full() const {
+        return true;
+    }
```
szhorvat: Spaces between functions:

```suggestion
    void init() {
        clear();
    }

    void clear() {
        found = 0;
    }

    size_t size() const {
        return found;
    }

    bool full() const {
        return true;
    }
```

src/spatial/beta_skeleton.cpp
```diff
@@ -0,0 +1,701 @@
...
+    return IGRAPH_SUCCESS;
+}
+
+static inline igraph_real_t calculate_r(igraph_real_t beta) {
+    if (beta < 1) {
+        return 0.5 / beta;
+    }
+    return 0.5 * beta;
+}
```
szhorvat, line 242: My IDE says this is not used in reachable code, is it wrong?

src/spatial/beta_skeleton.cpp
```diff
@@ -0,0 +1,701 @@
...
+// formula is a_centre = a + (r-1) * (a - b), similar for b_centre
+static igraph_error_t construct_lune_centres(igraph_vector_t *a_centre,
+                                      igraph_vector_t *b_centre,
+                                      igraph_integer_t a,
+                                      igraph_integer_t b,
+                                      igraph_real_t r,
+                                      const igraph_matrix_t *points) {
+    igraph_vector_t a_point, b_point;
+    IGRAPH_VECTOR_INIT_FINALLY(&a_point, 0);
+    IGRAPH_VECTOR_INIT_FINALLY(&b_point, 0);
```
szhorvat: Please avoid initializing new data structures in loops when not strictly required, as this is expensive. Use MATRIX indexing to access `points` instead of `VECTOR`.

Note that this function is called in a loop.

Please do this not just here but also in other similar functions.

src/spatial/beta_skeleton.cpp
```diff
@@ -0,0 +1,701 @@
...
+    igraph_vector_destroy(&a_point);
+    igraph_vector_destroy(&b_point);
+    IGRAPH_FINALLY_CLEAN(4);
+    return IGRAPH_SUCCESS;
+}
+
+// Iterate through the edges given, applying the provided filter.
+// Currently used with is_intersection_empty and is_union_empty.
+// TODO: specialize for multiple dimensions?
+template < igraph_error_t filter(igraph_bool_t *result, kdTree < -1 > &tree, igraph_integer_t a, igraph_integer_t b, const igraph_matrix_t *points, igraph_real_t beta) >
```
szhorvat: Don't forget to typedef filtering functions like these.

src/spatial/beta_skeleton.cpp
```diff
@@ -0,0 +1,701 @@
...
+ */
+
+// Test whether the intersection of the two circles as generated
+// by centre_positions is empty of points except for a and b.
+template <igraph_error_t centre_positions(igraph_vector_t *a_centre,
+          igraph_vector_t *b_centre,
+          igraph_integer_t a,
+          igraph_integer_t b,
+          igraph_real_t beta,
+          const igraph_matrix_t * points)>
```
szhorvat: Please use a consistent with pointers.

```suggestion
          const igraph_matrix_t *points)>
```

src/spatial/beta_skeleton.cpp
```diff
@@ -0,0 +1,701 @@
...
+                }
+            }
+        }
+        return true;
+    }
+};
+
+// Shrinks type signatures significantly.
+template <igraph_integer_t Dimension>
+using kdTree = nanoflann::KDTreeSingleIndexAdaptor <
```
szhorvat: Let's stick to a consistent style for class names: CamelCase. So this would be KDTree.

src/spatial/beta_skeleton.cpp
Zepeacedust: Yes, it is called in the filter functions.
They are themselves only called when passed as template parameters, so the IDE might have trouble detecting it.

src/spatial/beta_skeleton.cpp
Zepeacedust: After adding a template function that needed two or more template parameters, it seems to have issue with the commas.
I don't think it is a specific issue with IGRAPH_CHECK, and more an issue with macros in general.

tests/benchmarks/beta_skeletons.c
```diff
@@ -0,0 +1,150 @@
...
+    bench_gabriel(++test_nr, 100, 2, 2);
+    bench_gabriel(++test_nr, 1000, 2, 2);
+    bench_gabriel(++test_nr, 10000, 2, 2);
+    bench_gabriel(++test_nr, 100000, 2, 2);
+
+    bench_gabriel(++test_nr, 100, 3, 2);
+    bench_gabriel(++test_nr, 1000, 3, 2);
+    bench_gabriel(++test_nr, 10000, 3, 2);
+    bench_gabriel(++test_nr, 100000, 3, 2);
+    return 1;
```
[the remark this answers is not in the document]
szhorvat: @Zepeacedust Please do pay attention to this, I'll fix it here.

szhorvat:
@Zepeacedust Can you think through the time complexities of these, based on the complexity indicated for the Delaunay function, k-D tree search complexity, and the code here?

src/spatial/nanoflann_internal.hpp
```diff
@@ -0,0 +1,118 @@
...
+    std::vector<igraph_real_t> distances;
+
+    using DistanceType = igraph_real_t;
+    using IndexType = igraph_integer_t;
+
+    GraphBuildingResultSet(const igraph_integer_t max_neighbors, const igraph_real_t max_distance) :
+        max_distance(max_distance),
+        max_neighbors(max_neighbors),
+        neighbors(1),
+        distances(1) { }
```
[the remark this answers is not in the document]
szhorvat: @Zepeacedust The memory management with push_back() using a dummy 0 to extend the vector length, but otherwise directly indexing into the vector to add values is weird and difficult to follow. I am overwhelmed at the moment. Can you please have a look and see if this can be cleaned up?

src/spatial/beta_skeleton.cpp
```diff
@@ -0,0 +1,794 @@
...
+ * containing all the points.
+ */
+typedef igraph_error_t FilterFunc(
+    igraph_bool_t *result,
+    KDTree<-1> &tree,
+    igraph_integer_t a,
+    igraph_integer_t b,
+    const igraph_matrix_t *points,
+    igraph_real_t beta
+);
```
szhorvat: Can `&tree` be `const` here? If yes, make it so.

Can we get rid of the Cisms in this C++ functions? As far as I can see, it allocates igraph data structures strictly for internal use. Can we replace these with standard C++ containers, and implicitly rely on exceptions instead of igraph error codes?

Note that these filter functions are actually called without doing an `IGRAPH_CHECK`, so if we stayed with igraph style error handling, we'd need to add that. But I recommend moving on to standard C++ style, for as long as it doesn't complicate things. I expect it will _simplify_ things.

Finally, can we actually avoid these repeated allocations (these are called for every edge)? Perhaps rely on a pre-allocated vector passed to the filter function?

szhorvat:
@Zepeacedust There was a major renaming from igraph_integer_t -> igraph_int_t. I took care of adapting this PR. Make sure you pull before you do anything else.

src/spatial/nanoflann_internal.hpp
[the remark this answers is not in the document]
Zepeacedust: It is an artefact of some memory optimization resulting in strange behavior.
I think I will extract it into a helper function.

szhorvat:
@Zepeacedust Can you please remove spaces from around `<` and `>` when used in templates? I see some of the ones I removed got the spaces back. This is subjective, but we prefer no spaces for readability and no confusion with less-than / greater-than.

szhorvat:
@ntamas Okay to merge this? I'll get additional polish later. 

EDIT: Just let me know and I'll merge.

szhorvat:
@ntamas OK to merge this? It's been ready for a while. We're hoping to expose it in Python soon.

PR #2835, opened 2025-08-17
correct spelling errors

jgmbenoit:
Correct spelling errors as reported by lintian(1) in some binraries.

### DIFF
--- a/src/properties/neighborhood.c
+++ b/src/properties/neighborhood.c
@@ -91,7 +91,7 @@ igraph_error_t igraph_neighborhood_size(const igraph_t *graph, igraph_vector_int
     }
 
     if (mindist < 0) {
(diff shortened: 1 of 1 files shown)


szhorvat:
Thanks, as always! :-)

szhorvat:
@jgmbenoit Would it be feasible/useful to use lintian in igraph CI checks or would this produce too many warnings that are irrelevant for us?

jgmbenoit:
According to the manpage `lintian(1)`, `lintian` can run specific test with the option `-C`. However `lintian(1)`  checks against the Debian packages (the deb files). So the tricky part is to build the Debian packages. This can hardly be an automatic job. Most of the time some issues (easy to resolve or not by hand) pop up during the building and packaging of the upstream version due to changes in the dependencies, in the upstream source, in Debian policy, ... This is our job as Debian Developer. Having said that,  may be you can use the tools that `lintian(1)`  uses to check spelling errors in the librairies (namely the `.so' files) that you build.

PR #2840, opened 2025-08-19
Add support for regularization to infomap

### LINKED ISSUE #2811: Expose Infomap features
Let's expose extra Infomap features before 1.0. Most features should be very easy to _implement_ (just add parameters and write values into the config object), we just need someone with a good understanding of Infomap to decide which features are most relevant, and to write docs for them. Given that the implementation part is very easy, I'm making this a blocker for 1.0.

@vtraag @danieledler, can you please advise? A quick answer would be much appreciated due to the impending release deadline. My bottleneck is a sufficient understanding of Infomap.

Ideas:

 - Expose teleportation probability? My understanding is that this is not particularly useful, so perhaps not.
 - `--two-level` feature of the command-line tool as a Boolean parameter?
 - `--flow-model` feature as enum?
 - Do we want a `directed` flag to ignore edge directions? The user can just convert the graph.

Config object for reference:

https://github.com/igraph/igraph/blob/109f9950d2b19ecc9cd10c751f2768bda0b5ce01/vendor/infomap/src/io/Config.h#L69-L135

vtraag:
This PR fixes #2811 by adding relevant regularization parameters to infomap.

### DIFF
--- a/interfaces/functions.yaml
+++ b/interfaces/functions.yaml
@@ -1765,6 +1765,7 @@ igraph_community_infomap:
     PARAMS: |-
         GRAPH graph, OPTIONAL EDGE_WEIGHTS edge_weights,
         OPTIONAL VERTEX_WEIGHTS vertex_weights, INTEGER nb_trials=10,
+        OPTIONAL BOOLEAN is_regularized=false, OPTIONAL REAL regularization_strength=1,
         OPTIONAL OUT VECTOR_INT membership, OPTIONAL OUT REAL codelength
     DEPS: edge_weights ON graph, vertex_weights ON graph
 
--- a/interfaces/functions.yaml
+++ b/interfaces/functions.yaml
@@ -1765,7 +1765,7 @@ igraph_community_infomap:
     PARAMS: |-
         GRAPH graph, OPTIONAL EDGE_WEIGHTS edge_weights,
         OPTIONAL VERTEX_WEIGHTS vertex_weights, INTEGER nb_trials=10,
-        OPTIONAL BOOLEAN is_regularized=false, OPTIONAL REAL regularization_strength=1,
+        BOOLEAN is_regularized=False, REAL regularization_strength=1,
         OPTIONAL OUT VECTOR_INT membership, OPTIONAL OUT REAL codelength
     DEPS: edge_weights ON graph, vertex_weights ON graph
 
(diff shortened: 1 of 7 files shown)


vtraag:
Note that I assume that @ntamas is happy with the added parameters instead of a parameter struct.

interfaces/functions.yaml
```diff
@@ -1765,6 +1765,7 @@ igraph_community_infomap:
     PARAMS: |-
         GRAPH graph, OPTIONAL EDGE_WEIGHTS edge_weights,
         OPTIONAL VERTEX_WEIGHTS vertex_weights, INTEGER nb_trials=10,
+        OPTIONAL BOOLEAN is_regularized=false, OPTIONAL REAL regularization_strength=1,
```
szhorvat:
```suggestion
        BOOLEAN is_regularized=False, REAL regularization_strength=1,
```

We use _either_ OPTIONAL (when the type implies a default value) _or_ an explicit default value. This could be improved but I don't think Stimulus can do better now.

Correct me if I'm wrong @ntamas .

src/community/infomap.cpp
```diff
@@ -165,6 +171,11 @@ static bool infomap_allow_interruption() {
  *     equal weights. Weights are expected to be positive.
  * \param nb_trials The number of attempts to partition the network
  *     (can be any integer value equal to or larger than 1).
+ * \param is_regularized If true, adds a fully connected Bayesian prior network
+ *     to avoid overfitting due to missing links.
+ * \param regularization_strength Adjust relative strength of the Bayesian prior
+ *     network used for regularization. This multiplies the default strength, a
+ *     parameter of 1 hence uses the default regularization strength.
```
szhorvat:
```suggestion
 * \param regularization_strength Adjust relative strength of the Bayesian prior
 *     network used for regularization. This multiplies the default strength, a
 *     parameter of 1 hence uses the default regularization strength. Ignored
 *     when \p is_regularized is set to \c false.
```

szhorvat (changes requested):
@vtraag Please have a look and make sure that the doc suggestions I made are fine.

Then go ahead and squash-and-merge, and finally update the changelog in a new commit.

interfaces/functions.yaml
ntamas: `OPTIONAL` leaves the decision of an appropriate default value to the interface generator. Having an explicit default value makes it -- well, explicit. In this case I would use the suggestion from @szhorvat

ntamas (approved):
This is good to go but please take care of the pending suggestions before merging.

szhorvat:
@vtraag I merged this to make progress. If you decide to make further improvements or changes, please open a followup PR.

vtraag:
Thanks @szhorvat !

Issue #2576, opened 2024-04-17
Finalize some experimental functions before 1.0


szhorvat:
Experimental functions should be reviewed, and decided if they still stay experimental after 1.0.

szhorvat:
@vtraag @iosonofabio I only need feedback from you on the particular functions where I mentioned you in the list above, hopefully this won't take too much of your time. (Of course, feedback is welcome for anything else.)

szhorvat:
Reminder that we need feedback about how to proceed with some of these functions. If I don't hear back, I will keep UMAP functions experimental and will finalize `igraph_ecc()`.

vtraag:
I'm just checking this, you yourself implemted `igraph_ecc` @szhorvat . I'm not directly familiar with the definition, and I therefore also do not know the reason for having that `offset` parameter. If you are unsure about this, we can just keep the experimental label.

ntamas:
My two cents; I'm only reviewing the functions in the "To be discussed" section as the ones in "Drop experimental" are okay for me as is and the ones in "Not ready to be dropped" already have reasons for keeping them as experimental.

* `igraph_iea_game()` and `igraph_bipartite_iea_game()` are recent additions so I'd give it more time (but if @szhorvat feels that they are okay as is then they can be stabilized).
* `igraph_degree_correlation_vector()` can be stabilized if it is up to me -- please provide reasons if you want to keep it experimental.
* `igraph_is_complete()` -- no, I don't want to add a `directed` parameter. Right now the function works as expected; adding a `directed` parameter would only help in the obscure case when the user has a directed graph but _still_ wants to ignore edge directions. In that case they can just convert it to undirected and do the check that way. We don't need to cover 100% of the use-cases, 99% is good enough and this is really not likely to occur too often in practice.
* `igraph_ecc()` - up to @szhorvat to decide
* `igraph_graph_center()` - the list gives a reason why it should remain experimental so I'm okay with that
* `igraph_layout_umap, igraph_layout_umap_3d, igraph_layout_umap_compute_weights` - to be decided by @iosonofabio 
* `igraph_reachability()` - What would be the advantage of introducing an `igraph_bitset_matrix_t`? It seems like an obscure data structure for me that is unlikely to be used anywhere else, it takes quite a few hours to implement, and the advantages are unclear to me. I would not change the type and just mark the function as stable.
[...]

PR #2842, opened 2025-08-23
feat: add support for marking experimental functions in a structured way

### LINKED ISSUE #2841: Mark experimental functions in a structured way
Mark experimental functions in a structured way that allows identifying whether they are being used in a project.

szhorvat:
Closes #2841 

This PR changes how experimental functions are marked. Instead of only relying on the documentation, now there is an `IGRAPH_EXPERIMENTAL` macro, to be used in declarations in a similar way to `IGRAPH_DEPRECATED`.

If the user defines the macro `IGRAPH_WARN_EXPERIMENTAL` to a nonzero value, supported compilers will issue when this function is used and is not eliminated through dead code elimination.  Currently this is GCC-like compilers like GCC itself and Clang.

Thus is a project that uses igraph does not wish to rely on experimental functions, they can use `-DIGRAPH_WARN_EXPERIMENTAL` when compiling.

I proposed this in #2643.

I did not update the documentation yet—that will come after there's a decision both on this PR and on #2643.

This PR also makes bitset functions final (i.e. no longer experimental). I updated #2576 for discussing what (not) to keep experimental for 1.0, and I need comments in that issue.

### DIFF
--- a/include/igraph_operators.h
+++ b/include/igraph_operators.h
@@ -85,15 +85,15 @@ IGRAPH_EXPORT igraph_error_t igraph_induced_subgraph_edges(
 IGRAPH_EXPORT igraph_error_t igraph_subgraph_from_edges(const igraph_t *graph, igraph_t *res,
                                         igraph_es_t eids, igraph_bool_t delete_vertices);
 IGRAPH_EXPORT igraph_error_t igraph_reverse_edges(igraph_t *graph, igraph_es_t eids);
-IGRAPH_EXPORT igraph_error_t igraph_product(igraph_t *res,
+IGRAPH_EXPERIMENTAL IGRAPH_EXPORT igraph_error_t igraph_product(igraph_t *res,
                                             const igraph_t *g1,
                                             const igraph_t *g2,
                                             igraph_product_t type);
-IGRAPH_EXPORT igraph_error_t igraph_rooted_product(igraph_t *res,
+IGRAPH_EXPERIMENTAL IGRAPH_EXPORT igraph_error_t igraph_rooted_product(igraph_t *res,
                                                    const igraph_t *g1,
                                                    const igraph_t *g2,
                                                    const igraph_integer_t root);
-IGRAPH_EXPORT igraph_error_t igraph_mycielskian(const igraph_t *graph, igraph_t *res, igraph_integer_t k);
+IGRAPH_EXPERIMENTAL IGRAPH_EXPORT igraph_error_t igraph_mycielskian(const igraph_t *graph, igraph_t *res, igraph_integer_t k);
 
 IGRAPH_END_C_DECLS
 
--- a/include/igraph_structural.h
+++ b/include/igraph_structural.h
@@ -81,7 +81,7 @@ IGRAPH_EXPORT igraph_error_t igraph_is_perfect(const igraph_t *graph, igraph_boo
 /* Structural properties                              */
 /* -------------------------------------------------- */
 
-IGRAPH_EXPORT igraph_error_t igraph_is_complete(const igraph_t *graph, igraph_bool_t *res);
+IGRAPH_EXPERIMENTAL IGRAPH_EXPORT igraph_error_t igraph_is_complete(const igraph_t *graph, igraph_bool_t *res);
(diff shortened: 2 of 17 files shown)


include/igraph_operators.h
```diff
@@ -85,15 +85,15 @@ IGRAPH_EXPORT igraph_error_t igraph_induced_subgraph_edges(
...
                                             const igraph_t *g1,
                                             const igraph_t *g2,
                                             igraph_product_t type);
-IGRAPH_EXPORT igraph_error_t igraph_rooted_product(igraph_t *res,
+IGRAPH_EXPERIMENTAL IGRAPH_EXPORT igraph_error_t igraph_rooted_product(igraph_t *res,
                                                    const igraph_t *g1,
                                                    const igraph_t *g2,
                                                    const igraph_integer_t root);
-IGRAPH_EXPORT igraph_error_t igraph_mycielskian(const igraph_t *graph, igraph_t *res, igraph_integer_t k);
+IGRAPH_EXPERIMENTAL IGRAPH_EXPORT igraph_error_t igraph_mycielskian(const igraph_t *graph, igraph_t *res, igraph_integer_t k);
```
ntamas, line 96:
```suggestion
IGRAPH_EXPORT igraph_error_t igraph_mycielskian(const igraph_t *graph, igraph_t *res, igraph_integer_t k);
```

The `\experimental` qualifier was already removed from the documentation of `igraph_mycielskian` and it is also proposed to be promoted to a stable function in #2576

include/igraph_structural.h
```diff
@@ -115,21 +115,21 @@ IGRAPH_EXPORT igraph_error_t igraph_avg_nearest_neighbor_degree(const igraph_t *
...
                                                      const igraph_vector_t *weights);
-IGRAPH_EXPORT igraph_error_t igraph_degree_correlation_vector(
+IGRAPH_EXPERIMENTAL IGRAPH_EXPORT igraph_error_t igraph_degree_correlation_vector(
         const igraph_t *graph, const igraph_vector_t *weights,
         igraph_vector_t *knnk,
         igraph_neimode_t from_mode, igraph_neimode_t to_mode,
         igraph_bool_t directed_neighbors);
 
-IGRAPH_EXPORT igraph_error_t igraph_feedback_arc_set(
+IGRAPH_EXPERIMENTAL IGRAPH_EXPORT igraph_error_t igraph_feedback_arc_set(
```
ntamas:
```suggestion
IGRAPH_EXPORT igraph_error_t igraph_feedback_arc_set(
```

`igraph_feedback_arc_set` is not marked as experimental, neither in the docs nor in #2576

include/igraph_structural.h
```diff
@@ -115,21 +115,21 @@ IGRAPH_EXPORT igraph_error_t igraph_avg_nearest_neighbor_degree(const igraph_t *
...
     const igraph_t *graph, igraph_vector_int_t *result,
     const igraph_vector_t *weights, igraph_fas_algorithm_t algo);
 
 IGRAPH_EXPORT igraph_error_t igraph_feedback_vertex_set(
```
ntamas:
```suggestion
IGRAPH_EXPERIMENTAL IGRAPH_EXPORT igraph_error_t igraph_feedback_vertex_set(
```

`igraph_feedback_vertex_set` is marked as experimental, both in the docs and in #2576

ntamas:
I compared the list of functions modified here with the documentation and with the list in #2576; the differences are highlighted in the suggestions above.

include/igraph_operators.h
szhorvat: I think it was forgotten to be added to this one.

`igraph_mycielski_graph()` seems fine to me, I don't expect to want to make changes.

`igraph_mycielskian()` has a behaviour which some may debate, namely the handling of 1-2 vertex graphs.

szhorvat:
I think this should be clear now, i.e. consistent with the docs.

The decision about what to finalize for 1.0 is for after this PR. Some of these will certainly be finalized.

include/igraph_operators.h
ntamas: Then feel free to restore the experimental qualifier here and also add it to the docs and to #2576

ntamas (approved):

Issue #2849, opened 2025-08-27
Sugiyama edge waypoints


iosonofabio:
**What is the feature or improvement you would like to see?**

Sugiyama currently returns an extended graph with ghost nodes. In my package `iplotx`, I use edge waypoints for routing and it works well. So if we wanted to simplify the current return a bit we could instead return a list of lists, where:

- the outer list refers to edges, i.e. it will have length `m` and the i-th element refers to the waypoints of the i-th edge in the original graph. If an edge has no waypoints (straight from source to sink), the i-th list is empty.

We might need to:
- return the length of each list somewhere. Alternatively, we could return a mapping from integers (edge indices) to a pair (int, array of float pairs) with the waypoints only of the edges that require them. This mapping would be empty if no waypoints are required at all.
- for undirected graphs, store somewhere which of the two vertices comes first, i.e. connects to the first waypoint. This might be unnecessary as long as we are consistent.

**Use cases for the feature**

Simplifying the visualization of Sugiyama-sryle layouts in iplotx and other packages.

**References**

https://iplotx.readthedocs.io/en/latest/gallery/plot_waypoints.html#sphx-glr-gallery-plot-waypoints-py

szhorvat:
I asked @iosonofabio to look at the interface and see if there's an opportunity for improvement before 1.0, and I support this.

iosonofabio:
At the C level, one could also return an array of custom structs with the following fields:

- edge_index (int)
- coordinates (i x 2 matrix of doubles, where i is the number of waypoints for this edge)

This way one avoids a sea of empty lists but does not have to use mappings. Making this more friendly towards downstream providers at the Python/R level (e.g. `iplotx`) can be done in the python-igraph/rigraph codebase.

szhorvat:
We do have an `igraph_matrix_list_t`, so we have a means of returning a list of waypoint lists.  There is a well-defined "to" and "from" end even for undirected edges (i.e. igraph is consistent about which is the "to" and which is the "from"), so no need to return extra information about which vertex is at which end of the waypoint list.

This makes a simple interface:

 - Vertex coordinates are returned in an $n\times d$ matrix, where $d$ is the dimension and $n$ the vertex count. Rows are the coordinate vectors. This is what we use now and it works fine for vertices.
[...]

PR #2850, opened 2025-08-29
Sugiyama layout now returns control points in a list of matrices

ntamas:
As discussed in #2849


ntamas:
Sorted out the doubts I've had with the correctness of edge directions (i.e. control point ordering) in the new implementation; it is also unit-tested now, so I think this can be merged.

PR #2854, opened 2025-09-05
Count successful swaps in igraph_rewire

### LINKED ISSUE #2848: igraph_rewire() should count successful swaps
`igraph_rewire()`: add extra output parameter that records the number of successful rewiring trials (i.e. successful swap)

stevesajeev1:
`igraph_rewire` now has `successful_swaps` as an output parameter, which records the number of successful rewiring trials.

Closes #2848.

This is my first contribution to this repository, so any feedback is appreciated. I would also like some advice on how to include this parameter in the unit tests, since I believe that the number of successful swaps will differ between each run.

### DIFF
--- a/examples/simple/igraph_assortativity_degree.c
+++ b/examples/simple/igraph_assortativity_degree.c
@@ -26,7 +26,7 @@ int main(void){
         printf("Assortativity before rewiring = %g\n", assortativity);
 
         /* Randomize the graph while preserving the degrees. */
-        igraph_rewire(&g, NULL, 20 * igraph_ecount(&g), IGRAPH_SIMPLE_SW);
+        igraph_rewire(&g, 20 * igraph_ecount(&g), IGRAPH_SIMPLE_SW, NULL);
 
         /* Re-compute assortativity. Did it change? */
         igraph_assortativity_degree(&g, &assortativity, /* ignore edge directions */ IGRAPH_UNDIRECTED);
--- a/examples/simple/igraph_assortativity_degree.c
+++ b/examples/simple/igraph_assortativity_degree.c
@@ -26,7 +26,7 @@ int main(void){
         printf("Assortativity before rewiring = %g\n", assortativity);
 
         /* Randomize the graph while preserving the degrees. */
-        igraph_rewire(&g, 20 * igraph_ecount(&g), IGRAPH_SIMPLE_SW);
+        igraph_rewire(&g, NULL, 20 * igraph_ecount(&g), IGRAPH_SIMPLE_SW);
 
         /* Re-compute assortativity. Did it change? */
         igraph_assortativity_degree(&g, &assortativity, /* ignore edge directions */ IGRAPH_UNDIRECTED);
--- a/examples/simple/igraph_assortativity_nominal.c
+++ b/examples/simple/igraph_assortativity_nominal.c
@@ -34,7 +34,7 @@ int main(void) {
         printf("Assortativity before rewiring = %g\n", assortativity);
 
         /* Rewire graph */
-        igraph_rewire(&g, NULL, 10 * igraph_ecount(&g), IGRAPH_SIMPLE_SW);
+        igraph_rewire(&g, 10 * igraph_ecount(&g), IGRAPH_SIMPLE_SW, NULL);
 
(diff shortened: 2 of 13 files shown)


szhorvat:
First of all, thanks for contributing to igraph.

However, the most important advice for new contributors, which is spelt out [in the contribution guide](https://github.com/igraph/igraph/?tab=contributing-ov-file#pull-requests): _always ask **before** you start any work!_

Otherwise we may get into an unpleasant situation where we are unable to accept your work, even if it is good and you spent a lot of time on it. Unfortunately, this might end up being one of these situations as we are approaching a big release, and were already in a near-feature-freeze ...

I'll get back to you about how we will proceed soon.

----

Feedback on the work:

 - It looks good at first glance.
 - Failing CI checks should be fixed. This is simple to do here, just a minor update needed in benchmarks.
 - Interfaces (`interfaces/functions.yaml`) need to be fixed, but this is again something easy that we usually do ourselves and do not expect from first-time contributors.

szhorvat:
@ntamas Shall we try to squeeze this in for 1.0 or abort? 

The idea behind #2848 was that the acceptance rate might be useful in diagnosing sampler performance, but at this time I don't have a good enough understanding of this topic, especially as it applies to graph sampling (which is different from some more common uses of MCMC due to the huge sampling space). I opened #2848 during one of the GSoC meetings, hence the very brief description.

I don't see real disadvantages to including this parameter, other than a slightly more complex interface. Recording the acceptance rate is always going to be easy, so it won't create a maintenance burden.  What exactly counts as "acceptance" is not entirely clear due to some edge cases, but I do not see it as a major problem if we slightly redefine this in the future.

My suggestion is to go ahead, assuming that this PR can be finalized in time.

ntamas:
@szhorvat The question is: do we anticipate similar output parameters that we want to use for diagnostics purposes but that are mostly optional and rarely used by the end user? In that case I'd rather prepare an `igraph_rewiring_stats_t` struct even if it has a single member only, and pass a _pointer_ to that struct as an input argument (where `NULL` can mean "not interested"). This way we can extend the struct later without changing the function signature and making the argument list even longer.

Also, IMHO the argument should be moved to the end of the argument list as it is optional and will mostly be unused.

szhorvat:
Let's do that. I see potential for recording more detailed statistics once this function improves, so introducing such a struct is useful. We already have `igraph_maxflow_stats_t`, so this is a good naming convention.

Our guidelines do say that output arguments should come first, but in this case I also prefer it last.

Can you please make these changes @stevesajeev1 ? Then we can squeeze this into the upcoming big release. Let me know if you have questions about any details.

stevesajeev1:
Should now be updated.

szhorvat:
@stevesajeev1 Great independent work, thank you! It needs some minor polish that's easier for me to do than explain, then it's good to go.

@ntamas How about making this struct (and not `igraph_rewire()`) experimental?  This is outside of our current definition of "experimental", which so far applied only to functions. I'd need to research if it's possible to emit compiler warnings when using a struct.

Motivation/Pros: This would make it possible to add extra stats within the 1.x series, something we might actually want to do, as we are actively working on this functionality now. I am thinking of stats relating to 2-switches vs 3-switches. What

Cons: This is now seriously twisting the definition of "experimental", which was so far only for functions. Given the new structured way to mark functions as experimental, I'd be happy with this only if the compiler warning functionality can be retained for structs as well. I'm not sure this is possible, I'll need to check.

ntamas:
@szhorvat I don't think that you can mark structs as experimental and make the compiler emit a warning - but even if you do, we would be breaking our promise to let people integrating igraph into Linux distributions do away with experimental stuff by setting a compiler flag at compile time as we can't get rid of the struct itself. And technically, adding a new field to a struct is ABI-breaking.

We can pre-allocate a few extra members in the struct with an anticipated type (`igraph_integer_t` or `igraph_real_t`), and leave them unnamed or unused. If we want to make use of them later, we can _rename_ them in the struct without breaking ABI compatibility. How does that sound?

szhorvat:
Let's not complicate things, and leave it as-is ... I don't have the capacity for this tonight. I hopefully fixed the Stimulus issue now.

After merging, the changelog needs an update.

szhorvat:
@ntamas Can you please handle this before 1.0?

ntamas:
Thanks a lot!

PR #2858, opened 2025-09-10
refactor!: const-correctness in igraph_motifs_handler_t

szhorvat:


szhorvat:
@ntamas OK to merge this? (I don't see why not.)

Issue #2847, opened 2025-08-25
Make sure no abort() or exit() exists in tests/examples


szhorvat:
Make sure no abort() or exit() exists in **tests/examples** because failure modes can be very confusing with this. We just got bit while debugging GSoC code.

jgmbenoit:
There are also some `exit()` and similar in the library part. This issue is detected by `lintian(1)` with the tag [exit-in-shared-library](https://lintian.debian.org/tags/exit-in-shared-library.html). It would be a very good idea to get ride of them.

szhorvat:
@jgmbenoit I opened this issue during a meeting and it isn't written up properly. The `exit()` won't be fully removed from examples/tests _yet_.

But there should be no `exit()` in the library, and all calls to `abort()` should go through `igraph_abort()`.

I grepped the sources and found no `exit()` that isn't commented out. The one in `lsap.c` and `degree_sequence_vl` are commented out, the one in prpack is prevented from getting compiled in through `#ifdef`s. The first two need cleanup/replacement, while we'd like to keep prpack fairly close to the original at https://github.com/dgleich/prpack so it's easier to pick up fixes (although at this point it's just us who are making fixes).

Can you please help track down where you find the `exit()`?

szhorvat:
@jgmbenoit Is this problem new in igraph 1.0.0? I found one use of `exit()` in Qhull which I will replace with an assertion, as this code should never be reached. I cannot find any other uses of `exit()`, except in vendored libraries that you are using external copies of anyway.

szhorvat:
I see where it's coming from: it's the flex-generated code, `yy_fatal_error()`.  We do "override" this by defining the `YY_FATAL_ERROR` macro, but flex still emits a definition of `yy_fatal_error()` and I'm not sure how to prevent this. I would expect the linker not to include `yy_fatal_error()` in the shared library since it is never actually called. Is there another hidden `exit()` somewhere?

jgmbenoit:
It is reported by `lintian(1)` and, so far I can remember, there was no surch report in the previous version. I came to the same conclusion of you. I will have a look on the flex stuff this week-end.

szhorvat:
I found all of them now. Note to self: Qt Creator's "Find usages" feature works effectively for finding uses of `exit()`.

PR #2868, opened 2025-10-06
fix: eliminate references to exit() from igraph

szhorvat:
Eliminate references to `exit()` from igraph. These got added to igraph 1.0.0 through Infomap and Qhull.

Refs #2847

Thanks for reporting this @jgmbenoit! I think we're good now. `nm -g src/libigraph.dylib | grep exit` finds nothing on my machine (not even when building with all-vendored libraries).

Calls to `exit()` are either eliminated or replaced with `IGRAPH_FATAL()`, which will emit some useful information and tell the user to report the occurrence as a bug. These calls to `IGRAPH_FATAL()` should all be unreachable—they effectively serve as assertions of unreachability here.

While `flex` generates code that references `exit()`, this code is never called so the linker eliminates it from the shared library. We do not need to worry about the flex-generated code.

### DIFF
--- a/vendor/infomap/src/io/ProgramInterface.cpp
+++ b/vendor/infomap/src/io/ProgramInterface.cpp
@@ -10,6 +10,8 @@
 #include "ProgramInterface.h"
 #include "../utils/Log.h"
 
+#include "igraph_error.h"
+
 #include <iostream>
 #include <cstdlib>
 #include <map>
@@ -45,6 +47,9 @@ ProgramInterface::ProgramInterface(std::string name, std::string shortDescriptio
   addOptionArgument(m_printJsonParameters, "print-json-parameters", "Print Infomap parameters in JSON.", "About").setHidden(true);
 }
 
+/* Modification for igraph: Disable functions that call forbidden
+ * functions such as exit(). */
+#if 0
 void ProgramInterface::exitWithUsage(bool showAdvanced) const
 {
   Log() << "Name:\n";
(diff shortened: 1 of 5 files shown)


szhorvat:
If we do end up with an igraph 1.0.1, this should be included.

szhorvat:
@krlmlr Do you know what is the complete list of CRAN-forbidden symbols so we can check for all of them?

vendor/infomap/src/io/ProgramInterface.cpp
```diff
@@ -45,6 +47,9 @@ ProgramInterface::ProgramInterface(std::string name, std::string shortDescriptio
   addOptionArgument(m_printJsonParameters, "print-json-parameters", "Print Infomap parameters in JSON.", "About").setHidden(true);
 }
 
+/* Moficication for igraph: Disable functions that call forbidden
```
ntamas:
```suggestion
/* Modification for igraph: Disable functions that call forbidden
```

vendor/infomap/src/io/ProgramInterface.cpp
```diff
@@ -178,6 +191,9 @@ std::string toJson(const Option& opt)
                    << " }";
 }
 
+/* Moficication for igraph: Disable functions that call forbidden
```
ntamas:
```suggestion
/* Modification for igraph: Disable functions that call forbidden
```

vendor/infomap/src/io/ProgramInterface.cpp
```diff
@@ -298,12 +315,16 @@ void ProgramInterface::parseArgs(const std::string& args)
           }
         }
       }
+    /* Moficication for igraph: Disable calls to functions that
```
ntamas:
```suggestion
    /* Modification for igraph: Disable calls to functions that
```

vendor/infomap/src/io/ProgramInterface.h
```diff
@@ -392,10 +392,16 @@ class ProgramInterface {
   unsigned int numRequiredArguments() const { return m_nonOptionArguments.size() - m_numOptionalNonOptionArguments; }
 
 private:
+/* Moficication for igraph: Disable functions that call forbidden
```
ntamas:
```suggestion
/* Modification for igraph: Disable functions that call forbidden
```

ntamas (approved):

krlmlr:
Re forbidden symbols: perhaps here?

https://github.com/r-devel/r-svn/blob/d7a43e1d7c7db3dc36d04cb3fd6ef13e9f0ceb91/src/library/tools/R/sotools.R#L104

Thanks for checking. In the worst case, our GHA runs will catch that early in the future.

PR #2869, opened 2025-10-07
fix: eliminate references to std::cout

szhorvat:
Infomap contained a bunch of references to `std::cout` which CRAN forbids. This PR eliminates them. Unfortunately, the logging and debug system in Infomap is a bit messy, so a lot of patching was required, which reduces maintainability.

I tried to keep using `#if 0 ... #endif` to make the patches easier to identify when updating Infomap in the future. The only good option I can think of here is to help clean up Infomap and make it more suitable for inclusion, or ideally for dynamic linking. But I have no time for that now.

CC @krlmlr @vtraag

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Fixed
 
  - Eliminated references to `exit()` from the igraph shared library. These were accidentally introduces into igraph 1.0.0 through Qhull and Infomap.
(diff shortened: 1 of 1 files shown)


vtraag:
This sounds like the most reasonable approach for now at least.

Perhaps it can indeed be cleaned up somehow at Infomap itself to at least only need us to have only a single place to remove the printing of output from. It should not be too difficult to do so.

PR #2875, opened 2025-10-21
fix: more robust compilation on some Apple systems

szhorvat:
This defines `_DARWIN_C_SOURCE` to improve compilation robustness on misbehaving Apple systems.

On some old Apple systems, as well as with MacPorts's GCC, defining `_POSIX_C_SOURCE` can break some standard C++ headers (i.e. including them triggers errors). This is unfortunate as `_POSIX_C_SOURCE` is the standard way to enable POSIX features, which improve igraph functionality/performance. It seems that defining `_DARWIN_C_SOURCE` explicitly fixes these types of issues.  There are many discussions on GitHub about this, going back over 10 years. 

@ntamas Can you please try to build igraph with Homebrew's GCC _without_ this patch, and see if it works? I don't have Homebrew. If it does not work, can you test with this patch as well to see if it fixes it?

Let's not merge this yet, as it's unclear if defining `_DARWIN_C_SOURCE` can have negative effects in some circumstances. Let's wait for https://github.com/macports/macports-ports/pull/29711 first, and see how compilation goes.

### DIFF
--- a/etc/cmake/compilers.cmake
+++ b/etc/cmake/compilers.cmake
@@ -8,6 +8,17 @@ include(CheckCCompilerFlag)
 #  - https://www.gnu.org/software/libc/manual/html_node/Feature-Test-Macros.html
 add_compile_definitions(_POSIX_C_SOURCE=200809L)
 
+# On some Apple systems, with some compilers, defining _POSIX_C_SOURCE causes compilation
+# failures when including C++ headers. Defining _DARWIN_C_SOURCE explicitly usually fixes
+# this. Refs: https://trac.macports.org/ticket/60655 and https://trac.macports.org/ticket/73145
+# This is claimed to be due to _POSIX_C_SOURCE disabling APIs that C++ headers rely on
+# on some systems. _DARWIN_C_SOURCE re-enables these APIs, but _GNU_SOURCE alone does not 
+# necessarily do so. Using different global defined for C and C++ is problematic with CMake,
+# so instead of restricting _POSIX_C_SOURCE to C, we define _DARWIN_C_SOURCE.
+if(APPLE)
+  add_compile_definitions(_DARWIN_C_SOURCE)
+endif()
+
 if(MSVC)
   add_compile_options(/FS)
   add_compile_definitions(_CRT_SECURE_NO_WARNINGS) # necessary to compile for UWP


kencu:
some exploration on chatgpt suggests your troubles with missing features in libc when compiling c++ code probably result from this commit:

https://github.com/igraph/igraph/commit/c08151dd327418943beb502b9fb6a9df043f40cd


chatgpt:

When you run:
g++ prog.cpp
the compiler’s built-in macros (checkable via g++ -dM -E - < /dev/null | grep SOURCE) include:
#define _GNU_SOURCE 1
#define _ISOC99_SOURCE 1
#define _POSIX_C_SOURCE 200809L
These are injected before system headers are parsed, ensuring that glibc exposes all symbols the C++ library might use.
However, if you manually define _POSIX_C_SOURCE without _GNU_SOURCE:
g++ -D_POSIX_C_SOURCE=200809L prog.cpp
then _GNU_SOURCE is no longer pre-defined (your definition overrides the internal logic), and suddenly functions like quick_exit vanish from the C headers — which breaks <cstdlib>.
That’s why the comment says:
It’s not compatible for users to restrict libc to anything less than the default feature set.
Because doing so literally removes symbols the C++ library requires to compile.
💡 In Other Words
The C++ runtime depends on a “rich” libc feature set (not just POSIX).
GCC ensures that by predefining macros like _GNU_SOURCE.
When you override them with _POSIX_C_SOURCE, you tell libc:
“hide everything not strictly in POSIX,” and that breaks libstdc++.
So the compiler expects:
_GNU_SOURCE → all features
but you’re forcing:
_POSIX_C_SOURCE → limited feature set
and the two are incompatible for normal C++ compilation.
✅ The Right Takeaway
For C++, you should not restrict libc feature macros — ever — unless you fully control both the headers and the runtime (e.g. embedded systems with a minimal libc).
Instead:
If you need POSIX APIs like clock_gettime or pthread_*, you already get them under _GNU_SOURCE.
Don’t define _POSIX_C_SOURCE yourself — or, if you must (for mixed C codebases), isolate it in C source files, not in C++ ones.

szhorvat:
I'd prefer to stay conservative and use only `_DARWIN_C_SOURCE` on Apple systems, instead of using `_GNU_SOURCE` everywhere. CMake makes it easy to test for Apple systems (`if(APPLE)`), and so far we've only encountered this issue in limited circumstances on macOS.

Furthermore, `_GNU_SOURCE` does not work around https://trac.macports.org/ticket/73145 but `_DARWIN_C_SOURCE` does.

kencu:
The upshot is that it is simply invalid to set _POSIX_C_SOURCE yourself to some value when compiling c++ code at all.

Perhaps someone might look at how it came to be believed that the build needed to set _POSIX_C_SOURCE when compiling c++ code in the first place, and re-examine that decision.

kencu:
eg this one of many references that say the same thing:

https://docs.oracle.com/cd/E88353_01/html/E37853/c-plus--plus--7.html

"specifying feature test macros such as _POSIX_SOURCE, _POSIX_C_SOURCE, and _XOPEN_SOURCE can result in compilation errors due to conflicting requirements of standard C++ and these specifications."

szhorvat:
This is unfortunately an ugly situation. There is almost no official documentation on this incompatibility between C++ and POSIX or X/Open. The only one I could find is the page linked above from Oracle.

On the vast majority of systems there seems to be no problem. It is indeed silly not to be able to use POSIX functions (which are sometimes unavailable without the feature test macro) from C++. [The standard seems to say](https://pubs.opengroup.org/onlinepubs/007904875/functions/xsh_chap02_02.html) that while defining _POSIX_C_SOURCE can remove symbols, it doesn't need to remove any from reserved namespaces (whatever "namespace" may refer to in this context, does it apply to C++'s `std`?) It also says that other feature test macros can (re-)enable symbols.

Defining _DARWIN_C_SOURCE allows igraph to compile fine on MacPorts's infrastructure with a very large number of macOS versions.

Given that restricting _POSIX_C_SOURCE to C only is a mess with CMake, and inconsistencies in the symbols defined for C and C++ carry their own risks for future maintenance/development, I think the best solution at this time is to just define _DARWIN_C_SOURCE, as in this PR.

PR #2880, opened 2025-11-01
feat: Provide separate extra argument for each callback function

krlmlr:
Prepared by Copilot, builds on my system, couldn't run tests on my local machine (`make test`) or in the devcontainer (`ninja test`).

Callbacks in R are implemented using closures. To assist autogeneration, it will be helpful if each callback has their own "extra" argument.

We can work around, I wonder if this is something worth doing here anyway.


szhorvat:
Unfortunately we cannot accept this because it's API-breaking, and 1.0 is already released. The solution would be to pack multiple pieces of data into the single `extra` argument. You could use a `struct` for this, there should be a way to make autogen still work with a suitable wrapper function ...

CC @ntamas

krlmlr:
Agreed. We can evolve new interfaces in the downstream packages and then decide how to change the API.

Wrapper function in https://github.com/igraph/rigraph/pull/2465 .

PR #2886, opened 2026-01-12
Doc updation for igraph_vcount

aryarathoree:
Hi! I'm an aspiring contributor. While going through the docs I found out that there is no proper description of the cases wherein the specific function doesn't works. I have started with v_count() - that it doesn't handle NULL case. I would to firstly work on docs while going through this project.
Thanks

### DIFF
--- a/src/graph/type_indexededgelist.c
+++ b/src/graph/type_indexededgelist.c
@@ -822,7 +822,8 @@ igraph_error_t igraph_delete_vertices_map(
  * \return Number of vertices.
  *
  * \remark
- * The graph pointer must be non-NULL and refer to a properly initialized graph object. Behavior is undefined otherwise.
+ * The graph pointer must be non-NULL and refer to a properly initialized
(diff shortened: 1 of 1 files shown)


szhorvat:
It's good to see your interest in contributing to igraph. 

However, in this case I vote against adding this clarification. There is value in conciseness.

In igraph, null pointers are only allowed for parameters that are optional. We do not point out when a null pointer is disallowed. That is the usual situation. We only point out the opposite, when a parameter is optional, and may be `NULL`.

If you are interested in making other contributions to the documentation, we can talk about that. I will probably have capacity for reviewing PRs or proposing documentation projects in February the earliest.

I will leave this open for a while in case other maintainers wish to chime in.

aryarathoree:
Hi! Thankyou so much for taking out time to clarify about `NULL` in igraph . I'll find more relevant issues to work on right now.
Thanks again!

PR #2898, opened 2026-03-10
docs: categories

### LINKED ISSUE #2889: Add "category" field to Stimulus metadata?
The igraph C docs are organized by category: https://igraph.org/c/html/latest/

The igraph R docs are also organized by category: https://r.igraph.org/reference/index.html

We'd like to ask whether the categories could be a YAML field in https://github.com/igraph/igraph/blob/main/interfaces/functions.yaml

So https://github.com/igraph/igraph/blob/2329c0ed9bbf5e7d41397583849d5746b36b5e29/interfaces/functions.yaml#L2629-L2633

would become

```yaml

igraph_find_cycle:
    PARAMS: |-
        GRAPH graph, OPTIONAL OUT VERTEX_INDICES vertices, OPTIONAL OUT EDGE_INDICES edges,
        NEIMODE mode=OUT
    DEPS: vertices ON graph, edges ON graph
    CATEGORY: graph-cycles
```

We'd use this information for the docs grouping but in the R case, also for splitting our `R/aaa-auto.R` script, which contains the automatically generated wrappers, into several scripts, one per category. The current script is too big, for instance I know I can't probably navigate it using symbol search in my IDE.

What do y'all think? If you agree, can we (the R igraph team) help make it happen?

maelle:
Fix #2889

I now need to have a look at the R side of things: how do I map this to the pkgdown config.

@krlmlr @schochastics

### DIFF
--- a/interfaces/functions-categories.yaml
+++ b/interfaces/functions-categories.yaml
@@ -1,0 +1,1914 @@
+# Functions ordered by category
+igraph_add_edge:
+    category: basicigraph
+    subcategory: adding-and-deleting-vertices-and-edges
+
+igraph_add_edges:
+    category: basicigraph
+    subcategory: adding-and-deleting-vertices-and-edges
+
+igraph_add_vertices:
+    category: basicigraph
+    subcategory: adding-and-deleting-vertices-and-edges
+
+igraph_delete_edges:
+    category: basicigraph
+    subcategory: adding-and-deleting-vertices-and-edges
+
+igraph_delete_vertices:
+    category: basicigraph
+    subcategory: adding-and-deleting-vertices-and-edges
+
+igraph_delete_vertices_map:
+    category: basicigraph
+    subcategory: adding-and-deleting-vertices-and-edges
+
+igraph_degree:
+    category: basicigraph
+    subcategory: basic-query-operations
+
+igraph_ecount:
+    category: basicigraph
+    subcategory: basic-query-operations
+
+igraph_edge:
+    category: basicigraph
+    subcategory: basic-query-operations
+
+igraph_edges:
+    category: basicigraph
+    subcategory: basic-query-operations
+
+igraph_get_all_eids_between:
+    category: basicigraph
+    subcategory: basic-query-operations
+
+igraph_get_eid:
+    category: basicigraph
+    subcategory: basic-query-operations
+
+igraph_get_eids:
+    category: basicigraph
+    subcategory: basic-query-operations
+
+igraph_incident:
+    category: basicigraph
+    subcategory: basic-query-operations
+
+igraph_is_directed:
+    category: basicigraph
+    subcategory: basic-query-operations
+
+igraph_neighbors:
+    category: basicigraph
+    subcategory: basic-query-operations
+
+igraph_vcount:
+    category: basicigraph
+    subcategory: basic-query-operations
+
+igraph_copy:
+    category: basicigraph
+    subcategory: graph-constructors-and-destructors
+
+igraph_empty:
+    category: basicigraph
+    subcategory: graph-constructors-and-destructors
+
+igraph_empty_attrs:
+    category: basicigraph
+    subcategory: graph-constructors-and-destructors
+
+igraph_expand_path_to_pairs:
+    category: basicigraph
+    subcategory: misc-helper-functions
+
+igraph_invalidate_cache:
+    category: basicigraph
+    subcategory: misc-helper-functions
+
+igraph_is_same_graph:
+    category: basicigraph
+    subcategory: misc-helper-functions
+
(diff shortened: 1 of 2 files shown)


maelle:
My current thinking on the R side is that igraph.r2cdocs should also take care of `_pkgdown.yml`, so `roxygenize()` would update the reference index configuration.

interfaces/functions-categories.yaml
```diff
@@ -0,0 +1,1914 @@
...
+    category: basicigraph
+    subcategory: adding-and-deleting-vertices-and-edges
+
+igraph_degree:
+    category: basicigraph
+    subcategory: basic-query-operations
+
+igraph_ecount:
+    category: basicigraph
+    subcategory: basic-query-operations
```
krlmlr, line 32: Why not:


```suggestion
basicigraph:
  adding-and-deleting-vertices-and-edges:
    - igraph_add_edge
    - igraph_add_edges
    - igraph_add_vertices
    - igraph_delete_edges
    - igraph_delete_vertices
    - igraph_delete_vertices_map
  basic-query-operations:
    - igraph_degree
    - igraph_ecount
    - ...
```

krlmlr (commented):
Thanks, nice! Do we want to add a comment (perhaps the first 10 words or so) next to the function name?

krlmlr:
Re igraph.r2cdocs: Let's carefully evaluate the pros and cons of hosting this in a separate repo.

szhorvat:
I've been thinking about this since we've spoken, and again, I am not too keen on doing too much with this categorization, or creating an extra maintenance burden by requiring that these need to be kept up-to-date.

These categories are simply not that meaningful, and very subjective. I've just implemented a function to check if a graph is triangle free. Where does that fit?  Structural properties? Transitivity? Motifs (where some triangle functions are)? It really is useful for graph theory, not network science, so should we have a category for that?

It feels like putting a lot of effort into formalizing something that is far from being properly fleshed out, and far from ready to be set in stone.

You said that you'd prefer to use this to break up `rinterface.c` and `aaa-auto.R` into several files. The usual motivation for breaking up source files is maintainability. But these files are auto-generated, so their size has no impact on maintainability. In fact, there is an advantage to keeping all auto-generated code together to make it easier to search.

I see that this is generated with an R script. There are two possibilities here: 1. The file is initially generated, then manually kept up to date. This is extra burden that I am not in favour of. 2. The file is always generated. I absolutely do not want to add a dependence on R for the C/igraph build system.

Given that this information is auto-generated with an R script, and is intended to be used in R/igraph, I think this should go into the R/igraph system (i.e. not this repository).

What is your take on this @ntamas ? The relevant issue is https://github.com/igraph/igraph/issues/2889

To summarize the objections: 1. If manually maintained, this is extra burden. 2. This repo should have no R dependence. 3. Not currently useful for C/igraph, and the categorization is not well-founded.

maelle:
@krlmlr first 10 words of what?

I'll explore a bit more then we can discuss where to put that if we want to keep using those categories for the R scripts and pkgdown config.

@szhorvat I think part of the idea is that R/aaa-auto.R is so big that it's not nice to open and that it makes AI read too much, but I might be mistaken.

maelle:
For docs, it's fine if a function is in several categories, or only one, because site-wide search and cross-references from similar functions should help navigation.

For scripts (and in R, we align test files with code files), the splitting might not be perfect but it's better than nothing.

An automatic approach might be imperfect but it's ok. Now, it's true that in the R case, we already made a lot of efforts to create the pkgdown config.

interfaces/functions-categories.yaml
maelle: right, we can choose that structure too later.

maelle:
Reporting back. Generating part of the pkgdown reference config from the categories would work but some R functions are pure R so we'd need to decide on categories for those anyway. https://github.com/igraph/igraph.r2cdocs/pull/9

krlmlr:
Random thought: should the top-level category be the scientific field (physics, social sciences, graph theory, general-purpose graph algorithms)? It's fine if a function lives in multiple categories anyway.

Let's discuss separately where this lives and how it is maintained.

szhorvat:
> should the top-level category be the scientific field (physics, social sciences, graph theory, general-purpose graph algorithms)?

No, that will not work well here.

> It's fine if a function lives in multiple categories anyway.

OK, let's do this so that multiple categories are allowed, which makes things a lot easier (there can be a main one if necessary). Would it be easier if subcategories are not used?

maelle:
> Random thought: should the top-level category be the scientific field (physics, social sciences, graph theory, general-purpose graph algorithms)? It's fine if a function lives in multiple categories anyway.

@krlmlr this sounds like an idea for a `@family` in the R igraph package docs.

Issue #2911, opened 2026-05-22
simple_cycles() returns empty for a graph with cycles


KevCaz:
Dear maintainer, 

While working with igraph I found what I believe is a bug. Hopefully I did not miss something obvious here. This may be a bug for the C package, but I have no experience with it.

Thanks for all your work on this. 

### What happens, and what did you expect instead?

Here is the issue `igraph::simple_cycles()` returns an empty list on a directed graph that **does** contain a cycle (a 2-cycle), while `is_acyclic()` correctly reports the graph is not acyclic. The bug is **vertex-order-dependent**: permuting the vertex sequence makes `simple_cycles()` find the cycle.

### To reproduce

```R
library(igraph)
#> 
#> Attaching package: 'igraph'
#> The following objects are masked from 'package:stats':
#> 
#>     decompose, spectrum
#> The following object is masked from 'package:base':
#> 
#>     union
m <- matrix(0, 6, 6, dimnames = list(as.character(1:6), as.character(1:6)))
m["1","2"] <- 1; m["1","6"] <- 1; m["2","6"] <- 1
m["3","6"] <- 1; m["5","4"] <- 1; m["5","6"] <- 1
m["6","3"] <- 1; m["6","4"] <- 1
g <- graph_from_adjacency_matrix(m, mode = "directed", weighted = TRUE)

is_acyclic(g) 
#> [1] FALSE
# no cycles
simple_cycles(g)
#> $vertices
#> list()
#> 
#> $edges
#> list()

# Permutation
perm <- match(igraph::V(g)$name, c("1","2","6","3","5","4"))
g2 <- igraph::permute(g, perm)
igraph::V(g2)$name         
#> [1] "1" "2" "6" "3" "5" "4"
igraph::simple_cycles(g2) 
#> $vertices
#> $vertices[[1]]
#> + 2/6 vertices, named, from f460b4d:
#> [1] 6 3
#> 
#> 
#> $edges
#> $edges[[1]]
#> + 2/8 edges from f460b4d (vertex names):
#> [1] 6->3 3->6
```

### System information

```R
sessionInfo()
#> R version 4.6.0 (2026-04-24)
#> Platform: x86_64-pc-linux-gnu
#> Running under: Ubuntu 26.04 LTS
#> 
#> Matrix products: default
#> BLAS:   /usr/lib/x86_64-linux-gnu/blas/libblas.so.3.12.1 
[...]

PR #2913, opened 2026-05-24
Use subcomponents to more accurately determine where to search cycles for

GenieTim:
Fixes issue #2911.

I am not yet happy with the implementation, but it would be ok I guess; I submit here and now to show that I am working on it.

What I do currently is to replace the apparently too simplified way I used previously to detect components with a more real component detection. The alternative, using `igraph_subcomponent`, was decided against since it would require an additional pointer to the graph on the current search state; I don't remember why we don't have one, but tried to keep at it.

GitHub Copilot was used during the preparation of this PR.

- [X] By submitting this pull request, I assign the copyright of my contribution to _The igraph development team_.

### DIFF
--- a/src/cycles/simple_cycles.c
+++ b/src/cycles/simple_cycles.c
@@ -136,6 +136,12 @@ static igraph_error_t simple_cycles_unblock(
     return IGRAPH_SUCCESS;
 }
 
+/**
+ * Marks the strongly connected component of 
+ * 
+ * the start vertex as exhausted, using the existing subcomponent
+ * implementation from igraph.
+ */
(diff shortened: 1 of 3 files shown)


GenieTim:
If I'm not mistaken, now after re-reading the Johnson paper, what I missed is that Ak is only the adjlist of an SCC, whereas I just used the full graph. While technically not a problem, it's what made my alternative performance improvement necessary, yet that one was, as we saw in the example, not quite compliant with the rest of the algorithm.

GenieTim:
Closing in favor of #2916, which is a bit more elegant / less breaking changes and faster.

PR #2926, opened 2026-09-07
Perf: count_reachable should do O(components) popcounts instead of O(v)

jberg5:
By way of example,

```r
library(igraph)
set.seed(42)
g <- sample_gnm(n = 1000000, m = 3000000, directed = FALSE)
system.time(counts <- count_reachable(g))
```

goes from 3s (yay for speedy vectorized popcount) to 0.5s on my mac, 6x faster. On commodity x86 cloud hardware, popcount isn't vectorized in the default build, so the baseline is ~30s, speedup is around 50-70x.

(note that I put gpt6 through the wringer to get local builds + R working on my mac and on gcloud - wouldn't recommend doing it by hand).

Previously, `igraph_count_reachable` did one`igraph_bitset_popcount` per vertex to get the corresponding size of the connected component to which it belonged. However, there are only as many component sizes as there are components. In the worst case, for a fully connected graph, we were counting the same bits O(v) times.

This change makes a pass over `reach` to do one popcount per component first, and store those results in an intermediate int vector. Then the `counts` vector can be built by quickly looking up the size of the vertex's corresponding component, without re-consuming the bitset every time.

The worst case is that we have an edgeless graph, the number of components == the number of vertices, and we provision O(v) space to hold the temporary component size array. In this case we end up being marginally slower because of the additional allocation. To give a specific example, at 100,000 disconnected vertices, both intel and amd are around 0.3% slower. I pushed GPT6 to find all the regressions it could; I've put them in a table below.

I think doing fewer popcounts on average is worth a slight regression 

AI disclosure: GPT6 did the first pass for the code you see here. I tweaked a couple things but mostly LGTM.
It did a *lot* of work getting local builds working so it could sweep through different problem sizes.

#### Average-case workloads

| Graph | Vertices | Component size | Intel before -> after (ms) | Intel speedup | AMD before -> after (ms) | AMD speedup |
|---|---:|---:|---:|---:|---:|---:|
| Disjoint undirected paths | 65,536 | 2 | 148.746 -> 94.511 | 1.58x | 155.366 -> 89.812 | 1.73x |
| Disjoint undirected paths | 65,536 | 4 | 128.942 -> 47.230 | 2.75x | 143.047 -> 45.199 | 3.17x |
| Disjoint undirected paths | 65,536 | 16 | 114.149 -> 12.179 | 9.37x | 134.395 -> 12.164 | 11.06x |
| Disjoint undirected paths | 65,536 | 64 | 110.583 -> 3.762 | 29.52x | 132.559 -> 4.017 | 32.98x |
| Disjoint undirected paths | 65,536 | 256 | 109.792 -> 1.765 | 62.21x | 132.107 -> 2.047 | 65.04x |
| Disjoint undirected paths | 65,536 | 1,024 | 109.637 -> 1.323 | 82.84x | 131.887 -> 1.703 | 77.73x |
| One undirected path | 65,536 | 65,536 | 108.307 -> 0.163 | 666.30x | 130.305 -> 0.130 | 999.09x |

#### Identified Regressions

| CPU | Graph | Vertices | Measurement | Before | After | Paired time increase [95% interval] |
|---|---|---:|---|---:|---:|---:|
| AMD | Directed path, OUT | 8 | First call | 5.300 us | 7.350 us | +43.00% [34.95%, 49.07%] |
| AMD | Directed path, IN | 8 | First call | 5.240 us | 7.460 us | +40.35% [31.66%, 47.43%] |
| Intel | Directed path, IN | 8 | First call | 7.629 us | 9.064 us | +22.89% [12.85%, 28.80%] |
| AMD | Singleton | 1 | Repeated | 83.735 ns | 102.311 ns | +22.18% [22.12%, 22.35%] |
| Intel | Singleton | 1 | Repeated | 72.309 ns | 87.422 ns | +20.90% [20.78%, 21.23%] |
| Intel | 32 paths of size 4 | 128 | Repeated | 4.253 us | 4.469 us | +5.15% [4.85%, 5.72%] |
| Intel | 32 paths of size 4, permuted IDs | 128 | Repeated | 4.275 us | 4.487 us | +5.02% [4.61%, 5.24%] |
| AMD | 32 paths of size 4, permuted IDs | 128 | Repeated | 4.292 us | 4.357 us | +1.47% [0.45%, 2.57%] |
| AMD | Directed path, IN | 64 | Repeated | 15.659 us | 16.018 us | +2.37% [2.07%, 2.70%] |
| Intel | Directed out-star, IN | 128 | Repeated | 27.372 us | 28.253 us | +3.06% [2.96%, 3.44%] |
| Intel | Isolated vertices | 100,000 | Repeated | 809.602 ms | 811.875 ms | +0.35% [0.24%, 0.50%] |
| AMD | Isolated vertices | 100,000 | Repeated | 1062.531 ms | 1064.645 ms | +0.36% [0.19%, 0.43%] |


Tagl (approved):
Ahh, thanks for the improvement!

This is a silly omission on my part in translating my original version of this in C++ and Python, as I did do exactly this there.

I do agree that improving the speed in almost all cases is worth the slow down in the exceptional cases where almost all strong components are 1 vertex.

szhorvat:
Thank you!

Let me know how you'd like to be acknowledged in the changelog. Normally we use full name and GitHub handle.

jberg5:
Thanks @szhorvat ! Yes name and github handle (Jonathan Berg, @jberg5 ) works for me.

PR #2930, opened 2026-09-22
fix: more robust compatibiliy across libxml2 versions and patched libxml2 shipped with some Ubuntu and Fedora versions

szhorvat:
This PR eliminates reliance on knowledge of the layout of the `xmlEntity` struct. 

This struct has changed in libxml2 2.11, which we used to handle by checking the `LIBXML_VERSION` macro. However, it was also patched, for security reasons, in an old Fedora or RedHat version (I'm writing this from memory) [and now in new Ubuntu versions](https://bugs.launchpad.net/ubuntu/+source/libxml2/+bug/2141260?comments=all). There is no good way to check for this.


szhorvat:
This won't work because `xmlFreeEntity` is not exposed in older libxml2 versions, including version 2.9 which ships with Ubuntu.

szhorvat:
It appears that `xmlEntityPtr` is intended to be castable to `xmlNodePtr`, and we can use `xmlFreeNode()`.

